PR #71, opened 2020-05-17
Throw on invalid env key

### LINKED ISSUE #69: Invalid ENV declaration should return an error
If I do the following in my `build.earth` file:

```
ENV ${CODE_SRC}=/go/src/<project>

protobuf:
    COPY --dir api ${CODE_SRC}/backend
```

Then running `earth +protobuf` results in the files being copied into `/backend`:

`+protobuf | --> COPY [api] /backend`

I have worked around this by specifying the full path in the copy command:

```
ENV ${CODE_SRC}=/go/src/<project>

protobuf:
    COPY --dir api /go/src/<project>/backend
```

drews256:
Cool project. I was looking for something like this recently, wish I had seen it before I found a different solution.

Here is an attempt to fix #69 

I used regex to match invalid characters. It's specific to the error seen in #69. It matches against `[]{}$` characters. Though the more I think about it maybe it should be the other way around and ensure a match on uppercase A-Z. 

I figured I'd open this guy up first.

I have a log statement in there that should be taken out if this is the path you wanna go down.


earthfile2llb/converter.go
```diff
@@ -477,14 +478,21 @@ func (c *Converter) Volume(ctx context.Context, volumes []string) {
 }
 
 // Env applies the ENV command.
-func (c *Converter) Env(ctx context.Context, envKey string, envValue string) {
+func (c *Converter) Env(ctx context.Context, envKey string, envValue string) error {
 	envValue = c.expandArgs(envValue)
+	invalidEnvKey, err := regexp.MatchString(`[${}()]`, envKey)
```
vladaionescu, line 483: In would go for `[a-zA-Z_]+[a-zA-Z0-9_]*` as you suggested in the PR description. [Source](https://stackoverflow.com/a/2821201).

earthfile2llb/converter.go
```diff
@@ -477,14 +478,21 @@ func (c *Converter) Volume(ctx context.Context, volumes []string) {
...
+	logging.GetLogger(ctx).With("env-key", envKey).With("env-key-value", invalidEnvKey).Info("ENV Key Value")
+	if invalidEnvKey || err != nil {
+		logging.GetLogger(ctx).Info("Failed to apply ENV")
+		return fmt.Errorf("Invalid env key definition %s", envKey)
+	}
```
vladaionescu, line 488: Can you move this logic into `listener.go`? (ie the caller of this)

The overall idea is to put any syntax related checks before it gets to the converter.

vladaionescu, line 488: Side note: No need to do logging - just make sure `l.err` is set after you move to listener.

go.mod
```diff
@@ -23,8 +23,12 @@ require (
...
 	golang.org/x/crypto v0.0.0-20200302210943-78000ba7a073
 	golang.org/x/lint v0.0.0-20190930215403-16217165b5de // indirect
-	golang.org/x/sync v0.0.0-20190911185100-cd5d95a43a6e
-	golang.org/x/xerrors v0.0.0-20191011141410-1b5146add898 // indirect
+	golang.org/x/mod v0.3.0 // indirect
+	golang.org/x/sync v0.0.0-20200317015054-43a5402ce75a
+	golang.org/x/tools v0.0.0-20200515220128-d3bf790afa53 // indirect
+	golang.org/x/tools/gopls v0.4.1 // indirect
+	honnef.co/go/tools v0.0.1-2020.1.4 // indirect
+	mvdan.cc/xurls/v2 v2.2.0 // indirect
```
vladaionescu, line 31: These dependency changes seem unrelated? Mind reverting these? (Unless I'm missing something)

drews256:
@vladaionescu here's another option addressing the comments you made and with the regex you shared.

earthfile2llb/listener.go
```diff
@@ -595,7 +595,12 @@ func (l *listener) EnterEnvArgKey(c *parser.EnvArgKeyContext) {
 	if l.shouldSkip() {
 		return
 	}
-	l.envArgKey = c.GetText()
+	var envKey = c.GetText()
+	var matchedEnvKey = checkShellName(envKey)
+	if matchedEnvKey != envKey {
+		l.err = fmt.Errorf("Invalid env key definition %s", envKey)
```
vladaionescu, line 601: Best to add a return so that nothing else is applied.

```suggestion
		l.err = fmt.Errorf("Invalid env key definition %s", envKey)
		return
```

earthfile2llb/listener.go
```diff
@@ -662,6 +667,12 @@ func (ssf *StringSliceFlag) Set(arg string) error {
 	return nil
 }
 
+var shellNameStringRegexp = regexp.MustCompile("^[a-zA-Z_]+[a-zA-Z0-9_]*")
+
+func checkShellName(str string) string {
```
vladaionescu, line 672: Mind renaming this to `checkEnvVarName`? "Shell name" could be misunderstood as the shell itself.

earthfile2llb/listener.go
```diff
@@ -662,6 +667,12 @@ func (ssf *StringSliceFlag) Set(arg string) error {
...
+	return shellNameStringRegexp.FindString(str)
```
vladaionescu, line 673: Can you make this func return a `bool`? Or even better - an `error`? You'll probably need to use `regexp.MatchString`.

earthfile2llb/listener.go
```diff
@@ -662,6 +667,12 @@ func (ssf *StringSliceFlag) Set(arg string) error {
 	return nil
 }
 
+var shellNameStringRegexp = regexp.MustCompile("^[a-zA-Z_]+[a-zA-Z0-9_]*")
```
vladaionescu, line 670: Best to also end with a `$` to ensure that the entire strings matches.

alexcb:

PR #102, opened 2020-07-10
Fix prefixing for control character progress bars

alexcb:
This fixes the handling of control characters which are used to display
progress bars which use the \r control character.

### DIFF
--- a/conslogging/conslogging.go
+++ b/conslogging/conslogging.go
@@ -94,7 +94,7 @@ func (cl ConsoleLogger) PrintBytes(data []byte) {
 	cl.mu.Lock()
 	defer cl.mu.Unlock()
 
-	output := []byte{}
+	output := make([]byte, 0, len(data))
 	for len(data) > 0 {
 		r, size := utf8.DecodeRune(data)
 		ch := data[:size]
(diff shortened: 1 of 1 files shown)


conslogging/conslogging.go
```diff
@@ -90,20 +91,37 @@ func (cl ConsoleLogger) Printf(format string, args ...interface{}) {
...
-		// No prefix when it's not a complete line.
-		cl.w.Write(data)
-	} else {
-		adjustedData := bytes.TrimSuffix(data, []byte("\n"))
-		for _, line := range bytes.Split(adjustedData, []byte("\n")) {
-			cl.printPrefix()
-			cl.w.Write(line)
-			cl.w.Write([]byte("\n"))
+
+	output := []byte{}
```
vladaionescu, line 97: Should we do something like this here?

```suggestion
	output := make([]byte, 0, len(data))
```

... or what would be a better guess of the capacity?

conslogging/conslogging.go
```diff
@@ -90,20 +91,37 @@ func (cl ConsoleLogger) Printf(format string, args ...interface{}) {
...
+				}
+				cl.printPrefix()
+				cl.trailingLine = true
+			}
+			output = append(output, ch...)
 		}
 	}
+	if len(output) > 0 {
+		cl.w.Write(output)
+		output = []byte{}
```
vladaionescu, line 123: I think this line makes the inefassign linter complain.

conslogging/conslogging.go
alexcb: good plan, i'll make that change.

PR #104, opened 2020-07-10
Update docs for earthly config

alexcb:
Additionally removed the reference to GIT_URL_INSTEAD_OF which is no
longer used.

### DIFF
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -11,6 +11,7 @@
 * [Earthfile reference](earthfile/earthfile.md)
     * [Builtin args](earthfile/builtin-args.md)
 * [earth command reference](earth-command/earth-command.md)
+* [Configuration reference](earth-config/earth-config.md)
 * [Examples](examples/examples.md)
     * [Go](examples/go.md)
     * [Java](examples/java.md)
--- a/docs/earth-config/earth-config.md
+++ b/docs/earth-config/earth-config.md
@@ -36,7 +36,7 @@ git:
 
 ### site
 
-The git repo hostname; e.g. `github.com`, or `gitlab.com`
+The git repository hostname. For example `github.com`, or `gitlab.com`
 
 ### auth
 
--- a/docs/earth-config/earth-config.md
+++ b/docs/earth-config/earth-config.md
@@ -6,7 +6,7 @@ By default, earth reads the configuration file `~/.earthly/config.yaml`; however
 overridden with the `--config` command flag option.
 
 Currently only https authentication configuration is configured in the config file; ssh authentication
-is additionally supported (and the default when the no overrides have been configured)
+is additionally [supported](../guides/auth.md) (and the default when the no overrides have been configured).
 
 ## Format
 
--- a/docs/earth-config/earth-config.md
+++ b/docs/earth-config/earth-config.md
@@ -51,4 +51,4 @@ The https username to use when auth is set to `https`. This setting is ignored w
 
 ### password
 
-The https password to use when auth is set to https; this is ignored when auth is ssh.
+The https password to use when auth is set to `https`. This setting is ignored when auth is `ssh`.
--- a/docs/earth-config/earth-config.md
+++ b/docs/earth-config/earth-config.md
@@ -47,7 +47,7 @@ See the [Authentication guide](../guides/auth.md) for a guide on setting up auth
 
 ### user
 
-The https username to use when auth is set to https; this is ignored when auth is ssh.
+The https username to use when auth is set to `https`. This setting is ignored when auth is `ssh`.
 
 ### password
 
--- a/docs/earth-config/earth-config.md
+++ b/docs/earth-config/earth-config.md
@@ -1,0 +1,55 @@
+# Earthly configuration file
+
+Global configuration values for earth can be stored on disk in the configuration file.
+
+By default, earth reads the configuration file `~/.earthly/config.yaml`; however, it can also be
+overridden with the `--config` command flag option.
+
+Currently only https authentication configuration is configured in the config file; ssh authentication
+is additionally supported (and the default when the no overrides have been configured)
+
+## Format
+
+The earthly config file is a [yaml](https://yaml.org/) formatted file that looks like:
+
+```yaml
+git:
+    <site>:
+        auth: https|ssh
+        user: <username>
+        password: <password>
+    <site2>:
+        ...
+```
+
+Example:
+
+```yaml
+git:
+    github.com:
+        auth: https
+        user: alice
+        password: itsasecret
+```
+
+## Git configuration reference
+
+### site
+
+The git repo hostname; e.g. `github.com`, or `gitlab.com`
+
+### auth
+
+Either `https` or `ssh` (default). If https is specified, user and password fields are used
+to authenticate over https when pulling from git for the corresponding site.
+
+See the [Authentication guide](guides/auth.md) for a guide on setting up authentication.
+
+### user
+
+The https username to use when auth is set to https; this is ignored when auth is ssh.
+
+### password
+
+The https password to use when auth is set to https; this is ignored when auth is ssh.
+
--- a/docs/earth-config/earth-config.md
+++ b/docs/earth-config/earth-config.md
@@ -43,7 +43,7 @@ The git repo hostname; e.g. `github.com`, or `gitlab.com`
 Either `https` or `ssh` (default). If https is specified, user and password fields are used
 to authenticate over https when pulling from git for the corresponding site.
 
-See the [Authentication guide](guides/auth.md) for a guide on setting up authentication.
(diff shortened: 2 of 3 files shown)


docs/SUMMARY.md
```diff
@@ -11,6 +11,7 @@
 * [Earthfile reference](earthfile/earthfile.md)
     * [Builtin args](earthfile/builtin-args.md)
 * [earth command reference](earth-command/earth-command.md)
+* [earth config reference](earth-config/earth-config.md)
```
vladaionescu, line 14: I'd name this `Configuration reference` to make it more reader friendly.

docs/earth-config/earth-config.md
```diff
@@ -0,0 +1,34 @@
+# Earthly config file
+
+Global configuration values for earth can be stored on disk in the configuration file.
+
+By default, earth reads the configuration file `~/.earthly/config.yaml`; however, it can also be
+overridden with the `--config` command flag option.
+
+Currently only https authentication configuration is supported by the config file.
```
vladaionescu, line 8: Perhaps this should be clarified that only https auth is *configured* here, however both SSH and user-pass auth are supported. We should also point to the auth guide for more info.

docs/earth-config/earth-config.md
```diff
@@ -0,0 +1,34 @@
...
+
+```yaml
+git:
+    <site>:
+        auth: https
+        user: <username>
+        password: <password>
+    <site2>:
+        ...
+```
```
vladaionescu, line 22: Might be worth also adding a `github.com` example just below this format.

docs/earth-config/earth-config.md
```diff
@@ -0,0 +1,34 @@
...
+
+To configure https authentication, the auth mode must be set to https
+in addition to the user and password values.
+
+### SSH Authenticating
+
+By default earth uses ssh authentication. It can be explicitly set by setting `auth: ssh`,
+or implicitly set by omitting the site-specific configuration.
+
+See the [Authenticat guide](guides/auth.md) for more details.
```
vladaionescu, line 34: It feels like this is more like a guide rather than reference. I would keep the auth guide page to serve as the guide for these settings.

For this particular page, I would just keep it purely a reference page, with just format and reference. See the [docker-compose reference](https://docs.docker.com/compose/compose-file/) as an example.

Perhaps this could be the structure, roughly:

```markdown
## Format
...
## Global configuration reference (future)
...
### some-config-item
...
### another-config-item
...
## Git configuration reference
...
### auth
...
### user
...
### password
...
```

cmd/earth/main.go
```diff
@@ -192,13 +192,6 @@ func newEarthApp(ctx context.Context, console conslogging.ConsoleLogger) *earthA
 			Usage:       "The git password to use for git HTTPS authentication",
 			Destination: &app.gitPasswordOverride,
 		},
-		&cli.StringFlag{
-			Name:        "git-url-instead-of",
-			Value:       "https://gitlab.com/=git@gitlab.com:,https://github.com/=git@github.com:",
-			EnvVars:     []string{"GIT_URL_INSTEAD_OF"},
-			Usage:       "Rewrite git URLs of a certain pattern. Similar to git-config url.<base>.insteadOf (https://git-scm.com/docs/git-config#Documentation/git-config.txt-urlltbasegtinsteadOf). Multiple values can be separated by commas. Format: <base>=<instead-of>[,...]. For example: ''git@github.com:=https://github.com/'",
-			Destination: &app.buildkitdSettings.GitURLInsteadOf,
-		},
```
vladaionescu, line 201: Hmm.. this needs to be deprecated over time - we shouldn't remove it directly.

We should also add such a setting to the configuration format.

docs/guides/auth.md
```diff
@@ -38,16 +38,28 @@ Another key setting available, is `GIT_URL_INSTEAD_OF`. It allows for `https://g
...
 * `GIT_USERNAME`
 * `GIT_PASSWORD`
 
-And additionally overriding the setting `GIT_URL_INSTEAD_OF` to force `git@github.com` URLs to be interpreted as `https://github.com` URLs:
-
-```bash
-export GIT_URL_INSTEAD_OF='https://github.com/=git@github.com:'
-```
+However, environment variable authentication may be depreciated in the future as it doesn't allow specific
+site configuration and only works for github and gitlab.
```
vladaionescu, line 62: Perhaps we should say that these are deprecated *now*... and not merge this PR into master until the next release?

docs/guides/auth.md
```diff
@@ -38,16 +38,28 @@ Another key setting available, is `GIT_URL_INSTEAD_OF`. It allows for `https://g
...
 
 * `GIT_USERNAME`
 * `GIT_PASSWORD`
 
-And additionally overriding the setting `GIT_URL_INSTEAD_OF` to force `git@github.com` URLs to be interpreted as `https://github.com` URLs:
-
-```bash
-export GIT_URL_INSTEAD_OF='https://github.com/=git@github.com:'
-```
+However, environment variable authentication may be depreciated in the future as it doesn't allow specific
```
vladaionescu, line 61:
```suggestion
However, environment variable authentication may be deprecated in the future as it doesn't allow specific
```

docs/earth-config/earth-config.md
```diff
@@ -0,0 +1,34 @@
...
+Currently only https authentication configuration is supported by the config file.
+
+## Format
+
+The earthly config file is a [yaml](https://yaml.org/) formatted file that looks like:
+
+```yaml
+git:
+    <site>:
+        auth: https
```
vladaionescu, line 17:
```suggestion
        auth: https|ssh
```

cmd/earth/main.go
alexcb: I added this back in, and realised that I had accidentally broken it, which has been fixed in https://github.com/earthly/earthly/pull/106

docs/earth-config/earth-config.md
alexcb: excellent point

docs/earth-config/earth-config.md
alexcb: thanks for the example, I've fixed this up.

docs/earth-config/earth-config.md
```diff
@@ -0,0 +1,55 @@
...
+### site
+
+The git repo hostname; e.g. `github.com`, or `gitlab.com`
+
+### auth
+
+Either `https` or `ssh` (default). If https is specified, user and password fields are used
+to authenticate over https when pulling from git for the corresponding site.
+
+See the [Authentication guide](guides/auth.md) for a guide on setting up authentication.
```
vladaionescu, line 46:
```suggestion
See the [Authentication guide](../guides/auth.md) for a guide on setting up authentication.
```

docs/earth-config/earth-config.md
```diff
@@ -0,0 +1,55 @@
+# Earthly configuration file
+
+Global configuration values for earth can be stored on disk in the configuration file.
+
+By default, earth reads the configuration file `~/.earthly/config.yaml`; however, it can also be
+overridden with the `--config` command flag option.
+
+Currently only https authentication configuration is configured in the config file; ssh authentication
+is additionally supported (and the default when the no overrides have been configured)
```
vladaionescu, line 9:
```suggestion
is additionally [supported](../guides/auth.md) (and the default when the no overrides have been configured).
```

docs/earth-config/earth-config.md
```diff
@@ -0,0 +1,55 @@
...
+### auth
+
+Either `https` or `ssh` (default). If https is specified, user and password fields are used
+to authenticate over https when pulling from git for the corresponding site.
+
+See the [Authentication guide](guides/auth.md) for a guide on setting up authentication.
+
+### user
+
+The https username to use when auth is set to https; this is ignored when auth is ssh.
```
vladaionescu, line 50:
```suggestion
The https username to use when auth is set to `https`. This setting is ignored when auth is `ssh`.
```

docs/earth-config/earth-config.md
```diff
@@ -0,0 +1,55 @@
...
+
+### password
+
+The https password to use when auth is set to https; this is ignored when auth is ssh.
```
vladaionescu, line 54: The https password to use when auth is set to `https`. This setting is ignored when auth is `ssh`.

docs/earth-config/earth-config.md
```diff
@@ -0,0 +1,55 @@
...
+        auth: https
+        user: alice
+        password: itsasecret
+```
+
+## Git configuration reference
+
+### site
+
+The git repo hostname; e.g. `github.com`, or `gitlab.com`
```
vladaionescu, line 39:
```suggestion
The git repository hostname. For example `github.com`, or `gitlab.com`
```

PR #105, opened 2020-07-11
Move all build arg and env var related logic to variables package

vladaionescu:
Also, add an interface for command interpreter.


vladaionescu:
Investigating some potential cache issues...

PR #108, opened 2020-07-14
GIT_CONFIG env conflicts with using git

alexcb:
earth sets a variable named GIT_CONFIG which conflicts with what "git config" is looking for, renaming it to EARTH_GIT_CONFIG avoids the conflict.

### DIFF
--- a/buildkitd/buildkitd.go
+++ b/buildkitd/buildkitd.go
@@ -201,11 +201,11 @@ func Start(ctx context.Context, image string, settings Settings, reset bool) err
 	}
 
 	args = append(args,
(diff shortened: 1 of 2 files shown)


buildkitd/buildkitd.go
```diff
@@ -201,11 +201,11 @@ func Start(ctx context.Context, image string, settings Settings, reset bool) err
 	}
 
 	args = append(args,
-		"-e", "GIT_CONFIG",
+		"-e", "EARTH_GIT_CONFIG",
```
vladaionescu, line 204: Let's use `EARTHLY_` prefix, like all other such vars.

PR #110, opened 2020-07-14
CI should test using the latest binary and image

vladaionescu:


vladaionescu:
Fixed the GIT_URL_INSTEAD_OF issue via https://github.com/vladaionescu/buildkit/commit/829d92f1c64c86dce68da52a3fc239113a3af3c2

PR #113, opened 2020-07-15
Adding Scala example

agbell:
Adding scala example


examples/scala/build.earth
```diff
@@ -0,0 +1,20 @@
...
+    COPY project project
+    RUN sbt update
+    SAVE IMAGE
+
+build:
+    FROM +deps
+    COPY src src
+    RUN sbt assembly
+    SAVE ARTIFACT target/scala-2.13 /scala-example AS LOCAL build/bin
+    SAVE IMAGE
```
vladaionescu, line 15: This `SAVE IMAGE` is not necessary because this target is not used in any `FROM` command.

examples/scala/build.earth
```diff
@@ -0,0 +1,20 @@
...
+
+docker:
+    COPY build/bin build/bin
```
vladaionescu, line 18: Here you probably want to copy from the `+build` target rather than from the local dir, so that you can build `+docker` in one go.

```suggestion
    COPY +build/scala-example build/bin
```

examples/scala/build.earth
agbell: Fixed

examples/scala/build.earth
agbell: Ah, makes sense.  This line was actually the trickiest part for me.  Fixed.

PR #118, opened 2020-07-17
docs: target reference

alexcb:
when a user runs `earth --help` it shows a links to https://docs.earthly.dev/earth-command however on that page it doesn't define how the target-reference can be formatted.

### DIFF
--- a/docs/earth-command/earth-command.md
+++ b/docs/earth-command/earth-command.md
@@ -74,7 +74,7 @@ specified by `<local-path>`, which must start with `./`, `../`, or `/`.
 
 ##### Artifact Reference
 
-`<target-ref>/<artifact-path>` will reference a build target's artifact.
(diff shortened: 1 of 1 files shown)


docs/earth-command/earth-command.md
```diff
@@ -57,6 +57,31 @@ If the build phase does not succeed, not output is produced and no push instruct
...
+`<local-path>+<target-name>` will reference a local earthfile in a different directory as
+specified by `<local-path>`, which must start with `./`, `../`, or `/`.
+
+##### Remote Reference
+
+`<gitvendor>/<namespace>/<project>/path/in/project[:some-tag]+<target-name>` will access a remote git repository.
+
+##### Artifact Reference
+
+`<target-ref>/<artifact-path>` will reference a build target's artifact.
```
vladaionescu, line 77: Just to make it clear that this is a different kind of ref.

```suggestion
The `<artifact-ref>` can reference artifacts built by targets. `<target-ref>/<artifact-path>` will reference a build target's artifact.
```

Issue #116, opened 2020-07-17
Use a path outside of /tmp for the cache if available


dmarti:
On some systems, `/tmp` is a small tmpfs partition. This creates 2 problems for putting Earthly's cache there:

 * not enough space available for large builds 

 * contents are lost on reboot

Consider checking for the `/var/cache` directory and putting Eartlhy's cache there if it is available.

related issue: [Explore remote cache · Issue #11 · earthly/earthly](https://github.com/earthly/earthly/issues/11)

reference: [Features/tmp-on-tmpfs - Fedora Project Wiki](https://fedoraproject.org/wiki/Features/tmp-on-tmpfs)

vladaionescu:
Good points.

A temporary workaround is to create a symbolic link from `/tmp/earthly` to some other location on the system.

dmarti:
Yes, that's how I have it working, thank you.

alexcb:
@dmarti the next release of earthly will allow this to be changed by creating a file under `~/.earthly/config.yaml` containing:
```
global:
  cache_path: "/path/to/earthly/cache"
```

We also changed it to default to `/var/cache/earthly` as you suggested. Thanks for reporting the issue.

PR #121, opened 2020-07-21
Allow user to change earthly cache path

alexcb:
Rather than assume users want to cache data under tmp,
allow the user to override it.

The default value will be changed to /var/cache/earthly
so the cache persists between reboots.

### DIFF
--- a/buildkitd/buildkitd.go
+++ b/buildkitd/buildkitd.go
@@ -17,8 +17,6 @@ import (
 const (
 	// ContainerName is the name of the buildkitd container.
 	ContainerName = "earthly-buildkitd"
-	// TempDir is the directory used for buildkitd cache.
-	TempDir = "/tmp/earthly"
 )
 
 // Address is the address at which the daemon is available.
@@ -176,7 +174,7 @@ func Start(ctx context.Context, image string, settings Settings, reset bool) err
 		return errors.Wrap(err, "settings hash")
 	}
 	env := os.Environ()
-	cacheMount := fmt.Sprintf("%s:%s:delegated", TempDir, TempDir)
+	cacheMount := fmt.Sprintf("%s:/tmp/earthly:delegated", settings.TempDir)
 	args := []string{
 		"run",
 		"-d", "--rm",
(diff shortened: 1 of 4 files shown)


alexcb:
This PR addresses the issue reported in #116

buildkitd/buildkitd.go
```diff
@@ -176,11 +174,12 @@ func Start(ctx context.Context, image string, settings Settings, reset bool) err
 		return errors.Wrap(err, "settings hash")
 	}
 	env := os.Environ()
-	cacheMount := fmt.Sprintf("%s:%s:delegated", TempDir, TempDir)
+	cacheMount := fmt.Sprintf("%s:%s:delegated", settings.TempDir, settings.TempDir)
```
vladaionescu, line 177: We could always mount to the same place within the container (eg keep it as `/tmp/earthly` within buildkitd). I would only configure the source. That way we don't have to change the buildkit settings. What do you think?

config/config.go
```diff
@@ -11,7 +11,7 @@ import (
 var ErrInvalidTransport = fmt.Errorf("invalid transport")
 
 type GlobalConfig struct {
-	//TODO add support for global config as needed
+	CachePath string `yaml:"cache_path"`
```
vladaionescu, line 14: K8s uses camelCase, docker-compose uses snake_case. I guess docker-compose is more common to average dev. 👍

config/config.go
```diff
@@ -36,7 +36,11 @@ func ensureTransport(s, transport string) (string, error) {
 }
 
 func ParseConfigFile(yamlData []byte) (*Config, error) {
-	var config Config
+	config := Config{
+		Global: GlobalConfig{
+			CachePath: "/var/cache/earthly",
```
vladaionescu, line 41: Nice 👍

buildkitd/buildkitd.go
alexcb: I was thinking about doing it that way too. lets make that change, then it's less variables to pass into the container.

config/config.go
alexcb: excellent! I'm more of a snake_case fan :smile:

buildkitd/buildkitd.go
alexcb: fixed.

PR #126, opened 2020-07-21
Config file settings v2

alexcb:


cmd/earth/main.go
```diff
@@ -298,8 +299,28 @@ func (app *earthApp) parseConfigFile(context *cli.Context) error {
...
+	app.buildkitdSettings.GitCredentials = gitCredentials
+	return nil
+
+}
+
+func (app *earthApp) processDeprecatedCommandOptions(context *cli.Context, cfg *config.Config) error {
+
 	// command line overrides the config file
 	if app.gitUsernameOverride != "" || app.gitPasswordOverride != "" {
+		app.console.Warnf("Warning: the --git-username and --git-password command flags are deprecated and are now configured in the earthly config file under the `git` section\n")
```
vladaionescu, line 328: Might be worth mentioning where that config file exists (could be created) and also provide a link to docs about the config file.

cmd/earth/main.go
```diff
@@ -298,8 +299,33 @@ func (app *earthApp) parseConfigFile(context *cli.Context) error {
...
```
vladaionescu, line 328: Nit: I would also remove the backticks around `git` to make it a bit cleaner

cmd/earth/main.go
```diff
@@ -298,8 +299,33 @@ func (app *earthApp) parseConfigFile(context *cli.Context) error {
...
+		return err
+	}
+
+	gitConfig, gitCredentials, err := config.CreateGitConfig(cfg)
+	if err != nil {
+		return errors.Wrapf(err, "failed to create git config from %s", app.configPath)
+	}
+
+	// command line option overrides the config which overrides the default value
+	if !context.IsSet("buildkit-image") && cfg.Global.BuildkitImage != "" {
```
vladaionescu, line 313: I would check if the underlying variable is `== ""` rather than `context.IsSet`, to avoid typo of the flag name.

cmd/earth/main.go
```diff
@@ -319,14 +345,32 @@ func (app *earthApp) parseConfigFile(context *cli.Context) error {
...
+	} else {
+		if gitGlobal, ok := cfg.Git["global"]; ok {
+			if gitGlobal.GitURLInsteadOf != "" {
+				app.buildkitdSettings.GitURLInsteadOf = gitGlobal.GitURLInsteadOf
+			}
+		}
+	}
+
+	if context.IsSet("no-loop-device") {
+		app.console.Warnf("Warning: the --no-loop-device command flag is deprecated and is now configured in the earthly config file under the `no_loop_device` setting\n")
```
vladaionescu, line 359: Should warnings only have the prefix colored (eg `Warning`) and leave rest of the text as default color? I'm thinking making the whole line red might be too much?

config/config.go
```diff
@@ -11,10 +11,17 @@ import (
 var ErrInvalidTransport = fmt.Errorf("invalid transport")
 
 type GlobalConfig struct {
-	CachePath string `yaml:"cache_path"`
+	CachePath         string `yaml:"cache_path"`
+	DisableLoopDevice bool   `yaml:"no_loop_device"`
+	BuildkitCacheSize string `yaml:"buildkit_cache_size"`
```
vladaionescu, line 16: I would call this just `cache_size_mb`:
   * the fact that it's buildkit underneath shouldn't matter to the user
   * `_mb` would make the unit of measure clear, so they don't need to go and read the docs. Though I also like being able to say `10G`... I'm torn :-)

What do you think?

cmd/earth/main.go
alexcb: I don't think that'll work because the underlying variable is set to the default value.

cmd/earth/main.go
alexcb: I was unsure about that too, if it's just the "warning", then the eye is drawn to that more so than equally drawn to the entire line. 

Here's what it currently looks like:

image

config/config.go
alexcb: that's a fair point, because most users don't expect "10GB" to work. I'll go back to just using ints (then it's also one less dependency we have).

cmd/earth/main.go
alexcb: Then compare with only the prefix:

image

maybe it's fine just as-is with the prefix only highlighted.

cmd/earth/main.go
vladaionescu: Ah ok :+1:

cmd/earth/main.go
vladaionescu: Ok sounds good

config/config.go
```diff
@@ -11,10 +11,17 @@ import (
 var ErrInvalidTransport = fmt.Errorf("invalid transport")
 
 type GlobalConfig struct {
-	CachePath string `yaml:"cache_path"`
+	CachePath           string `yaml:"cache_path"`
+	DisableLoopDevice   bool   `yaml:"no_loop_device"`
+	BuildkitCacheSizeMB int    `yaml:"buildkit_cache_size_mb"`
```
vladaionescu, line 16: Did you mean to remove `buildkit_` prefix here?

alexcb: I'm not sure to be honest. Should we keep it as `buildkit_cache_size_mb` or rename it to `cache_size_mb` (which makes it more in line with the `cache_path` config).

cmd/earth/main.go
alexcb: good point.

cmd/earth/main.go
alexcb: good point. fixed.

config/config.go
alexcb: in the end I removed the prefix so it (`cache_size_mb`) aligns with `cache_path`.

PR #128, opened 2020-07-21
Docs config file v3

alexcb:


docs/earth-command/earth-command.md
```diff
@@ -170,8 +170,6 @@ For example:
 * `--git-url-instead-of='git@github.com:=https://github.com/'` forces use of SSH-based URLs for GitHub (default)
 * `--git-url-instead-of='https://github.com/=git@github.com:'` forces use of HTTPS-based URLs for GitHub
 
-Currently, only one such pattern is allowed to be specified.
-
```
vladaionescu, line 174: Perhaps we should remove these options from here. Maybe we just mention that the setting has moved to the config file and not give out any other details here.

docs/earth-config/earth-config.md
```diff
@@ -12,7 +12,12 @@ The earthly config file is a [yaml](https://yaml.org/) formatted file that looks
 ```yaml
 global:
   cache_path: <cache_path>
+  cache_size_mb: <cache_size_mb>
+  no_loop_device: true|false
+  buildkit_image: <buildkit_image>
 git:
+    global:
+	    url_instead_of: <url_instead_of>
```
vladaionescu, line 20: Unintended tab?

docs/earth-config/earth-config.md
```diff
@@ -24,7 +29,11 @@ git:
 Example:
 
 ```yaml
+global:
+    cache_size_mb: 20000
 git:
+    global:
+        url_instead_of:
```
vladaionescu, line 36: Did you mean to add an example value here?

docs/earth-config/earth-config.md
```diff
@@ -37,23 +46,65 @@ git:
...
 
+### cache_size_mb
+
+Specifies the total size of the buildkit cache, in MB. The buildkit daemon will allocate disk space for this size. Size less than `1000` (1GB) is not recommended. The default size if this option is not set is `10000` (10GB).
+
+This setting is only used when the buildkit daemon is started (or restarted). In order to apply the setting immediately, issue the command
+
+```bash
+earth --buildkit-cache-size-mb <cache-size-mb> prune --reset
+```
```
vladaionescu, line 57:
```suggestion
This setting is only used when the cache is initialized for the first time. In order to apply the setting immediately, issue the following command after changing the configuration

```bash
earth prune --reset
```
```

File: docs/earth-config/earth-config.md [hunk 5]
```diff
@@ -37,23 +46,65 @@ git:
...
+
+### no_loop_device
+
+When set to true, disables the use of a loop device for storing the cache. By default, Earthly uses a file mounted as a loop device, so that it can control the type of filesystem used for the cache, in order to ensure that overlayfs can be mounted on top of it. If you are already using a filesystem compatible with overlayfs, then you can disable the loop device.
+
+### buildkit_image
+
+Instructs earth to use an alternate image for buildkitd. The default image used is `earthly/buildkitd:<earth-version>`.
```
vladaionescu, line 65: We were meaning to keep this as a hidden setting?

docs/earth-config/earth-config.md
```diff
@@ -37,23 +46,65 @@ git:
...
+
+
 ## Git configuration reference
 
-### site
+git configuration is split up into global confi options, or site-specific options.
```
vladaionescu, line 70:
```suggestion
The git configuration is split up into global config options, or site-specific options.
```

docs/earth-config/earth-config.md
```diff
@@ -37,23 +46,65 @@ git:
...
+### global options
+
+The global git options 
+
+#### url_instead_of
+
+Rewrites git URLs of a certain pattern. Similar to [`git-config url.<base>.insteadOf`](https://git-scm.com/docs/git-config#Documentation/git-config.txt-urlltbasegtinsteadOf).
+Multiple values can be separated by commas. Format: `<base>=<instead-of>[,...]`.
+
+This setting allows rewriting all git URLs of the form `https://...` into `git@github.com:...`, or vice-versa.
```
vladaionescu, line 81: We should perhaps give a non-github example here, because for github we recommend using the auth type setting. Maybe this point is also worth mentioning.

docs/earth-command/earth-command.md
alexcb: I agree, that'll simplify things.

docs/earth-config/earth-config.md
alexcb: thanks :)

docs/earth-config/earth-config.md
alexcb: I missed that, thanks. how about something non-github related like:
`url_instead_of: "git@example.com:=https://localmirror.example.com/"`

docs/earth-config/earth-config.md
alexcb: fixed.

docs/earth-config/earth-config.md
alexcb: how about something a bit more exotic like this?
```
* `--git-url-instead-of='git@example.com:=https://example.com/'` forces use of SSH-based URLs rather than HTTPS
* `--git-url-instead-of='https://localmirror.example.com/=git@example.com:'` forces use of HTTPS-based local mirror for ssh-based example.com repositories
```

alexcb: I changed the github references to example.com

docs/earth-config/earth-config.md
alexcb: fixed. I also changed the `Hidden` flag in the code to be true for this.

docs/earth-config/earth-config.md
vladaionescu: Looks good to me

docs/earth-config/earth-config.md
vladaionescu: Perfect

docs/earth-command/earth-command.md
```diff
@@ -32,9 +30,8 @@
         [--push] [--no-cache] [--allow-privileged|-P]
         [--ssh-auth-sock <path-to-sock>]
         [--git-username <git-user>] [--git-password <git-pass>]
-        [--git-url-instead-of <git-instead-of>]
         [--buildkit-host <bk-host>] [--buildkit-cache-size-mb <cache-size-mb>]
-        [--buildkit-image <bk-image>] [--no-loop-device]
+        [--no-loop-device]
```
vladaionescu, line 34: Probably worth removing all the other options we plan not to support here, so that the command looks cleaner. Just mentioning them below (that they have moved) should be enough.

docs/earth-command/earth-command.md
vladaionescu: I think it's still worth mentioning that the options have moved. Example:

```markdown
##### `--git-url-instead-of <git-instead-of>`	

Also available as an env var setting: `GIT_URL_INSTEAD_OF=<git-instead-of>`.

This option is now deprecated. Please use <insert-page-name-and-link-here> instead.
```

docs/earth-config/earth-config.md
```diff
@@ -37,23 +46,58 @@ git:
...
 Specifies the location where build data is stored. The default location is `/var/cache/earthly`.
 
+### cache_size_mb
+
+Specifies the total size of the buildkit cache, in MB. The buildkit daemon will allocate disk space for this size. Size less than `1000` (1GB) is not recommended. The default size if this option is not set is `10000` (10GB).
+
+This setting is only used when the cache is initialized for the first time. In order to apply the setting immediately, issue the following command after changing the configuration
+
+```bash
+earth prune --reset
```
vladaionescu, line 56: The closing back ticks are missing here?

alexcb: fixed, this was rather sloppy of me :(

docs/earth-command/earth-command.md
```diff
@@ -144,6 +138,8 @@ Sets the git username to use for non-SSH git authentication. For more informatio
 
 To prevent the need to specify this option on every `earth` invocation, it is recommended to specify the env var form in a file like `.bashrc` or `.profile`.
 
+This option is now deprecated. Please use the [configuration file](../earth-config/earth-config.md) instead.
+
```
vladaionescu, line 142: Did you mean to remove this part too:

```
Sets the git username to use for non-SSH git authentication. For more information see the [Authentication page](../guides/auth.md).

To prevent the need to specify this option on every `earth` invocation, it is recommended to specify the env var form in a file like `.bashrc` or `.profile`.
```

Similar for others too.

alexcb: do you think it's better to just cut it down to this?

```
##### `--git-username <git-user>`

Also available as an env var setting: `GIT_USERNAME=<git-user>`.

This option is now deprecated. Please use the [configuration file](../earth-config/earth-config.md) instead.

```

vladaionescu: Yes exactly - that's what I meant. We should do this for all the deprecated options.

PR #131, opened 2020-07-22
Docs for Mac cache location update

vladaionescu:


docs/earth-config/earth-config.md
```diff
@@ -44,7 +44,7 @@ git:
 
 ### cache_path
 
-Specifies the location where build data is stored. The default location is `/var/cache/earthly`.
+Specifies the location where build data is stored. The default location is `/var/cache/earthly` on Linux and `~/Library/Caches/earthly` on OS X.
```
rjriel, line 47: I think it's `macOS` now instead of `OS X`

vladaionescu: Oh yeah - good point.

PR #137, opened 2020-07-23
Remove socket file rm hack

alexcb:
An issue in buildkit caused a "tar/archive: sockets not supported" error
while building the dot net example; this issue has been fixed upstream
in buildkit ( see https://github.com/moby/buildkit/pull/1581 ), and we
can now remove this work-around.

### DIFF
--- a/examples/dotnet/Earthfile
+++ b/examples/dotnet/Earthfile
@@ -13,11 +13,6 @@ build:
 
     # make sure you have /bin and /obj in .earthignore, as their content from context might cause problems
     RUN dotnet publish --no-restore src/HelloEarthly -o publish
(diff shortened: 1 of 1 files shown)


vladaionescu:
> This requires a release, right?

Or not? How did the CI pass?

alexcb:
we'll want to merge it at the same time as when we do the release; otherwise users might use v0.2.0 against examples in master which would fail.

CI passed since it uses the latest built earth version from master to run against the examples.

vladaionescu:
> CI passed since it uses the latest built earth version from master to run against the examples.

Oh yeah - we changed that 👍

PR #143, opened 2020-07-24
limit amount of status percentage that is output

alexcb:
only output percentage every 5 seconds or when it finishes, this avoids
the issue of lots of stdout being written.

Additionally display failed commands in red, and avoid showing too much
of context canceled errors which are a side effect of the main failure
(and can confuse users).

### DIFF
--- a/builder/solver.go
+++ b/builder/solver.go
@@ -7,8 +7,6 @@ import (
 	"io"
 	"os"
 	"os/exec"
-	"regexp"
-	"strings"
 
 	"github.com/earthly/earthly/conslogging"
 	"github.com/earthly/earthly/earthfile2llb/image"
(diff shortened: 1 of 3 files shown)


vladaionescu:
Looks good - I've been playing with adding this to our integration tests:

Screenshot from 2020-07-24 18-50-33

The error definitely stands out more than before, though it's quite a few scrolls from the end.

I'm wondering also if we should repeat the output of the failed target at the end to make it more clear to the user. Either that, or have another separate command (eg `earth last-failed` or `earth lf` for short) whereby earth prints back the output of the last failure. What do you think?

Also, at the end, there's still a cryptic message like this `Error: solve side effects: build error group: solve: failed to solve: rpc error: code = Unknown desc = executor failed running [/bin/sh -c  /bin/sh -c 'sleep 10; false']: buildkit-runc did not terminate successfully`. Maybe we want something a bit more informative, where it says which target failed and which command (without the `/bin/sh -c` wrappers)?

Anyway... these are thoughts for a separate PR.

PR #146, opened 2020-07-28
Improve vertex names... and colors

vladaionescu:
#### Improvements

* Use different colors if the build args are different
* Use different colors for different build contexts
* Improved, slightly cleaner printout of Earthfile commands
* An unrelated change: Add some handy build targets (eg `earth +for-linux`) for faster development

#### Example output

Screenshot from 2020-07-27 18-38-35

* `+earth` for mac and linux have different colors
* The different contexts have different colors so as to be able to tell them apart
* Other improvements (eg eliminating the `[ ]` around the `RUN` command arguments)


alexcb:
nice work, and great idea on the `earth +for-linux` speedup

PR #153, opened 2020-07-29
let the OS provide an open port for the debugger

alexcb:
Rather than use a hardcoded port, let the OS provide a free port. The
address to connect to is now passed as a secret to prevent the port from
affecting the cache.

### DIFF
--- a/cmd/debugger/main.go
+++ b/cmd/debugger/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net"
 	"os"
 	"os/exec"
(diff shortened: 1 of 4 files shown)


cmd/debugger/main.go
```diff
@@ -54,21 +55,36 @@ func interactiveMode(remoteConsoleAddr string) {
...
 }
 
-func main() {
-	remoteConsoleAddr := os.Getenv("EARTHLY_REMOTE_CONSOLE_ADDR")
-	if remoteConsoleAddr == "" {
-		remoteConsoleAddr = "127.0.0.1:8543"
+func getRemoteDebuggerAddr() string {
+	remoteConsoleAddr, err := ioutil.ReadFile("/run/secrets/earthly_remote_console_addr")
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Warning: failed to read earthly_remote_console_addr: %v", err)
```
vladaionescu, line 61: There's a lot of fprintf for errors scattered around. Perhaps it's better that all errors are bubbled up to main and then there's only one place in main that prints errors.

alexcb: I think that's a good point if we were immediately exiting after the errors, but in this instance we're treating it as a warning and not returning an error.

And the other cases are for a message before we drop into the interactive shell.

Issue #152, opened 2020-07-29
automatically remove exited earthly-buildkit image


alexcb:
When earth starts, it checks that the earthly-buildkit container is Running; if not it attempts to start it. 

If an existing container exists but is in an exited state the Start() command fails. This should be fixed.

Additionally we will remove the `--rm` flag so when it does fail, we can look at the logs.

alexcb:
![image](https://user-images.githubusercontent.com/1806823/88754267-18454c80-d113-11ea-94a5-726137cdf65a.png)

PR #154, opened 2020-07-30
Remove earthly-buildkitd container if exited

alexcb:
If the earthly-buildkitd container is in a stopped or exited state,
remove it before attempting to start a new instance of it.

This fix is to allow for the removal of the --rm flag from the
docker run command, so that logs persist in the cause of a failure.

This fixes https://github.com/earthly/earthly/issues/152

### DIFF
--- a/buildkitd/buildkitd.go
+++ b/buildkitd/buildkitd.go
@@ -167,17 +167,41 @@ func MaybeRestart(ctx context.Context, console conslogging.ConsoleLogger, image 
 	return nil
 }
 
+// RemoveExited removes any stopped or exited buildkitd containers
+func RemoveExited(ctx context.Context) error {
+	started, err := IsStarted(ctx)
+	if err != nil {
+		return err
+	}
+	if started {
+		return nil
+	}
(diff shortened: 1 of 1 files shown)


buildkitd/buildkitd.go
```diff
@@ -167,17 +167,41 @@ func MaybeRestart(ctx context.Context, console conslogging.ConsoleLogger, image
 	return nil
 }
 
+// RemoveExited removes any stopped or exited buildkitd containers
+func RemoveExited(ctx context.Context) error {
+	started, err := IsStarted(ctx)
```
vladaionescu, line 172: I don't think we need to check here too because we already check in the caller of `Start`.

alexcb: I agree with that; I added in this check in case for some reasons `RemoveExited` was called in a different code path. If the container is up and running RemoveExited will return an error saying the container is running and needs to be stopped (or we use the `--force` flag).

This is probably fine since it'll return an error if it's used in a different code path, and we could change the behaviour then.

PR #155, opened 2020-07-30
Pin debugger image to a specific version

alexcb:
Our build has a circular dependency where the debugger image is required
before our CI can pass; however CI is responsible for creating the image
after the tests pass. We will pin the version to get around this problem
and come up with a better solution down the road.

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -94,7 +94,7 @@ earth:
 	ARG EARTHLY_TARGET_TAG
 	ARG VERSION=$EARTHLY_TARGET_TAG
 	ARG DEFAULT_BUILDKITD_IMAGE=earthly/buildkitd:$VERSION
-	ARG DEFAULT_DEBUGGER_IMAGE=earthly/debugger:$VERSION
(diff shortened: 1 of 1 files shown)


Earthfile
```diff
@@ -94,7 +94,7 @@ earth:
 	ARG EARTHLY_TARGET_TAG
 	ARG VERSION=$EARTHLY_TARGET_TAG
 	ARG DEFAULT_BUILDKITD_IMAGE=earthly/buildkitd:$VERSION
-	ARG DEFAULT_DEBUGGER_IMAGE=earthly/debugger:$VERSION
+	ARG DEFAULT_DEBUGGER_IMAGE=earthly/debugger-docker-image:asdf@sha256:b3d0c0db9b08c3e63b105f6efb9bd028632955d337128188406c0256d95f39b2
```
vladaionescu, line 97: Hmm.. it doesn't look quite right. The image is `earthly/debugger`, right?

alexcb: that's embarrassing, I don't know how that happened. it should have been `debugger:docker-image@...`. I fixed this now.

PR #158, opened 2020-07-31
Create a pseudo tty to provide an upgraded shell

alexcb:
This interactive reverse debugger shell uses pty which allows bash to
use tabs, arrows, etc.

### DIFF
--- a/cmd/debugger/main.go
+++ b/cmd/debugger/main.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/creack/pty"
 	"github.com/hashicorp/yamux"
+	"github.com/pkg/errors"
 )
 
 var (
@@ -36,7 +37,7 @@ func getShellPath() (string, bool) {
 	return "", false
 }
 
-func interactiveMode(remoteConsoleAddr string) error {
+func interactiveMode(ctx context.Context, remoteConsoleAddr string) error {
 	conn, err := net.Dial("tcp", remoteConsoleAddr)
 	if err != nil {
 		return err
@@ -69,23 +70,28 @@ func interactiveMode(remoteConsoleAddr string) error {
 	if !ok {
 		return ErrNoShellFound
 	}
-	c := exec.Command(shellPath)
+	c := exec.CommandContext(ctx, shellPath)
 
-	ptmx, e := pty.Start(c)
-	if e != nil {
-		fmt.Fprintf(os.Stderr, "failed to start pty: %v\n", e)
-		return e
+	ptmx, err := pty.Start(c)
+	if err != nil {
+		return errors.Wrap(err, "failed to start pty")
 	}
 	defer func() { _ = ptmx.Close() }() // Best effort.
 
-	ctx, cancel := context.WithCancel(context.Background())
+	ctx, cancel := context.WithCancel(ctx)
 
 	go func() {
-		_, _ = io.Copy(ptmx, ptyStream)
+		_, err := io.Copy(ptmx, ptyStream)
+		if err != nil && err != io.EOF {
+			fmt.Fprintf(os.Stderr, "failed copying pty to ptyStream: %v\n", err)
+		}
 		cancel()
 	}()
 	go func() {
-		_, _ = io.Copy(ptyStream, ptmx)
+		_, err := io.Copy(ptyStream, ptmx)
+		if err != nil && err != io.EOF {
+			fmt.Fprintf(os.Stderr, "failed copying pty to ptyStream: %v\n", err)
+		}
 		cancel()
 	}()
 	go func() {
@@ -142,6 +148,8 @@ func main() {
 		return
 	}
 
+	ctx := context.Background()
+
 	remoteConsoleAddr := getRemoteDebuggerAddr()
 
 	cmd := exec.Command(args[0], args[1:]...)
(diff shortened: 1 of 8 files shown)


cmd/debugger/main.go
```diff
@@ -1,58 +1,129 @@
...
+	}
+	return "", false
+}
+
+func interactiveMode(remoteConsoleAddr string) error {
+	conn, err := net.Dial("tcp", remoteConsoleAddr)
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "failed to connect to earth remote console: %v\n", err)
-		return
+		return err
```
vladaionescu, line 42: It might be a good idea to `errors.Wrap` all errors coming from other packages. We've done this in all other places in our code. It helps with tracking issues.

In this case, we should additionally specify the address we were trying to connect to.

cmd/debugger/main.go
```diff
@@ -1,58 +1,129 @@
...
+	if !ok {
+		return ErrNoShellFound
 	}
-	err = cmd.Wait()
-	if err != nil {
-		fmt.Fprintf(os.Stderr, "failed to wait for debugger shell: %v\n", err)
-		return
+	c := exec.Command(shellPath)
+
+	ptmx, e := pty.Start(c)
```
vladaionescu, line 74: Nit: `err`, for consistency?

cmd/debugger/main.go
```diff
@@ -1,58 +1,129 @@
...
+	if e != nil {
+		fmt.Printf("failed to start pty: %v\n", e)
+		return e
 	}
+	defer func() { _ = ptmx.Close() }() // Best effort.
+
+	ctx, cancel := context.WithCancel(context.Background())
```
vladaionescu, line 81: Ideally we should create `context.Background()` only in main and pass it over here.

cmd/debugger/main.go
```diff
@@ -1,58 +1,129 @@
...
+	if e != nil {
+		fmt.Fprintf(os.Stderr, "failed to start pty: %v\n", e)
+		return e
 	}
+	defer func() { _ = ptmx.Close() }() // Best effort.
+
+	ctx, cancel := context.WithCancel(context.Background())
+
+	go func() {
+		_, _ = io.Copy(ptmx, ptyStream)
```
vladaionescu, line 84: Should we log these errors?

cmd/debugger/main.go
```diff
@@ -1,58 +1,129 @@
...
-		stdin.Write(b[:n])
+	shellPath, ok := getShellPath()
+	if !ok {
+		return ErrNoShellFound
 	}
-	err = cmd.Wait()
-	if err != nil {
-		fmt.Fprintf(os.Stderr, "failed to wait for debugger shell: %v\n", err)
-		return
+	c := exec.Command(shellPath)
```
vladaionescu, line 72: Should we pass in a context here too, for cancellation purposes? (Just in case shell hangs)

debugger/server/debugger_server.go
```diff
@@ -1,40 +1,139 @@
...
+			s.server.sigs <- syscall.SIGWINCH
+		default:
+			return fmt.Errorf("unsupported stream code %v", buf[0])
+		}
+	}
+}
+
+func (s *session) handlePtyStream(conn net.Conn) error {
+	go func() {
+		_, _ = io.Copy(os.Stdout, conn)
```
vladaionescu, line 78: Should we log these errors?

debugger/server/debugger_server.go
```diff
@@ -1,40 +1,139 @@
...
+		s.cancel()
+	}()
+	go func() {
+		_, _ = io.Copy(conn, os.Stdin)
+		s.cancel()
+	}()
+
+	<-s.ctx.Done()
+	return nil
```
vladaionescu, line 87: What do you think about separating session and DebugServer in different files?

debugger/server/debugger_server.go
```diff
@@ -55,8 +157,35 @@ func (ds *DebugServer) Start() (string, error) {
...
+func (ds *DebugServer) Stop() {
+	ds.cancel()
+}
+
+// NewDebugServer returns a new deubgging server
+func NewDebugServer(console conslogging.ConsoleLogger) *DebugServer {
+	sigs := make(chan os.Signal, 100)
+	signal.Notify(sigs, syscall.SIGWINCH)
+
+	ctx, cancel := context.WithCancel(context.Background())
```
vladaionescu, line 181: We should pass context from main and not use `Background` here.

debugger/server/debugger_server.go
```diff
@@ -1,40 +1,139 @@
...
+func (ds *DebugServer) handleRequest(conn net.Conn) error {
 	defer conn.Close()
-	b := make([]byte, 256)
+
+	yaSession, err := yamux.Server(conn, nil)
+	if err != nil {
+		return err
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
```
vladaionescu, line 98: We shouldn't use `Background` here. Should this perhaps use `ds.ctx` instead?

cmd/debugger/main.go
alexcb: good catch, that's some leftover copy and past from examples I was following :)

debugger/server/debugger_server.go
alexcb: sounds good, there's still a bit of interactions between the two due to the signal handling, but i think it's still a good idea.

debugger/server/debugger_server.go
alexcb: probably a good idea, I'll ignore any EOF errors though.

debugger/server/debugger_server.go
alexcb: good catch, I'll reference ds.ctx so we'll be able to cancel things from the main function's context.

cmd/debugger/main.go
alexcb: good point, i'll try to remember this in the future.

PR #166, opened 2020-08-06
Pipe password through stdin to avoid interactive login error

abennett:
I want to see green checkmarks for my last PR. 😁 

Following the [docker docs](https://docs.docker.com/engine/reference/commandline/login/#provide-a-password-using-stdin).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
         fi
         git checkout -b "$branch" || true
     - name: Docker Login
-      run: docker login --username vladaionescu --password "$DOCKERHUB_TOKEN"
+      run: echo "$DOCKERHUB_TOKEN" | docker login --username vladaionescu --password-stdin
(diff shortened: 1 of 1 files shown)


abennett:
Never mind. That didn't work as intended.

abennett:
Ah! Perhaps [this](https://docs.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets#using-encrypted-secrets-in-a-workflow) is why.
>With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository.

vladaionescu:
This doesn't make a difference actually, as you may have noticed.

We don't use GH Actions for external contributions (due to secrets, as you point out). We should disable it altogether.

The Circle failure is something unrelated - it's because the branch name is used as a docker tag and and the branch it used in this case was something like `pull/166`, which is an invalid docker tag. We'll fix this for better contributing experience.

PR #169, opened 2020-08-06
Extend verbose debugging into earth_debugger

alexcb:


debugger/server/debugger_server.go
```diff
@@ -79,17 +79,48 @@ func (ds *DebugServer) windowResizeHandler() error {
...
 	if err != nil {
-		return "", errors.Wrap(err, fmt.Sprintf("failed listening on %s", ds.addr))
+		return "", errors.Wrap(err, fmt.Sprintf("failed listening on %s", addr))
 	}
 
 	go ds.windowResizeHandler()
 
 	go func() {
-		ds.console.Printf("interactive debugger listening\n")
+		ds.console.Printf("interactive debugger listening on %v\n", l.Addr())
```
vladaionescu, line 123: Nit (for consistency)

```suggestion
		ds.console.Printf("Interactive debugger listening on %v\n", l.Addr())
```

Thinking that we should also add a prefix for this one, same way we use a prefix for `buildkitd` messages. What do you think? eg `ds.console.WithPrefix('debugger').Printf('....')`

alexcb:
Thanks for the comments @vladaionescu ; Due to the merge conflict and the change from tcp to socket-based connections for the stream, I've incorporated your comments and cherry picked bits off of this PR and opened up a new PR under https://github.com/earthly/earthly/pull/173

Let me know if I accidentally missed anything.

PR #170, opened 2020-08-06
Produce docker-compatible tags

vladaionescu:
This addresses issues with branch names containing things like `/`.

Wondering if we should be put in the existing variable `EARTHLY_TARGET_TAG` rather than a new one `EARTHLY_TARGET_TAG_DOCKER`.


vladaionescu:
Sorry for the whitespace changes in the Earthfile - it was very inconsistent before :disappointed_relieved: 

I can move that to a different PR if necessary.

PR #173, opened 2020-08-07
Bind interactive debugger to socket

alexcb:
Rather than using tcp to run the interactive debugger, use a
socket-based connection. This is more secure as it doesn't expose any
ports on the network (and also prevents firewalls from disallowing
connections).

### DIFF
--- a/buildkitd/buildkitd.go
+++ b/buildkitd/buildkitd.go
@@ -192,7 +192,7 @@ func Start(ctx context.Context, image string, settings Settings, reset bool) err
 	}
 	env := os.Environ()
 	cacheMount := fmt.Sprintf("%s:/tmp/earthly:delegated", settings.TempDir)
-	runMount := fmt.Sprintf("%s:/run/earthly:delegated", settings.RunDir)
+	runMount := fmt.Sprintf("%s:/run/earthly:consistent", settings.RunDir)
 	args := []string{
 		"run",
 		"-d",
--- a/buildkitd/buildkitd.go
+++ b/buildkitd/buildkitd.go
@@ -192,10 +192,12 @@ func Start(ctx context.Context, image string, settings Settings, reset bool) err
 	}
 	env := os.Environ()
 	cacheMount := fmt.Sprintf("%s:/tmp/earthly:delegated", settings.TempDir)
+	runMount := fmt.Sprintf("%s:/run/earthly:delegated", settings.RunDir)
 	args := []string{
 		"run",
 		"-d",
 		"-v", cacheMount,
+		"-v", runMount,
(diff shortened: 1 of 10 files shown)


debugger/server/debugger_server.go
```diff
@@ -107,16 +109,17 @@ func (ds *DebugServer) Start() (string, error) {
 			}
 		}
 	}()
-	return l.Addr().String(), nil
+	return nil
 }
 
 // Stop stops the server
 func (ds *DebugServer) Stop() {
+	fmt.Printf("stopping debug server")
```
vladaionescu, line 117: We should use `conslogging` and a prefix (eg `WithPrefix("debugger")`) for messages like this, so that it looks a little cleaner.

... though I wonder if the user even needs to know about the debug server. Maybe we don't say anything unless there's an error?

earthfile2llb/converter.go
```diff
@@ -672,11 +673,21 @@ func (c *Converter) internalRun(ctx context.Context, args []string, secretKeyVal
 		}
 	}
 
+	finalOpts = append(finalOpts,
+		llb.AddMount("/usr/bin/earth_debugger",
+			llb.Image(c.debuggerImage, llb.MarkImageInternal, llb.ResolveModePreferLocal, llb.Platform(llbutil.TargetPlatform)),
```
vladaionescu, line 678: I have an idea... if we host-bind the debugger binary too, then we wouldn't bust the cache if the debugger binary changes. What do you think? (This can be a separate PR though)

buildkitd/buildkitd.go
```diff
@@ -192,10 +192,12 @@ func Start(ctx context.Context, image string, settings Settings, reset bool) err
 	}
 	env := os.Environ()
 	cacheMount := fmt.Sprintf("%s:/tmp/earthly:delegated", settings.TempDir)
+	runMount := fmt.Sprintf("%s:/run/earthly:delegated", settings.RunDir)
```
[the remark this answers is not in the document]
alexcb: good idea, i'll switch it to `consistent`; we'll need to do a bit more macOS testing for this.

debugger/server/debugger_server.go
alexcb: whoops, this was some left-over debugging code that slipped in. I'll remove it.

earthfile2llb/converter.go
alexcb: that would also make distributing the debugger binary easier. Currently we have to manually push it to docker hub, this would greatly improve that process.

alexcb:
Good find on the macOS socket issue. It would be really hacky, but maybe we could abuse the ssh agent socket and look for a special non-ssh handshake which would then forward data to the debugger shell, otherwise we would forward data to the regular ssh agent.  This is a last-resort sort of idea though.

PR #186, opened 2020-08-15
Use buildkit Earthfile

vladaionescu:
Also, fix run mounts not working in Dockerfiles.

- [x] Should update this PR to use `earthly-master`, when https://github.com/earthly/buildkit/pull/6 gets merged


alexcb:
Can you help me understand why we want to specify the buildkit version in the CI configs rather than in the Earthfile? By moving it into the CI configs, won't that mean we need to update the version in two different places, as well as when we run it locally it won't necessarily use the same version as which is used in CI.

alexcb:
Ah I understand now, thanks for explaining that. It made much more sense once I noticed the Earthfile in the buildkit repo.

PR #190, opened 2020-08-16
Get rid of vfs in docker-in-docker (experiment)

vladaionescu:
This gets us closer to being able to use docker-in-docker safely.

Changes:

* Use a separate dir for docker data for each target and bind that dir to the host.
* In our case, the host is the buildkitd container, so binding to the host, really means binding to a dir within the buildkitd container.
* The dir used is `/tmp/earthly/dind/<hash-of-target>` within buildkitd container, which will end up being part of the same docker volume as the rest of the cache.
* Between runs, the same data dir is used. Meaning that docker data will be reused between runs. This is a temporary unintended consequence.

Things that are no longer a limitation due to this:

* The docker daemon no longer uses vfs driver, but rather overlayfs2, which is far more performant (both in terms of speed and disk usage).

New limitations (for now):

* Between runs, the same docker dir is used, causing an unintended leaky abstraction.
* Because the docker dir is not part of the regular cache, there may be surprising effects if only part of the build executes (eg partly cached).
* The data generated by dind runs is not garbage-collected automatically (it's separate from the buildkit cache). However, because it's in the same volume as the cache it would be cleaned up by a `earth prune --reset`. This needs to be addressed separately.

This requires more thinking.

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -164,6 +164,3 @@ examples:
     BUILD ./examples/readme/go3+build
     BUILD ./examples/readme/proto+docker
     BUILD github.com/earthly/hello-world+hello
-
-test-experimental:
-    BUILD ./examples/tests+experimental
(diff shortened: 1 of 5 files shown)


vladaionescu:
The ideas from this experiment were used in the new `WITH DOCKER` clause. This is now obsolete.

PR #206, opened 2020-08-19
Remove extraneous checkout step in Actions example

cirego:
GitHub Actions checkout@v2 no longer leaves the branch in a detached head state,
so there is no longer a need to run `git checkout` as a step.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,15 +16,6 @@ jobs:
     if: github.repository == 'earthly/earthly'
     steps:
     - uses: actions/checkout@v2
-    - name: Put back the git branch into git (Earthly uses it for tagging)
-      run: |
-        branch=""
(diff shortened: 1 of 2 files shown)


vladaionescu:
Very nice!

Mind removing this from the Earthly CI build as well? https://github.com/earthly/earthly/blob/master/.github/workflows/ci.yml#L19

It'll help with testing this.

cirego:
Hi @vladaionescu -- I just updated the PR to include the GitHub Action as well!

vladaionescu:
Actually it won't run here because of missing creds for PRs. I'll merge as is.

Thanks for the contribution!

cirego:
> Actually it won't run here because of missing creds for PRs. I'll merge as is.
> 
> Thanks for the contribution!

Thanks for Earthly! I find it super useful!

Issue #204, opened 2020-08-19
Should fail on RUN error


ctison:
Earth version: `v0.3.0`

#### Earthfile
```
FROM alpine:3.12.0@sha256:90baa0922fe90624b05cb5766fa5da4e337921656c2f8e2b13bd3c052a0baac1
shouldfail:
  RUN false
  RUN echo KO
```

#### Result
```
buildkitd | Found buildkit daemon as docker container (earthly-buildkitd)
+base | --> FROM alpine:3.12.0@sha256:90baa0922fe90624b05cb5766fa5da4e337921656c2f8e2b13bd3c052a0baac1
+base | resolve docker.io/library/alpine:3.12.0@sha256:90baa0922fe90624b05cb5766fa5da4e337921656c2f8e2b13bd3c052a0baac1 100%
+shouldfail | --> RUN false
+shouldfail | Command /bin/sh -c  /bin/sh -c 'false' failed with exit code 1
+shouldfail | +shouldfail | --> RUN echo KO
+shouldfail | KO
+shouldfail | Target +shouldfail built successfully
=========================== SUCCESS ===========================
```

alexcb:
Thank you very much for the report @ctison , we'll get this fix out ASAP.

vladaionescu:
Hi @ctison - FYI, we just released `v0.3.1` of Earthly which contains a fix for this.

PR #209, opened 2020-08-19
test for capturing run errors

alexcb:
* validate that RUN commands that return an error code are correctly
caught by earth, which causes earth to fail.

This test is to prevent #204 from reoccurring.

### DIFF
--- a/examples/tests/Earthfile
+++ b/examples/tests/Earthfile
@@ -19,6 +19,7 @@ all:
     BUILD +dockerfile-test
     BUILD +fail-test
     BUILD +fail-push-test
+    BUILD +push-test
 
 experimental:
     # TODO: These are troublesome due to vfs driver used in dind. They quickly exhaust disk space.
--- a/examples/tests/Earthfile
+++ b/examples/tests/Earthfile
@@ -17,6 +17,8 @@ all:
     BUILD +host-bind-test
     BUILD +star-test
     BUILD +dockerfile-test
+    BUILD +fail-test
+    BUILD +fail-push-test
 
 experimental:
     # TODO: These are troublesome due to vfs driver used in dind. They quickly exhaust disk space.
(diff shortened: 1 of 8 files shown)


examples/tests/fail.earth
```diff
@@ -0,0 +1,4 @@
+FROM alpine:3.11
+test:
+    # intentionally cause a failure (this test is to ensure failures are caught)
+    RUN false
```
vladaionescu, line 4: Wondering if we can also put some output in here (eg `RUN echo "expected output"; false`) and then check that the output is printed (`.... +test | grep "expected output"`).

examples/tests/Earthfile
```diff
@@ -200,3 +200,8 @@ dockerfile-test:
         --entrypoint \
         --mount=type=tmpfs,target=/tmp/earthly \
         -- --no-output +test
+
+fail-test:
```
vladaionescu, line 204: Remember to add this to `+all`

examples/tests/fail.earth
alexcb: good plan, I changed it to a `ls /this-will-fail` then I can grep for that fake path shows up in the error message.

examples/tests/Earthfile
```diff
@@ -17,6 +17,8 @@ all:
     BUILD +host-bind-test
     BUILD +star-test
     BUILD +dockerfile-test
+    BUILD +fail-test
+    BUILD +fail-push-test
```
vladaionescu, line 21: Add also +push-test?

examples/tests/Earthfile
```diff
@@ -200,3 +203,23 @@ dockerfile-test:
         --entrypoint \
         --mount=type=tmpfs,target=/tmp/earthly \
         -- --no-output +test
+
+fail-test:
+    COPY fail.earth ./Earthfile
+    # test that an error code is correctly returned
+    RUN --privileged ! /usr/bin/earth-buildkitd-wrapper.sh +test
```
vladaionescu, line 210: Oh - another detail. I've noticed it uses less cache if the buildkit dir is not preserved (`--mount=type=tmpfs,target=/tmp/earthly`). See other tests.

alexcb: I was wondering why those other tests had that, that makes a lot of sense.

PR #228, opened 2020-08-22
Bash auto completion

alexcb:
To enable auto completion, a user must enter:

  complete -o filenames -C "/path/to/earth" earth

### DIFF
--- a/autocomplete/complete_test.go
+++ b/autocomplete/complete_test.go
@@ -1,0 +1,23 @@
+package autocomplete
+
+import (
+	"testing"
+
+	. "github.com/stretchr/testify/assert"
+)
+
+func TestFlagCompletion(t *testing.T) {
+	matches, err := GetPotentials("--fl", []string{"flag", "fleet", "fig"}, nil)
+	NoError(t, err, "GetPotentials failed")
+	Equal(t, []string{"--flag", "--fleet"}, matches)
+}
+func TestCommandCompletion(t *testing.T) {
+	matches, err := GetPotentials("pru", nil, []string{"prune", "foo"})
+	NoError(t, err, "GetPotentials failed")
+	Equal(t, []string{"prune"}, matches)
+}
+func TestPathCompletion(t *testing.T) {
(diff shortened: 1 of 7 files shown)


autocomplete/complete_test.go
```diff
@@ -0,0 +1,25 @@
...
+	matches, err := GetPotentials("earth pru", 9, nil, []string{"prune", "foo"})
+	NoError(t, err, "GetPotentials failed")
+	Equal(t, []string{"prune "}, matches)
+}
+
+func TestPathCompletion(t *testing.T) {
+	matches, err := GetPotentials("earth .", 7, nil, nil)
+	NoError(t, err, "GetPotentials failed")
+	Equal(t, []string{"./", "../"}, matches)
+}
```
vladaionescu, line 25: Very nice

cmd/earth/main.go
```diff
@@ -79,7 +81,66 @@ var (
...
+	}
+
+	commands := []string{}
+	for _, cmd := range app.cliApp.Commands {
+		commands = append(commands, cmd.Name)
+	}
+
+	potentials, err := autocomplete.GetPotentials(compLine, int(compPointInt), flags, commands)
+	if err != nil {
+		//panic(err) // can't display error or it will show up under tab-completion
```
vladaionescu, line 116: Would stderr work for errors?

I assume that the answer is "no" given that you probably tried `panic`.

cmd/earth/main.go
```diff
@@ -79,7 +81,66 @@ var (
 	GitSha string
 )
 
+// to enable autocomplete, enter
+// complete -o nospace -C "/path/to/earth" earth
+func autoComplete(app *earthApp) {
```
vladaionescu, line 86: How about

```suggestion
func (app *earthApp) autoComplete() {
```

cmd/earth/main.go
alexcb: unfortunately not; if the panic is there it'll suggestion the output as options.

PR #230, opened 2020-08-22
Implementation for WITH DOCKER

vladaionescu:
Replaces the old syntax of

```Dockerfile
DOCKER LOAD ...
DOCKER PULL ...
RUN --with-docker docker run ...
RUN --with-docker docker run ...
```

with

```Dockerfile
WITH DOCKER
  DOCKER LOAD ...
  DOCKER PULL ...
  RUN docker run ... && \
    docker run ...
END
```

(The old syntax still works, however it has the same performance drawbacks as before. It will be removed in a future version.)

Docker data is no longer persisted as layers, but rather it is stored as a temporary dir via a host bind (binds to a dir in buildkitd - that dir happens to be as part of the same cache volume as buildkit data). Because we use this host bind that maps to a docker volume, we can now use overlayfs for the inner docker. This fixes the performance and disk usage issues of the previous docker-in-docker approach.

Because data is not part of the layers, it cannot be used as partial cache. The behavior now is that after a `WITH DOCKER...END`, the docker data is wiped. So each `WITH DOCKER` clause gets its own single-use docker daemon. However, `DOCKER LOAD`'s and `DOCKER PULL`'s are cached at the buildkit layer, which makes the whole thing reasonably fast.

In the future, the syntax might change to something more generic like:

```Dockerfile
WITH LAYER
  DOCKER LOAD ...
  DOCKER PULL ...
  RUN --with-docker docker run ...
  RUN --with-docker docker run ...
END
```

which makes the entire operation be part of a single layer. This is TBD however.


examples/tests/config.yaml
```diff
@@ -0,0 +1,2 @@
+global:
+    debugger_port: 4999
```
alexcb, line 2: ah ha, this explains #231

vladaionescu: Exactly :D.

Couldn't get earth within earth via `WITH DOCKER` working just yet, however. It seems that `docker run --privileged` does not work (we need that to start buildkitd). Anyway, most people don't need that so we should be ok for now.

PR #237, opened 2020-08-25
package up binaries in docker for pre-release testing

alexcb:
The Idea with this new target, is we could have it built on master, then we can use a script like this:

```
set -e

function get_latest_binary {
    docker pull alexcb132/earthlybinaries:latest
    docker create --name binaryhack alexcb132/earthlybinaries
    
    earthbinpath=/earth-linux-amd64
    if [ "$(uname)" == "Darwin" ]; then
    earthbinpath=/earth-darwin-amd64
    fi
    
    docker cp binaryhack:$earthbinpath ~/bin/earth-bin
    docker rm binaryhack
}

LAST=`cat /tmp/last-earth-check 2>/dev/null || echo 0`
NOW=`date +%s`
since=$(expr $NOW - $LAST)

if [ $since -ge 3600 ]; then
    echo pulling latest earth version
    get_latest_binary
    echo $NOW > /tmp/last-earth-check
fi

~/bin/earth-bin $@

```

to periodically check for latest pre-releases.

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -42,7 +42,7 @@ code:
 
 lintscripts:
     FROM +deps
-    COPY ./buildkitd/entrypoint.sh ./shell_scripts/.
+    COPY ./earth ./buildkitd/entrypoint.sh ./shell_scripts/.
     RUN shellcheck shell_scripts/*
 
 lint:
--- a/Earthfile
+++ b/Earthfile
@@ -153,6 +153,15 @@ earth-docker:
     COPY --build-arg VERSION=$TAG +earth/earth /usr/bin/earth
     SAVE IMAGE --push earthly/earth:$TAG
 
+# we abuse docker here to distribute our binaries
+prerelease-docker:
+    COPY +earth/earth /earth-linux-amd64
+    COPY --build-arg GOOS=darwin \
+        --build-arg GOARCH=amd64 \
+        --build-arg GO_EXTRA_LDFLAGS= \
+    +earth/earth /earth-darwin-amd64
+    SAVE IMAGE --push earthly/earthlybinaries:latest
+
 for-linux:
     BUILD +buildkitd
     BUILD +earth
@@ -165,6 +174,7 @@ all:
     BUILD +buildkitd
     BUILD +earth-all
     BUILD +earth-docker
+    BUILD +prerelease-docker
 
 test:
     BUILD +lint
--- a/earth
+++ b/earth
@@ -1,0 +1,29 @@
+#!/bin/bash
+set -e
+
+LAST_CHECK_STATE_PATH=/tmp/last-earth-prerelease-check
+
+function get_latest_binary {
+    docker pull alexcb132/earthlybinaries:latest
+    docker create --name earthly_binary alexcb132/earthlybinaries
+
+    earthbinpath=/earth-linux-amd64
+    if [ "$(uname)" == "Darwin" ]; then
+        earthbinpath=/earth-darwin-amd64
+    fi
+
+    docker cp earthly_binary:$earthbinpath ~/bin/earth-bin
+    docker rm earthly_binary
+}
+
+LAST=`cat $LAST_CHECK_STATE_PATH 2>/dev/null || echo 0`
+NOW=`date +%s`
+since=$(expr $NOW - $LAST)
+
+if [ $since -ge 3600 ]; then
+    echo checking for latest earth pre-release binaries
+    get_latest_binary
+    echo $NOW > $LAST_CHECK_STATE_PATH
+fi
+
+~/bin/earth-bin $@
--- a/earth
+++ b/earth
@@ -4,7 +4,7 @@ set -e
 last_check_state_path=/tmp/last-earth-prerelease-check
 
 function get_latest_binary {
-    docker pull alexcb132/earthlybinaries:latest
+    docker pull earthly/earthlybinaries:latest
     docker create --name earthly_binary earthly/earthlybinaries
 
     earth_bin_path=/earth-linux-amd64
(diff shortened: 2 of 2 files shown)


Earthfile
```diff
@@ -146,6 +146,16 @@ earth-docker:
     COPY --build-arg VERSION=$TAG +earth/earth /usr/bin/earth
     SAVE IMAGE --push earthly/earth:$TAG
 
+# we abuse docker here to distribute our binaries
+prerelease-docker:
+    ARG EARTHLY_GIT_HASH
```
vladaionescu, line 151: This is not necessary, right?

Earthfile
```diff
@@ -146,6 +146,16 @@ earth-docker:
...
+    COPY +earth/earth /earth-linux-amd64
+    COPY --build-arg GOOS=darwin \
+        --build-arg GOARCH=amd64 \
+        --build-arg GO_EXTRA_LDFLAGS= \
+    +earth/earth /earth-darwin-amd64
+    SAVE IMAGE --push alexcb132/earthlybinaries:latest
```
vladaionescu, line 157: I'll make an earthly repo for this.

vladaionescu: I've created this: https://hub.docker.com/repository/docker/earthly/earthlybinaries

Earthfile
alexcb: good catch, I used to be using it in the tag, but no longer am and will remove it.

Earthfile
alexcb: thank you kindly :)

alexcb:
I added the `prerelease-docker` target under `all`, so the github action will automatically handle it.

alexcb:
good point regarding the scrip, I added it to the root of the repo, let me know if that's what you meant by `./earth`

earth
```diff
@@ -0,0 +1,29 @@
...
+NOW=`date +%s`
+since=$(expr $NOW - $LAST)
+
+if [ $since -ge 3600 ]; then
+    echo checking for latest earth pre-release binaries
+    get_latest_binary
+    echo $NOW > $LAST_CHECK_STATE_PATH
+fi
+
+~/bin/earth-bin $@
```
vladaionescu, line 29: You probably want

```suggestion
~/bin/earth-bin "$@"
```

just in case.

earth
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+set -e
+
+LAST_CHECK_STATE_PATH=/tmp/last-earth-prerelease-check
+
+function get_latest_binary {
+    docker pull alexcb132/earthlybinaries:latest
+    docker create --name earthly_binary alexcb132/earthlybinaries
```
vladaionescu, line 8: Update to use `earthly/...` image?

earth
```diff
@@ -0,0 +1,29 @@
...
+    earthbinpath=/earth-linux-amd64
+    if [ "$(uname)" == "Darwin" ]; then
+        earthbinpath=/earth-darwin-amd64
+    fi
+
+    docker cp earthly_binary:$earthbinpath ~/bin/earth-bin
+    docker rm earthly_binary
+}
+
+LAST=`cat $LAST_CHECK_STATE_PATH 2>/dev/null || echo 0`
```
vladaionescu, line 19:
```suggestion
LAST=`cat "$LAST_CHECK_STATE_PATH" 2>/dev/null || echo 0`
```

It's a good idea to always put variable expansions in double-quotes. Bash scripts tend to be kinda dangerous.

earth
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+set -e
+
+LAST_CHECK_STATE_PATH=/tmp/last-earth-prerelease-check
```
vladaionescu, line 4: We've generally used lower-case variable names for local vars and upper-case for exported vars in bash files. Might be good to maintain style.

earth
```diff
@@ -0,0 +1,29 @@
...
+
+function get_latest_binary {
+    docker pull alexcb132/earthlybinaries:latest
+    docker create --name earthly_binary alexcb132/earthlybinaries
+
+    earthbinpath=/earth-linux-amd64
```
vladaionescu, line 10:
```suggestion
    earth_bin_path=/earth-linux-amd64
```

Nit: style

earth
```diff
@@ -0,0 +1,29 @@
...
+    if [ "$(uname)" == "Darwin" ]; then
+        earthbinpath=/earth-darwin-amd64
+    fi
+
+    docker cp earthly_binary:$earthbinpath ~/bin/earth-bin
```
vladaionescu, line 15:
```suggestion
    docker cp earthly_binary:"$earthbinpath" ~/bin/earth-bin
```

Just in case this script gets modified in the future.

alexcb:
thanks for all those bash recommendations, my bash is rather sloppy.
I switched it over to use `earthly/earthlybinaries:latest` however nothing has been pushed yet, but i'm hopeful to see the github action setup perform the push.

earth
```diff
@@ -0,0 +1,29 @@
...
+}
+
+last=`cat "$last_check_state_path" 2>/dev/null || echo 0`
+now=`date +%s`
+since=$(expr $now - $last)
+
+if [ $since -ge 3600 ]; then
+    echo checking for latest earth pre-release binaries
+    get_latest_binary
+    echo $now > $last_check_state_path
```
vladaionescu, line 26:
```suggestion
    echo "$now" > "$last_check_state_path"
```

earth
```diff
@@ -0,0 +1,29 @@
...
+    docker cp earthly_binary:"$earth_bin_path" ~/bin/earth-bin
+    docker rm earthly_binary
+}
+
+last=`cat "$last_check_state_path" 2>/dev/null || echo 0`
+now=`date +%s`
+since=$(expr $now - $last)
+
+if [ $since -ge 3600 ]; then
+    echo checking for latest earth pre-release binaries
```
vladaionescu, line 24:
```suggestion
    echo "checking for latest earth pre-release binaries"
```

Nit: style

PR #241, opened 2020-08-26
add shellcheck into lint

alexcb:
I only added `./buildkitd/entrypoint.sh` to start, but we could expand it to more scripts if we like this direction?

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -14,6 +14,7 @@ RUN apk add --update --no-cache \
     less \
     make \
     openssl \
+    shellcheck \
     util-linux
 
 WORKDIR /earthly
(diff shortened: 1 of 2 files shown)


Earthfile
```diff
@@ -36,10 +37,13 @@ code:
     COPY --dir earthfile2llb/antlrhandler earthfile2llb/dedup earthfile2llb/image \
         earthfile2llb/imr earthfile2llb/variables earthfile2llb/*.go earthfile2llb/
     COPY ./earthfile2llb/parser+parser/*.go ./earthfile2llb/parser/
+    COPY ./earthfile2llb/parser+parser/*.go ./earthfile2llb/parser/
+    COPY ./buildkitd/entrypoint.sh ./shell_scripts/.
```
vladaionescu, line 41: I would maybe move this to its own target to avoid recompiling go code on every shell script change.

alexcb: good idea, I just created a `lintscripts` target, how does that look?

PR #247, opened 2020-08-26
Fix actual failure reason sometimes masked by context canceled error

vladaionescu:


builder/solver_monitor.go
```diff
@@ -194,8 +194,6 @@ Loop:
 				}
 
 			}
-		case <-ctx.Done():
```
alexcb, line 197: is this intentional?

vladaionescu: Yes, the status stream also ends when ctx is done, however there may be a few more important log messages that need to be printed before that.

PR #248, opened 2020-08-26
Support for .env files

### LINKED ISSUE #179: Support for .env files
A feature request is the support for `.env` files (similar to how compose uses them https://docs.docker.com/compose/env-file/).

~It might work like this: if an `.env` file is present in the current dir, or any dir above the current dir (dirs closer to current dir have higher precendence),~ load the vars and pass them to the build as both build-args AND secrets. EDIT: Using the `.env` in the current dir only seems to be the the behavior most similar to how other tools work.

CC @zenyui

vladaionescu:
Fixes #179

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 *.tokens
 *.interp
 .antlr/
+.env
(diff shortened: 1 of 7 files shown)


zenyui:
this is awesome thanks guys.

PR #249, opened 2020-08-26
Perform proper cleanup on sigint and sigterm

### LINKED ISSUE #244: Pressing ctrl+c mid-build causes leftover temp files
Earthly uses a series of temporary dirs and files during the build. These exist in `/tmp/earthly-*` and also in `./.tmp-out-*`. Earthly normally cleans these up at the end of a build, however there is currently no intelligent sigint handling to clean these up in case of ctrl+c.

vladaionescu:
This added waiting gives `cleanupCollection.Close()` a chance to finish and remove any temp files and dirs lying around.

Fixes #244

### DIFF
--- a/debugger/terminal/terminal.go
+++ b/debugger/terminal/terminal.go
@@ -7,6 +7,7 @@ import (
 	"net"
 	"os"
 	"os/signal"
+	"sync"
 	"syscall"
 
 	"github.com/earthly/earthly/debugger/common"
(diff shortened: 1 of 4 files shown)


vladaionescu:
> The interactive debugger catches sigwinchange to resize the terminal; it might be good to double check this doesn't interfere with that code. I also wonder if we need to worry about a ctrl-c being directed to the debugger vs this code?

Good thinking - I'll test manually.

debugger/terminal/terminal.go
```diff
@@ -163,7 +158,41 @@ func ConnectTerm(ctx context.Context, addr string) error {
...
-
 	fmt.Fprintf(os.Stderr, "exiting interactive debugger shell\n")
+	err = ts.restore()
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+type termState struct {
```
alexcb, line 169: very nice!

PR #260, opened 2020-08-28
Experiment with darwin CI via GH actions hosted

vladaionescu:


vladaionescu:
It works.... but it's suuuuuper slow. Unusable.

PR #262, opened 2020-08-30
Add buildkite definition for Windows CI

vladaionescu:
... also remove GitHub Actions windows CI definition.


earth-buildkitd-wrapper.sh
```diff
@@ -15,7 +15,7 @@ i=1
 while [ ! -S "/run/buildkit/buildkitd.sock" ]; do
     sleep 1
     i=$((i+1))
-    if [ "$i" -gt "10" ]; then
+    if [ "$i" -gt "30" ]; then
         kill -9 "$buildkitd_pid" >/dev/null 2>&1
         echo "Buildkitd did not start within 10 seconds"
```
alexcb, line 20: `echo "Buildkitd did not start within 10 seconds"` also needs updating as it's now 30 seconds.

How about adding this timeout value as a (const) variable e.g. `BUILDKITD_STARTUP_TIMEOUT=30`

vladaionescu: Good point - I called it just `timeout`, given that it's a local variable.

PR #266, opened 2020-09-01
Auto install bash autocompletion

alexcb:
Create a bootstrap command which automatically installs bash
autocompletion.

### DIFF
--- a/cmd/earth/main.go
+++ b/cmd/earth/main.go
@@ -130,6 +130,38 @@ func (app *earthApp) autoComplete() {
 	os.Exit(0)
 }
 
+func (app *earthApp) insertBashCompleteEntry() error {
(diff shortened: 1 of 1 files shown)


cmd/earth/main.go
```diff
@@ -167,7 +199,7 @@ func main() {
 	if fileExists(dotEnvPath) {
 		err := godotenv.Load(dotEnvPath)
 		if err != nil {
-			fmt.Printf("Error loading dot-env file %s: %s", dotEnvPath, err.Error())
+			fmt.Printf("Error loading dot-env file %s: %s\n", dotEnvPath, err.Error())
```
vladaionescu, line 202: Thanks :D

PR #269, opened 2020-09-01
extend autocompletion to work with ~/ prefix

alexcb:
* fix autocompletion to work with ~/ prefixes

### DIFF
--- a/autocomplete/complete.go
+++ b/autocomplete/complete.go
@@ -68,14 +68,14 @@ func getPotentialTarget(prefix string) ([]string, error) {
 	}
 	dirPath := splits[0]
 
-	user, err := user.Current()
+	currentUser, err := user.Current()
 	if err != nil {
 		return nil, err
 	}
 
 	realDirPath := dirPath
 	if strings.HasPrefix(prefix, "~/") {
-		realDirPath = user.HomeDir + "/" + dirPath[2:]
+		realDirPath = currentUser.HomeDir + "/" + dirPath[2:]
 	}
 
 	targets, err := earthfile2llb.GetTargets(path.Join(realDirPath, "Earthfile"))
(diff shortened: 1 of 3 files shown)


autocomplete/complete.go
```diff
@@ -67,7 +68,17 @@ func getPotentialTarget(prefix string) ([]string, error) {
 	}
 	dirPath := splits[0]
 
-	targets, err := earthfile2llb.GetTargets(path.Join(dirPath, "Earthfile"))
+	user, err := user.Current()
```
vladaionescu, line 71: Might be worth using a different variable name for `user` to avoid overlap with the package name `user`?

autocomplete/complete.go
```diff
@@ -3,6 +3,7 @@ package autocomplete
 import (
 	"io/ioutil"
 	"os"
+	"os/user"
```
vladaionescu, line 6: TIL

autocomplete/complete.go
```diff
@@ -67,7 +68,17 @@ func getPotentialTarget(prefix string) ([]string, error) {
...
 
-	targets, err := earthfile2llb.GetTargets(path.Join(dirPath, "Earthfile"))
+	user, err := user.Current()
+	if err != nil {
+		return nil, err
+	}
+
+	realDirPath := dirPath
+	if strings.HasPrefix(prefix, "~/") {
+		realDirPath = user.HomeDir + "/" + dirPath[2:]
```
vladaionescu, line 78: Should we use `user.HomeDir` in all the places where we've been using `os.Getenv("HOME")`?

autocomplete/complete.go
alexcb: ha, that's a really good point. I didn't notice I did that :)

autocomplete/complete.go
alexcb: It's probably a good idea to be consistent. I'll make the switch.

PR #275, opened 2020-09-02
automatically delete zcompdump cache files when adding zsh autocomple…

alexcb:
…tion

### DIFF
--- a/cmd/earth/main.go
+++ b/cmd/earth/main.go
@@ -165,26 +165,26 @@ func (app *earthApp) deleteZcompdump() error {
 	if !found {
 		currentUser, err := user.Current()
 		if err != nil {
-			return err
+			return errors.Wrapf(err, "failed to lookup current user")
 		}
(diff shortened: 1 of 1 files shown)


cmd/earth/main.go
```diff
@@ -159,6 +159,38 @@ func (app *earthApp) insertBashCompleteEntry() error {
...
+	files, err := ioutil.ReadDir(homeDir)
+	if err != nil {
+		return err
+	}
+	for _, f := range files {
+		if strings.HasPrefix(f.Name(), ".zcompdump") {
+			path := filepath.Join(homeDir, f.Name())
+			err := os.Remove(path)
+			if err != nil {
+				return err
```
vladaionescu, line 187: We should wrap this with `errors.Wrap` and also include the filename in the error.

vladaionescu (approved):

vladaionescu (approved):

PR #279, opened 2020-09-03
Use $HOME rather than user.Current().HomeDir

alexcb:
the brew install overrides $HOME to point to a sandbox; however when
using user.Current().HomeDir it doesn't pick this setting up ( see
https://github.com/golang/go/issues/26463 ).


vladaionescu (approved):
Interesting

PR #280, opened 2020-09-03
Adding Python example

agbell:
Adding python example using wheel files to install package dependencies.  

The build target is unnecessary for now but mirrors the structure of examples in https://docs.earthly.dev/guides/basics.


alexcb:
Looks like a great python example @agbell 

can you also add an entry to our main Earthfile under the examples target (see https://github.com/earthly/earthly/blob/master/Earthfile#L195 )

This way it'll automatically be tested via CI and during our release process.

alexcb:
ha, you were too quick, you already got it added in before I hit enter on my comment :+1:

alexcb (approved):

vladaionescu (approved):
Yay! Perhaps we can build a pypi package instead for `+build`? I don't know if it makes sense though.

PR #281, opened 2020-09-03
display all text, but return error code if search string doesnt exist

alexcb:


vladaionescu (approved):
Ooh perl :D

alexcb:
I don't know what it means, stackoverflow told me to use it :grin:

PR #287, opened 2020-09-04
Python examples on basics tutorial 

agbell:
View here https://app.gitbook.com/@earthly/s/earthly-staging/v/python-basic-tutorial/guides/basics

### DIFF
--- a/docs/guides/basics.md
+++ b/docs/guides/basics.md
@@ -597,6 +597,7 @@ Markdown==3.2.2
 The code of the app would now look like this
 ```python
 # src/hello.py
+
 from markdown import markdown
 
 def hello():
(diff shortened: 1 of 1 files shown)


alexcb (approved):

vladaionescu:
Woohoo! Might be worth also adding a dedicated section for Python under examples, in case someone is scanning the TOC looking for it. It can be very simple, like we have for Java: https://docs.earthly.dev/examples/java

agbell:
> Woohoo! Might be worth also adding a dedicated section for Python under examples, in case someone is scanning the TOC looking for it. It can be very simple, like we have for Java: https://docs.earthly.dev/examples/java

Great idea.  raised here https://github.com/earthly/earthly/pull/289

PR #292, opened 2020-09-09
pass ssh agent socket to RUN commands

### LINKED ISSUE #58: SSH agent socket passing
Hi, I am trying to download a go module from a private github repository within an earthly build and I cannot figure out how this is supposed to work. In the earthly documentation I read about [SSH agent socket passing](https://docs.earthly.dev/guides/auth), but I think this is only intended for the GIT CLONE command and won't work for `go mod download`?

So I tried to do it like described in the [buildkit doc](https://docs.docker.com/develop/develop-images/build_enhancements/), but then docker complains about the missing mount point:

```
RUN git config --global url."ssh://git@github.com/some-org/".insteadOf https://github.com/some-org/ 
RUN mkdir -p -m 0600 ~/.ssh && ssh-keyscan github.com >> ~/.ssh/known_hosts
RUN --mount=type=ssh GOPRIVATE=github.com/some-org go mod download
```
Could you please point me in the right direction?

alexcb:
This fixes #58

Here's what I used for testing this:

```
FROM ubuntu:20.10

## for apt to be noninteractive
ENV DEBIAN_FRONTEND noninteractive
ENV DEBCONF_NONINTERACTIVE_SEEN true

RUN apt-get update && apt-get install -y ssh

test:
  RUN echo hello world
  RUN ssh -o "StrictHostKeyChecking=no" git@github.com
```

### DIFF
--- a/cmd/earth/main.go
+++ b/cmd/earth/main.go
@@ -873,10 +873,9 @@ func (app *earthApp) actionBuild(c *cli.Context) error {
 	}
 
 	var withSSH bool
-	sshPath := defaultSSHAuthSock()
-	if sshPath != "" {
+	if app.buildkitdSettings.SSHAuthSock != "" {
 		ssh, err := sshprovider.NewSSHAgentProvider([]sshprovider.AgentConfig{{
-			Paths: []string{sshPath},
+			Paths: []string{app.buildkitdSettings.SSHAuthSock},
 		}})
 		if err != nil {
 			return errors.Wrap(err, "ssh agent provider")
--- a/cmd/earth/main.go
+++ b/cmd/earth/main.go
@@ -872,7 +872,6 @@ func (app *earthApp) actionBuild(c *cli.Context) error {
 		authprovider.NewDockerAuthProvider(os.Stderr),
 	}
 
-	var withSSH bool
 	if app.buildkitdSettings.SSHAuthSock != "" {
 		ssh, err := sshprovider.NewSSHAgentProvider([]sshprovider.AgentConfig{{
 			Paths: []string{app.buildkitdSettings.SSHAuthSock},
@@ -881,7 +880,6 @@ func (app *earthApp) actionBuild(c *cli.Context) error {
 			return errors.Wrap(err, "ssh agent provider")
 		}
 		attachables = append(attachables, ssh)
-		withSSH = true
 	}
 
 	var enttlmnts []entitlements.Entitlement
@@ -916,7 +914,6 @@ func (app *earthApp) actionBuild(c *cli.Context) error {
 			ArtifactBuilderFun: b.MakeArtifactBuilderFun(),
 			CleanCollection:    cleanCollection,
 			VarCollection:      varCollection,
-			WithSSH:            withSSH,
 		})
 	if err != nil {
 		return err
--- a/cmd/earth/main.go
+++ b/cmd/earth/main.go
@@ -40,6 +40,7 @@ import (
 	"github.com/moby/buildkit/session"
 	"github.com/moby/buildkit/session/auth/authprovider"
 	"github.com/moby/buildkit/session/secrets/secretsprovider"
+	"github.com/moby/buildkit/session/sshforward/sshprovider"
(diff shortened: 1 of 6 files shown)


cmd/earth/main.go
```diff
@@ -870,6 +871,20 @@ func (app *earthApp) actionBuild(c *cli.Context) error {
 		secretsprovider.FromMap(secretsMap),
 		authprovider.NewDockerAuthProvider(os.Stderr),
 	}
+
+	var withSSH bool
+	sshPath := defaultSSHAuthSock()
```
vladaionescu, line 876: I would perhaps use the same SSH flag we had before to configure this as well. It allows the user the override the SSH socket if they want to.

The one defined as:

```
&cli.StringFlag{
			Name:        "ssh-auth-sock",
			Value:       defaultSSHAuthSock(),
			EnvVars:     []string{"EARTHLY_SSH_AUTH_SOCK"},
			Usage:       "The SSH auth socket to use for ssh-agent forwarding",
			Destination: &app.buildkitdSettings.SSHAuthSock,
		},
```

earthfile2llb/converter.go
```diff
@@ -724,6 +727,9 @@ func (c *Converter) internalRun(ctx context.Context, args []string, secretKeyVal
 	runEarthlyMount := llb.AddMount("/run/earthly", llb.Scratch(),
 		llb.HostBind(), llb.SourcePath("/run/earthly"))
 	finalOpts = append(finalOpts, debuggerSecretMount, debuggerMount, runEarthlyMount)
+	if c.withSSH {
```
vladaionescu, line 730: Wondering if we should have this as a setting that needs to be enabled in the Earthfile, similarly to how buildkit requires [`--mount=type=ssh`](https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/experimental.md#run---mounttypessh).

Wondering how this might translate to the cloud runner eg `--mount=type=ssh,id=+secrets/<org>/path/ssh-key`.. or something like that.

cmd/earth/main.go
alexcb: great catch, this has been fixed now.

earthfile2llb/converter.go
alexcb: That's a good question.

So if a user has the following in their Earthfile
`RUN --ssh <cmd that needs ssh>` we could proxy through to the specified (or default) ssh socket agent; but obviously that won't work for the cloud runner.

So then if we allow a value to be specified like:
`RUN --ssh=+secrets/<org>/path/ssh-key <cmd that needs ssh>` then it would grab the ssh key instead.

and then if there's a basic RUN command without an ssh flag, then the ssh socket won't be made available to the command?

vladaionescu: Yes, something along those lines.

(BTW, adding a new flag to RUN is easy: https://github.com/earthly/earthly/blob/master/earthfile2llb/listener.go#L266 )

> and then if there's a basic RUN command without an ssh flag, then the ssh socket won't be made available to the command?

I would think so.

alexcb: I added in a `--ssh` flag, and down the road that'll give us the flexibility to hook it into the secrets.

earthfile2llb/earthfile2llb.go
```diff
@@ -40,6 +40,8 @@ type ConvertOpt struct {
 	VarCollection *variables.Collection
 	// A cache for image solves. depTargetInputHash -> context containing image.tar.
 	SolveCache map[string]llb.State
+	// WithSSH will pass along the ssh auth agent socket when true
+	WithSSH bool
```
vladaionescu, line 44: This is no longer necessary, right?

vladaionescu (approved):
One small comment remaining.

earthfile2llb/earthfile2llb.go
alexcb: good eyes, I missed that one. it's gone now.

vladaionescu (approved):

PR #298, opened 2020-09-11
./earth command should check that ~/bin exists

alexcb:
* ./earth command will fail if ~/bin doesn't exist; this will now print
an error message if it doesn't exist.

* remove earthly_binary container if it is already running.

### DIFF
--- a/earth
+++ b/earth
@@ -1,9 +1,18 @@
 #!/bin/bash
 set -e
 
+bindir="$HOME/bin"
+
+if [ ! -d "$bindir" ]; then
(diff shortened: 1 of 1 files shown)


agbell (approved):

vladaionescu:
BTW we don't really need to use a bin. An alternative might be a random /tmp location? Or even `~/.earthly/earth-prerelease`.

alexcb:
I like the idea of using `~/.earthly/earth-prerelease`, I'll change that.

PR #303, opened 2020-09-15
fix earth failure when ssh agent sock doesnt exist

alexcb:
* fixes issue where "ssh agent provider: stat /run/host-services/ssh-auth.sock: no such file or directory"
error occurs when trying to access the ssh-auth.sock during CI testing (or
cases where the user doesnt have an ssh agent running)

### DIFF
--- a/cmd/earth/main.go
+++ b/cmd/earth/main.go
@@ -1015,10 +1015,19 @@ func processSecrets(secrets []string, dotEnvMap map[string]string) (map[string][
 }
 
 func defaultSSHAuthSock() string {
+	var sshPath string
 	if runtime.GOOS == "darwin" {
-		return "/run/host-services/ssh-auth.sock"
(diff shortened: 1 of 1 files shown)


vladaionescu (approved):

vladaionescu (changes requested):
Oh wait... this might not work as intended. That file doesn't exist on the mac, but rather within the linux VM created by the docker app. So the existence check will always fail....

vladaionescu:
> That file doesn't exist on the mac, but rather within the linux VM created by the docker app.

I now realize that this might also affect the logic of the `RUN --ssh` feature.

alexcb:
oh I see, because we need to check if that exists on the inside of the docker engine. Hmm, any other ways we can make a guess if it exists or not?

vladaionescu (approved):

PR #309, opened 2020-09-15
add docs for RUN --ssh flag

alexcb:


docs/earthfile/earthfile.md
```diff
@@ -185,6 +185,21 @@ release:
...
+
+Here is an example:
+
+```Dockerfile
+git-clone:
+    RUN --ssh git clone git@host.com:user/private-repo.git
+```
+
+```bash
+earth --secret GH_TOKEN="the-actual-secret-token-value" +release
```
vladaionescu, line 200: This line seems unrelated?

docs/earthfile/earthfile.md
```diff
@@ -185,6 +185,21 @@ release:
 earth --secret GH_TOKEN="the-actual-secret-token-value" +release
 ```
 
+##### `--ssh`
```
vladaionescu, line 188: We should also add `[--ssh]` in the synopsis of `RUN` above.

docs/earthfile/earthfile.md
alexcb: ekk, that accidentally snuck in.

docs/earthfile/earthfile.md
```diff
@@ -181,6 +181,17 @@ release:
...
+
+Allows a command to access the ssh authentication client running on the host via the socket which is referenced by the environment variable `SSH_AUTH_SOCK`.
+
+Here is an example:
+
+```Dockerfile
+git-clone:
+    RUN --ssh git clone git@host.com:user/private-repo.git
```
vladaionescu, line 192: I just realized that we give an example of an operation that we generally recommend to be done via our own `GIT CLONE` command. Perhaps replace with `go mod download`?

```suggestion
    RUN --ssh go mod download
```

vladaionescu (approved):

docs/earthfile/earthfile.md
alexcb: that's a really good point. If we want to use go, then we might also have to change it to:
`RUN --ssh git config --global url."git@github.com:".insteadOf "https://github.com/" && go mod download`

vladaionescu: SGTM

vladaionescu (approved):

PR #310, opened 2020-09-16
initial debugging guide

alexcb:


docs/guides/debugging.md
```diff
@@ -0,0 +1,226 @@
+# Debugging techniques
+
+Errors will happen, be it during builds, unit tests, integration testing, or somewhere else in your pipeline. Earthly provides an interactive debugger to aid debugging command failures.
+
+Debugging an error can be challenging during a traditional image build, since the container state disappears as soon as the build error occurs. A developer may then inject various print commands through out the build to help reason about the state of the system before the failure occurs.
```
vladaionescu, line 5: I would add another sentence saying something along the lines of... Earthly has an interactive mode, which addresses this problem as we will show in this guide.

(Just so the reader doesn't get the impression that there's no solution to this problem and stops reading)

docs/guides/debugging.md
```diff
@@ -0,0 +1,226 @@
...
++build | Traceback (most recent call last):
++build |   File "generate_phrase.py", line 3, in <module>
++build |     text = open('sherlock.txt').read()
++build | FileNotFoundError: [Errno 2] No such file or directory: 'sherlock.txt'
++build | Command /bin/sh -c python3 generate_phrase.py failed with exit code 1
++build | Entering interactive debugger (**Warning: only a single debugger per host is supported**)
++build | root@buildkitsandbox:/code#
+```
+
+This time rather than exiting, earth will drop us into an interactive root shell, where we can type in arbitrary commands to figure out the problem:
```
vladaionescu, line 62: I would add here "into an interactive root shell, within the container of build environment" - to clarify that this shell is within the build itself.

(We used the term "build environment" in other places to describe this place)

docs/guides/debugging.md
```diff
@@ -0,0 +1,226 @@
...
++test | Hello, world!
++test | Target +test built successfully
+=========================== SUCCESS ===========================
+```
+
+With the use of the interactive debugger; we were able to examine the state of the embeded containerized 
+
+## Final tips
+
+If you ever want to jump into an interactive debugging session at any point in your Earthfile, you can simply add a command that will fail such as:
```
vladaionescu, line 218: We should also remind them to use `-i`, besides inserting `RUN false` - just in case they scan this guide and miss that they need to add that flag.

docs/guides/debugging.md
```diff
@@ -0,0 +1,231 @@
...
+```
+
+and our python code:
+```Python
+# generate_phrase.py
+
+import markovify
+text = open('sherlock.txt').read()
+text_model = markovify.Text(text)
+print(text_model.make_sentence())
```
agbell, line 32: Fun example :)

docs/guides/debugging.md
```diff
@@ -0,0 +1,231 @@
...
+```
+
+
+Now we can run it with `earth +test`, and we'll see a failure has occured:
```
agbell, line 36: this target was called `build` above

agbell (changes requested):

docs/guides/debugging.md
alexcb: great catch.

docs/guides/debugging.md
alexcb: good point, that's been added now: `and run earth with the `--interactive` (or `-i`) flag.`

agbell (approved):

PR #313, opened 2020-09-18
add hidden flag to print out completion entry so it can be referenced…

alexcb:
… via homebrew install

### DIFF
--- a/cmd/earth/main.go
+++ b/cmd/earth/main.go
@@ -79,6 +79,7 @@ type cliFlags struct {
 	gitPasswordOverride  string
 	interactiveDebugging bool
 	sshAuthSock          string
+	homebrewSource       string
 }
 
 var (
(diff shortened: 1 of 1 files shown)


cmd/earth/main.go
```diff
@@ -716,6 +728,15 @@ func (app *earthApp) run(ctx context.Context, args []string) int {
 
 func (app *earthApp) actionBootstrap(c *cli.Context) error {
 
+	switch app.homebrewSource {
+	case "bash":
+		fmt.Printf(bashCompleteEntry)
+		os.Exit(0)
```
vladaionescu, line 734: Can we just return `nil` here so that os.Exit is only used in main?

vladaionescu (approved):

cmd/earth/main.go
alexcb: that's a good point; for some reason I was worried it would keep running but that's not the case. I'll return nil.

PR #316, opened 2020-09-21
First version of Docker-in-Earthly guide. Promote WITH DOCKER to beta.

vladaionescu:


docs/guides/docker-in-earthly.md
```diff
@@ -0,0 +1,129 @@
+
+# Using Docker in Earthly
+
+This guide walks through using Docker commands in Earthly as well as some possible use-cases, such as integration testing.
+
+## Basic usage
+
+In order to use Docker commands (such as `docker run`), Earthly makes available isolated Docker daemons which are started and stopped on-demand. The reason for using isolated instances of Docker daemons is such that no pre-existing Docker state (eg images, containers, networks, volumes) can influence the way the build executes, thus having no way to create inconsistencies when the same build is run in a different environment.
```
alexcb, line 8:
```suggestion
In order to use Docker commands (such as `docker run`), Earthly makes available isolated Docker daemons which are started and stopped on-demand. The reason for using isolated instances of Docker daemons is such that no pre-existing Docker state (e.g. images, containers, networks, volumes) can influence the way the build executes, thus having no way to create inconsistencies when the same build is run in a different environment.
```

docs/guides/docker-in-earthly.md
```diff
@@ -0,0 +1,129 @@
...
+        DOCKER PULL hello-world
+        RUN docker run hello-world
+    END
+```
+
+Let's break it down.
+
+`FROM docker:19.03.13-dind` inherits from an official docker-in-docker (dind) image. This is mandatory, because `WITH DOCKER` requires all the Docker binaries (not just the client) to be present in the build environment.
+
+`WITH DOCKER ... END` starts a Docker daemon for the purpose of running Docker commands against it. At the end of the execution, this also terminates the daemon and permanently deletes all of its data (eg daemon cached images).
```
alexcb, line 25:
```suggestion
`WITH DOCKER ... END` starts a Docker daemon for the purpose of running Docker commands against it. At the end of the execution, this also terminates the daemon and permanently deletes all of its data (e.g. daemon cached images).
```

docs/guides/docker-in-earthly.md
```diff
@@ -0,0 +1,129 @@
...
+  END
+  ```
+* `docker build` cannot be used to build Dockerfiles. However, the Earthly command `FROM DOCKERFILE` can be used instead. See [alternative to docker build](#alternative-to-docker-build) below.
+* The state of the Docker daemon within Earthly cannot be inspected on the host (eg for debugging purposes). For example, if a `docker-compose` stack fails, you cannot execute commands like `docker-compose logs` or `docker logs` on the host. However, you may use the interactive mode to drop into a shell within the build environment and execute such commands there. For more information, see the [debugging guide](./debugging.md).
+
+## Alternatives to Docker in Earthly
+
+It is not always necessary to execute docker commands within an Earthly build. Certain operations can be replicated with Earthly constructs.
+
+### Alternative to docker run
```
alexcb, line 89: nice section :+1:

alexcb (approved):
looks great, just a minor nitpick.

agbell (approved):

PR #319, opened 2020-09-22
Increase timeout for buildkit wrapper start

vladaionescu:
Needed for windows.


alexcb (approved):
maybe 3600 to be safe? :laughing:

(just joking, looks good).

Issue #324, opened 2020-09-23
Proposal: WITH DOCKER-COMPOSE


vladaionescu:
### Background

A common pattern of integration testing and e2e testing is to have something like

```Dockerfile
RUN apk add docker-compose
WITH DOCKER
  DOCKER PULL img1
  DOCKER PULL img2
  DOCKER PULL img3
  ...
  DOCKER LOAD +custom-img img13
  RUN docker-compose up -d && \
    (perform testing) && \
    docker-compose down
END
```

### Proposal

The proposal is to simplify this as

```Dockerfile
WITH DOCKER-COMPOSE
  DOCKER LOAD +custom-img img13
  RUN (perform testing)
END
```

Behind the scenes, this command ensures the following:

* docker-compose utility is available
* `DOCKER PULL` is performed for all the images specified in the compose spec AND not specified in any `DOCKER LOAD` line
* The run command is wrapped by `docker-compose up -d` and `docker-compose down`

### Extra options

The following additional options could be specified:

```
WITH DOCKER-COMPOSE [-f compose-file.yml...] [service-name...]
  ...
END
```

* `-f ...` works the same as `docker-compose -f ...`, including having the ability to string together multiple specs.
* `service-name` works the same as the `docker-compose up service-name`, including being able to list more than one service

### Benefits

The core benefit is reducing the brittleness of the `DOCKER PULL` statements. If an image tag is changed in a docker-compose.yml, the user might forget to update the `DOCKER PULL` statement, causing inefficiencies that are not easy to spot.

The other benefits are secondary.

### Alternative proposal

Another option might be to only change the `DOCKER PULL` command to accept a docker-compose specification. eg

```Dockerfile
WITH DOCKER
  DOCKER PULL --compose docker-compose.yml
  DOCKER LOAD +custom-img img13
  RUN docker-compose up -d && \
    (perform testing) && \
    docker-compose down
END
```

Although a little more verbose, this has the benefit of reusing more of the existing syntax, making it less surprising (magical) to the user as to what exactly takes place behind the scenes.

agbell:
I support the proposal. 

It removes the need not just for the `DOCKER PULL` but also for starting and stopping docker-compose.

alexcb:
I like the idea; I'm unsure about the connection between the docker-compose and docker load spanning multiple lines.

What if we could combine them with:

```
WITH DOCKER-COMPOSE --image img13=+custom-img
    RUN ...
END
```

[...]

PR #335, opened 2020-09-28
Implement WITH DOCKER --compose

vladaionescu:
Fixes https://github.com/earthly/earthly/issues/324

This also fixes two bugs, which were tangential:
* In certain circumstances `FROM DOCKERFILE` would cause a nil deref panic.
* `ARG` before `WITH DOCKER` caused a cryptic failure

There are still a few limitations:
* Ports exposed to the host can conflict with each other, causing issues with parallelism. We need to find a way to isolate these to the specific targets they run in.
* `WITH DOCKER --compose` installs docker-compose automatically as a separate layer from the RUN execution. However, if there are COPY's before the WITH DOCKER command, there will be noticeable re-installation of docker-compose every time those files change. A possible future solution for this is to require a specific Earthly-provided dind image, in which we put in all the necessary dependencies, including docker-compose.

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -189,7 +189,6 @@ test:
 test-all:
     BUILD +examples
     BUILD +test
-    BUILD ./examples/tests+experimental
 
 examples:
     BUILD ./examples/go+docker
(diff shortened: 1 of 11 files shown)


alexcb (approved):
looks great.

PR #336, opened 2020-09-29
Fix pegged CPU bug

alexcb:
* fix bug where close() was being called continually
 * add --profiler flag to enable http-based CPU profiler

### DIFF
--- a/builder/solver.go
+++ b/builder/solver.go
@@ -138,12 +138,10 @@ func (s *solver) solveDockerTar(ctx context.Context, localDirs map[string]string
 		return nil
 	})
 	go func() {
-		for {
-			select {
-			case <-ctx.Done():
-				// Close read pipe on cancels, otherwise the whole thing hangs.
-				pipeR.Close()
-			}
+		select {
+		case <-ctx.Done():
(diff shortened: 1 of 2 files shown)


alexcb:
This allows us to run `earth --profiler ...`, then in another window we can run `go tool pprof http://localhost:6060/debug/pprof/profile?seconds=15` followed by `web .` to generate a summary of the CPU usage.

builder/solver.go
```diff
@@ -143,6 +143,7 @@ func (s *solver) solveDockerTar(ctx context.Context, localDirs map[string]string
 			case <-ctx.Done():
 				// Close read pipe on cancels, otherwise the whole thing hangs.
 				pipeR.Close()
+				return
```
vladaionescu, line 146: BTW, do we even need the `for` loop? The select basically waits for ctx.Done() indefinitely.

vladaionescu (approved):

builder/solver.go
alexcb: hah, that's a great point! we probably dont. I'll simplify this.

PR #349, opened 2020-09-30
feat(autocomplete): implemented functionality to autocomplete all the…

avasapollo:
… users


alexcb:
Looks like a good start, I just gave it a test and it was able to complete `earth ~` into the available usernames; however when I tried typing `earth ~a<tab><tab>`, it didn't autocomplete into `earth ~alex`. I recorded a little video showing what I saw: https://asciinema.org/a/jD1YU3uoZEVOPM714h2t9lghD

.gitignore
```diff
@@ -3,4 +3,5 @@
 *.tokens
 *.interp
 .antlr/
-.env
\ No newline at end of file
+.env
+.idea/
```
vladaionescu, line 7: We recommend that devs add this to their personal .gitignore, rather than the project's.

autocomplete/complete.go
```diff
@@ -94,27 +96,69 @@ func getPotentialTarget(prefix string) ([]string, error) {
...
-	if prefix == "." {
-		return []string{"./", "../"}, nil
+func lookupUsers(prefix string) ([]string, error) {
+	u, err := user.Current()
+	if err != nil {
+		return nil, err
 	}
-	currentUser, err := user.Current()
+
+	lPath := strings.Split(u.HomeDir, "/")
```
vladaionescu, line 105: Wonder if we should use `filepath.Separator` for all the `"/"`

avasapollo: the filepath.Separator is an int32, do you prefer to have a var to replace the "/"? something like this 
```
var pathSeparator = fmt.Sprintf("%c", os.PathSeparator)

```

.gitignore
avasapollo: yes right

avasapollo (commented):
I add the check if ~a<tab><tab> we get
andrea

alexcb:
That's looking better, I just gave it another test and the user expansion is working, but it doesn't continue to expand into the user's directory. Here's a video of my latest test: https://asciinema.org/a/NE31c94OmuCGYR8SFVN5PUbnk

keep it up :)

autocomplete/complete.go
vladaionescu: Oh, interesting. Didn't realize it's int32.

I think this should be fine as it is. We anyway don't support auto-complete on non-unix systems.

avasapollo (commented):
hey @alexcb, I found out how to test I ll try today or tomorrow thanks

avasapollo (commented):
hey @alexcb I think we are close, can you take a look please? let me know if it is missing something

alexcb:
It looks like it's autocompleting directories now, but there's an issue where it didn't correctly autocomplete the targets under the root Earthfile. I recorded another screen capture under https://asciinema.org/a/eH6COSKcmV988OULBFG2Yqv6H

Here's some paths you can try during testing, assuming that you have earthly checked out to `~/path/to/earthly`:

```
earth /home/avasapollo/path/to/earthly+for-linux
earth ~/path/to/earthly+for-linux
earth ~avasapollo/path/to/earthly+for-linux
earth ~avasapollo/path/to/earthly/examples/go+build
```

additionally, one should still be able to use the relative expansion like:
```
cd ~/path/to/earthly
earth +for-linux
earth ./examples/go+build
earth ./examples/tests/dockerfile+test
```

Thanks for keeping at it, it looks like you're getting close :)

avasapollo (commented):
ciao @alexcb, I pushed another commit I worked on the getPotentialTarget function that was connected to the current user.
now I am getting this result, is it correct?
Screenshot from 2020-10-05 17-01-30

alexcb:
The targets in the screenshot look good, are you able to access the same targets via `~andreavasapollo/hactoberfest/earthly+`?

avasapollo:
I tried +for-l and I got +for-linux

avasapollo:
@alexcb ok I fixed and I tested with both :)
Screenshot from 2020-10-05 21-37-11

alexcb:
looks like it's getting closer, but there's still a few places where tab-completing isn't working; checkout the latest video in https://asciinema.org/a/gHdo2Hr4peDrvXe0NKrn4mipA

for instance, `~/` isn't working anymore, and when I try `~alex/testing/avasapollo/earthly` and hit tab-tab, it doesn't show `~alex/testing/avasapollo/earthly+` as an option, but instead just tries to immediately jump into `~alex/testing/avasapollo/earthly/`

avasapollo:
Ok I ll check tomorrow just to understand better,
-when you write '~/'  what do you expect? Do you expect the current home user path?
- for this ~alex/testing/avasapollo/earthly should autocomplet ~alex/testing/avasapollo/earthly because the suffix is earthly?

alexcb:
`~/` should be equivalent to `~username/` (just like how it works when you do `ls ~` or `ls ~username` (or `ls ~/some/path` or `ls ~username/some/path`).

the `+` autocompletion should appear whenever there's an Earthfile in the directory that it's referencing. this `+` represents a target and is covered in our target reference guide under https://docs.earthly.dev/guides/target-ref

avasapollo:
hey @alexcb, I did some fix. I tested the functionalities that you mentioned and they seems ok, sorry If I am wasting your time with these tests

alexcb:
It's no waste of my time at all, thank YOU for helping us out. it's looking really good.

Is there a way we can have the existing functionality of `~/` stay as is? in the master version if I type `~/pat<tab><tab>` it completes to `~/path`; however in this version now it's replacing it from `~/pat<tab><tab>` to `~username/path`. The autocompletion functionality should never change what the user has already typed, sorry if I was unclear with that in my earlier message.

Ideally we want all tab completion for anything related to a file path to be comparable to the way regular bash commands like cd, ls, etc work. here's a brief example of using ls: https://asciinema.org/a/IUERZxlaZiBxdACFKzRChVFDh

avasapollo:
@alexcb ok I think now the ~/ works fine, let me know what you think about that

autocomplete/complete.go
```diff
@@ -94,64 +104,188 @@ func getPotentialTarget(prefix string) ([]string, error) {
...
+	return user.Lookup(username)
+}
+
+func getUsers() ([]string, error) {
+	u, err := user.Current()
+	if err != nil {
+		return nil, err
+	}
+
+	lPath := strings.Split(u.HomeDir, "/")
```
alexcb, line 118: Instead of splitting this, would it be possible to use `filepath.Dir(u.HomeDir)` from the `path/filepath` package instead?

autocomplete/complete.go
```diff
@@ -94,64 +104,188 @@ func getPotentialTarget(prefix string) ([]string, error) {
...
+		if strings.HasPrefix(s, path.Join(p, filter)) {
+			res = append(res, s)
 		}
-		return "~/" + s[expandedHomeLen:]
 	}
+	return res, nil
+}
 
-	files, err := ioutil.ReadDir(dir)
+func printDirectory(prefix string) ([]string, error) {
```
alexcb, line 234: nitpick, this function doesn't print anything; perhaps it should be named something like `listDirectory` or `getPotentialDirectoryCompletions` if you're not into the whole brevity thing?

alexcb (changes requested):
I gave it a test and it looks like all the paths I could think of testing are working great. This is looking amazing! :muscle: :tada: 

I left a couple very minor suggestions in the code, very nice work, thanks you very much for the contribution.

alexcb (changes requested):
.

autocomplete/complete.go
avasapollo: done ;)

autocomplete/complete.go
avasapollo: yes make sense I changed to listDirectory

avasapollo:
@alexcb, nice I made the changes that you have requested

alexcb (approved):

alexcb:
@avasapollo woohoo! thanks again for your contribution :raised_hands:

avasapollo:
@alexcb it was a pleasure happy to help 😃

agbell:
Hi @avasapollo!  Thanks for your contributions.  Would you like us to send you an earthly sticker?  If so please send me your address to adam@earthly.dev

avasapollo:
@agbell I sent you an email with my address thank you very much

PR #354, opened 2020-10-01
Invalid env key

JosemyDuarte:
This PR tries to fix issue #69.

Following @agbell advice I'm starting from PR #71 and will apply all the suggestions given. I'm opening this PR to kick off the discussion.


earthfile2llb/listener.go
```diff
@@ -891,7 +891,12 @@ func (l *listener) EnterEnvArgKey(c *parser.EnvArgKeyContext) {
 	if l.shouldSkip() {
 		return
 	}
-	l.envArgKey = c.GetText()
+	var envKey = c.GetText()
+	var matchedEnvKey = checkShellName(envKey)
+	if matchedEnvKey != envKey {
+		l.err = fmt.Errorf("Invalid env key definition %s", envKey)
```
vladaionescu, line 897: There should be a return just after this, so that the parsing stops.

earthfile2llb/listener.go
```diff
@@ -962,6 +967,12 @@ func (ssf *StringSliceFlag) Set(arg string) error {
 	return nil
 }
 
+var shellNameStringRegexp = regexp.MustCompile("^[a-zA-Z_]+[a-zA-Z0-9_]*")
+
+func checkShellName(str string) string {
```
vladaionescu, line 972: This function would be better suited as returning an error if something is wrong with the env var passed.

Also, I would call it `checkEnvVarName`, as it's not really related to the shell.

earthfile2llb/listener.go
vladaionescu, line 897: The error string should start with lower-case for consistency.

vladaionescu (commented):
Hi @JosemyDuarte - thanks for taking this on 🙌 

I left some comments.

alexcb:
I just gave it a quick test and functionality wise it's looking good:

```
alex@mah:~/testing/JosemyDuarte/earthly$ ./build/linux/amd64/earth ~/broken/envtest+protobuf 
buildkitd | ...Done
Error: parse: apply implicit FROM +base: apply build +base: earthfile2llb for +base: parse: Invalid env key definition ${CODE_SRC}
```

earthfile2llb/listener.go
JosemyDuarte: With "something is wrong with the env var passed" you mean if it doesn't match against the regex, right? Or you mean something else?

vladaionescu: Yes - exactly

JosemyDuarte: I added the return of error, but I am not sure about what I did. From [documentation](https://golang.org/pkg/regexp/#Regexp.FindString), I got that `FindString` returns an empty string if it doesn't match. That's the reasoning behind my change. Let me know your thoughts and if you want me to change something.

alexcb: would it be possible to change `checkEnvVarName(str string) (string, error)` to only return an error? right now it's returning a string that's not being modified or used by the caller.

then it'll simplify the calling code to:
```
	var err = checkEnvVarName(envKey)
	if err != nil {
		l.err = err
		return
	}
	l.envArgKey = envKey
```

JosemyDuarte: Of course! Consider it done.

earthfile2llb/listener.go
```diff
@@ -962,6 +968,16 @@ func (ssf *StringSliceFlag) Set(arg string) error {
 	return nil
 }
 
+var shellNameStringRegexp = regexp.MustCompile("^[a-zA-Z_]+[a-zA-Z0-9_]*")
+
+func checkEnvVarName(str string) error {
+	matchedString := shellNameStringRegexp.FindString(str)
```
vladaionescu, line 974: Should rather use [Match](https://golang.org/pkg/regexp/#Regexp.Match) for this

earthfile2llb/listener.go
```diff
@@ -962,6 +968,16 @@ func (ssf *StringSliceFlag) Set(arg string) error {
 	return nil
 }
 
+var shellNameStringRegexp = regexp.MustCompile("^[a-zA-Z_]+[a-zA-Z0-9_]*")
```
vladaionescu, line 971:
```suggestion
var shellNameStringRegexp = regexp.MustCompile("^[a-zA-Z_]+[a-zA-Z0-9_]*$")
```

Make sure we match on the entire string only.

earthfile2llb/listener.go
```diff
@@ -891,7 +891,13 @@ func (l *listener) EnterEnvArgKey(c *parser.EnvArgKeyContext) {
...
 		return
 	}
-	l.envArgKey = c.GetText()
+	var envKey = c.GetText()
+	var err = checkEnvVarName(envKey)
+	if err != nil {
+		l.err = err
+		return
+	}
+	l.envArgKey = envKey
```
vladaionescu, line 900:
```suggestion
	l.envArgKey = c.GetText()
	err := checkEnvVarName(l.envArgKey)
	if err != nil {
		l.err = err
		return
	}
```

Simplification

vladaionescu (commented):
Looks good - almost there.

earthfile2llb/listener.go
JosemyDuarte: Any reason to use Match instead of [MatchString](https://golang.org/pkg/regexp/#Regexp.MatchString)?

earthfile2llb/listener.go
```diff
@@ -962,6 +967,16 @@ func (ssf *StringSliceFlag) Set(arg string) error {
 	return nil
 }
 
+var shellNameStringRegexp = regexp.MustCompile("^[a-zA-Z_]+[a-zA-Z0-9_]*$")
+
+func checkEnvVarName(str string) error {
+	itMatch := shellNameStringRegexp.MatchString(str)
```
vladaionescu, line 973:
```suggestion
var envVarNameRegexp = regexp.MustCompile("^[a-zA-Z_]+[a-zA-Z0-9_]*$")

func checkEnvVarName(str string) error {
	itMatch := envVarNameRegexp.MatchString(str)
```

vladaionescu (commented):
One last nit! Thanks for being patient :-)

earthfile2llb/listener.go
vladaionescu: You are right indeed. MatchString is correct.

vladaionescu (approved):

vladaionescu:
Thank you for the contribution @JosemyDuarte! 🙌

JosemyDuarte:
Thanks for the help and for building such a great tool!

agbell:
Hi @JosemyDuarte !  Thanks for your contributions.  Would you like us to send you an earthly sticker?  If so please send me your address to adam@earthly.dev

PR #362, opened 2020-10-01
Add Elixir example

dchw:


vladaionescu (approved):
Woohoo! Looks good! 🎉

vladaionescu:
Oh, can you also add the example to this list? https://github.com/earthly/earthly/blob/master/Earthfile#L193

It'll be covered by CI that way.

vladaionescu (approved):

Issue #341, opened 2020-09-29
Add example rust application to examples


agbell:
We need an example rust application added to the examples folder:
https://github.com/earthly/earthly/tree/master/examples

An example Earthfile should also be included which can build and produce a docker file of the example. 

The [Go example](https://github.com/earthly/earthly/tree/master/examples/go) can serve as a reference point for how to structure an example.

vbhv007:
Hey, Can I pick this up under hacktoberfest?

agbell:
Hey @vbhv007, you certainly can.

agarwalBinay18:
can I work on this issue?

agbell:
You can @agarwalBinay18

vbhv007:
Hey @agbell  
I tried running the python example and I'm getting this error

`Error: new file stream build.earth: open build.earth: no such file or directory`

Is this some known issue, I couldn't find anything on the internet

agbell:
Hi @vbhv007, that is strange.    What command are you running?  Try earth +build in the python folder. It runs as part of the build, so I believe it should be working.   Also, make sure you have the latest version of earthly.

vbhv007:
Yup, I think it was a version issue, working with v0.3.8

kevinkjt2000:
Is this done already?

agbell:
Hey @kevinkjt2000, I believe it is. Good catch.

PR #365, opened 2020-10-02
added example rust app

vbhv007:
PR for the issue https://github.com/earthly/earthly/issues/341

Added sample rust application. Testing steps

```
earth +build

earth +docker

docker run rust-earthly
```

### DIFF
--- a/examples/rust/Earthfile
+++ b/examples/rust/Earthfile
@@ -1,0 +1,13 @@
+FROM rust:1.24.0
+
+WORKDIR /usr/src/app
+
+build:
+    COPY . .
+    RUN cargo install
+    SAVE ARTIFACT /usr/src/app /usr/src/app
+
+docker:
+    COPY +build/usr/src/app /usr/src/app
+    CMD ["./target/release/rust-earthly"]
+    SAVE IMAGE rust-earthly:latest
--- a/Earthfile
+++ b/Earthfile
@@ -208,6 +208,7 @@ examples:
     BUILD ./examples/readme/go3+build
     BUILD ./examples/readme/proto+docker
     BUILD github.com/earthly/hello-world+hello
+    BUILD ./example/rust+docker
 
 test-fail:
     RUN false
--- a/Earthfile
+++ b/Earthfile
@@ -208,7 +208,7 @@ examples:
     BUILD ./examples/readme/go3+build
     BUILD ./examples/readme/proto+docker
     BUILD github.com/earthly/hello-world+hello
-    BUILD ./example/rust+docker
+    BUILD ./examples/rust+docker
 
 test-fail:
     RUN false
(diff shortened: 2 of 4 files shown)


agbell (changes requested):
Great work!  Can you add this to the examples builds https://github.com/earthly/earthly/blob/master/Earthfile#L193

vbhv007:
Sure, will add

examples/rust/Earthfile
```diff
@@ -0,0 +1,13 @@
+FROM rust:1.24.0
+
+WORKDIR /usr/src/app
+
+build:
+    COPY . .
+    RUN cargo install
+    SAVE ARTIFACT /usr/src/app /usr/src/app
```
vladaionescu, line 8: You should only save the binary here, not the entire directory with all the sources.

examples/rust/Earthfile
```diff
@@ -0,0 +1,13 @@
+FROM rust:1.24.0
+
+WORKDIR /usr/src/app
```
vladaionescu, line 3: We should use `/example-rust` for simplicity and consistency with other examples.

examples/rust/Earthfile
```diff
@@ -0,0 +1,13 @@
...
+
+build:
+    COPY . .
+    RUN cargo install
```
vladaionescu, line 7: Can the downloading of dependencies be moved to a separate target? (eg `deps`)

It helps show typical usage of Earthly with a real app, where the Earthly doesn't need to download all dependencies on each code change.

examples/rust/Earthfile
```diff
@@ -0,0 +1,13 @@
...
+    SAVE ARTIFACT /usr/src/app /usr/src/app
+
+docker:
+    COPY +build/usr/src/app /usr/src/app
+    CMD ["./target/release/rust-earthly"]
```
vladaionescu, line 12: It would be better to use `ENTRYPOINT` for this - it's not clear what the entrypoint is for the base image (if an Earthly user tries using a different base image, it might not work).

examples/rust/Earthfile
```diff
@@ -0,0 +1,13 @@
+FROM rust:1.24.0
+
+WORKDIR /usr/src/app
+
+build:
+    COPY . .
```
vladaionescu, line 6: We generally copy files one by one in the examples to make it clearer for the reader.

vladaionescu (changes requested):
Hi @vbhv007 - thank you for the contribution! I left some comments.

agbell:
Hi @vbhv007 ,
Can you pull in and use a rust package to show how that would work. You could use https://github.com/BurntSushi/termcolor. to print out "hello world" in colors perhaps?

alexcb:

PR #368, opened 2020-10-03
Implement TTY Progress bar

### LINKED ISSUE #84: Improve logs (introduce a progress-bar for tty devices)
Running the Go example, earth floods the terminal with lines like this:

```
+base | sha256:61b11419476186415e86da00c53c938fd32dc5904a889e1c3d40e238d69d168e 11%
+base | sha256:61b11419476186415e86da00c53c938fd32dc5904a889e1c3d40e238d69d168e 11%
+base | sha256:61b11419476186415e86da00c53c938fd32dc5904a889e1c3d40e238d69d168e 12%
+base | sha256:61b11419476186415e86da00c53c938fd32dc5904a889e1c3d40e238d69d168e 12%
+base | sha256:61b11419476186415e86da00c53c938fd32dc5904a889e1c3d40e238d69d168e 12%
+base | sha256:61b11419476186415e86da00c53c938fd32dc5904a889e1c3d40e238d69d168e 12%
+base | sha256:61b11419476186415e86da00c53c938fd32dc5904a889e1c3d40e238d69d168e 12%
+base | sha256:61b11419476186415e86da00c53c938fd32dc5904a889e1c3d40e238d69d168e 12%
```
Though it is a lot more of that than I've included here. I would expect this as a progress bar replacing the line on update rather than drawing a new line.

`go build` is silent on success so this is a lot messier.

Gituser143:
Made a TTY progress bar to display progress in one bar instead of multiple lines of output.

Fixes #84

### DIFF
--- a/builder/solver.go
+++ b/builder/solver.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"os"
 	"os/exec"
+	"sync"
 
 	"github.com/earthly/earthly/earthfile2llb/image"
 	"github.com/earthly/earthly/llbutil"
@@ -28,6 +29,8 @@ type solver struct {
 	enttlmnts   []entitlements.Entitlement
 	remoteCache string
 }
+
+var printingMutex sync.Mutex
 
 func (s *solver) solveDocker(ctx context.Context, localDirs map[string]string, state llb.State, img *image.Image, dockerTag string, push bool) error {
 	dt, err := state.Marshal(ctx, llb.Platform(llbutil.TargetPlatform))
@@ -53,7 +56,7 @@ func (s *solver) solveDocker(ctx context.Context, localDirs map[string]string, s
 		return nil
 	})
 	eg.Go(func() error {
-		return s.sm.monitorProgress(ctx, ch)
+		return s.sm.monitorProgress(ctx, ch, &printingMutex)
 	})
 	eg.Go(func() error {
 		defer pipeR.Close()
@@ -111,7 +114,7 @@ func (s *solver) solveDockerTar(ctx context.Context, localDirs map[string]string
 		return nil
 	})
 	eg.Go(func() error {
-		return s.sm.monitorProgress(ctx, ch)
+		return s.sm.monitorProgress(ctx, ch, &printingMutex)
 	})
 	eg.Go(func() error {
 		file, err := os.Create(outFile)
@@ -174,7 +177,7 @@ func (s *solver) solveArtifacts(ctx context.Context, localDirs map[string]string
 		return nil
 	})
 	eg.Go(func() error {
-		return s.sm.monitorProgress(ctx, ch)
+		return s.sm.monitorProgress(ctx, ch, &printingMutex)
 	})
 	err = eg.Wait()
 	if err != nil {
@@ -228,7 +231,7 @@ func (s *solver) solveSideEffects(ctx context.Context, localDirs map[string]stri
 		return nil
 	})
 	eg.Go(func() error {
-		return s.sm.monitorProgress(ctx, ch)
+		return s.sm.monitorProgress(ctx, ch, &printingMutex)
 	})
 	err = eg.Wait()
 	if err != nil {
--- a/builder/solver.go
+++ b/builder/solver.go
@@ -7,7 +7,6 @@ import (
 	"io"
 	"os"
 	"os/exec"
-	"sync"
 
 	"github.com/earthly/earthly/earthfile2llb/image"
 	"github.com/earthly/earthly/llbutil"
@@ -29,8 +28,6 @@ type solver struct {
 	enttlmnts   []entitlements.Entitlement
 	remoteCache string
 }
-
-var printingMutex sync.Mutex
 
 func (s *solver) solveDocker(ctx context.Context, localDirs map[string]string, state llb.State, img *image.Image, dockerTag string, push bool) error {
 	dt, err := state.Marshal(ctx, llb.Platform(llbutil.TargetPlatform))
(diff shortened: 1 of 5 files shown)


Gituser143:
@vladaionescu I've made appropriate changes to create a progress bar for every activity in `ss.Statuses`. I noticed during the CI tests that the progress bars might get overwritten mid way. I was thinking I could use a `WaitGroup` or a `Mutex` to handle the issue of parallel printing between all functions that log output to the screen. 

Do you think this is a fair approach? I'm currently unable to think of any other solution.

Gituser143:
@vladaionescu I tried using a mutex to ensure only one function (in console logging) prints at a time but it seems like there are more places where text is printed on terminal. Could you tell me where they are so that I can try fixing that?

vladaionescu:
It's probably best that the mutex is kept within the console logger itself (like it was before). I suggest we insert the progress bar functionality within the console logger, so that the same mutex can be reused somehow.

Note that there may be an issue with creating too many progress bars. The way it works is that the monitor receives update messages and each of the update messages contain multiple progress bar updates. The way it is right now, it looks like a completely new progress bar is created on each update message (instead of updating potentially already existing progress bars).

It might be better to store progress bar state information in the `vertexMonitor` struct. That will help with correlating updates with previously created progress bars.

Gituser143:
@vladaionescu I made a receiver function called `printProgress` which uses the console logger mutex to print the progress bar. I have introduced `progressBar` as a variable of the `ConsoleLogger` struct. I tried this out locally with examples and it seems to work correctly. Let me know if this is fine or if I have to make any changes!

conslogging/conslogging.go
```diff
@@ -107,21 +113,20 @@ func (cl ConsoleLogger) WithFailed(isFailed bool) ConsoleLogger {
...
 	cl.color(successColor).Fprintf(cl.w, "=========================== SUCCESS ===========================\n")
+	cl.mu.Unlock()
 }
 
 // PrintFailure prints the failure message.
 func (cl ConsoleLogger) PrintFailure() {
 	cl.mu.Lock()
-	defer cl.mu.Unlock()
 	cl.color(warnColor).Fprintf(cl.w, "=========================== FAILURE ===========================\n")
+	cl.mu.Unlock()
```
vladaionescu, line 124: What's the reason for this change?

conslogging/conslogging.go
```diff
@@ -8,6 +8,8 @@ import (
 	"sync"
 	"unicode/utf8"
 
+	"github.com/cheggaaa/pb/v3"
+
```
vladaionescu, line 12: Nit: Unnecessary blank link

builder/solver.go
```diff
@@ -29,6 +30,8 @@ type solver struct {
 	remoteCache string
 }
 
+var printingMutex sync.Mutex
```
vladaionescu, line 33: This should no longer be necessary, right?

vladaionescu (commented):
Great - this is a little better. However, it still doesn't take into account if there are multiple vertices to update in parallel. There is only a single progress bar which could be jumping back and forth depending on what vertices are being updated. We still need to adapt this to take into account my previous suggestion.

conslogging/conslogging.go
Gituser143: This was me reverting from the attempt at using the global `printingMutex`. I'll revert these back to the defer calls. Sorry about this!

builder/solver.go
Gituser143: Yes, I seem t have forgotten to remove that! I'll clean these minor fixes by the next commit.

conslogging/conslogging.go
Gituser143: My bad, will clean by next commit.

Gituser143:
@vladaionescu 

> However, it still doesn't take into account if there are multiple vertices to update in parallel. There is only a single progress bar which could be jumping back and forth depending on what vertices are being updated.

What I noticed was each `vertexMonitor` has it's own console logger named as `console` in the vertex monitor struct. Thus multiple vertices will have their own progress bars right (each progress bar is associated with a console logger)? Am I missing something here?

vladaionescu:
Ah yes - there's a different console logger object for each vertex - that means that there is a different progress bar for each vertex as well. This should be good then. I'll figure out a way to test this. It seems that my internet connection is too fast and progress bar always shows as 100% :-) #firstworldproblems

Gituser143:
> It seems that my internet connection is too fast and progress bar always shows as 100% :-) #firstworldproblems

@vladaionescu I wish I could relate :P 

Do let me know if I have to make any changes...

vladaionescu:
Hey @Gituser143 - so we tested on a couple of machines but the progress bar is always directly 100%. Do you think you can make a video or gif or asciinema cast of how the progress bars work?

Gituser143:
@vladaionescu do you want this GIF to be a demonstration of just the bar or as an example in an earthly build/test? If it's the former, I hope this suffices. 

ezgif com-video-to-gif

I was unable to get an actual example from the builds/tests as for me too they are always directly at 100%. Do let me know if there's anything I else I can do for the same.

Gituser143:
@vladaionescu Is there anything I need to do?

vladaionescu:
We definitely need to validate that this works correctly. I have a suspicion that the progress bar would not work correctly if it didn't jump to 100% directly. I don't think the library is smart enough to go back in the console history and update the progress bar.

Perhaps you could test Earthly by simulating slow internet connection via something like https://wiki.linuxfoundation.org/networking/netem and show multiple progress bars being updated in parallel somehow.

Gituser143:
> I have a suspicion that the progress bar would not work correctly if it didn't jump to 100% directly.

@vladaionescu you might be right. I'll look into this.

Gituser143:
@vladaionescu I've run into a dead end with this. Any pointers?

alexcb:

Issue #357, opened 2020-10-01
C Example


adamgordonbell:
We need an example c application added to the examples folder:
https://github.com/earthly/earthly/tree/master/examples

An example Earthfile should also be included which can build and produce a docker file of the example.

The Go example can serve as a reference point for how to structure an example.

kzuri:
Hello @agbell, I would like to take up this issue. 😄

vladaionescu:
Sold

Gauravsaha-97:
I would like to work on this issue.

AnshMishra2001:
Can I work on this issue, pls?

alexcb:
Go for it @AnshMishra2001 , if it's possible we would really like to see an example that makes use of the cunit library for an example of performing unittesting of a simple function. A suggestion could be to write a program that implements a fibonacci function along with a second program that uses cunit to test that function.

sea-grass:
Since there hasn't been any activity on this issue in a while, is it okay if I pick this up?

vladaionescu:
Sure @sea-grass - feel free. Seems that many people offered but nobody has shown anything yet.

sea-grass:
Thanks! I just created a PR that adds a C example: https://github.com/earthly/earthly/pull/1235. Let me know if there's someone I should tag as a reviewer.

vladaionescu:
Awesome! Added @alexcb as reviewer.

PR #369, opened 2020-10-03
Added a C example according to the issue 'C Example #357'

Gauravsaha-97:


alexcb:
Thanks for the contribution, ideally we're looking for an example that would have two or more files that get linked together along with a library. We would then need a corresponding Earthfile to be written which would compile the program. Take a look at our existing c++ example: https://github.com/earthly/earthly/tree/master/examples/cpp

What would be really cool is if we could have two binaries build: one which is the main example, and a second binary which does some unit testing.

Gauravsaha-97:
I have tried to implement what you have mentioned before. I am not very confident as I am doing this type of task for the first time. I would be happy to rectify in case of any issue.

examples/c/src/main.c
```diff
@@ -0,0 +1,11 @@
+#include <stdio.h>
+
+#include "sum.h"
+
+int main(int argc, char *argv[]) {
+	int n1 = 100;
+    int n2 = 200;
```
alexcb, line 7: It looks like there's a mix between tabs and spaces in this file; can you change this to consistently use one or the other.

examples/c/src/sum.c
```diff
@@ -0,0 +1,7 @@
+#include "sum.h"
+
+int sum(int n1, int n2)
+{
```
alexcb, line 4: The style of this function is different from the style used in the main function (i.e. the main function has the curly brace on the same line as the function, but this one is on the second line). Can you pick a single style and use it consistently.

alexcb (changes requested):
Thanks for the changes, they are looking good, left a few nitpicky issues related to the consistently using the same code style.

Would you be able to also write a test for the `sum` function using the cunit library, which can be installed via `apt-get install libcunit1-dev`.

examples/c/src/main.c
Gauravsaha-97: Okay..I will fix them.

Gauravsaha-97:
@alexcb Added a new file which contains the unit tests.

alexcb:
This is looking great! Thanks for adding a sample unit test. How does one run the unit test? Can we add those steps to the Earthfile under a `test` target?

Gauravsaha-97:
@alexcb Added the testing procedures for all modes.

alexcb:
I'm not sure those additions to the Earthfile are going to work; did you test them by running `earth +test`?

Gauravsaha-97:
@alexcb I have added the necessary Header Files to the directory. If something is not right, can you please suggest some resources to learn about it?

alexcb:
https://docs.earthly.dev/guides/basics is a good reference to learn about earthly. It shows some basic hello world examples for different languages.

I don't think adding the cunit source code is needed; instead it should be possible to install the `libcunit1-dev` package via a `RUN` command in the Earthfile.

Issue #344, opened 2020-09-29
Ruby on rails example


agbell:
We need an example rails application added to the examples folder:
https://github.com/earthly/earthly/tree/master/examples

An example Earthfile should also be included which can build, test, and produce a docker file of the example. 

The [Go example](https://github.com/earthly/earthly/tree/master/examples/go) can serve as a reference point for how to structure an example.

RajivSah:
Can I take this?

agbell:
Hi @RajivSah, you certainly can.  Let me know if you have any questions or jump on our Gitter channel:
https://gitter.im/earthly-room/community

RajivSah:
Thanks.

RajivSah:
Wouldn't setting up new rails app in the example folder create multiple files as the new rails project contains a lot of files?

zedtux:
Well, for Python, there's no featured framework, why would it be for Ruby?

I guess @agbell meant Ruby, not Rails.

agbell:
Hi @zedtux and @RajivSah ,
I was thinking an actual ruby on rails example.  I also put a ticket in for a django example.  The problem we are trying to solve here is somebody has a rails project and is considering using early to run the tests and containerize it, we want to give them an example that they can work off of.  I think a small ruby example would also be useful.

zedtux:
@RajivSah will you do both, Ruby or Rails?

RajivSah:
The ruby example has been merged, I am working on rails as well.

Gauravsaha-97:
I would like to work on this issue.

RajivSah:
> I would like to work on this issue.

You can take it if you want.

Gauravsaha-97:
@RajivSah Created a PR

zedtux:
@Gauravsaha-97 I believe you messed up with your PR #371 as it contains the same code (The C example) than the PR #369.

Can you please double check?

Gauravsaha-97:
Actually, I messed up with the commit message. The code in the two PRs are
different. I dont know how to edit the commit message.

zedtux:
Okay, sorry, you're right. The PR #371 has both Ruby and C examples because its branch is a child of #369.

Have look at [Changing a commit message](https://docs.github.com/en/free-pro-team@latest/github/committing-changes-to-your-project/changing-a-commit-message) from Github's documentation.

Gauravsaha-97:
I have changed the commit message.

PR #371, opened 2020-10-03
Added a Ruby example according to the issue 'Ruby on rails example #344'

Gauravsaha-97:


examples/c/src/sum.c
```diff
@@ -0,0 +1,14 @@
...
+
+  printf("Enter two numbers to add: \n");
+  scanf("%d %d", &x, &y);
+
+  z = x + y;
+
+  printf("Sum of the numbers : %d\n", z);
+
+  return 0;
+}
```
vladaionescu, line 14: This seems unrelated?

Gauravsaha-97: @vladaionescu I have added 'prime_check.rb' file to the 'ruby' directory. This is the C file that you are viewing.

RajivSah:
There is already a ruby example in the repo.

agbell:
Hi @Gauravsaha-97, You have two commits here but only should be part of this PR.  Please remove 6e5621572d70747f7b3ca5d07b653add5286b35e. You could do this with git rebase -i, or just start new branch off master and reintroduce your changes.  Also we are looking for a more fully-featured ruby on rails example.

examples/c/src/sum.c
vladaionescu: Right.. the question is why is this in this PR

PR #374, opened 2020-10-04
corrected the sentence formation

pixelcode009:


alexcb:
Thanks for the contribution, but we don't want to change the README.

PR #380, opened 2020-10-05
Remove old --with-docker implementation. Set up for future auto-install script

vladaionescu:


alexcb (approved):
nice code reduction :)

PR #381, opened 2020-10-05
Auto-install dockerd if not found and using WITH DOCKER

vladaionescu:
Also, add official earthly/dind image, which contains docker-compose already installed. This is the recommended image to use for Docker-in-Earthly as it is the most cache-efficient.

CC @agbell

### DIFF
--- a/buildkitd/docker-auto-install.sh
+++ b/buildkitd/docker-auto-install.sh
@@ -2,6 +2,8 @@
 
 set -eu
 
+distro=$(sed -n -e 's/^ID=\(.*\)/\1/p' /etc/os-release)
+
 detect_dockerd() {
     set +e
     command -v dockerd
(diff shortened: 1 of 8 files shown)


buildkitd/docker-auto-install.sh
```diff
@@ -2,4 +2,111 @@
...
+            install_dockerd_debian_like
+            ;;
+
+        debian)
+            install_dockerd_debian_like
+            ;;
+
+        *)
+            echo "Warning: Distribution $distro not yet supported for Docker-in-Earthly."
+            echo "Will attempt to treat like Debian."
```
alexcb, line 60: How about expanding this to request the user creates a ticket under https://github.com/earthly/earthly/issues for additional OS support?

alexcb (approved):
looks good!

buildkitd/docker-auto-install.sh
vladaionescu: Good idea

PR #386, opened 2020-10-07
initial secrets integration

alexcb:


Earthfile
```diff
@@ -54,7 +54,7 @@ lint:
             echo "$output" ; \
             exit 1 ; \
         fi
-    RUN output="$(goimports -d . 2>&1)" ; \
+    RUN output="$(goimports -d . 2>&1)" | grep -v '.*\.pb\.go' ; \
```
vladaionescu, line 57: Should the grep be within the `$(...)`?

cmd/earth/main.go
```diff
@@ -387,6 +395,72 @@ func newEarthApp(ctx context.Context, console conslogging.ConsoleLogger) *earthA
 				},
 			},
 		},
+		{
+			Name:   "org",
+			Usage:  "earthly organization administration",
```
vladaionescu, line 400:
```suggestion
			Usage:  "Earthly organization administration",
```

Start with capital letter to stay consistent.

Same with a few other places below.

secretsclient/client.go
```diff
@@ -0,0 +1,388 @@
...
+		_, authToken, err := c.getAuthToken()
+		if err != nil {
+			return 0, "", err
+		}
+		req.Header.Add("Authorization", authToken)
+	}
+
+	client := &http.Client{}
+
+	resp, err := client.Do(req)
```
vladaionescu, line 107: We should probably bake in some auto-retry logic. We'll probably have to be very careful with `4xx` vs `5xx` status codes on the server side.

In any case, this could be a `TODO` item.

secretsclient/api/api.pb.go
```diff
@@ -0,0 +1,543 @@
+// Code generated by protoc-gen-go. DO NOT EDIT.
```
vladaionescu, line 1: How is this file generated / updated? Should we place the protos in a more common location where both the server and earth can access?

vladaionescu (commented):
Looks great! Left a few comments...

secretsclient/api/api.pb.go
alexcb: that's a good plan, right now I simply copy and pasted this out of the server repo :(

Earthfile
alexcb: I'm not sure how this was working, it doesn't work inside the braces; i'll change it to use find instead.

secretsclient/client.go
alexcb: that's a good point. i'll make a TODO.

cmd/earth/main.go
```diff
@@ -387,6 +395,72 @@ func newEarthApp(ctx context.Context, console conslogging.ConsoleLogger) *earthA
...
+			Action:      app.actionRegister,
+			Flags: []cli.Flag{
+				&cli.StringFlag{
+					Name:        "email",
+					Usage:       "email address",
+					Destination: &app.email,
+				},
+				&cli.StringFlag{
+					Name:        "token",
+					Usage:       "email verification token",
```
vladaionescu, line 447: Forgot about these ones

vladaionescu (approved):

cmd/earth/main.go
alexcb: whoops, thanks for pointing that out. fixed.

vladaionescu (approved):

PR #391, opened 2020-10-09
Add debug flag, some refactoring

vladaionescu:
Debug flag now takes the function of the previous verbose flag. Verbose flag also prints output of internal operations (and in a following PR it will also print git clone details).

This PR also includes a lot of refactoring:
* states.go, image and dedup have been moved outside of earthfile2llb and into their own package, `states`
* Some renames of states-related things, to no longer include the word "states" now that the package name is named as such.
* `SideEffects*` have been renamed to `Main*` to be more suggestive


alexcb (approved):
nice simplification in naming

vladaionescu:
Merged in https://github.com/earthly/earthly/pull/392

Issue #387, opened 2020-10-08
Git-related failures are not clear


vladaionescu:
Git-related failures are not shown. Git clones resulting in an error should be displayed in detail to the user.

PR #392, opened 2020-10-09
Show git output on error or on verbose flag

vladaionescu:
Fixes https://github.com/earthly/earthly/issues/387

We now reuse the same builder (and solver monitor) when building the git context for fetching the Earthfile. This ensures that a we print out the logs of a failed internal operation (eg the git clone), making it easier for the user to debug git failures. Also, the git logs are visible if turning on the verbose flag.


buildcontext/git.go
```diff
@@ -8,24 +8,24 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/earthly/earthly/conslogging"
+	"github.com/earthly/earthly/cleanup"
```
alexcb, line 11: is this rename intentional? are you wanting to circle back to cleanup this package later?

alexcb (approved):

PR #394, opened 2020-10-09
C "hello world" example

kzuri:
In reference to #357


kzuri:
@agbell kindly have a look :)

alexcb:
Thanks for the PR @kzuri , it's looking like a good start. Can you expand on this example to have `main()` reference a function in a different c file, and include a basic unit test of that function using the cunit library?

To install cunit, you should be able to make the following change to the Earthfile:

`RUN apt-get update && apt-get install -y build-essential cmake libcunit1-dev`

kzuri:
Cool @alexcb . I will try 😄

kzuri:
@alexcb how to do i integrate cunit files in `test` folder and `Earthfile`?

alexcb:
I'm not sure what the best way to approach it is, it seems like you'll need another `add_executable` line to tell cmake to reference the `test/*.c` files. Then once cmake compiles a test binary, it could be run via an Earthfile test target.

kzuri:
@alexcb And to be clear, there would be only one Earthfile right?

alexcb:
yeah the single Earthfile that you have could be changed to something like:

```
RUN --mount=type=cache,target=/code/CMakeFiles make
RUN ./fibonacci-test
SAVE ARTIFACT fibonacci AS LOCAL "fibonacci"
```

That way if `./fibonacci-test` fails, then it'll stop the build thus indicating something went wrong with the test.

alexcb (changes requested):
It's no rush, but just wanted to check in to see if you had any luck getting the `fibonacci-test` binary compiled?

kzuri:
This change you suggest I not working well with cunit

`RUN apt-get update && apt-get install -y build-essential cmake libcunit1-dev`

I will update you soon 👍

alexcb:

PR #395, opened 2020-10-09
Add Terraform Example

dchw:


examples/terraform/start_localstack.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+docker run -d -p 4566:4566 -e SERVICES=s3 --name localstack --network="host" localstack/localstack:0.11.5
```
vladaionescu, line 3: We could reuse this file within the Earthfile if we wanted to. Or we could remove it from this example perhaps.

examples/terraform/Earthfile
```diff
@@ -0,0 +1,19 @@
+
+build:
```
vladaionescu, line 2: We should add this example to the root Earthfile, under the `+examples` target. That'll cause the CI to pick it up.

vladaionescu (commented):
Looks good! Left two minor comments.

agbell (approved):
Great example!

examples/terraform/Earthfile
dchw: Yep, realized I spaced that. Fixed.

examples/terraform/Earthfile
```diff
@@ -0,0 +1,19 @@
...
+    FROM hashicorp/terraform:light
+
+    ARG AWS_ACCESS_KEY_ID
+    ARG AWS_SECRET_ACCESS_KEY
```
vladaionescu, line 6: We should probably add a comment here saying that in prod environments we should use a secret. However in this case this is not really a secret so it's ok.

vladaionescu (approved):
:+1:

examples/terraform/Earthfile
dchw: Yeah, I was also looking at some of the terraform and some comments are probably in order.

PR #398, opened 2020-10-09
default to first public key during registration

alexcb:


cmd/earth/main.go
```diff
@@ -1009,14 +1009,16 @@ func (app *earthApp) actionRegister(c *cli.Context) error {
 			fmt.Printf("%d) %s\n", i+1, key.String())
 		}
 		keyNum := promptInput("enter key number: ")
```
vladaionescu, line 1011: I would also change this to

```suggestion
		keyNum := promptInput("enter key number (1=default): ")
```

alexcb: good catch, I lost this change when I rebased.

vladaionescu (approved):

PR #400, opened 2020-10-09
Attempt to address text file busy error on windows CI

vladaionescu:


alexcb (approved):
wow! that's a really creative work-around. This makes it seem like we sometimes have multiple tasks running on windows in parallel?

vladaionescu:
> This makes it seem like we sometimes have multiple tasks running on windows in parallel?

It seems like the agent correctly waits for other tasks to finish before starting a new one. No idea what's up with this.

Separately, I'll also switch the shell to PowerShell (I suspect CMD hangs for too long) and also switch to using SIGKILL rather than SIGTERM to cancel jobs.

PR #409, opened 2020-10-14
allow changing secrets server via ENV variable

alexcb:


secretsclient/client.go
```diff
@@ -132,6 +132,11 @@ func NewClient() (Client, error) {
 		return nil, fmt.Errorf("SSH_AUTH_SOCK is not set; is ssh-agent running?")
 	}
 
+	secretServer, ok := os.LookupEnv("EARTHLY_SECRETS_SERVER")
+	if !ok {
+		secretServer = "https://api.earthly.dev"
```
alexcb, line 137: this doesn't exist yet, but what are your thoughts on a new subdomain for when we enable it @dchw ?

vladaionescu: Perhaps this should be a command line option + env var?

vladaionescu (approved):

secretsclient/client.go
alexcb: I was wondering about that; it's only meant to be used for internal dev. It could be a hidden argument otherwise? I don't feel strongly about it one way or the other. I can make the change if you think it should be different.

vladaionescu: Yeah, hidden option sounds good - I think it's good to keep all knobs in a single place, even if some are hidden. Perhaps we can also call it `EARTHLY_API_SERVER` to make it more generic?

vladaionescu (approved):

secretsclient/client.go
alexcb: I like it, what if we take it one more step and just calling it `EARTHLY_SERVER` then? I would expect all servers to have an API.

vladaionescu: Good idea

dchw: @alexcb That is the exact subdomain I have written down for when I can take traffic

dchw (approved):

PR #412, opened 2020-10-14
Add contribution guide

vladaionescu:


CONTRIBUTING.md
```diff
@@ -0,0 +1,63 @@
...
+* Mac
+    ```bash
+    ./earth +for-darwin
+    ```
+
+This builds the earth binary in `./build/<platform>/amd64/earth` and also the buildkitd image.
+
+The buildkitd image is tagged with your current branch name and also the built binary defaults to using that built image. The built binary will always check on startup whether it has the latest buildkitd running for its configured image name and will restart buildkitd automatically to update. So if during your development you end up making changes to just the buildkitd image, the binary will pick up the change on its next run.
+
+For development purposes, you may use the built `earth` binary to rebuild itself. It's usually faster than switching between the built binary and the prerelease binary because it avoids constant buildkitd restarts. So after you've built the first time, you'll end up using:
```
alexcb, line 24: very minor nitpick, starting a sentence with `So` seemed a bit awkward, maybe something like this?
```suggestion
For development purposes, you may use the built `earth` binary to rebuild itself. It's usually faster than switching between the built binary and the prerelease binary because it avoids constant buildkitd restarts. After the first initial build, you'll end up using:

```

CONTRIBUTING.md
```diff
@@ -0,0 +1,63 @@
...
+
+### Running a full build stops after `SUCCESS`, but does not exit
+
+After starting a run of `earth +all` the console displays a green line which states
+
+```
+=========================== SUCCESS ===========================
+```
+
+It does not denote the end of the build process however and some additional work might still be required. However it might take a while to be displayed.
```
alexcb, line 64: Another minor nit, reading however in each sentence seemed a bit cumbersome. Maybe we can make it more concise with something like this?
```suggestion
The message does not denote the end of the build process as some additional work might be required before displaying the final output and cleanly exiting with a status code of 0.

```

alexcb (approved):
looks good, left a couple very minor nitpicky suggestions, feel free to ignore them.

PR #413, opened 2020-10-14
Make cache size limit 0 by default

vladaionescu:
Also, 0 now means to use buildkit's default, which automatically adapts to backing storage size.

This should improve performance as less GC is necessary.

### DIFF
--- a/buildkitd/entrypoint.sh
+++ b/buildkitd/entrypoint.sh
@@ -67,8 +67,6 @@ if [ "$CACHE_SIZE_MB" -gt "0" ]; then
 fi
 export CACHE_SETTINGS
 cat /etc/buildkitd.toml.template | envsubst >/etc/buildkitd.toml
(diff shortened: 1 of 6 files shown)


buildkitd/entrypoint.sh
```diff
@@ -56,13 +59,22 @@ if [ -n "$GIT_URL_INSTEAD_OF" ]; then
...
 # Set up buildkit cache.
-BUILDKIT_ROOT_DIR="$EARTHLY_TMP_DIR"/buildkit
+export BUILDKIT_ROOT_DIR="$EARTHLY_TMP_DIR"/buildkit
 mkdir -p "$BUILDKIT_ROOT_DIR"
+CACHE_SETTINGS=
+if [ "$CACHE_SIZE_MB" -gt "0" ]; then
+    CACHE_SETTINGS="$(cat /etc/buildkitd.cache.template | envsubst)"
+fi
+export CACHE_SETTINGS
+cat /etc/buildkitd.toml.template | envsubst >/etc/buildkitd.toml
```
alexcb, line 69: nice change

alexcb (approved):

PR #414, opened 2020-10-14
WITH DOCKER --load: infer image name from loaded target

vladaionescu:
CC @agbell


alexcb (approved):
nice change, loots pretty small too :)

PR #423, opened 2020-10-15
lazy init of ssh agent to avoid unneeded warnings

alexcb:
- prevent initializing the ssh agent client unless it's needed. This is
to prevent warning messages occuring if the SSH_AUTH_SOCK isn't set,
which shouldn't matter unless the secrets server is required.

### DIFF
--- a/secretsclient/lazy_ssh_agent.go
+++ b/secretsclient/lazy_ssh_agent.go
@@ -1,0 +1,107 @@
+package secretsclient
+
+import (
+	"net"
+
+	"github.com/pkg/errors"
+	"golang.org/x/crypto/ssh"
+	"golang.org/x/crypto/ssh/agent"
+)
+
+type lazySSHAgent struct {
+	sockPath string
+	sshAgent agent.ExtendedAgent
+}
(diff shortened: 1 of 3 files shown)


secretsclient/lazy_ssh_agent.go
```diff
@@ -0,0 +1,107 @@
...
+
+	"github.com/pkg/errors"
+	"golang.org/x/crypto/ssh"
+	"golang.org/x/crypto/ssh/agent"
+)
+
+type lazySSHAgent struct {
+	sockPath string
+	sshAgent agent.ExtendedAgent
+}
```
vladaionescu, line 14: Wonder if we can leverage https://golang.org/pkg/sync/#Once for this. I don't know if it simplifies this lazy proxy code though.

vladaionescu (approved):

secretsclient/lazy_ssh_agent.go
alexcb: I don't think we would be able to handle errors using this.

Issue #422, opened 2020-10-15
Some images don't work well with `WITH DOCKER` on WSL2


vladaionescu:
Hard to figure out what the cause is - something network-related apparently. Docker daemon does not start up and it spits an error that looks like this:

```
failed to start docker daemon: Error initializing network controller: error obtaining controller instance: failed to create NAT chain DOCKER: iptables failed: iptables --wait -t nat -N DOCKER: iptables v1.8.2 (nf_tables):  CHAIN_ADD failed (No such file or directory): chain PREROUTING
 (exit status 4)
```

Base image used: `hseeberger/scala-sbt:8u265_1.4.0_2.12.12`
Earthfile: https://github.com/earthly/earthly/blob/b1f86ebdebaba99f256db6c994e6e48346f6f58e/examples/integration-test/Earthfile
Full example: https://github.com/earthly/earthly/tree/b1f86ebdebaba99f256db6c994e6e48346f6f58e/examples/integration-test

Suggested workaround: use `earthly/dind` as a base image instead.

CC @agbell

ingwarsw:
Maybe `INSTALL_DIND` could check iptables version and revert to old one if needed?

ingwarsw:
I see it was partially fixed in https://github.com/earthly/earthly/pull/3254
But should be also fixed for other systems as well..
Im not sure how to "check" what version of iptables is supported? and whats present but that should be done in similar way as for WSL

ingwarsw:
I did small test
And I found thats trickier than I thought its..
```
test-me:
  FROM debian:bullseye
  RUN iptables --version || true
  DO github.com/earthly/lib+INSTALL_DIND
  # Bit hacky hack to fix docker on new systems
  RUN iptables --version || true
  RUN update-alternatives --set iptables /usr/sbin/iptables-legacy
  RUN iptables --version || true
  WITH DOCKER
    RUN iptables --version || true
  END
```

And results on `linux/arm64` and `linux/amd64`
```
            +test-me | --> RUN iptables --version || true 
            +test-me | /bin/sh: 1: iptables: not found # iptables not installed on base image
            +test-me | --> COPY +install-dind-script/install-dind.sh /tmp/install-dind.sh
 ...
            # INSTALL_DIND install both iptables and nftables itself
            +test-me | Setting up nftables (0.9.8-3.1+deb11u1) ...
            +test-me | Setting up iptables (1.8.7-1) ...
[...]

PR #425, opened 2020-10-15
Reverting integration example to use docker in docker 

agbell:
Reverting integration example to use docker in docker to work around #422

### DIFF
--- a/docs/guides/integration.md
+++ b/docs/guides/integration.md
@@ -133,23 +133,42 @@ We start with a simple Earthfile that can build and create a docker image for ou
 
 We start from an appropriate docker image and set up a working directory. 
 ``` Dockerfile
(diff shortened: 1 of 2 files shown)


vladaionescu:
Should the equivalent docs also be updated?

agbell:
> Should the equivalent docs also be updated?

updated, though I am sad to lose the example of auto-installing docker tools.  I will revert this once #422 is addressed in windows.

vladaionescu (approved):
Looks good - might need to wait for master CI to be fixed and rebase to make sure tests pass.

vladaionescu:
> Looks good - might need to wait for master CI to be fixed and rebase to make sure tests pass.

Master is fixed now. I'll merge as is.

PR #426, opened 2020-10-15
add command to list orgs

alexcb:


secretsclient/client.go
```diff
@@ -458,3 +465,33 @@ func (c *client) Invite(path, user string, write bool) error {
 	}
 	return nil
 }
+
+func (c *client) ListOrgs() ([]*OrgDetail, error) {
+	status, body, err := c.doCall("GET", fmt.Sprintf("/api/v0/admin/organizations"), withPublicKeyAuth())
```
vladaionescu, line 470: Sprintf is not necessary here, right?

vladaionescu (approved):

secretsclient/client.go
alexcb: good catch.

PR #431, opened 2020-10-18
Implements single- and double quotes

BjoernAkAManf:
This pull request incorporates the past couple of days, which were basically changing some code and letting it rerun for a couple of hours, due to the full build being very slow (even with --no-output). Might be a problem with WSL2.

However this Solution still suffers from the it wrongly detecting 
```BUILD --build-arg ACTUAL='$(whoami)' --build-arg EXPECTED='^\$\(whoami\)$' +build```
as
```ACTUAL='$(whoami)' --build-arg```
which obviously fails due to --build-arg not being a shell argument.
Not sure why that is though, as the ANTLR grammar worked almost perfectly for that example.

Any suggestions how to proceed would be nice :)

This PR is solving #86 once the above issues have been resolved.

PS: I'm also not sure why the Build arguments have been preprocessed / parsed before, as Earthly already shells out, which makes any preprocessing here pointless, right?

### CODE REVIEW

File: earthfile2llb/converter.go [hunk 1]
```diff
@@ -614,7 +614,7 @@ func (c *Converter) internalRun(ctx context.Context, args []string, secretKeyVal
 			continue
 		}
 		if ba.IsConstant() {
-			extraEnvVars = append(extraEnvVars, fmt.Sprintf("%s=\"%s\"", buildArgName, ba.ConstantValue()))
+			extraEnvVars = append(extraEnvVars, fmt.Sprintf("%s=%s", buildArgName, ba.ConstantValue()))
```
vladaionescu, line 617: This change isn't right probably, in case the build arg value contains a space.

earthfile2llb/parser/EarthLexer.g4
```diff
@@ -81,7 +81,9 @@ WS_R: WS -> type(WS);
 mode COMMAND_ARGS;
 
 Atom: (RegularAtomPart | QuotedAtomPart)+;
-fragment QuotedAtomPart: ('"' (~'"' | '\\"')* '"');
+fragment QuotedAtomPart: SingleQuotedAtomPart | DoubleQuotedAtomPart;
+fragment SingleQuotedAtomPart: ('\'' (~'\'' | '\\\'')* '\'');
+fragment DoubleQuotedAtomPart: ('"' (~'"' | '\\"')* '"');
```
vladaionescu, line 86: This part seems to make sense, however, you probably also need to eliminate `'` from the `RegularAtomPart` fragment. So something like

```
fragment RegularAtomPart: ~([ \t\r\n\\"\\']) | EscapedAtomPart;
```

(Note that there's also `fragment RegularAtomPart_CAKV` which would need updating).

earthfile2llb/listener.go
```diff
@@ -434,9 +434,6 @@ func (l *listener) ExitBuildStmt(c *parser.BuildStmtContext) {
 		return
 	}
 	fullTargetName := l.expandArgs(fs.Arg(0))
-	for i, arg := range buildArgs.Args {
-		buildArgs.Args[i] = l.expandArgs(arg)
-	}
```
vladaionescu, line 439: This isn't right.

We definitely need to perform expansion here - not all build args are used in a shell command. So not all would be automatically expanded by the shell. For example you could have things like `FROM $SOME_BUILD_ARG` and other such commands.

.gitignore
```diff
@@ -1,3 +1,5 @@
+*.iml
+.idea
```
vladaionescu, line 2: This should be in your user's gitignore, rather than the project's.

vladaionescu (commented):
This is good progress - if my suggestion doesn't fix it, then I'm not sure how to address this either. It'd require more investigation.

.gitignore
BjoernAkAManf: Sure thing. I'll revert that.

earthfile2llb/converter.go
BjoernAkAManf: But the ConstantValue will already include either ' or " due to the ConstantValue being an Atom, so if it has a space, it doesn't matter?

earthfile2llb/listener.go
BjoernAkAManf: The Problem with expanding here however is, that once you do that, every ' and " will be gone. So maybe we should either use the shell expansion solely or implement the execution manually. Otherwise it might get hard to "quote" each quotes.

earthfile2llb/parser/EarthLexer.g4
BjoernAkAManf: That seems like a good catch! I'll take a look.

earthfile2llb/converter.go
vladaionescu: Oh... does it? If that's the case, you may be right

BjoernAkAManf:
I think there are still some implementation questions to be resolved to be honest, however i will take a look another look once i find the time and dedication 😄 .

earthfile2llb/listener.go
vladaionescu: I think that's intended and also, we may need to actually escape all of the `'` and `"` as you suggest. Otherwise we can't support both commands like `FROM` and also commands like `RUN` with build args.

You could use something like https://github.com/alessio/shellescape to do the heavy lifting.

alexcb:

PR #432, opened 2020-10-19
Optimize images

JeffResc:
This pull request optimizes the images found in the repository using lossless compression. This means that the image quality will not be degraded. This will mean less space taken up by images along with faster loading times when viewing them. The overall file size has been reduced by around 18%.


vladaionescu (approved):
Interesting, thanks!

Can you share what you did to reduce size? Perhaps it's something we can take into account for future images.

JeffResc:
@vladaionescu I use ImageMagick, but for a streamlined solution you could use something like [ImgBot](https://github.com/marketplace/imgbot) to have it done automatically :)

PR #438, opened 2020-10-22
Client retry

alexcb:


secretsclient/client.go
```diff
@@ -102,6 +114,36 @@ func (c *client) doCall(method, url string, opts ...requestOpt) (int, string, er
...
+			}
+		}
+		if duration > maxSleepBeforeRetry {
+			duration = maxSleepBeforeRetry
+		}
+		time.Sleep(duration)
+		duration *= 2
+	}
+	return status, body, err
+}
```
vladaionescu, line 144: Wondering if we should use some library for this. Although this is pretty small and simple too.

secretsclient/client.go
```diff
@@ -143,16 +185,18 @@ type client struct {
 	sshKey                string
 	lastUsedPublicKeyPath string
 	sshAgent              agent.ExtendedAgent
+	warnFunc              func(string, ...interface{})
```
vladaionescu, line 188: Maybe it's simpler to just pass the entire `ConsoleLogger` object? But not a strong opinion.

vladaionescu (approved):

secretsclient/client.go
alexcb: I went with this as it gives us an option to move this client into an outside library down the road.

vladaionescu: Good idea

secretsclient/client.go
alexcb: I was half tempted to use one, but figured less (dependencies) is more.

PR #439, opened 2020-10-22
Use a build context provider to incrementally feed local dirs into the solve

vladaionescu:
Depends on https://github.com/earthly/buildkit/pull/10

### DIFF
--- a/cmd/earth/main.go
+++ b/cmd/earth/main.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/earthly/earthly/autocomplete"
 	"github.com/earthly/earthly/buildcontext"
+	"github.com/earthly/earthly/buildcontext/provider"
 	"github.com/earthly/earthly/builder"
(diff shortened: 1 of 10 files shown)


cmd/earth/main.go
```diff
@@ -1326,9 +1327,19 @@ func (app *earthApp) actionBuild(c *cli.Context) error {
 
 	sc := secretsclient.NewClient(app.apiServer, app.sshAuthSock, app.publicKey)
 
+	cacheLocalDir, err := ioutil.TempDir("/tmp", "earthly-cache")
```
alexcb, line 1330: if the first arg to TempDir is `""`, then it'll use the OS's default tmp dir location (which will most likely be /tmp, but you never know...)

alexcb (approved):

vladaionescu:
Will merge as part of https://github.com/earthly/earthly/pull/442

PR #443, opened 2020-10-23
Add ability to align and control size of target name column in logs

dchw:
Willing and ready for suggestions on this idea. I think this improves readability of the logs. Would love to detect max width of target names; and size down if possible on top of this, to make it feel less dead with small target names. But, this is a start toward that for now.

Still needs documentation pending decision on if this is worth doing.

Some examples of what this accomplishes:

Old (also now, but with `FULL_TARGETS`):
```
Deprecation: using SAVE IMAGE with no arguments is no longer necessary and can be safely removed
+base | --> FROM golang:1.13-alpine3.11
context | --> local context ./examples/go
+base | resolve docker.io/library/golang:1.13-alpine3.11@sha256:ec6dcf15073c307fbcfc3149efe8835f3ec2bd0a0cb49aaaee4949cfc4c86b65 100%
context | transferring ./examples/go: 100%
+base | *cached* --> RUN apk add --update --no-cache bash bash-completion binutils ca-certificates coreutils curl findutils g++ git grep less make openssl shellcheck util-linux
+base | *cached* --> WORKDIR /earthly
./examples/go+base | *cached* --> WORKDIR /go-example
./examples/go+deps | *cached* --> COPY go.mod go.sum ./
./examples/go+deps | *cached* --> RUN go mod download
./examples/go+build | *cached* --> COPY main.go .
./examples/go+build | *cached* --> RUN go build -o build/go-example main.go
./examples/go+build | *cached* --> SAVE ARTIFACT build/go-example ./examples/go+build/go-example
./examples/go+docker | *cached* --> COPY ./examples/go+build/go-example .
+examples | Target github.com/earthly/earthly:corey/align-logs+examples built successfully
=========================== SUCCESS ===========================
Loaded image: go-example:latest
./examples/go+docker | Image github.com/earthly/earthly/examples/go:corey/align-logs+docker as go-example:latest
./examples/go+build | *cached* --> SAVE ARTIFACT build/go-example ./examples/go+build/go-example AS LOCAL build/go-example
./examples/go+build | Artifact github.com/earthly/earthly/examples/go:corey/align-logs+build/go-example as local build/go-example
./examples/go+deps | *cached* --> SAVE ARTIFACT go.mod ./examples/go+deps/go.mod AS LOCAL go.mod
./examples/go+deps | Artifact github.com/earthly/earthly/examples/go:corey/align-logs+deps/go.mod as local go.mod
./examples/go+deps | *cached* --> SAVE ARTIFACT go.sum ./examples/go+deps/go.sum AS LOCAL go.sum
./examples/go+deps | Artifact github.com/earthly/earthly/examples/go:corey/align-logs+deps/go.sum as local go.sum

```

New default of 20 width:
```
❯ ./testearth -P +examples
           buildkitd | ...Done
Deprecation: using SAVE IMAGE with no arguments is no longer necessary and can be safely removed
             context | --> local context ./examples/go
               +base | --> FROM golang:1.13-alpine3.11
               +base | resolve docker.io/library/golang:1.13-alpine3.11@sha256:ec6dcf15073c307fbcfc3149efe8835f3ec2bd0a0cb49aaaee4949cfc4c86b65 100%
             context | transferring ./examples/go: 100%
               +base | *cached* --> RUN apk add --update --no-cache bash bash-completion binutils ca-certificates coreutils curl findutils g++ git grep less make openssl shellcheck util-linux
               +base | *cached* --> WORKDIR /earthly
  ./examples/go+base | *cached* --> WORKDIR /go-example
  ./examples/go+deps | *cached* --> COPY go.mod go.sum ./
  ./examples/go+deps | *cached* --> RUN go mod download
 ./examples/go+build | *cached* --> COPY main.go .
 ./examples/go+build | *cached* --> RUN go build -o build/go-example main.go
 ./examples/go+build | *cached* --> SAVE ARTIFACT build/go-example ./examples/go+build/go-example
./examples/go+docker | *cached* --> COPY ./examples/go+build/go-example .
           +examples | Target github.com/earthly/earthly:corey/align-logs+examples built successfully
=========================== SUCCESS ===========================
Loaded image: go-example:latest
./examples/go+docker | Image github.com/earthly/earthly/examples/go:corey/align-logs+docker as go-example:latest
 ./examples/go+build | *cached* --> SAVE ARTIFACT build/go-example ./examples/go+build/go-example AS LOCAL build/go-example
 ./examples/go+build | Artifact github.com/earthly/earthly/examples/go:corey/align-logs+build/go-example as local build/go-example
  ./examples/go+deps | *cached* --> SAVE ARTIFACT go.mod ./examples/go+deps/go.mod AS LOCAL go.mod
  ./examples/go+deps | Artifact github.com/earthly/earthly/examples/go:corey/align-logs+deps/go.mod as local go.mod
  ./examples/go+deps | *cached* --> SAVE ARTIFACT go.sum ./examples/go+deps/go.sum AS LOCAL go.sum
  ./examples/go+deps | Artifact github.com/earthly/earthly/examples/go:corey/align-logs+deps/go.sum as local go.sum

```

### DIFF
--- a/cmd/earth/main.go
+++ b/cmd/earth/main.go
@@ -171,7 +171,21 @@ func main() {
 		color.NoColor = true
 	}
 
-	app := newEarthApp(ctx, conslogging.Current(colorMode))
+	padding := conslogging.DefaultPadding
+	customPadding, ok := os.LookupEnv("TARGET_PADDING")
+	if ok {
+		targetPadding, err := strconv.Atoi(customPadding)
+		if err == nil {
+			padding = targetPadding
+		}
+	}
+
+	_, fullTarget := os.LookupEnv("FULL_TARGET")
+	if fullTarget {
+		padding = conslogging.NoPadding
+	}
+
+	app := newEarthApp(ctx, conslogging.Current(colorMode, padding))
 	app.autoComplete()
 
 	os.Exit(app.run(ctx, os.Args))
--- a/conslogging/conslogging.go
+++ b/conslogging/conslogging.go
@@ -24,6 +24,14 @@ const (
 	ForceColor
 )
 
+const (
+	// NoPadding means the old behavior of printing the full target only.
+	NoPadding int = -1
+	// DefaultPadding always prints 20 characters for the target, right
+	// justified. If it is longer, it prints the right 20 characters.
+	DefaultPadding int = 20
+)
+
 var currentConsoleMutex sync.Mutex
 
 // ConsoleLogger is a writer for consoles.
@@ -42,15 +50,17 @@ type ConsoleLogger struct {
 	nextColorIndex *int
 	w              io.Writer
 	trailingLine   bool
+	prefixPadding  int
 }
 
 // Current returns the current console.
-func Current(colorMode ColorMode) ConsoleLogger {
+func Current(colorMode ColorMode, prefixPadding int) ConsoleLogger {
 	return ConsoleLogger{
 		w:              os.Stdout,
 		colorMode:      colorMode,
 		saltColors:     make(map[string]*color.Color),
 		nextColorIndex: new(int),
+		prefixPadding:  prefixPadding,
 		mu:             &currentConsoleMutex,
 	}
 }
@@ -65,6 +75,7 @@ func (cl ConsoleLogger) clone() ConsoleLogger {
 		saltColors:     cl.saltColors,
 		colorMode:      cl.colorMode,
 		nextColorIndex: cl.nextColorIndex,
+		prefixPadding:  cl.prefixPadding,
 		mu:             cl.mu,
 	}
 }
@@ -193,7 +204,7 @@ func (cl ConsoleLogger) printPrefix() {
 		*cl.nextColorIndex = (*cl.nextColorIndex + 1) % len(availablePrefixColors)
 	}
 	c = cl.color(c)
-	c.Fprintf(cl.w, "%s", cl.prefix)
+	c.Fprintf(cl.w, cl.prettyPrefix())
 	if cl.isFailed {
 		cl.w.Write([]byte(" *"))
 		cl.color(warnColor).Fprintf(cl.w, "failed")
@@ -220,4 +231,20 @@ func (cl ConsoleLogger) color(c *color.Color) *color.Color {
 		return c
 	}
 	return noColor
+}
+
+func (cl ConsoleLogger) prettyPrefix() string {
+	if cl.prefixPadding == NoPadding {
+		return cl.prefix
+	}
+
+	formatString := fmt.Sprintf("%%%vv", cl.prefixPadding)
+	substringStart := len(cl.prefix) - cl.prefixPadding
+
+	clampedPadding := substringStart
+	if clampedPadding < 0 {
+		clampedPadding = 0
+	}
+
+	return fmt.Sprintf(formatString, cl.prefix[clampedPadding:])
 }
--- a/cmd/debugger/main.go
+++ b/cmd/debugger/main.go
@@ -190,7 +190,7 @@ func main() {
 		return
 	}
 
-	conslogger := conslogging.Current(conslogging.ForceColor)
+	conslogger := conslogging.Current(conslogging.ForceColor, conslogging.NoPadding)
 	color.NoColor = false
 
 	debuggerSettings, err := getSettings(fmt.Sprintf("/run/secrets/%s", common.DebuggerSettingsSecretsKey))
--- a/docs/guides/ci-integration.md
+++ b/docs/guides/ci-integration.md
@@ -61,7 +61,11 @@ docker login --username '<username>' --password '<password>'
 Make sure that secrets (like `<password>` above) are not exposed in plain text. You may need to configure an environment variable with your CI vendor.
 {% endhint %}
 
-## Step 4: (Optional) Force or disable color output
+## Step 4: (Optional) Choose log formatting options
+
+There are a few options you can configure to make the log formatting look best in your CI environment.
+
+### Force or disable color output
 
 The CLI `earth` automatically detects the presence of a TTY for the purpose of deciding whether to use colorized output or not. In some CI environments, this kind of detection is not enough in order to infer support for colorized output. However, two environment variables can be used to either disable or force it:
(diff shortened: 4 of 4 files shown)


cmd/earth/main.go
```diff
@@ -171,7 +171,21 @@ func main() {
 		color.NoColor = true
 	}
 
-	app := newEarthApp(ctx, conslogging.Current(colorMode))
+	padding := conslogging.DefaultPadding
+	customPadding, ok := os.LookupEnv("TARGET_PADDING")
+	if ok {
+		targetPadding, err := strconv.Atoi(customPadding)
+		if err == nil {
+			padding = targetPadding
```
dchw, line 179: Only assign when its not nil, and just silently use the default if the env var is not set correctly. Didn't seem worth logging a message about, would likely be apparent what happened?

conslogging/conslogging.go
```diff
@@ -221,3 +232,19 @@ func (cl ConsoleLogger) color(c *color.Color) *color.Color {
...
+	formatString := fmt.Sprintf("%%%vv", cl.prefixPadding)
+	substringStart := len(cl.prefix) - cl.prefixPadding
+
+	clampedPadding := substringStart
+	if clampedPadding < 0 {
+		clampedPadding = 0
+	}
+
+	return fmt.Sprintf(formatString, cl.prefix[clampedPadding:])
+}
```
dchw, line 250: I am torn here. I do not like calculating this on each log right now... Should I move this up into the main struct alongside regular prefix, and calculate it once? I'm probably going to do that.

cmd/earth/main.go
```diff
@@ -171,7 +171,21 @@ func main() {
 		color.NoColor = true
 	}
 
-	app := newEarthApp(ctx, conslogging.Current(colorMode))
+	padding := conslogging.DefaultPadding
+	customPadding, ok := os.LookupEnv("TARGET_PADDING")
```
alexcb, line 175: rather than pull out the env variable here, one can define it as a flag under main. e.g. https://github.com/earthly/earthly/blob/7d72007a596506b3d3925aa2a4bed91e49cc7c4f/cmd/earth/main.go#L320-L326

the flag can also have an env name associated with it.

conslogging/conslogging.go
```diff
@@ -221,3 +232,19 @@ func (cl ConsoleLogger) color(c *color.Color) *color.Color {
 	}
 	return noColor
 }
+
+func (cl ConsoleLogger) prettyPrefix() string {
+	if cl.prefixPadding == NoPadding {
+		return cl.prefix
+	}
+
+	formatString := fmt.Sprintf("%%%vv", cl.prefixPadding)
```
dchw, line 241: Use the padding number as the number in a format string like this one: "%20v". We just have to format the format string to get there.

dchw (commented):
Few thoughts I had while doing this

alexcb:
looks like a nice improvement, I'm sort of torn between left vs right align. If it's left aligned it would look like:
```
context              | transferring ./examples/go: 100%
+base                | *cached* --> RUN apk add --update --no-cache bash bash-completion binutils ca-certificates coreutils curl findutils g++ git grep less make openssl shellcheck util-linux
+base                | *cached* --> WORKDIR /earthly
./examples/go+base   | *cached* --> WORKDIR /go-example
./examples/go+deps   | *cached* --> COPY go.mod go.sum ./
./examples/go+deps   | *cached* --> RUN go mod download
./examples/go+build  | *cached* --> COPY main.go .
./examples/go+build  | *cached* --> RUN go build -o build/go-example main.go
./examples/go+build  | *cached* --> SAVE ARTIFACT build/go-example ./examples/go+build/go-example
./examples/go+docker | *cached* --> COPY ./examples/go+build/go-example .
```

which would also keep all the example targets grouped together so the majority of the `+`s also stay aligned.

vs right aligned:

```
             context | transferring ./examples/go: 100%
               +base | *cached* --> RUN apk add --update --no-cache bash bash-completion binutils ca-certificates coreutils curl findutils g++ git grep less make openssl shellcheck util-linux
               +base | *cached* --> WORKDIR /earthly
  ./examples/go+base | *cached* --> WORKDIR /go-example
  ./examples/go+deps | *cached* --> COPY go.mod go.sum ./
  ./examples/go+deps | *cached* --> RUN go mod download
 ./examples/go+build | *cached* --> COPY main.go .
 ./examples/go+build | *cached* --> RUN go build -o build/go-example main.go
 ./examples/go+build | *cached* --> SAVE ARTIFACT build/go-example ./examples/go+build/go-example
./examples/go+docker | *cached* --> COPY ./examples/go+build/go-example .
```

but the right aligned one limits the amount of space between short targets and their output. (insert shrug emoji here)

cmd/earth/main.go
dchw: Roger

dchw: Actually, those look like they get set _after_; which is also why color is doing it the same way?

vladaionescu:
Looks great! I'll let Alex finish off the review

cmd/earth/main.go
dchw: Digging deeper, those could all be handled in an `app.before`

dchw:
> but the right aligned one limits the amount of space between short targets and their output. (insert shrug emoji here)

Yep, which is exactly why I went with the right aligned. I also worried about short targets getting lost in the sheer W I D T H. I don't have that strong of opinions though, willing to go with whatever the veterans think is best

cmd/earth/main.go
alexcb: ah, I didn't realize the other logging options were set via a regular call to LookupEnv.

Do we even want to support customizing this behaviour? what if We just switched to your new alignment for all cases, then we don't have to document these new options. It might be a bit too much choice?

alexcb (approved):
looks good.

cmd/earth/main.go
dchw: I think we should for CI logs; which is also why the color options exist it looks like: https://docs.earthly.dev/guides/ci-integration#step-4-optional-force-or-disable-color-output

We can leave the 20 char width as the default; and add documentation to the same section?

PR #445, opened 2020-10-23
Add column width documentation, generalize options for log formatting

dchw:


docs/guides/ci-integration.md
```diff
@@ -73,7 +77,13 @@ The following environments are known to require additional settings:
 * GitHub Actions: requires `FORCE_COLOR=1`
 * Jenkins: requires `NO_COLOR=1`
 
-## Step 5: Run the build
+### Adjust target name column
+
+The CLI defaults to a right-aligned, 20-character width to display the name of the target that the log originated from. When a target name exceeds 20 characters, the name is truncated and the last 20 characters are used.
+
+ * `TARGET_PADDING=n` will set the column to the width of `n` characters.
+ * `FULL_TARGET=1` will always print the full target name, and nothing more. This may result in unaligned logs.
```
vladaionescu, line 85: I would prefix these with `EARTHLY_` like we did with most env vars. (We didn't prefix `NO_COLOR` and `FORCE_COLOR` as they are more standard env var names)

docs/guides/ci-integration.md
```diff
@@ -73,7 +77,13 @@ The following environments are known to require additional settings:
 * GitHub Actions: requires `FORCE_COLOR=1`
 * Jenkins: requires `NO_COLOR=1`
 
-## Step 5: Run the build
+### Adjust target name column
+
+The CLI defaults to a right-aligned, 20-character width to display the name of the target that the log originated from. When a target name exceeds 20 characters, the name is truncated and the last 20 characters are used.
+
+ * `TARGET_PADDING=n` will set the column to the width of `n` characters.
```
vladaionescu, line 84: Is the default behavior that it'll try to pad and if it exceeds this size, it'll truncate? I'd maybe go the other way: try to pad, and if longer, just exceed and cause misalignment only for that longer target.

I'm thinking if you build `earth github.com/earthly/earthly+earth` for example, most prefixes would be pretty long and the output might not make sense if it all exceeds.

vladaionescu: Another thought: maybe we can come up with some decent truncation technique. Eg `github.com/earthly/earthly+earth` could be truncated like `g/e/earthly+earth`

agbell: +1 on clever truncating.  That would be cool

dchw: @vladaionescu Yeah, that one does have a lot that exceed. They make enough sense since with the last 20 characters, in fact its the "finger in the wind" I used to find a limit.

I am 100% also down for clever truncation. I found myself really wanting to go further and figure out the max width of all log prefixes that would be used... but that wasn't going to happen as quick as I wanted it to.

I chose not to exceed because I wanted alignment to be the default, and ragged full names can happen if you specify the `FULL_TARGET` env var... its the old style entirely. Im not super opinionated on the exceeding, just wanted to keep it aligned because I really like aligning things.

I'll do the clever truncation for sure, plus the ragged potential on too long of names.

alexcb (approved):

vladaionescu:
Wondering if we should document this here. I'd keep the CI integration guide as simple as possible to make it easier to digest for the reader. If everything has sane defaults, most people would not need to tweak it. Perhaps just putting this in the earth command line reference page is enough?

dchw:
It seemed like the best place, because the option it best matches also was only documented there. Perhaps an environment variable section in the CLI reference would be bet?

vladaionescu:
> It seemed like the best place, because the option it best matches also was only documented there. Perhaps an environment variable section in the CLI reference would be bet?

Yeah, that would be best

vladaionescu (approved):

PR #450, opened 2020-10-26
track exit codes via segment.io

alexcb:


alexcb:
Here's a preview of tracking the exit codes on our dashboard
image

cmd/earth/main.go
```diff
@@ -47,11 +47,14 @@ import (
 	"github.com/moby/buildkit/session/auth/authprovider"
 	"github.com/moby/buildkit/session/sshforward/sshprovider"
 	"github.com/moby/buildkit/util/entitlements"
+	uuid "github.com/nu7hatch/gouuid"
 	"github.com/pkg/errors"
 	"github.com/seehuhn/password"
 	"github.com/urfave/cli/v2"
 	"golang.org/x/crypto/ssh"
 	"golang.org/x/sync/errgroup"
+
```
vladaionescu, line 56: Nit: Unintended blank line?

cmd/earth/main.go
```diff
@@ -188,7 +191,10 @@ func main() {
 	app := newEarthApp(ctx, conslogging.Current(colorMode, padding))
 	app.autoComplete()
 
-	os.Exit(app.run(ctx, os.Args))
+	startTime := time.Now()
+	exitCode := app.run(ctx, os.Args)
+	app.collectAnalytics(exitCode, time.Since(startTime))
+	os.Exit(exitCode)
```
vladaionescu, line 197: I would also make sure that the user is running `build` here and not some other command.

vladaionescu: ... or perhaps the command itself can be a "property".

cmd/earth/main.go
```diff
@@ -1450,6 +1456,36 @@ func (app *earthApp) newBuildkitdClient(ctx context.Context, opts ...client.Clie
...
 }
 
+func (app *earthApp) collectAnalytics(exitCode int, realtime time.Duration) {
+	if app.cfg.Global.DisableAnalytics {
+		return
+	}
+	installID, err := getInstallID()
+	if err != nil {
+		installID = "unknown"
+	}
```
vladaionescu, line 1466: It seems that most CIs have the env var `CI` set to true:

* https://circleci.com/docs/2.0/env-vars/
* https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables
* https://docs.travis-ci.com/user/environment-variables/#default-environment-variables
* https://buildkite.com/docs/pipelines/environment-variables#bk-env-vars-ci

Jenkins seems like the odd one out, but we could perhaps detect `BUILD_ID` there: https://wiki.jenkins.io/display/JENKINS/Building+a+software+project

Perhaps we can add a property like `"ci": "true"` when we detect a CI environment.

vladaionescu (commented):
Woohoo

cmd/earth/main.go
alexcb: that's a good idea, that way we'll avoid lots of unique installIDs since the uuid won't be saved between CI runs.

vladaionescu: Yes, exactly

cmd/earth/main.go
alexcb: good catch :)

cmd/earth/main.go
alexcb: I added a command property; I couldn't figure out a great way to do it, so I had to make use of the `After` function to record which command was run, but I can't put the call to collectAnalytics in `After` since the returned error wasn't available there.

vladaionescu: I wonder if `Command.Name` ends up being just `get` in the case of `earth secrets get`? In any case, I think it's a good start regardless.

cmd/earth/main.go
```diff
@@ -1450,6 +1462,42 @@ func (app *earthApp) newBuildkitdClient(ctx context.Context, opts ...client.Clie
...
+	if !ci {
+		var err error
+		installID, err = getInstallID()
+		if err != nil {
+			installID = "unknown"
+		}
+	}
+	segmentClient := analytics.New("RtwJaMBswcW3CNMZ7Ops79dV6lEZqsXf")
+	segmentClient.Enqueue(analytics.Track{
+		Event:  "exit",
```
vladaionescu, line 1479: Reading a little more on the amplitude docs, I think it's better to use the command name here, as the event (possibly with `"earth "` as a prefix), because this is the actual action that the user performed.

cmd/earth/main.go
alexcb: I was wrong with this initial idea, it's just showing up as an empty string, I'll need to figure out another way to get to it. I initially thought including all of argv would work, but it could capture some sensitive user information so that's not an option.

vladaionescu: A simple idea could be to set the command name ourselves at the beginning of each `action*` fun

alexcb: I'll try that

cmd/earth/main.go
alexcb: sounds like a good change, I was hoping we could configure it somewhere in the pipeline but perhaps not.

cmd/earth/main.go
```diff
@@ -1450,6 +1470,41 @@ func (app *earthApp) newBuildkitdClient(ctx context.Context, opts ...client.Clie
...
+	if !ci {
+		var err error
+		installID, err = getInstallID()
+		if err != nil {
+			installID = "unknown"
+		}
+	}
+	segmentClient := analytics.New("RtwJaMBswcW3CNMZ7Ops79dV6lEZqsXf")
+	segmentClient.Enqueue(analytics.Track{
+		Event:  app.commandName,
```
vladaionescu, line 1487: Also add something like `"earth "` or `"CLI "` in front? These events will be mixed in together with UI / website events in the analytics.

vladaionescu (approved):

vladaionescu (approved):

PR #452, opened 2020-10-27
Evaluate symlink when detecting gitRelDir

### LINKED ISSUE #449: Does not work if any elements of the current directory path are symlinks
When I run the earth command, and the directory I am in has a symlink in it, I get the following error:
```shell
$ earth +docker
Error: resolve build context for target +docker: Unexpected non-relative path within git dir
```
(I'm using the `Go` example from here: [a simple earthfile](https://docs.earthly.dev/guides/basics#a-simple-earthfile))

I store my repositories in a `REPOS` directory, let's call it `$HOME/Work/company/foo/bar/REPOS`, and I have a symlink to it in `$HOME/REPOS`. If I run the `earth` command from `$HOME/REPOS`, then I get the error, but if run from the destination directory, the command works correctly.

OS: Linux (Pop!_OS 20.10)
```shell
$ earth --version
earth version v0.3.10
```

alexcb:
fixes #449

### DIFF
--- a/buildcontext/detectgit.go
+++ b/buildcontext/detectgit.go
@@ -232,11 +232,15 @@ func gitRelDir(basePath string, path string) (string, bool, error) {
 	if err != nil {
 		return "", false, errors.Wrapf(err, "get abs path for %s", path)
 	}
+	absPath2, err := filepath.EvalSymlinks(absPath)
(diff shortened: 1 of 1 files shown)


vladaionescu (approved):
Awesome

Issue #99, opened 2020-07-09
Associated buildkit solve's


vladaionescu:
Currently Earthly builds rely on multiple solve submissions to buildkit. Basically one for the side effects, one for each artifact, one for each image and one for each target with `RUN --push` instructions.

Although this works reasonably well, there are some possible issues and limitations due to this approach:

* [Performance](https://github.com/earthly/earthly/issues/85) is impacted
* Race conditions are possible as the builds are not associated with each other via a single session. It is possible for a GC to kick in between such related solves and end up needing to rebuild parts that have already been built. This is very rare due to the cache typically not evicting very recent builds - but maybe the GC strategy of buildkit may change.
* Other undiscovered side-effects.

In order to address this, we would need to switch to using the [`Build`](https://godoc.org/github.com/moby/buildkit/client#Client.Build) API of buildkit, rather than [`Solve`](https://godoc.org/github.com/moby/buildkit/client#Client.Solve). This would allow to associate multiple solve's within the same session. However, this approach has the following limitations / possible challenges:

* Local dirs need to be specified at the beginning of the build. Possible ways to address this:
  * Add a way in buildkit to incrementally add local dirs in the middle of the session
  * Use two passes when processing earthfiles: one pass to collect all dirs, and another pass to actually take the commands and convert them to LLB (which may also include intermediate solve's). However, this approach prevents us from using variable args for expanding external targets, should we want to do this in the future.
* Exports can become gnarly to track and match back to the originating target as they are returned as a single big blob (?). There may be unexpected hoops we'd need to jump through to make it work for the Earthly use-case.
* The experimental docker-in-docker related features rely on invoking a build in the middle of a build (in order to output a docker image which can be used within a docker-in-docker). This can't be merged into a single build, because the `Build` API doesn't have the ability to turn a state into a docker .tar in the middle of a build (without exporting to the client).

Issue #85, opened 2020-06-24
Slow Performance


brandonkal:
Steps should be taken to make this as fast as possible. As it is now, I really like this project but something is adding a lot of latency here.

Changing the string printed in the Go Hello World example.
Local build is 0.7 seconds whereas earth build is 14.6 seconds.

```
brandon@dev:~/go-earth$ vim main.go # making edit
brandon@dev:~/go-earth$ time go build

real	0m0.732s
user	0m0.413s
sys	0m0.398s
brandon@dev:~/go-earth$ time earth +build
buildkitd | Found buildkit daemon as docker container (earthly-buildkitd)
context | --> local context .
+base | --> FROM golang:1.13-alpine3.11
+base | resolve docker.io/library/golang:1.13-alpine3.11@sha256:5237fe9c1c891aee9ffb22258a412cc7449f310f8f2f65d4389c516218f90443 0%
+base | resolve docker.io/library/golang:1.13-alpine3.11@sha256:5237fe9c1c891aee9ffb22258a412cc7449f310f8f2f65d4389c516218f90443 100%
context | transferring .: 0%
context | transferring .: 0%
context | transferring .: 100%
+base | *cached* --> WORKDIR /go-example
+build | --> COPY [main.go] .
+build | --> RUN [go build -o build/go-example main.go]
+build | Target <redacted>/go-earth:master+build built successfully
=========================== SUCCESS ===========================
+build | Artifact <redacted>/go-earth:master+build/go-example as local build/go-example

real	0m14.589s
user	0m0.824s
sys	0m0.754s
```

Even with no change and everything cached, the delay that earth adds is several seconds:

```
brandon@dev:~/go-earth$ time earth +build
buildkitd | Found buildkit daemon as docker container (earthly-buildkitd)
context | --> local context .
+base | --> FROM golang:1.13-alpine3.11
+base | resolve docker.io/library/golang:1.13-alpine3.11@sha256:5237fe9c1c891aee9ffb22258a412cc7449f310f8f2f65d4389c516218f90443 0%
+base | resolve docker.io/library/golang:1.13-alpine3.11@sha256:5237fe9c1c891aee9ffb22258a412cc7449f310f8f2f65d4389c516218f90443 100%
context | transferring .: 0%
context | transferring .: 0%
context | transferring .: 100%
+base | *cached* --> WORKDIR /go-example
+build | *cached* --> COPY [main.go] .
+build | *cached* --> RUN [go build -o build/go-example main.go]
[...]

Issue #35, opened 2020-03-29
SAVE ARTIFACT empty wildcard ends up as an error after success


vladaionescu:
When we have something like

```
SAVE ARTIFACT *.txt AS LOCAL build/
```

... and say that there are no *.txt files found, the system detects this as an error only after it has deemed the build as success. It should detect the error before that and not declare it as a success.

Ideas:
* Tie artifact independent states to final side effects with fake deps? This may have performance implications.

Issue #18, opened 2020-03-22
Should chains of buildkit solve commands reuse a session?


vladaionescu:
Currently, as part of a single build, there are multiple Solve commands issued to buildkit and they share nothing other than some copied state. The copied state allows buildkit to reuse a lot of cache.

It's possible that between these Solve's, the cache is purged and then the system unexpectedly has to build the same thing twice.

Is there a better way to do this? Is there a way for us to reuse the session? Would that help with telling buildkit to hold onto the cache for a while?

PR #453, opened 2020-10-27
Execute builds in a single BuildKit session (most pieces anyway), get rid of fakedep

vladaionescu:
Depends on https://github.com/earthly/buildkit/pull/11

* The net benefit of this PR is performance: for a one line go change in our codebase, the build used to take ~68s. Now it takes ~15s (~4.5X improvement).
* Merging all solves in a single session also fixes a number of obscure edge case bugs.

Changes:
* Perform main build, image outputs and artifact outputs within a single BuildKit session
* Eliminate fakedep and instead add all deps as refs to the response of the build function. Fakedep was extremely slow (assuming due to the need to rehash the FS over and over again)
* Order of printout of the outputs is now consistent between runs

Fixes https://github.com/earthly/earthly/issues/99
Fixes https://github.com/earthly/earthly/issues/85
Fixes https://github.com/earthly/earthly/issues/35
Fixes https://github.com/earthly/earthly/issues/18

TODO in future PRs:
* Some git operations are still using the old solves.
* Building a generated Dockerfile via `FROM DOCKERFILE` also still uses the old solves.
* WITH DOCKER --load and --pull also use the old solves.
* Execution of RUN --push also uses the old solves.


Earthfile
```diff
@@ -24,6 +24,8 @@ deps:
     RUN go get golang.org/x/lint/golint
     RUN go get github.com/gordonklaus/ineffassign
     COPY go.mod go.sum ./
+    # @#
```
alexcb, line 27: is this comment intentional?

vladaionescu: Fixed :-)

buildkitd/Earthfile
```diff
@@ -1,6 +1,6 @@
 
 buildkitd:
-    ARG BUILDKIT_BASE_IMAGE=github.com/earthly/buildkit:earthly-master+build
+    ARG BUILDKIT_BASE_IMAGE=github.com/earthly/buildkit:vlad/exporters+build
```
alexcb, line 3: can this be switched back now that the buildkit PR was merged?

vladaionescu: Yes done

alexcb (approved):
very nice!

vladaionescu:
One of the tests flakes with `./e/tests+build-earth-test *failed* | ERROR: () failed to send last stat: EOF`. I encountered this issue locally too - I'll attempt to fix.

vladaionescu:
> One of the tests flakes with ./e/tests+build-earth-test *failed* | ERROR: () failed to send last stat: EOF. I encountered this issue locally too - I'll attempt to fix.

I believe this has been resolved now.

PR #454, opened 2020-10-27
Add artifact test for no-cache case where same artifact would be written twice

dchw:


vladaionescu (approved):

alexcb (approved):
nice test!

alexcb:
I had to move this into https://github.com/earthly/earthly/pull/458 since we've renamed master to main

PR #455, opened 2020-10-27
override install-id to use CI value

alexcb:


.buildkite/windows.yml
```diff
@@ -6,6 +6,7 @@ steps:
     env:
       GIT_URL_INSTEAD_OF: "https://github.com/=git@github.com:"
       FORCE_COLOR: 1
+      CI: buildkite-earthly-win
```
vladaionescu, line 9: Need to add this to WSLENV below

vladaionescu (commented):
How does this work? The CIs already provide a CI=true env var, right?

alexcb:
I noticed that CI=true in some cases, but we're still getting lots of uuids coming through, which I can only guess is from these builds.

alexcb:
on second thought, maybe we should change this env variable to `EARTHLY_INSTALL_ID` and let it override the auto-generated one. and then fall back to using the CI value if it's set?

alexcb:
I put in a new PR to allow for `EARTHLY_INSTALL_ID` under https://github.com/earthly/earthly/pull/456
then once that's in, we can set these vars.

vladaionescu (approved):
Do we still need this? I think we should be fine now that we come up with a sensible install ID even for CIs.

alexcb:
the only reason is that we could use it to identify our usage versus other usage, but perhaps it's fine to just track our own usage as if it's someone elses, that way we'll follow the same codepath as other users.

PR #456, opened 2020-10-27
allow overriding install id

alexcb:
How about we allow overriding `EARTHLY_INSTALL_ID`, so we can exclude our own builds.
Then we'll track if it was triggered via a CI (assuming the CI env is set), which we could then use to filter out CI vs CLI executions.

### DIFF
--- a/cmd/earth/main.go
+++ b/cmd/earth/main.go
@@ -36,9 +36,9 @@ import (
 	"github.com/earthly/earthly/domain"
 	"github.com/earthly/earthly/earthfile2llb"
 	"github.com/earthly/earthly/earthfile2llb/variables"
+	"github.com/earthly/earthly/fileutils"
 	"github.com/earthly/earthly/llbutil"
 	"github.com/earthly/earthly/secretsclient"
-	"github.com/earthly/earthly/utils"
 
 	"github.com/fatih/color"
 	"github.com/joho/godotenv"
@@ -155,7 +155,7 @@ func main() {
 
 	// Load .env into current global env's. This is mainly for applying Earthly settings.
 	// Separate call is made for build args and secrets.
-	if utils.FileExists(dotEnvPath) {
+	if fileutils.FileExists(dotEnvPath) {
 		err := godotenv.Load(dotEnvPath)
 		if err != nil {
 			fmt.Printf("Error loading dot-env file %s: %s\n", dotEnvPath, err.Error())
@@ -615,7 +615,7 @@ func (app *earthApp) before(context *cli.Context) error {
 		}
 	}
 
-	if !utils.DirExists(app.cfg.Global.RunPath) {
+	if !fileutils.DirExists(app.cfg.Global.RunPath) {
 		err := os.MkdirAll(app.cfg.Global.RunPath, 0755)
 		if err != nil {
 			return errors.Wrapf(err, "failed to create run directory %s", app.cfg.Global.RunPath)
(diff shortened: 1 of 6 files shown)


cmd/earth/main.go
```diff
@@ -1480,7 +1476,7 @@ func (app *earthApp) collectAnalytics(exitCode int, realtime time.Duration) {
 			Set("version", Version).
 			Set("gitsha", GitSha).
 			Set("exitcode", exitCode).
-			Set("ci", ci).
+			Set("ci", os.Getenv("CI")).
```
vladaionescu, line 1479: BTW, I would also add a special case for Jenkins here.. (detect presence of `BUILD_ID` or something like that).

Also, I would make this explicitly `"false"` when the var doesn't exist, rather than `""` - it'll make filtering easier.

vladaionescu (approved):

cmd/earth/main.go
alexcb: ah that's a really good point. let me fix that.

cmd/earth/main.go
```diff
@@ -1480,7 +1543,8 @@ func (app *earthApp) collectAnalytics(exitCode int, realtime time.Duration) {
 			Set("version", Version).
 			Set("gitsha", GitSha).
 			Set("exitcode", exitCode).
-			Set("ci", ci).
+			Set("ci", detectCI()).
+			Set("repo", getRepoHash()).
```
vladaionescu, line 1547: `repohash` for clarity?

```suggestion
			Set("repohash", getRepoHash()).
```

cmd/earth/main.go
```diff
@@ -1460,17 +1462,78 @@ func (app *earthApp) newBuildkitdClient(ctx context.Context, opts ...client.Clie
...
+	return "unknown"
+}
+
+func getRepoHash() string {
+	repo := getRepo()
+	if repo == "unknown" || repo == "" {
+		return repo
+	}
+	return fmt.Sprintf("%x", sha256.Sum256([]byte(repo)))
+}
```
vladaionescu, line 1528: It's probably worth moving all of these into a new `analytics.go` file. `main.go` is getting pretty big.

cmd/earth/main.go
```diff
@@ -1532,6 +1596,11 @@ func defaultConfigPath() string {
 }
 
 func getInstallID() (string, error) {
+	installID, ok := os.LookupEnv("EARTHLY_INSTALL_ID")
+	if ok {
+		return installID, nil
+	}
```
vladaionescu, line 1602: I'm thinking we could also try to craft an install ID from repo + some build name, when CI is true. Build name would be workflow name / pipeline name / build name - something that stays constant for each CI build pipeline.

And when CI is false, we can use the generated file in `.earthly` dir.

cmd/earth/main.go
alexcb: i was on the fence about that, but yeah that's a good idea.

cmd/earth/main.go
alexcb: excellent plan, I just pushed up a new version to do this.

utils/utils.go
```diff
@@ -0,0 +1,23 @@
+package utils
```
vladaionescu, line 1: Rename package to `fileutils`? It's better to have multiple focused utility / common packages, rather than a single utils with many unrelated functions.

vladaionescu (approved):

utils/utils.go
alexcb: that's fair, I couldn't think of a good name.

vladaionescu (approved):

PR #457, opened 2020-10-27
fix segfault when running --version

alexcb:


cmd/earth/main.go
```diff
@@ -1461,7 +1461,7 @@ func (app *earthApp) newBuildkitdClient(ctx context.Context, opts ...client.Clie
 }
 
 func (app *earthApp) collectAnalytics(exitCode int, realtime time.Duration) {
-	if app.cfg.Global.DisableAnalytics {
+	if app.cfg == nil || app.cfg.Global.DisableAnalytics {
```
vladaionescu, line 1464:
```suggestion
	if app.cfg != nil && app.cfg.Global.DisableAnalytics {
```

This should make the default behavior more consistent

vladaionescu (approved):

PR #464, opened 2020-10-27
Add early ECR example

dchw:
The current state of the world when pushing to ECR. Did not add it to the main Earthfile for running all the tests, because the only option would be localstack... and that seems to be flaky inside a `WITH DOCKER` (see the terraform example).

### DIFF
--- a/examples/ecr-push/Readme.md
+++ b/examples/ecr-push/Readme.md
@@ -1,0 +1,28 @@
+# Push an image to ECR with Earthly
+
+We are still making this easier, we hope it gets much shorter!
+
+The steps to build your image go in the `+some-thing` target. The `+ecr-push` target will push it to ECS. Assuming you keep your AWS credentials in the standard environment variables, you can build your image and push it to ECR like this:
+
+```
+earth -P \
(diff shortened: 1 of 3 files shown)


examples/ecr-push/Readme.md
```diff
@@ -0,0 +1,28 @@
...
+
+If you do not have the environment variables, you can set the AWS credentials like this via the already existing AWS CLI setup:
+
+```
+...
+    --secret AWS_ACCESS_KEY_ID=$(aws configure get default.aws_access_key_id)
+    --secret AWS_SECRET_ACCESS_KEY=$(aws configure get default.aws_secret_access_key)
+    --secret AWS_SESSION_TOKEN=$(aws configure get default.aws_session_token)
+...
+```
```
vladaionescu, line 28: Rename this file to `README.md` for consistency?

vladaionescu (commented):
I would note in the README that there's also the alternative of doing the fetching of aws creds + docker login on the host before running `earth`. That approach has the advantage that it also works for pulls, not just for pushes.

vladaionescu (approved):
Nice

PR #465, opened 2020-10-28
Ignore analytics for --version command

alexcb:
When --version is run, urfav/cli does not invoke the app.Before
function, this causes the config not to be parsed.

If a user had a config option to disable analytics, and ran the
--version flag, it would cause analytics to be reported.

### DIFF
--- a/cmd/earth/main.go
+++ b/cmd/earth/main.go
@@ -196,7 +196,9 @@ func main() {
 	app.autoComplete()
 
 	exitCode := app.run(ctx, os.Args)
-	if app.cfg == nil || !app.cfg.Global.DisableAnalytics {
+	// app.cfg will be nil when a user runs `earth --version`;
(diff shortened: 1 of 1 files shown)


vladaionescu (approved):

vladaionescu (changes requested):
Wait... this isn't right.. if the user does not have a config file, it means that we don't collect analytics at all?

alexcb:
we'll default to collecting metrics. The issue is because https://github.com/earthly/earthly/blob/main/cmd/earth/main.go#L579 is not invoked until app.Before() is called. app.Before() is called in regular cases, except for when the `--version` flag is detected.

The other approach would be to run `ParseConfigFile` in `main`. maybe that's the best approach?

alexcb:
oh right, but we can't read the config file without allowing for the `--config` override flag.

vladaionescu (approved):
:+1:

PR #467, opened 2020-10-28
expose earthly env variable

alexcb:
This will prevent having earth create a new installID when running the examples and other earth-in-earth tests.

### DIFF
--- a/analytics/analytics.go
+++ b/analytics/analytics.go
@@ -21,7 +21,6 @@ func detectCI() (string, bool) {
 	for k, v := range map[string]string{
 		"GITHUB_WORKFLOW": "github-actions",
 		"CIRCLECI":        "circle-ci",
-		"EARTHLY":         "earthly",
 		"JENKINS_HOME":    "jenkins",
 		"BUILDKITE":       "buildkite",
 		"DRONE_BRANCH":    "drone",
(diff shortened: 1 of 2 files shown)


vladaionescu (commented):
Oh good find.

This won't work unfortunately because the builtin build arg needs to be predeclared by the user using ARG EARTHLY. For our tests it's probably better to add EARTHLY=true in the +docker-earth target, which is used in integration tests.

alexcb:
Right that's a good point. This question might be outside of the scope of this PR, but is there a reason we want to be so explicit with predeclaring these builtin EARTHLY_* args?

vladaionescu:
Yes... because otherwise the varying git hash would bust the cache on every run...

alexcb:
ah right of course. What about always exposing env variables we know won't ever change, like `EARTHLY=true`? That might be a weird case as then we'll have to explain the various edge-cases and that makes documentation more complex.

another idea is to set the ID to unknown when the repohash is unknown. ie.:
```
		if ci {
			installID = fmt.Sprintf("%x", sha256.Sum256([]byte(ciName+repoHash)))
		} else if repoHash == "unknown" {
			installID = "unknown"
		} else {
			installID, err = getInstallID()
			if err != nil {
				installID = "unknown"
			}
		}
```
```

vladaionescu:
I think marking install_id unknown like you suggest is the best idea... it'll also catch any other weird edge cases we're not thinking of.

vladaionescu (approved):

PR #468, opened 2020-10-28
experimental code for generating an Earthfile from a Dockerfile

alexcb:


cmd/earth/main.go
```diff
@@ -423,6 +424,13 @@ func newEarthApp(ctx context.Context, console conslogging.ConsoleLogger) *earthA
 				},
 			},
 		},
+		{
+			Name:        "docker",
```
vladaionescu, line 428: I would call this something more obvious / verbose - eg `dockerfile2earthfile`. It's anyway not something used every day.

dockerconverter/convert.go
```diff
@@ -0,0 +1,103 @@
...
+		return errors.Wrap(err, "failed to open ./Dockerfile")
+	}
+	defer in.Close()
+
+	targets := [][]string{
+		{},
+	}
+	target := 0
+	scanner := bufio.NewScanner(in)
+	for scanner.Scan() {
```
vladaionescu, line 39: It might be less brittle to use the dockerfile libraries themselves - like they are used here: https://github.com/moby/buildkit/blob/f4ebd6d352ac1fd5f6a691818a44b4dbbc4b09ba/frontend/dockerfile/dockerfile2llb/convert.go#L84-L105

Watch for that escape char though - Dockerfile has a way in which the user can override the escape character (it's used mostly for windows containers).

alexcb: Great suggestion, I didn't realise it was available before it was converted into LLB.

docker2earth/convert.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+// Docker2Earth converts an existing Dockerfile in the current directory and writes out an Earthfile in the current directory
+// and error is returned if an Earthfile already exists.
+func Docker2Earth() error {
+	if fileutils.FileExists("Earthfile") {
+		return fmt.Errorf("Earthfile already exists; please delete it if you wish to continue")
+	}
+
+	in, err := os.Open("Dockerfile")
```
vladaionescu, line 28: It might be nice to have this work via stdin & stdout, so it can be piped like typical unix commands. This could be a future thing though.

docker2earth/convert.go
```diff
@@ -0,0 +1,116 @@
...
+				targets[n+1] = append(targets[n+1], fmt.Sprintf("SAVE ARTIFACT %s %s\n", parts[2], artifactName))
+			}
+			if strings.HasPrefix(l, "ADD ") {
+				return fmt.Errorf("earthly does not support ADD, please convert to COPY instead")
+			}
+			targets[i+1] = append(targets[i+1], l)
+		}
+	}
+	i := len(targets) - 1
+	targets[i] = append(targets[i], "SAVE IMAGE myimage:latest")
```
vladaionescu, line 89: Perhaps it's better if the name for the final target is specified by the user when invoking the command and the names for any other intermediate targets are of some form representing some identifier of the stage (eg `stage-1:latest` or `stage-<stage-name>:latest`, if it has a name given by `FROM ... AS <stage-name>`).

docker2earth/convert.go
```diff
@@ -0,0 +1,116 @@
...
+		})
+		if stage.Name == "" {
+			names[fmt.Sprintf("%d", i)] = i
+		} else {
+			names[stage.Name] = i
+		}
+
+		for _, cmd := range stage.Commands {
+			l := fmt.Sprintf("%v", cmd)
+			if strings.HasPrefix(l, "COPY ") && strings.Contains(l, "--from") {
```
vladaionescu, line 67: Nice

vladaionescu (approved):
🎉

docker2earth/convert.go
alexcb: just opened up a PR for this: https://github.com/earthly/earthly/pull/474

PR #478, opened 2020-10-31
Use CNI rather than host networking for buildkit

vladaionescu:
The net benefit of this is that `WITH DOCKER` can now run in parallel with other `WITH DOCKER`'s. No more networking-related race conditions. This also removes the lock that was preventing the race conditions before.

### DIFF
--- a/.buildkite/tests.sh
+++ b/.buildkite/tests.sh
@@ -46,9 +46,8 @@ echo "Build latest earth using released earth"
 echo "Execute tests"
 ./build/"$EARTH_OS"/amd64/earth --no-output -P +test
 
-# Temporarily disable until failure is addressed.
-#echo "Execute experimental tests"
-#./build/"$EARTH_OS"/amd64/earth --no-output -P ./examples/tests+experimental
+echo "Execute experimental tests"
+./build/"$EARTH_OS"/amd64/earth --no-output -P ./examples/tests+experimental
 
 echo "Execute fail test"
 bash -c "! ./build/$EARTH_OS/amd64/earth --no-output +test-fail"
--- a/.buildkite/tests.sh
+++ b/.buildkite/tests.sh
@@ -46,8 +46,9 @@ echo "Build latest earth using released earth"
 echo "Execute tests"
 ./build/"$EARTH_OS"/amd64/earth --no-output -P +test
 
-echo "Execute experimental tests"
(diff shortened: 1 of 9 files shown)


alexcb (approved):
sweet!

vladaionescu:
Tests failing with some very random `failed to mount: no such file or directory` in various places

```
=========================== FAILURE ===========================
./e/t/with-docker+docker-load-test *failed* | --> WITH DOCKER (install deps)
./e/t/with-docker+docker-load-test *failed* | [no output]
./e/t/with-docker+docker-load-test *failed* | ERROR: Command exited with non-zero code: WITH DOCKER (install deps)
Error: build target: build main: bkClient.Build: failed to solve: rpc error: code = Unknown desc = executor failed running [/bin/sh -c EARTHLY_START_COMPOSE="false" EARTHLY_COMPOSE_FILES="" EARTHLY_COMPOSE_SERVICES="" /var/earthly/docker-auto-install.sh]: failed to mount: no such file or directory
```

```
=========================== FAILURE ===========================
./e/tests+dotenv-test *failed* | --> RUN echo "TEST_ENV_1=abracadabra" >.env
./e/tests+dotenv-test *failed* | [no output]
./e/tests+dotenv-test *failed* | ERROR: Command exited with non-zero code: RUN echo "TEST_ENV_1=abracadabra" >.env
Error: build target: build main: bkClient.Build: failed to solve: rpc error: code = Unknown desc = executor failed running [/bin/sh -c  /usr/bin/earth_debugger /bin/sh -c 'echo "TEST_ENV_1=abracadabra" >.env']: failed to mount: no such file or directory
```

I suspect that the increased performance has led to parallelism not seen before, which is hitting some system limits related to mounts. Running tests more sequentially doesn't cause this. Also, the tests seem to work fine on win and mac, which have worse performance.

vladaionescu:
Limiting parallelism in buildkit seems to mitigate the issue - but it has to be a really low limit (3) which slows everything down. So that doesn't sound like a great solution.

> I suspect that the increased performance has led to parallelism not seen before, which is hitting some system limits related to mounts.

This seem to be incorrect - seems that even when turning off dind-related tests, the random error still happens. New theory: CNI is the main cause of this, and the error `failed to mount: no such file or directory` is caused by some CNI internals.

vladaionescu:
The cause of the race condition has been identified and fixed in buildkit https://github.com/moby/buildkit/pull/1775

PR #488, opened 2020-11-05
only testing so far, please ignore.

alexcb:


alexcb:
Hi @vladaionescu , This PR isn't ready for merging, but I wanted to get your thoughts on this higher-level refactor of the domain.Target struct before I continue.

Consider a target github.com/earthly/earthly/examples/go:main+docker, previously this would get split up to have

    Target.Registry = "github.com", and Target.ProjectPath="earthly/earthly/examples/go".

I changed this to have

    Target.GitURL=github.com/earthly/earthly and projectPath="examples/go".

This way we can work towards supporting a repo of the form `host.com/my/path/to/repo` (because previously the code would just attempt to clone `git@host.com/my/path.git` and would assume `/to/repo` was within the git repo.

This PR paves the way for supporting other git users when sshing too, and eventually the GitLookup will be populated by user-specific entries in addition to standard ones for github, gitlab, etc.

Issue #498, opened 2020-11-10
parser failure: nil pointer dereference


jazzdan:
Repro steps:

```
$ earth --version
earth version v0.3.12
$ git clone https://github.com/jazzdan/earthly-nil-pointer.git
$ cd earthly-nil-pointer
$ cat Earthfile
FROM node:14.15.0
WORKDIR /usr/build

src:
    COPY package.json package-lock.json ./
    RUN npm install --production
    COPY . .

standalone-experiment-setup:
    FROM +src
    RUN mv src/build/Dockerfile.standalone Dockerfile

standalone-experiment:
    FROM DOCKERFILE +standalone-experiment-setup/
    SAVE IMAGE recordreplay-standalone
$ earth +standalone-experiment
        buildkitd | Found buildkit daemon as docker container (earthly-buildkitd)
        node:14.15.0 | --> Load metadata
Error: build target: build main: bkClient.Build: failed to solve: rpc error: code = Unknown desc = parser failure: runtime error: invalid memory address or nil pointer dereference
```

I think I ran in to this before and it was some typo, but as of right now I'm still looking for what caused it so I figured I'd report it. :)

vladaionescu:
Thanks for the repro Earthfile - I'll take a look

vladaionescu:
BTW - I see a few COPY commands - any chance that you could share the files that are being copied so that I have a fully working reproduction of the issue?

jazzdan:
@vladaionescu I don't think it matters in this case because even if I run that Earthfile in an empty directory (Like my [linked repro case](https://github.com/jazzdan/earthly-nil-pointer.git)) I get the same error.

Maybe I can reduce the repro case further, let me see

jazzdan:
@vladaionescu I reduced it further to:

```Earthfile
FROM node:14.15.0
WORKDIR /usr/build

foo:
    RUN touch Dockerfile

bar:
    FROM DOCKERFILE +foo/
```

jazzdan:
Then `earth +bar` :)

Hmm maybe the issue is that I didn't save the output of `foo` as an artifact?

vladaionescu:
Yes... exactly!

We should improve that error

vladaionescu:
Thanks for the report, @jazzdan - improved error message coming in the next release!

PR #499, opened 2020-11-10
Slightly better error message when no SAVE ARTIFACT command has been issued

vladaionescu:
The error is improved from

```
        node:14.15.0 | --> Load metadata
Error: build target: build main: bkClient.Build: failed to solve: rpc error: code = Unknown desc = parser failure: runtime error: invalid memory address or nil pointer dereference
```

to the slightly more useful

```
        node:14.15.0 | --> Load metadata
Error: build target: build main: bkClient.Build: failed to solve: rpc error: code = Unknown desc = parse: from dockerfile: artifact +foo/Dockerfile not found; no SAVE ARTIFACT command was issued in +foo
```

The Earthfile used is

```
FROM node:14.15.0
WORKDIR /usr/build

foo:
    RUN touch Dockerfile

bar:
    FROM DOCKERFILE +foo/
```

Fixes https://github.com/earthly/earthly/issues/498

### DIFF
--- a/earthfile2llb/converter.go
+++ b/earthfile2llb/converter.go
@@ -679,6 +679,10 @@ func (c *Converter) internalRun(ctx context.Context, args []string, secretKeyVal
 }
 
 func (c *Converter) readArtifact(ctx context.Context, mts *states.MultiTarget, artifact domain.Artifact) ([]byte, error) {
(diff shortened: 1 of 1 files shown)


alexcb (approved):
that's a nicer error message :)

Issue #502, opened 2020-11-11
Cannot COPY files containing +


vladaionescu:
Dan Miller (@jazzdan) reports on Slack

> I want to COPY a file in an Earthfile that has a + in it (like COPY lib/7z lib/7z.so lib/libc++.dylib lib/libsystem_c.dylib /), but when I do Earth complains:
> ```
> Error: build target: build main: bkClient.Build: failed to solve: rpc error: code = Unknown desc = parse: Combining artifacts and build context arguments in a single COPY command is not allowed: [lib/7z lib/7z.so lib/libc++.dylib lib/libsystem_c.dylib]
> ```
> Is there a way to escape the + or something?

We should support escaping the `+` in COPY commands (eg `\+`).

We should test:
* COPY a file containing a `+` from build context
* COPY an artifact containing a `+` from another target
* Three tests where each of the SAVE ARTIFACT params contains a `+`

vladaionescu:
This will be in the next release @jazzdan

PR #503, opened 2020-11-11
Support COPYing files and artifacts containing + in them

vladaionescu:
Fixes https://github.com/earthly/earthly/issues/502

### DIFF
--- a/domain/domain_test.go
+++ b/domain/domain_test.go
@@ -2,6 +2,8 @@ package domain
 
 import (
 	"testing"
+
+	. "github.com/stretchr/testify/assert"
 )
 
 var targetTests = []struct {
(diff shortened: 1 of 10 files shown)


domain/domain_test.go
```diff
@@ -0,0 +1,105 @@
...
+	{"../rel/local/dir-with-\\+-in+target", Target{Target: "target", LocalPath: "../rel/local/dir-with-+-in"}},
+	{"github.com/foo/bar/dir-with-\\+-in+target", Target{Target: "target", Registry: "github.com", ProjectPath: "foo/bar/dir-with-+-in"}},
+	{"github.com/foo/bar:tag-with-\\+-in+target", Target{Target: "target", Registry: "github.com", ProjectPath: "foo/bar", Tag: "tag-with-+-in"}},
+}
+
+func TestTargetParser(t *testing.T) {
+	for _, tt := range targetTests {
+		t.Run(tt.in, func(t *testing.T) {
+			out, err := ParseTarget(tt.in)
+			if err != nil {
```
alexcb, line 31: If you use the testify package, this can be shortened into:
```
	NoError(t, err, "ParseTarget failed")
	Equal(t, tt.out, out)
```
like what was done in https://github.com/earthly/earthly/blob/main/autocomplete/complete_test.go#L74-L75

alexcb (approved):
very nice testing!

domain/domain_test.go
vladaionescu: :+1:

PR #504, opened 2020-11-13
Compile earth for arm5 (raspberry pi)

alexcb:


Earthfile
```diff
@@ -130,6 +131,14 @@ earth:
     SAVE ARTIFACT ./build/ldflags
     SAVE ARTIFACT build/earth AS LOCAL "build/$GOOS/$GOARCH/earth"
 
+earth-arm5:
+    COPY \
+        --build-arg GOARCH=arm \
+        --build-arg GOARM=5 \
+        --build-arg GO_EXTRA_LDFLAGS= \
+        +earth/* ./
```
vladaionescu, line 139: Within the earth target, should the `AS LOCAL` also use `build/linux/arm5/earth` rather than `build/linux/arm/earth`?

vladaionescu (approved):

Earthfile
alexcb: that's a good point, I'll give that a try.

vladaionescu (approved):

PR #510, opened 2020-11-17
Print overridden args at the beginning of each target and also on failure

vladaionescu:
Some example of how this looks like (dark gray text):

Screenshot from 2020-11-17 16-02-45
Screenshot from 2020-11-17 16-03-02
Screenshot from 2020-11-17 16-09-32

Verbose mode prints the overridden build args before every single command.

Screenshot from 2020-11-17 16-23-54

The information is passed via the buildkit vertex name, with any overridden build args encoded in brackets as `+target(<base64 build args string>)` and then interpreted by the solver monitor and printed accordingly.

### DIFF
--- a/builder/solver_monitor.go
+++ b/builder/solver_monitor.go
@@ -202,7 +202,7 @@ func (sm *solverMonitor) printHeader(vm *vertexMonitor) {
 	if !seen {
 		sm.saltSeen[vm.salt] = true
 	}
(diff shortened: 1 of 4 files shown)


alexcb (approved):

vladaionescu:
Merged as part of https://github.com/earthly/earthly/pull/514

PR #516, opened 2020-11-18
Fix for arg expansion in copy command

vladaionescu:
A fix for the COPY issue @zenyui reported over on Gitter.

The fix is on line 233. The rest also addresses situations where files or artifacts with `+` are passed via build args.

### DIFF
--- a/examples/tests/escape.earth
+++ b/examples/tests/escape.earth
@@ -3,13 +3,31 @@ WORKDIR /test
 
 all:
     BUILD +test-copy-build-context
+    # TODO: FILE_IN_RUN shuoldn't need to be different. This is a bug.
+    BUILD \
+        --build-arg FILE="file-with-\+.txt" \
+        --build-arg FILE_IN_RUN="file-with-+.txt" \
+        +test-copy-build-arg
+    BUILD --build-arg FILE="regular-file.txt" +test-copy-build-arg
     BUILD +test-copy-artifact1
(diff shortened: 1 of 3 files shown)


examples/tests/escape.earth
```diff
@@ -3,14 +3,32 @@ WORKDIR /test
 
 all:
     BUILD +test-copy-build-context
+    # TODO: FILE_IN_RUN shuoldn't need to be different. This is a bug.
```
alexcb, line 6:
```suggestion
    # TODO: FILE_IN_RUN shouldn't need to be different. This is a bug.
```

examples/tests/escape.earth
```diff
@@ -3,14 +3,32 @@ WORKDIR /test
...
+    BUILD \
+        --build-arg FILE="file-with-\+.txt" \
+        --build-arg FILE_IN_RUN="file-with-+.txt" \
+        +test-copy-build-arg
+    BUILD --build-arg FILE="regular-file.txt" +test-copy-build-arg
     BUILD +test-copy-artifact1
     BUILD +test-copy-artifact2
     BUILD +test-copy-artifact3
+    # TODO: FILE_IN_RUN shuoldn't need to be different. This is a bug.
```
alexcb, line 15:
```suggestion
    # TODO: FILE_IN_RUN shouldn't need to be different. This is a bug.
```

alexcb (approved):

PR #518, opened 2020-11-18
use buildkit ssh mount for git clones

alexcb:
Make use of the new ssh mount for performing git clones, which was implemented in https://github.com/moby/buildkit/pull/1782


buildcontext/git.go
```diff
@@ -128,7 +123,10 @@ func (gr *gitResolver) resolveGitProject(ctx context.Context, gwClient gwclient.
 		llb.WithCustomNamef("[internal] GIT CLONE %s", gitURL),
 		llb.KeepGitDir(),
 	}
-	gitState := llbgit.Git(gitURL, ref, gitOpts...)
+	if keyScan != "" {
+		gitOpts = append(gitOpts, llb.KnownSSHHosts(keyScan))
```
vladaionescu, line 127: Should this be used in both of the `llb.Git` calls?

buildcontext/gitlookup.go
```diff
@@ -0,0 +1,179 @@
...
+func (gl *GitLookup) DisableSSH() {
+	for i, m := range gl.matchers {
+		if m.protocol == "ssh" {
+			gl.matchers[i].protocol = "https"
+		}
+	}
+	if gl.catchAll.protocol == "ssh" {
+		gl.catchAll.protocol = "https"
+	}
+}
```
vladaionescu, line 83: So if no SSH agent is found or no keys are registered, we automatically switch to trying out https instead, right? Some questions:

1. If user does have valid SSH key, but their key is not registered with GitHub, is there a way in which we could try HTTPS in that situation?
2. What happens if the user's Earthly config for `github.com` says `auth: ssh`, but user doesn't have an SSH agent or no SSH key? Does the HTTPS fallback still kick in in this case?

buildkitd/buildkitd.go
```diff
@@ -217,32 +216,11 @@ func Start(ctx context.Context, image string, settings Settings, reset bool) err
...
 
-	for i, data := range settings.GitCredentials {
-		args = append(args,
-			"-e", fmt.Sprintf("GIT_CREDENTIALS_%d", i),
-		)
-		env = append(env,
-			fmt.Sprintf("GIT_CREDENTIALS_%d=%s", i, base64.StdEncoding.EncodeToString([]byte("password="+data))),
-		)
-	}
-
```
vladaionescu, line 245: So satisfying!

buildkitd/settings.go
```diff
@@ -11,14 +11,12 @@ import (
...
-	GitURLInsteadOf string   `json:"gitUrlInsteadOf"`
-	CacheSizeMb     int      `json:"cacheSizeMb"`
-	GitConfig       string   `json:"gitConfig"`
-	GitCredentials  []string `json:"gitCredentials"`
-	RunDir          string   `json:"runDir"`
-	Debug           bool     `json:"debug"`
-	DebuggerPort    int      `json:"debuggerPort"`
+	CacheSizeMb    int      `json:"cacheSizeMb"`
+	GitConfig      string   `json:"gitConfig"`
+	GitCredentials []string `json:"gitCredentials"`
```
vladaionescu, line 16: Can remove `GitConfig` and `GitCredentials` too?

cmd/earth/main.go
```diff
@@ -320,13 +321,13 @@ func newEarthApp(ctx context.Context, console conslogging.ConsoleLogger) *earthA
...
-			Usage:       "Rewrite git URLs of a certain pattern. Similar to git-config url.<base>.insteadOf (https://git-scm.com/docs/git-config#Documentation/git-config.txt-urlltbasegtinsteadOf). Multiple values can be separated by commas. Format: <base>=<instead-of>[,...]. For example: 'https://github.com/=git@github.com:'",
-			Destination: &app.buildkitdSettings.GitURLInsteadOf,
-		},
+		//&cli.StringFlag{
+		//	Name:        "git-url-instead-of",
+		//	Value:       "",
+		//	EnvVars:     []string{"GIT_URL_INSTEAD_OF"},
+		//	Usage:       "Rewrite git URLs of a certain pattern. Similar to git-config url.<base>.insteadOf (https://git-scm.com/docs/git-config#Documentation/git-config.txt-urlltbasegtinsteadOf). Multiple values can be separated by commas. Format: <base>=<instead-of>[,...]. For example: 'https://github.com/=git@github.com:'",
+		//	Destination: &app.buildkitdSettings.GitURLInsteadOf,
+		//},
```
vladaionescu, line 330: Does our CI work if we remove this? I assume that with the HTTPS fallback it should work.

Wondering if we should still support something like this for `GIT CLONE` command. But perhaps if that's the only need, the user can always just use a build arg, so instead-of is no longer necessary.

cmd/earth/main.go
```diff
@@ -612,38 +613,28 @@ func (app *earthApp) before(context *cli.Context) error {
...
+	//	// EvalSymlinks evaluates "" as "." which then breaks docker volume mounting
+	//	realSSHSocketPath, err := filepath.EvalSymlinks(app.buildkitdSettings.SSHAuthSock)
+	//	if err != nil {
+	//		if runtime.GOOS != "darwin" {
+	//			app.console.Warnf("failed to evaluate potential symbolic links in ssh auth socket %q: %v\n", app.buildkitdSettings.SSHAuthSock, err)
+	//		} // else ignore the error on mac
+	//	} else {
+	//		app.buildkitdSettings.SSHAuthSock = realSSHSocketPath
+	//	}
+	//}
```
vladaionescu, line 637: Might be better to just remove altogether

cmd/earth/main.go
```diff
@@ -1403,6 +1346,26 @@ func (app *earthApp) actionBuild(c *cli.Context) error {
...
+			if !strings.Contains(host, ".") {
+				host += ".com"
+			}
+			pattern = host + "/[^/]+/[^/]+"
+		}
+		err := gitLookup.AddMatcher(k, pattern, v.User, v.Password, v.Suffix, v.Auth, v.KeyScan)
+		if err != nil {
+			return errors.Wrap(err, "gitlookup")
+		}
+	}
```
vladaionescu, line 1367: Mind moving this part to a separate func tasked with constructing `gitLookup`? `actionBuild` is getting pretty big.

cmd/earth/main.go
```diff
@@ -1411,6 +1374,9 @@ func (app *earthApp) actionBuild(c *cli.Context) error {
 			return errors.Wrap(err, "ssh agent provider")
 		}
 		attachables = append(attachables, ssh)
+	} else {
+		app.console.Printf("No ssh auth socket detected; all git clone commands will use https\n")
+		gitLookup.DisableSSH()
```
vladaionescu, line 1379: :+1:

earthfile2llb/converter.go
```diff
@@ -528,7 +527,7 @@ func (c *Converter) GitClone(ctx context.Context, gitURL string, branch string,
 			"%sGIT CLONE (--branch %s) %s", c.vertexPrefixWithURL(gitURL), branch, gitURL),
 		llb.KeepGitDir(),
 	}
-	gitState := llbgit.Git(gitURL, branch, gitOpts...)
+	gitState := llb.Git(gitURL, branch, gitOpts...)
```
vladaionescu, line 530: Yay!

vladaionescu (commented):
This looks great!

Had a few questions...

buildcontext/git.go
alexcb: ah yes, good catch. I'll fix that now.

buildcontext/gitlookup.go
alexcb: 1) We only switch over to https when no agent is found, so unfortunately not; I think we would have to manually try a ssh to the host to see if it's accepted or not. Maybe we could try this in a future PR?

2) it will automatically get flipped back to https, so that might be confusing to a user. Maybe we should only auto-fall back when no auth is specified in the config?

buildkitd/buildkitd.go
alexcb: They'll be a bit of code to remove from the entrypoint.sh script too I suspect, but I want to leave that for another PR :)

cmd/earth/main.go
alexcb: The fallback works for our CI; however, if someone is using this to inject some other auth, then that'll have to be updated to use the ~/.earthly/config.yml file. Maybe we can reach out to a few users to check in first to get an idea of how their CI is configured for auth.

buildcontext/gitlookup.go
vladaionescu: 1. Yeah, that makes sense. I think that we should still keep instead-of in this case then - otherwise the experience of the user is less than what they have today for this specific edge case. Though this means that we have to implement the instead-of ourselves on the client side...
2. Yeah. Perhaps "auto" is the default and so if nothing is specified, "auto" allows us to flip to either.

cmd/earth/main.go
vladaionescu: I'm thinking we should support commands like git has for editing config: `git config --global ...`. It'll help for configuring containerized CIs like GH actions, circle etc.

eg
```
earth config set ssh/github.com/ssh=https
```

or something like that. Anyway... that's a separate/future matter.

buildcontext/gitlookup.go
alexcb: sounds great, just pushed up some new code for the auto mode.

buildkitd/buildkitd.go
```diff
@@ -219,6 +219,7 @@ func Start(ctx context.Context, image string, settings Settings, reset bool) err
 
 	args = append(args,
 		"-e", fmt.Sprintf("CACHE_SIZE_MB=%d", settings.CacheSizeMb),
+		"-e", fmt.Sprintf("GIT_URL_INSTEAD_OF=%s", settings.GitURLInsteadOf),
```
vladaionescu, line 222: I *think* this no longer works in the updated buildkit, because we use `git config` and the buildkit git commands no longer use the global config?

alexcb: I put in a change to disable ssh configs (see https://github.com/moby/buildkit/blob/master/source/git/gitsource.go#L580 ); but I'm unsure of any change that prevents the git config from being read. I'll need to test to see if the `git config` commands we use still work or not.

alexcb: I just gave it a test with git, and it looks like it's being ignored by buildkit now. for instance I ran:

    GIT_URL_INSTEAD_OF="git@fitlab.com=git@gitlab.com:" ./build/linux/amd64/earth -VD gitlab.com/alexcb/hello+docker

which worked, and caused this to get fetched:

    internal | --> GIT CLONE git@gitlab.com:alexcb/hello.git

however when I run git directly in the buildkit docker container, I can see

    $ docker exec -ti earthly-buildkitd cat /root/.gitconfig
    [url "git@fitlab.com"]
    	insteadOf = git@gitlab.com:

and the clone fails with:

    $ docker exec -ti earthly-buildkitd git clone git@gitlab.com:alexcb/hello.git
    Cloning into 'hello'...
    fatal: 'git@fitlab.comalexcb/hello.git' does not appear to be a git repository
    fatal: Could not read from remote repository.

vladaionescu: Perhaps a short-term solution is to not ignore the config for now in the earthly buildkit fork and deal with this later.

cmd/earth/main.go
alexcb: I really like that idea! that'll be a great way to allow us to remove the depricated env flags once that's in place.

buildkitd/buildkitd.go
alexcb: I added in HOME into the buildkit exec, so git will now pick up the global config: https://github.com/earthly/buildkit/commit/c1d00f6ed86584948c4563beab90964d19119046

alexcb:
I just rebased this against main, which now includes our private repo integration test; seeing it pass makes me more confident it won't break private repos.

PR #521, opened 2020-11-19
Allow accepting the Terms/Conditions and Privacy Policy via CLI arg

dchw:
Could also provide an interactive one during registration too; but this is a start...


cmd/earth/main.go
```diff
@@ -1172,6 +1172,30 @@ func (app *earthApp) actionRegister(c *cli.Context) error {
...
+	if !app.termsConditions {
+		rawAccept := promptInput("Do you accept the Earthly Privacy Policy? [y/N]")
+		if rawAccept == "" {
+			rawAccept = "n"
+		}
+		accept := strings.ToLower(rawAccept)[0]
+
+		interactiveTermsConditions = accept == 'y'
+	}
+	privacyPolicy := app.privacyPolicy || interactivePrivacyPolicy
```
vladaionescu, line 1197: I would make this a single prompt to simplify - `Do you accept Earthly's T&Cs <link> and Privacy Policy <link>?` Perhaps even the flag can be a single flag.

Storing separate columns in the DB makes sense though.

dchw: Yeah? I was just thinking that individual ones would let us re-prompt in the future for just that one. But thats not super common so we can cross that bridge when we get there

dchw: Also do we want to consolidate to one CLI arg/ env var?

vladaionescu: I think re-prompting would anyway require some special logic... we might do it on login instead of register. We might just do it over email and that's it. We'll figure it out when we get there.

vladaionescu: > Also do we want to consolidate to one CLI arg/ env var?

Yeah, I would do that for simplicity.

dchw: :+1:

cmd/earth/main.go
```diff
@@ -1158,6 +1165,19 @@ func (app *earthApp) actionRegister(c *cli.Context) error {
...
+	var interactiveAccept bool
+	if !app.termsConditionsPrivacy {
+		// TODO: add link when we have one
+		rawAccept := promptInput("Do you accept Earthly's T&Cs <link> and Privacy Policy <link>? [y/N]")
+		if rawAccept == "" {
+			rawAccept = "n"
+		}
+		accept := strings.ToLower(rawAccept)[0]
+
+		interactiveAccept = accept == 'y'
```
alexcb, line 1177: I like how this will work with "yup", "yeah", "yar", and "yucky".

alexcb (approved):

cmd/earth/main.go
dchw: Yeah I didn't want to whitelist affirmative answers so we are here. I can hardcode it to yes if you want

alexcb: might be best to check for "y" or "yes". But I don't feel strongly one way or the other.

vladaionescu (approved):

PR #524, opened 2020-11-20
Don't retry auth failures

alexcb:
Signed-off-by: Alex Couture-Beil <alex@mofo.ca>

### DIFF
--- a/secretsclient/client.go
+++ b/secretsclient/client.go
@@ -121,17 +121,17 @@ func (c *client) doCall(method, url string, opts ...requestOpt) (int, string, er
 	duration := time.Millisecond * 100
 	for attempt := 0; attempt < maxAttempt; attempt++ {
 		status, body, err = c.doCallImp(r, method, url, opts...)
-		if err == nil && status < 500 {
+		if (err == nil && status < 500) || err == ErrNoAuthorizedPublicKeys {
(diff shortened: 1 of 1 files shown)


secretsclient/client.go
```diff
@@ -121,17 +121,17 @@ func (c *client) doCall(method, url string, opts ...requestOpt) (int, string, er
 	duration := time.Millisecond * 100
 	for attempt := 0; attempt < maxAttempt; attempt++ {
 		status, body, err = c.doCallImp(r, method, url, opts...)
-		if err == nil && status < 500 {
+		if (err == nil && status < 500) || err == ErrNoAuthorizedPublicKeys || err == ErrUnauthorized {
```
vladaionescu, line 124: Aren't these 4xx errors already?

vladaionescu (approved):

secretsclient/client.go
alexcb: the ErrNoAuthorizedPublicKeys occurs once https://github.com/earthly/earthly/blob/main/secretsclient/client.go#L398-L408 exhausts all potential keys to try authenticating with.

alexcb: ErrUnauthorized isn't needed here, I can remove that.

PR #526, opened 2020-11-20
Add help usage text to secrets commands

alexcb:

### DIFF
--- a/cmd/earth/main.go
+++ b/cmd/earth/main.go
@@ -454,20 +454,28 @@ func newEarthApp(ctx context.Context, console conslogging.ConsoleLogger) *earthA
 			Hidden: true,
 			Subcommands: []*cli.Command{
 				{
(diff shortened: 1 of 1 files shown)


cmd/earth/main.go
```diff
@@ -454,20 +454,28 @@ func newEarthApp(ctx context.Context, console conslogging.ConsoleLogger) *earthA
 			Hidden: true,
 			Subcommands: []*cli.Command{
 				{
-					Name:   "create",
-					Action: app.actionOrgCreate,
+					Name:      "create",
+					Usage:     "create a new organization",
```
vladaionescu, line 458: Should use the same style as other usage fields: `Create a new organization`

alexcb: good call, fixed.

vladaionescu (approved):

PR #528, opened 2020-11-20
automatically prefix secrets paths when listing permissions

alexcb:


vladaionescu (approved):

vladaionescu:
Wondering if we shouldn't do this. Maybe relative paths could have some special meaning in the future?

alexcb:
Perhaps we should explicitly require the user to name the org in the same way they grant permissions and set/get secrets. I'll change the output of list-orgs to prefix the org names with a / to make it consistent the other way then.

Issue #529, opened 2020-11-20
Add timestamps to output


agbell:
It would be helpful if the verbose flag added timestamps to the output to detect where there are performance improvements needed in an earthfile.  

The equivalent of this
```earth --verbose +target | ts -s ```

which produces output like the:
```
00:00:00            buildkitd | Found buildkit daemon as docker container (earthly-buildkitd)
00:00:00 h/elixir:1.11.1-erlang-21.3.8.18-alpine-3.12.0 | --> Load metadata
00:00:00               +setup | --> FROM hexpm/elixir:1.11.1-erlang-21.3.8.18-alpine-3.12.0
00:00:00              context | --> local context .
00:00:00               +setup | resolve docker.io/hexpm/elixir:1.11.1-erlang-21.3.8.18-alpine-3.12.0@sha256:78342f807374a20c717bae7fcbe5daea3ba9eb38a1771c719e276abf7ebd125c 100%
00:00:01              context | transferring .: 100%
00:00:01               +setup | *cached* --> WORKDIR /src
00:00:01               +setup | *cached* --> RUN apk add --no-progress --update git
00:00:01               +setup | *cached* --> COPY mix.exs .
00:00:01               +setup | *cached* --> COPY mix.lock .
00:00:01               +setup | *cached* --> COPY .formatter.exs .
00:00:01               +setup | *cached* --> COPY package.json .
00:00:01               +setup | *cached* --> RUN mix local.rebar --force
00:00:01               +setup | *cached* --> RUN mix local.hex --force
00:00:01               +setup | *cached* --> RUN mix deps.get
```

Issue #496, opened 2020-11-10
Build performance metrics


jazzdan:
When I'm writing an Earthfile for a project I'm always on the look out for what the biggest build bottlenecks are. Should I break out that build script in to smaller targets? Should I optimize that Dockerfile by using a smaller base image?

If Earthly could show me a detailed breakdown of how long each of my build steps take that would help me figure out what's important to focus on. As of right now I have to rely on timing each step with my eyeballs, and it can be really hard to tell when there are multiple builds happening in parallel.

vladaionescu:
This should be relatively easy - BuildKit already exposes some of this data. We just need to take it and report it to the user neatly.

PR #532, opened 2020-11-21
Print timing info on verbose

vladaionescu:
Fixes https://github.com/earthly/earthly/issues/529
Fixes https://github.com/earthly/earthly/issues/496

Prints elapsed time after completing each command and also a summary at the end for total time spent on each target.

Screenshot from 2020-11-20 16-47-01
Screenshot from 2020-11-20 16-47-14

CC @agbell @dchw

### DIFF
--- a/builder/solver_monitor.go
+++ b/builder/solver_monitor.go
@@ -302,6 +302,11 @@ func (sm *solverMonitor) recordTiming(targetStr, targetBrackets, salt string, ve
 	sm.timingTable[key] += dur
 }
 
+type durationAndKey struct {
+	dur time.Duration
+	key timingKey
+}
+
 func (sm *solverMonitor) PrintTiming() {
 	if !sm.verbose {
 		return
@@ -311,24 +316,17 @@ func (sm *solverMonitor) PrintTiming() {
 		Printf("Summary of timing information\n" +
 			"Note that the times do not include the expansion of commands like BUILD, FROM, COPY (artifact).")
 	var total time.Duration
-	durs := make([]time.Duration, 0, len(sm.timingTable))
-	durMap := make(map[time.Duration][]timingKey)
+	durs := make([]durationAndKey, 0, len(sm.timingTable))
 	for key, dur := range sm.timingTable {
-		durs = append(durs, dur)
-		durMap[dur] = append(durMap[dur], key)
+		durs = append(durs, durationAndKey{
+			d:   dur,
+			key: key,
+		})
 		total += dur
 	}
 	sort.Slice(durs, func(i, j int) bool {
-		return durs[i] > durs[j]
+		return durs[i].dur > durs[j].dur
 	})
-	for _, dur := range durs {
-		for _, key := range durMap[dur] {
-			sm.console.
-				WithPrefixAndSalt(key.targetStr, key.salt).
-				WithMetadataMode(true).
-				Printf("(%s) %s\n", key.targetBrackets, dur)
-		}
-	}
 	sm.console.
 		WithMetadataMode(true).
 		Printf("===============================================================\n")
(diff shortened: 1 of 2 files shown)


builder/solver_monitor.go
```diff
@@ -264,6 +287,60 @@ func (sm *solverMonitor) printHeader(vm *vertexMonitor) {
...
+	dur := vertex.Completed.Sub(*vertex.Started)
+	if dur == 0 {
+		return
+	}
+	key := timingKey{
+		targetStr:      targetStr,
+		targetBrackets: targetBrackets,
+		salt:           salt,
+	}
+	sm.timingTable[key] = sm.timingTable[key] + dur
```
alexcb, line 303: not that it really matters, but is it not possible to do this instead?
```suggestion
	sm.timingTable[key] += dur
```

builder/solver_monitor.go
```diff
@@ -264,6 +287,60 @@ func (sm *solverMonitor) printHeader(vm *vertexMonitor) {
...
+	if !sm.verbose {
+		return
+	}
+	sm.console.
+		WithMetadataMode(true).
+		Printf("Summary of timing information\n" +
+			"Note that the times do not include the expansion of commands like BUILD, FROM, COPY (artifact).")
+	var total time.Duration
+	durs := make([]time.Duration, 0, len(sm.timingTable))
+	durMap := make(map[time.Duration][]timingKey)
```
alexcb, line 316: I don't understand why the code is bucketing this by duration? is there a reason we can't just sort sm.timingTable directly (or make a copy of it, then sort it)? Then when it's printed to the console, one would only need to iterate over a single for loop rather than have two nested loops.

vladaionescu: Not really... maps in Go are not sorted. So we have to collect values in a slice and sort that slice and then reverse lookup. Don't know if you have something else in mind

builder/solver_monitor.go
vladaionescu: :+1:

builder/solver_monitor.go
vladaionescu: The double for loop is only for cases where multiple `timingKey`s have the exact same duration (very rare). It's not really inefficient.

alexcb: Ah, I didn't realise `sm.timingTable` was a map; I was wondering if we could do something like this instead? https://github.com/earthly/earthly/pull/533/files

I found the use of durMap hard to follow.

alexcb: full disclosure, I haven't tested that code; I'm also fine if you want to keep your code as-is.

vladaionescu: Yeah, much nicer. I would also move

```
type durationAndKey struct {
	dur time.Duration
	key timingKey
}
```

within the function itself since it's only used there.

alexcb:
Sorry I missed a wee bit in that example branch of mine:
```
alex@mah:~/gh/earthly/earthly$ git diff
diff --git a/builder/solver_monitor.go b/builder/solver_monitor.go
index a04a27e..096c73c 100644
--- a/builder/solver_monitor.go
+++ b/builder/solver_monitor.go
@@ -328,6 +328,12 @@ func (sm *solverMonitor) PrintTiming() {
        sort.Slice(durs, func(i, j int) bool {
                return durs[i].dur > durs[j].dur
        })
+       for _, x := range durs {
+               sm.console.
+                       WithPrefixAndSalt(x.key.targetStr, x.key.salt).
+                       WithMetadataMode(true).
+                       Printf("(%s) %s\n", x.key.targetBrackets, x.dur)
+       }
        sm.console.
                WithMetadataMode(true).
                Printf("===============================================================\n")
```

vladaionescu:
It should be good now

alexcb (approved):

PR #534, opened 2020-11-22
integration test clonning remote private repos

alexcb:
- makes use of secrets service to fetch shared-ssh key
- uses shared ssh key to clone from a private repo

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,6 +65,21 @@ jobs:
         run: earth +for-linux
       - name: Build examples
         run: ./build/linux/amd64/earth --no-output --allow-privileged +examples
+  secrets:
+    name: secrets-integration
+    runs-on: ubuntu-latest
+    env:
+      GIT_URL_INSTEAD_OF: "https://github.com/=git@github.com:"
+      FORCE_COLOR: 1
+      EARTHLY_TOKEN: "${{ secrets.EARTHLY_TOKEN }}"
+    steps:
+      - uses: actions/checkout@v2
+      - name: Download released earth
+        run: "sudo /bin/sh -c 'wget https://github.com/earthly/earthly/releases/latest/download/earth-linux-amd64 -O /usr/local/bin/earth && chmod +x /usr/local/bin/earth'"
+      - name: Build latest earth using released earth
+        run: earth +for-linux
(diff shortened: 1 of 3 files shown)


examples/tests/private-repo.sh
```diff
@@ -0,0 +1,75 @@
...
+
+# Test a private repo can be cloned using https, as setup via args
+echo === Test 4 ===
+
+docker image rm -f other-test-private:latest
+SSH_AUTH_SOCK="" $earth -VD --git-username=cinnamonthecat --git-password="$GITHUB_PASSWORD" github.com/cinnamonthecat/another-test-private:main+docker
+docker run --rm another-test-private:latest | grep "Hola Mundo"
+
+
+echo === All private-repo.sh tests have passed ===
```
vladaionescu, line 75: Add this script to shellcheck?

examples/tests/private-repo.sh
```diff
@@ -0,0 +1,75 @@
...
+$earth account login
+
+# fetch shared secret key (this step assums your personal user has access to the /earthly-technologies/ secrets org
+ID_RSA=$($earth secrets get -n /earthly-technologies/github/other-service+github-cinnamon@earthly.dev/id_rsa)
+GITHUB_PASSWORD=$($earth secrets get -n /earthly-technologies/github/other-service+github-cinnamon@earthly.dev/password)
+
+# now that we grabbed the cinnamonthecat credentials, unset our token, to ensure that we're only testing using cinnamonthecat's credentials
+unset EARTHLY_TOKEN
+
+echo starting new instance of ssh-agent, and loading c
```
vladaionescu, line 21: credentials? :-)

vladaionescu (approved):
Really cool stuff!

examples/tests/private-repo.sh
alexcb: lol, good guess!

examples/tests/private-repo.sh
alexcb: I fought with this far too long, turns out that `.earthignore` ignores all of the examples directory, which is preventing me from copying it into the `+lint-scripts` target; any thoughts on where I should move this test to, or if we should change the `.earthignore` to include examples (probably a bad idea since the examples are rather large).

PR #543, opened 2020-11-25
Account mgmt

alexcb:
This implements commands for managing public keys and auth tokens

```
COMMANDS:
   register      Register for an earthly account
   list-keys     List associated public keys used for authentication
   add-key       Associate a new public key with your account
   remove-key    Removes an existing public key from your account
   list-tokens   List associated tokens used for authentication
   create-token  Create a new authentication token for your account
   remove-token  Remove an authentication token from your account
```


cmd/earth/main.go
```diff
@@ -541,38 +535,128 @@ func newEarthApp(ctx context.Context, console conslogging.ConsoleLogger) *earthA
...
-			Name:        "register",
-			Usage:       "Register for an earthly account",
-			Description: "Register for an earthly account",
-			UsageText: "first, request a token with:\n" +
-				"     earth [options] register --email <email>\n" +
-				"\n" +
-				"   then check your email to retrieve the token, then continue by running:\n" +
-				"     earth [options] register --email <email> --token <token>",
+			Name:   "account",
+			Usage:  "Create or manage an earthly account",
```
vladaionescu, line 539:
```suggestion
			Usage:  "Create or manage an Earthly account",
```

cmd/earth/main.go
```diff
@@ -955,7 +1039,7 @@ func (app *earthApp) actionOrgCreate(c *cli.Context) error {
 		return errors.New("invalid number of arguments provided")
 	}
 	org := c.Args().Get(0)
-	sc := secretsclient.NewClient(app.apiServer, app.sshAuthSock, app.publicKey, app.console.Warnf)
+	sc, _ := secretsclient.NewClient(app.apiServer, app.sshAuthSock, app.console.Warnf)
```
vladaionescu, line 1042: Add a comment about why error is ignored?

cmd/earth/main.go
```diff
@@ -1163,13 +1243,17 @@ func (app *earthApp) actionRegister(c *cli.Context) error {
...
+	if app.sshAuthSock != "" {
+		var err error
+		publicKeys, err = sc.GetPublicKeys()
+		if err != nil {
+			return err
+		}
 	}
+	//if len(publicKeys) == 0 {
+	//	return fmt.Errorf("failed to find any public keys; did you forget to ssh-add your key?")
+	//}
```
vladaionescu, line 1256: Remove altogether?

cmd/earth/main.go
```diff
@@ -1237,6 +1323,249 @@ func (app *earthApp) actionRegister(c *cli.Context) error {
...
+	for _, k := range c.Args().Slice() {
+		err := sc.RemovePublickKey(k)
+		if err != nil {
+			return errors.Wrap(err, "failed to add public key to account")
+		}
+	}
+	return nil
+}
+func (app *earthApp) actionAccountListTokens(c *cli.Context) error {
+	app.commandName = "accounListTokens"
```
vladaionescu, line 1401:
```suggestion
	app.commandName = "accountListTokens"
```

cmd/earth/main.go
```diff
@@ -1237,6 +1323,249 @@ func (app *earthApp) actionRegister(c *cli.Context) error {
...
+			fmt.Fprintf(w, "\tr")
+		}
+		fmt.Fprintf(w, "\t%s UTC", token.Expiry.UTC().Format("2006-01-02T15:04"))
+		fmt.Fprintf(w, "\n")
+	}
+	w.Flush()
+	return nil
+}
+func (app *earthApp) actionAccountCreateToken(c *cli.Context) error {
+	app.commandName = "accounCreateToken"
```
vladaionescu, line 1424:
```suggestion
	app.commandName = "accountCreateToken"
```

cmd/earth/main.go
```diff
@@ -1237,6 +1323,249 @@ func (app *earthApp) actionRegister(c *cli.Context) error {
...
+			return err
+		}
+		fmt.Printf("Logged in as %q using password auth\n", email)
+		fmt.Printf("Warning unencrypted password has been stored under ~/.earthly/auth.token; consider using ssh-based auth to prevent this.\n")
+	}
+	return nil
+}
+
+func (app *earthApp) actionAccountLogout(c *cli.Context) error {
+	app.commandName = "actionAccountLogout"
```
vladaionescu, line 1557:
```suggestion
	app.commandName = "accountLogout"
```

(consistency)

secretsclient/client.go
```diff
@@ -629,3 +671,411 @@ func (c *client) ListOrgs() ([]*OrgDetail, error) {
...
+			return errors.Wrap(err, "base64 decode failed")
+		}
+	case "token":
+		c.authToken = authData
+	default:
+		c.warnFunc("unable to handle cached auth type %s", authType)
+		break
+	}
+	return nil
+}
```
vladaionescu, line 893: We should encrypt password / token so they are not stored in the clear on the disk. In any case, this can be a future thing.

(Docker uses this credentialsStore: https://docs.docker.com/engine/reference/commandline/login/#credentials-store which attempts to use common tools / key chains)

For now, we can recommend users to use SSH key + passphrase if they want to avoid this.

vladaionescu (commented):
This looks pretty good. It's pretty linear - I don't think a split is necessary.

cmd/earth/main.go
vladaionescu: Can you also double check that these are correct in all other places? Don't know if I missed any.

alexcb: thanks for pointing that out, found a few bad copy/paste places that caused this.

secretsclient/client.go
alexcb: As discussed in our chat, we'll do our best to prompt the user to add an ssh key wherever possible, and print a warning in red whenever we revert to caching the password on disk. In a future PR we can also extend the server to return some sort of signed auth token similar to the credentials-store.

cmd/earth/main.go
alexcb: errors are now correctly handled.

vladaionescu (approved):

PR #551, opened 2020-12-02
Print warning messages to stderr rather than stdout

alexcb:
- when fetching secrets, if a warning occurs it's printed to stdout
rather than stderr, which breaks accessing secrets via the cli.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,7 +44,7 @@ jobs:
       - name: "Experimental tests"
         run: ./build/linux/amd64/earth -P ./examples/tests+experimental
       - name: Execute test similar to homebrew test in https://github.com/Homebrew/homebrew-core/blob/master/Formula/earthly.rb
-        run: ./build/linux/amd64/earth --buildkit-host 127.0.0.1 ./examples/tests/with-docker+all | grep 'Error while dialing invalid address 127.0.0.1'
+        run: ./build/linux/amd64/earth --buildkit-host 127.0.0.1 ./examples/tests/with-docker+all 2>&1 | grep 'Error while dialing invalid address 127.0.0.1'
       - name: Docker Login
         run: docker login --username "${{ secrets.DOCKERHUB_USERNAME }}" --password "${{ secrets.DOCKERHUB_TOKEN }}"
         if: github.event.pull_request.head.repo.full_name == github.repository
--- a/conslogging/conslogging.go
+++ b/conslogging/conslogging.go
@@ -51,7 +51,8 @@ type ConsoleLogger struct {
 	mu             *sync.Mutex
 	saltColors     map[string]*color.Color
 	nextColorIndex *int
-	w              io.Writer
+	outW           io.Writer
+	errW           io.Writer
 	trailingLine   bool
 	prefixPadding  int
 }
@@ -59,7 +60,8 @@ type ConsoleLogger struct {
 // Current returns the current console.
 func Current(colorMode ColorMode, prefixPadding int) ConsoleLogger {
 	return ConsoleLogger{
-		w:              os.Stdout,
+		outW:           os.Stdout,
+		errW:           os.Stderr,
 		colorMode:      colorMode,
 		saltColors:     make(map[string]*color.Color),
 		nextColorIndex: new(int),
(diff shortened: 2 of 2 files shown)


conslogging/conslogging.go
```diff
@@ -130,17 +133,17 @@ func (cl ConsoleLogger) WithFailed(isFailed bool) ConsoleLogger {
...
-	cl.color(successColor).Fprintf(cl.w, "=========================== SUCCESS ===========================\n")
+	cl.color(successColor).Fprintf(cl.outW, "=========================== SUCCESS ===========================\n")
 }
 
 // PrintFailure prints the failure message.
 func (cl ConsoleLogger) PrintFailure() {
 	cl.mu.Lock()
 	defer cl.mu.Unlock()
-	cl.color(warnColor).Fprintf(cl.w, "=========================== FAILURE ===========================\n")
+	cl.color(warnColor).Fprintf(cl.outW, "=========================== FAILURE ===========================\n")
```
alexcb, line 143: we could optionally print this to errW (stderr) too, what do you think?

conslogging/conslogging.go
```diff
@@ -222,17 +225,17 @@ func (cl ConsoleLogger) printPrefix() {
 		*cl.nextColorIndex = (*cl.nextColorIndex + 1) % len(availablePrefixColors)
 	}
 	c = cl.color(c)
-	c.Fprintf(cl.w, cl.prettyPrefix())
+	c.Fprintf(cl.outW, cl.prettyPrefix())
 	if cl.isFailed {
-		cl.w.Write([]byte(" *"))
-		cl.color(warnColor).Fprintf(cl.w, "failed")
-		cl.w.Write([]byte("*"))
+		cl.outW.Write([]byte(" *"))
```
alexcb, line 230: same comment as earlier regarding printing this failure to stderr too?

.github/workflows/ci.yml
```diff
@@ -44,7 +44,7 @@ jobs:
       - name: "Experimental tests"
         run: ./build/linux/amd64/earth -P ./examples/tests+experimental
       - name: Execute test similar to homebrew test in https://github.com/Homebrew/homebrew-core/blob/master/Formula/earthly.rb
-        run: ./build/linux/amd64/earth --buildkit-host 127.0.0.1 ./examples/tests/with-docker+all | grep 'Error while dialing invalid address 127.0.0.1'
+        run: ./build/linux/amd64/earth --buildkit-host 127.0.0.1 ./examples/tests/with-docker+all 2>&1 | grep 'Error while dialing invalid address 127.0.0.1'
```
alexcb, line 47: luckily homebrew already combines stderr and stdout in https://github.com/homebrew/homebrew-core/blob/master/Formula/earthly.rb#L41

conslogging/conslogging.go
```diff
@@ -151,7 +154,7 @@ func (cl ConsoleLogger) Warnf(format string, args ...interface{}) {
 
 	for _, line := range strings.Split(text, "\n") {
 		cl.printPrefix()
-		c.Fprintf(cl.w, "%s\n", line)
+		c.Fprintf(cl.errW, "%s\n", line)
```
vladaionescu, line 157: `printPrefix` should take an argument for the writer to use so that warnings with a prefix get both the warning and the prefix written to stderr.

conslogging/conslogging.go
vladaionescu: Not sure what you mean here. I think printing the failed output to stderr could be interesting. 🤔

alexcb: I need to have calls to `Warnf()` go to stderr rather than stdout. Because without that change, I can't do:

```
MY_KEY=$(earth secrets get /user/secret_key)
```

I made the change in this PR to fix that issue, and I was wondering if we should also switch the errors that occur during a regular build to also go to stderr too. I'm fine leaving it as stdout as I just need to fix `warnf` to unblock progress with accessing secrets.

conslogging/conslogging.go
alexcb: good catch, I just pushed up a fix for this.

conslogging/conslogging.go
vladaionescu: Sounds good to me either way. No strong opinion.

vladaionescu (approved):

PR #553, opened 2020-12-03
Optionally use cloud secrets for Docker Hub in integration tests

vladaionescu:
Also,

* Use `docker login` in GH actions as much as possible
* Allow optional secrets via empty secret ref
* Fix an issue with propagating build args

------------------
TODO:

- [x] Test/fix all CIs
- [x] If external contribution, allow GH actions to run without credentials
- [x] Add doc mention for optional secret


alexcb (approved):
very nice addition

PR #557, opened 2020-12-03
Switch windows pipeline to use wsl2 buildkite agent

vladaionescu:
CC @alexcb FYI

We now use the Linux buildkite agent in Windows WSL2, rather than the Windows buildkite agent, outside of wsl2.


alexcb (approved):

dchw:
Lets go!

PR #562, opened 2020-12-04
Ignore unknown json fields when unmarshalling

alexcb:
- without this change, when the server returns a new field, the client
will reject it, thus breaking backwards-compatibility.


dchw (approved):
LGTM

PR #566, opened 2020-12-04
Display if auth token is read-only

alexcb:


alexcb:
This PR requires our server to be updated before we merge.

vladaionescu (approved):

PR #573, opened 2020-12-04
Faster detection when dockerd fails to start. Increase timeout to make it more stable on WSL

vladaionescu:


alexcb (approved):
nice speedup

PR #592, opened 2020-12-07
Fix interactive mode in WITH DOCKER

vladaionescu:
There was a conflict between host docker's IP address and the inner docker IP address. This caused the interactive debugger to not be able to find the right IP when it wanted to connect to the debug proxy.


alexcb (approved):
nice find!

Issue #593, opened 2020-12-07
Proposal: Global ARGs


vladaionescu:
## Background

@ido-namely reports on Gitter:

> I have different targets building different images, and I'd like them all to use the same docker registry address.

> I was hoping that ARG can be shared

```
ARG TST="hello world"
FROM golang:1.13-alpine3.11

test-target:
    ARG TST
    RUN echo $TST
```

## Proposal

* All args declared in the `+base` target are "global" ARGs. Meaning that they are passed throughout the other targets even if they don't inherit from `+base`. So the following should print "hello world":

```
ARG TST="hello world"
FROM golang:1.13-alpine3.11

test-target:
    RUN echo $TST
```

## Possible future extension

Could perhaps prevent this behavior if needed with a `ARG --local TST="hello world"`

## Alternatives considered

* A possibility might be that `ARG`s declared in the `+base` target are global and shared with the rest of the targets, if there is no `FROM` and they inherit automatically from `+base`. This would be inconsistent when chaining targets.
* Another option could be to always propagate build args too on a `FROM`. This may be deviating too much from `Dockerfile` behavior.
* Yet another option is to have a special arg `ARG --global` for such situations.

vladaionescu:
In `v0.4.0`

PR #596, opened 2020-12-08
Implementation for global ARGs

vladaionescu:
Re https://github.com/earthly/earthly/issues/593


variables/collection.go
```diff
@@ -134,7 +137,7 @@ func (c *Collection) SortedOverridingVariables() []string {
 
 // AddActive adds and activates a variable in the collection. It returns the effective variable. The
 // effective variable may be different from the one being added, when override is false.
-func (c *Collection) AddActive(name string, variable Variable, override bool) Variable {
+func (c *Collection) AddActive(name string, variable Variable, override bool, global bool) Variable {
```
alexcb, line 140: nit: ```suggestion
func (c *Collection) AddActive(name string, variable Variable, override, global bool) Variable {
```

alexcb (approved):

PR #599, opened 2020-12-08
Pretty print some errors

### LINKED ISSUE #537: Error messages could be less verbose and more clear
This error:
>Error: build target: build main: bkClient.Build: failed to solve: rpc error: code = Unknown desc = syntax error: line 1:0 no viable alternative at input 'build:'
would be more understandable if it were merely
> Error 1:0 no viable alternative at input 'build:

cc @JensRantil

dchw:
This changes how the earth command outputs the gRPC errors.

Example 1:

Original: `Error: build target: build main: bkClient.Build: failed to solve: rpc error: code = Unknown desc = syntax error: line 1:0 no viable alternative at input 'no-artifact:'`
Cleaned: `Error: line 1:0 no viable alternative at input 'no-artifact:'`

Example 2:

Original: `Error: build target: build main: bkClient.Build: failed to solve: rpc error: code = Unknown desc = parse: Invalid command LOL`
Cleaned: `Error: Invalid command LOL`

Fixes #537

### DIFF
--- a/cmd/earth/main.go
+++ b/cmd/earth/main.go
@@ -975,7 +975,7 @@ func (app *earthApp) run(ctx context.Context, args []string) int {
 				"Check your git auth settings.\n" +
 					"Did you ssh-add today? Need to configure ~/.earthly/config.yml?\n" +
 					"For more information see https://docs.earthly.dev/guides/auth\n")
-		} else if rpcRegex.MatchString(err.Error()) {
+		} else if !app.verbose && rpcRegex.MatchString(err.Error()) {
 			baseErr := errors.Cause(err)
 			baseErrMsg := rpcRegex.ReplaceAll([]byte(baseErr.Error()), []byte(""))
 
--- a/cmd/earth/main.go
+++ b/cmd/earth/main.go
@@ -963,6 +963,9 @@ func (app *earthApp) insertZSHCompleteEntry() error {
 
 func (app *earthApp) run(ctx context.Context, args []string) int {
 	err := app.cliApp.RunContext(ctx, args)
+
+	rpcRegex := regexp.MustCompile(`rpc error: code = .+ desc = .+:\s`)
+
 	if err != nil {
 		if strings.Contains(err.Error(), "security.insecure is not allowed") {
 			app.console.Warnf("Error: --allow-privileged (-P) flag is required\n")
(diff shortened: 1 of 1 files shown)


cmd/earth/main.go
```diff
@@ -972,6 +975,11 @@ func (app *earthApp) run(ctx context.Context, args []string) int {
 				"Check your git auth settings.\n" +
 					"Did you ssh-add today? Need to configure ~/.earthly/config.yml?\n" +
 					"For more information see https://docs.earthly.dev/guides/auth\n")
+		} else if rpcRegex.MatchString(err.Error()) {
+			baseErr := errors.Cause(err)
+			baseErrMsg := rpcRegex.ReplaceAll([]byte(baseErr.Error()), []byte(""))
+
+			app.console.Warnf("Error: %v\n", string(baseErrMsg))
 		} else {
 			app.console.Warnf("Error: %v\n", err)
```
dchw, line 984: I see maybe we could also do `errors.Cause(err)` here.

cmd/earth/main.go
```diff
@@ -972,6 +975,11 @@ func (app *earthApp) run(ctx context.Context, args []string) int {
 				"Check your git auth settings.\n" +
 					"Did you ssh-add today? Need to configure ~/.earthly/config.yml?\n" +
 					"For more information see https://docs.earthly.dev/guides/auth\n")
+		} else if rpcRegex.MatchString(err.Error()) {
```
vladaionescu, line 978: Wondering if using something like this would be better for unwrapping gRPC errors: https://stackoverflow.com/a/52972944

dchw:
Per @vladaionescu's comments, I will not truncate things when `--verbose` is supplied. Change for that is still incoming.

vladaionescu (commented):
Looks good - can you show a sample output?

cmd/earth/main.go
dchw: Maybe? All the status are `unknown`, at least that I can seem to trigger? Is there a gRPC error parser to pull out the error message?

We would still need to know that it is a gRPC error in the first place; perhaps then just checking for `rpc error:` would be best, assuming theres a built in way to pull out the error text.

vladaionescu: Based on the SO link, I think you can do
```
e, ok := status.FromError(err)
if ok {
  ...
}
```
to check if it's gRPC error and then
```
e.Message()
```
to grab the message.

dchw: Yeah I tried that. Doesn't pick it up as a gRPC message; FWIW I am passing in the error here from `errors.Cause(err)` to ensure its the gRPC error.

```
               if st, ok := status.FromError(errors.Cause(err)); ok {
			app.console.Warnf("YES GRPC:\n")
			app.console.Warnf("Error: %v\n", st.Message())
		} else if st, ok := status.FromError(errors.Cause(err)); !ok {
			app.console.Warnf("NO GRPC:\n")
			app.console.Warnf("Error: %v\n", st.Message())
		} 
```

Ends up printing this on my current test case:

```
NO GRPC:
Error: rpc error: code = Unknown desc = syntax error: line 1:0 no viable alternative at input 'no-artifact:'
```

vladaionescu: Ah ok. Then I think this looks good.

cmd/earth/main.go
vladaionescu: That could be a good idea, if `--verbose` is not passed

cmd/earth/main.go
dchw: OK. Also tried it with removing the `rpc error:` prefix... still no dice

vladaionescu (approved):

PR #601, opened 2020-12-08
Allow no newlines before targets

### LINKED ISSUE #538: Error if no base and no blank line
The following earthfile produces an error:
```
build:
    FROM golang:1.13-alpine3.11
    WORKDIR /demo
    COPY go.mod go.sum main.go ./
    RUN go build -o demo .
    ENTRYPOINT ["/demo/demo"]
    SAVE IMAGE example:latest
```
If a blank line is added at the top of the file, the error goes away.

dchw:
For Earthfiles without any `+base`; a (newline/comment) was required before the first target. This eliminates that requirement. This is now valid:

```
no-artifact:
    FROM alpine
    WORKDIR /usr/build
    RUN touch hi
    SAVE ARTIFACT hi
```

Fixes #538


alexcb (approved):
looks like an easy fix

earthfile2llb/parser/EarthParser.g4
```diff
@@ -4,7 +4,7 @@ options {
 	tokenVocab = EarthLexer;
 }
 
-earthFile: NL* stmts? NL+ targets? NL* EOF;
+earthFile: NL* stmts? NL* targets? NL* EOF;
```
vladaionescu, line 7: I think we still need to force a NL if base `stmts` are present.

```suggestion
earthFile: NL* (stmts NL)? NL* targets? NL* EOF;
```

dchw:
@alexcb The hard part was ANTLR. Never used it before.

vladaionescu (commented):
Good find

earthfile2llb/parser/EarthParser.g4
dchw: This is to enforce a new line between `+base` and `+my-target` right? Just checking for my own understanding

```
FROM xyz
RUN foo

my-target:
    RUN do-it.sh
```

vladaionescu: not quite.. it's to prevent this (no new line at all):

```
FROM xyz
RUN foo my-target:
    RUN do-it.sh
```

vladaionescu:
Do you want to also add a super simple integration test? Take a look at `examples/tests/Earthfile` and add one more.

earthfile2llb/parser/EarthParser.g4
dchw: Hm I tried that with the old way and it still failed. Willing to go along with it though

dchw:
> Do you want to also add a super simple integration test? Take a look at `examples/tests/Earthfile` and add one more.

Sure I can add another

vladaionescu (approved):

PR #603, opened 2020-12-09
Return an error whe the artifact to save locally doesnt exist

### LINKED ISSUE #542: `earth --artifact` doesn't fail when copying artifacts that don't exist
Full repro case [here](https://github.com/jazzdan/earthly-artifact-command)

## Repro steps
1. Clone [this repo](https://github.com/jazzdan/earthly-artifact-command) and `cd` in to it
2. `earth --artifact +no-artifact/foo /tmp/stuff`

### What I would expect to happen
I would expect to get some error that says that foo doesn't exist

### What happens instead
The command appears to succeed, but there's nothing in `/tmp/stuff`

dchw:
Fixes #542

We never checked if the thing we wanted to copy actually existed before we tried. The loop was never entered, so it was really just a no-op.

I don't know everything this affects, but seems like the root place to do this kind of check?


vladaionescu (commented):
Looks good! Add integration test too?

alexcb (approved):

vladaionescu (approved):

dchw:
Made an additional change to handle `SAVE LOCAL` in our sample remote target used in some integration tests.

alexcb (approved):

PR #610, opened 2020-12-11
Nobody asked for this, but it exists now

dchw:
I saw an article about a new GNU Cobol release and thought it was a travesty that we weren't on the COBOL train yet


vladaionescu (approved):
You've got to be kidding

alexcb:
lolololol, nice work @dchw

PR #617, opened 2020-12-12
Show status progress updates as progress bars

vladaionescu:
* If the progress related to a vertex and an update ID that we have just printed, go up one line and overwrite it
* In such cases, also don't delay the update (unless ANSI escape sequences are not supported)
* Use visual progress bars for these updates

#### BuildKite

Screenshot from 2020-12-12 09-13-53

#### GitHub actions

Screenshot from 2020-12-12 09-16-52

#### Local

Screenshot from 2020-12-12 12-01-09

The progress bar uses some special unicode characters for sub-character progress. Here are the different states of the progress bar:

Screenshot from 2020-12-12 08-45-30

### DIFF
--- a/builder/solver_monitor.go
+++ b/builder/solver_monitor.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
+	"os"
 	"regexp"
 	"sort"
 	"strings"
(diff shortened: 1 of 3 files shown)


alexcb (approved):
very nice!

builder/solver_monitor.go
```diff
@@ -435,3 +482,30 @@ func parseVertexName(vertexName string) (string, string, string, string) {
 func shortDigest(d digest.Digest) string {
 	return d.Hex()[:12]
 }
+
+var progressChars = []string{
+	" ", "▏", "▎", "▍", "▌", "▋", "▊", "▉", "█",
```
alexcb, line 487: ah ha, neat chars!

PR #618, opened 2020-12-14
Renaming earth to earthly

alexcb:
- when a user installs via brew, they type `brew install earthly`, so one would expect to find a command named `earthly` rather than `earth`
 - this git repo is located at `github.com/earthly/earthly.git` and not `github.com/earthly/earth.git`
 - Googling for `earthly` manages to find our documentation, where as the search-space for `earth` is much more crowded
 - `earth` easily tabs completes into `earthly`

### DIFF
--- a/buildcontext/excludes.go
+++ b/buildcontext/excludes.go
@@ -8,12 +8,12 @@ import (
 	"github.com/pkg/errors"
 )
 
-// EarthIgnoreFile is the name of the earth ignore file.
+// EarthIgnoreFile is the name of the earthly ignore file.
 const EarthIgnoreFile = ".earthignore"
 
 // ImplicitExcludes is a list of implicit patterns to exclude.
 var ImplicitExcludes = []string{
-	".tmp-earth-out/",
+	".tmp-earthly-out/",
 	"build.earth",
 	"Earthfile",
 	EarthIgnoreFile,
--- a/earthly
+++ b/earthly
@@ -6,7 +6,7 @@ if [ ! -d "$bindir" ]; then
   mkdir -p "$bindir"
 fi
 
-last_check_state_path=/tmp/last-earth-prerelease-check
+last_check_state_path=/tmp/last-earthly-prerelease-check
 
 get_latest_binary() {
     docker rm --force earthly_binary 2>/dev/null || true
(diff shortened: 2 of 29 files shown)


earthly
```diff
@@ -1,12 +1,12 @@
 #!/bin/bash
 set -e
 
-bindir="$HOME/.earthly"
+bindir="$HOME/.earth"
```
vladaionescu, line 4: Why not use `.earthly` for this too?

buildcontext/excludes.go
```diff
@@ -8,12 +8,12 @@ import (
 	"github.com/pkg/errors"
 )
 
-// EarthIgnoreFile is the name of the earth ignore file.
+// EarthIgnoreFile is the name of the earthly ignore file.
 const EarthIgnoreFile = ".earthignore"
```
alexcb, line 12: we're going to need some backwards compatibility if we want to rename this to `.earthlyignore`

README.md
```diff
@@ -319,7 +319,7 @@ release:
 ```
 
 ```bash
-earth --secret GITHUB_TOKEN --push +release
+earthly --secret GITHUB_TOKEN --push +release
```
vladaionescu, line 322: Make changes in README on `next` branch instead?

README.md
```diff
@@ -145,8 +145,8 @@ See also the [full documentation](https://docs.earthly.dev).
 Reference pages
 
 * 📑 [Earthfile reference](https://docs.earthly.dev/earthfile)
-* #️⃣ [Earth command reference](https://docs.earthly.dev/earth-command)
-* ⚙️ [Configuration reference](https://docs.earthly.dev/earth-config)
+* #️⃣ [Earth command reference](https://docs.earthly.dev/earthly-command)
+* ⚙️ [Configuration reference](https://docs.earthly.dev/earthly-config)
```
vladaionescu, line 149: Note that these will require redirects in gitbook. See https://docs.gitbook.com/integrations/github/content-configuration

vladaionescu (commented):
* Also rename `docker2earth`?
* `earthApp` -> `earthlyApp` in main.go
* Some other (maybe not very important) variables names still have names like `earthTargetName` (eg `artifact.go`). I guess it's ok to skip them for now.
* Docs will be done in separate PR branched off `next`, right?
* Changes to add symlink via bootstrap will be done in separate PR, right?

buildcontext/excludes.go
vladaionescu: The other possible one is `Earthfile`... Perhaps let's leave them like this for now.

README.md
alexcb: that's a good call, I left out the docs renaming for a different PR but let the README changes slip in, i'll save these for a future PR.

earthly
alexcb: good catch, this was a mistaken search/replace.

alexcb:
I renamed earthApp to earthlyApp in this PR; I'm hoping to do a follow up PR for docs, readmes, and other variable renames and more significant code changes like the bootstraping logic.

vladaionescu (approved):

PR #622, opened 2020-12-14
Add buldkitd start argument passthrough, compatibility checks

### LINKED ISSUE #598: Support for environments where userns-remap is enabled on docker daemon
Hi folks

In my build servers (Buildkite) environments, I have userns-remap enabled on  docker daemon configurations

```
> cat /etc/docker/daemon.json
{
  "storage-driver": "overlay2",
  "userns-remap": "buildkite-agent",
  "experimental": true
}
```

This causes buildkitd to crash when attempting to start with privileged mode
```
> earth --push +app
           buildkitd | Starting buildkit daemon as a docker container (earthly-buildkitd)...
           buildkitd | Is docker installed and running? Are you part of the docker group?
Error: buildkitd new client: buildkitd new client (own): maybe start buildkitd: start: docker run earthly/buildkitd:v0.3.19: docker: Error response from daemon: privileged mode is incompatible with user namespaces.  You must run the container in the host namespace when running privileged mode.
See 'docker run --help'.
: exit status 125
```

The problem goes away when I  remove the `userns-remap` flag 
```
cat /etc/docker/daemon.json
{
  "storage-driver": "overlay2",
  "experimental": true
}
```

To my understanding, this is related to [User namespace known limitations](https://docs.docker.com/engine/security/userns-remap/#user-namespace-known-limitations)
> The following standard Docker features are incompatible with running a Docker daemon with user namespaces enabled:
>  - sharing PID or NET namespaces with the host (--pid=host or --network=host).
>  - external (volume or storage) drivers which are unaware or incapable of using daemon user mappings.
>  - Using the --privileged mode flag on docker run without also specifying --userns=host.

It seems that docker is suggesting  adding `--userns=host` to buildkitd run command https://github.com/earthly/earthly/blob/fbf986db89dd8e53cf0546cf31df8f0ad716c793/buildkitd/buildkitd.go#L199 

Would there be any unintended side effect if we were to add `---userns=host` flag ?

I'm using the following versions:
```
> docker -v
Docker version 19.03.13, build 4484c46
> earth --version
earth version v0.3.19
```

dchw:
Fixes #598.

Because this effectively removes all isolation that was happening with the remapping, make the user turn it on to at least acknowledge that it needs to be done.

An example usage: `earth --bkd-arg "--userns=host" +some-target`

Though it would be better to use the environment variable to configure it to avoid accidential buildkit restarts.


buildkitd/buildkitd.go
```diff
@@ -186,6 +186,11 @@ func RemoveExited(ctx context.Context) error {
 
 // Start starts the buildkitd daemon.
 func Start(ctx context.Context, image string, settings Settings, reset bool) error {
+	err := CheckCompatibility(ctx, settings)
+	if len(settings.AdditionalArgs) == 0 && err != nil {
```
dchw, line 190: Assume that providing buildkit args means you are aware you have compatibility issues and go ahead anyways.

buildkitd/buildkitd.go
```diff
@@ -356,3 +362,44 @@ func GetAvailableImageID(ctx context.Context, image string) (string, error) {
 	}
 	return string(output), nil
 }
+
+// CheckCompatibility runs all avaliable compatibility checks before starting the buildkitd daemon.
+func CheckCompatibility(ctx context.Context, settings Settings) error {
```
dchw, line 367: Check for various docker setups we know we have issues with in one place

cmd/earth/main.go
```diff
@@ -794,6 +802,8 @@ func (app *earthApp) before(context *cli.Context) error {
 
 	app.buildkitdSettings.DebuggerPort = app.cfg.Global.DebuggerPort
 	app.buildkitdSettings.RunDir = app.cfg.Global.RunPath
+	app.buildkitdSettings.AdditionalArgs = app.buildkitdAdditionalArgs.Value()
```
vladaionescu, line 805: We should put this in the `config.yml` because any change to it requires buildkit restarts.

buildkitd/buildkitd.go
```diff
@@ -356,3 +362,44 @@ func GetAvailableImageID(ctx context.Context, image string) (string, error) {
...
+	isNamespaced, err := isNamespacedDocker(ctx)
+	if isNamespaced {
+		return errors.New(`user namespaces are enabled, run with --bkd-arg "--userns=host" to disable`)
+	} else if err != nil {
+		return errors.Wrap(err, "failed compatibilty check")
+	}
+
+	isRootless, err := isRootlessDocker(ctx)
+	if isRootless {
+		return errors.New(`rootless docker detected. Compatibility is limited. Run with --bkd-arg "" to give it a shot`)
```
vladaionescu, line 377: The error message seems incomplete?

buildkitd/buildkitd.go
```diff
@@ -356,3 +362,44 @@ func GetAvailableImageID(ctx context.Context, image string) (string, error) {
...
+		"docker", "info", "--format={{.SecurityOptions}}")
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		return false, errors.Wrap(err, "get docker security info")
+	}
+
+	return strings.Contains(string(output), "name=userns"), nil
+}
+
+func isRootlessDocker(ctx context.Context) (bool, error) {
```
dchw, line 396: Playing with it; it kinda works sometimes. Clearly needs more work but looks possible. I got it to do a simple build, but `WITH DOCKER` is a no-go. Also you will run into stuff like this when manipulating the buildkitd container:

```
docker: Error response from daemon: OCI runtime create failed: container_linux.go:370: starting container process caused: process_linux.go:459: container init caused: write sysctl key net.ipv4.ping_group_range: write /proc/sys/net/ipv4/ping_group_range: invalid argument: unknown.
: exit status 125

```

cmd/earth/main.go
dchw, line 805: HAve to bridge from the CLI framework slice to a legit slice; may as well shove it into the `before`

dchw (commented):
Some comments; I anticipate some changes. All in all, I think compatibility checks by default, plus ability to pass through any CLI args gives us the right kind of flexibility; while keeping people on rails.

alexcb (approved):

buildkitd/buildkitd.go
dchw: Hm. The intent wasn't to make it seem empty. Specifying it is how you override the compatibility check; see my comments in my commentary review. That said, I just realized that it will sill have `len(additionalArgs)=0`, so we need something we can pass in there thats more or less a no-op :thinking:

cmd/earth/main.go
dchw: What about the other args we pass through to buildkitd? It seemed 50-50 so I just went for one.

buildkitd/buildkitd.go
vladaionescu: It's possible to use rootless overlay snapshotter in buildkit. But I think it requires some additional stuff in the buildkitd container. See eg the buildkit `rootless` build: https://github.com/moby/buildkit/blob/master/Dockerfile#L288

cmd/earth/main.go
vladaionescu: I think all that's left in flag form is either deprecated settings or internal ones.

Config is better especially for CI environments like Jenkins where we want to avoid different pipelines from causing restarts.

dchw: I see. Th other ones are passed in via different route; wont cause a restart since its not in the `docker run`. I'll move this then. Do we do any clever config merging from file -> env?

buildkitd/buildkitd.go
dchw: Seems orthogonal to the intent here. Don't want to shave that yak in this PR. Felt reasonable to include a check for it though

cmd/earth/main.go
vladaionescu: Actually whatever is in `buildkitSettings` will cause a restart - we compare a hash of its JSON. Not sure what you mean by `file -> env`? Don't think we need env setting in this case.

buildkitd/buildkitd.go
```diff
@@ -356,3 +362,44 @@ func GetAvailableImageID(ctx context.Context, image string) (string, error) {
 	}
 	return string(output), nil
 }
+
+// CheckCompatibility runs all avaliable compatibility checks before starting the buildkitd daemon.
+func CheckCompatibility(ctx context.Context, settings Settings) error {
+	isNamespaced, err := isNamespacedDocker(ctx)
+	if isNamespaced {
+		return errors.New(`user namespaces are enabled, run with --bkd-arg "--userns=host" to disable`)
```
vladaionescu, line 370: Won't this fail even if the setting is enabled?

cmd/earth/main.go
dchw: MIve used a lot of config packages that build a config starting from a file, then layering env vars, then CLI args. Looks like we dont. I have already made the changes.

dchw: We have two remaining args that come in via the CLI - `GIT_URL_INSTEAD_OF` and `EARTHLY_BUILDKIT_CACHE_SIZE_MB`. Those should be moved out sometime. Probably another PR another time.

vladaionescu: :+1:

I believe both can be removed with @alexcb's latest ssh-related changes.

buildkitd/buildkitd.go
dchw: We dont even check compatibility if you specify additional buildkit args; we assume you know what you are doing and are non-standard. It didn't seem worth setting a precedent for either handling each arg as a first-class citizen, or parsing the passed-in args.

vladaionescu: Ah yes. Good thinking

dchw: Excuse me; we check, but it won't cause a failure. Lines 189-190.

cmd/earth/main.go
alexcb: `GIT_URL_INSTEAD_OF` is schedule to be removed, it no longer has an effect.

vladaionescu (approved):

vladaionescu:
Compatibility check is pretty cool

PR #628, opened 2020-12-15
Rename earth to earthly

alexcb:


vladaionescu (approved):
Do you also want to do replacements in `README.md` here?

alexcb:
good point regarding the READMEs, I just pushed a new version.

vladaionescu (approved):

PR #632, opened 2020-12-15
Symlink earth to earthly during bootstrap

alexcb:
- if earth exists, delete it and symlink it to new earthly install

### DIFF
--- a/cmd/earthly/main.go
+++ b/cmd/earthly/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"bufio"
+	"bytes"
 	"context"
 	"crypto/rand"
 	"encoding/base64"
@@ -13,6 +14,7 @@ import (
 	"os"
 	"os/signal"
 	"os/user"
+	"path"
 	"path/filepath"
 	"regexp"
 	"runtime"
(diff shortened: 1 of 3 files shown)


cmd/earthly/main.go
```diff
@@ -1023,6 +1026,58 @@ func (app *earthlyApp) run(ctx context.Context, args []string) int {
...
+		return false
+	}
+	if !bytes.Contains(data, []byte("api.earthly.dev")) {
+		return false
+	}
+	if !bytes.Contains(data, []byte("Earthfile")) {
+		return false
+	}
+	return true
+}
```
vladaionescu, line 1045: Interesting heuristic :D

termutil/is_tty.go
```diff
@@ -0,0 +1,11 @@
...
+
+import "os"
+
+// IsTTY returns true if a terminal is detected
+func IsTTY() bool {
+	if fileInfo, _ := os.Stdout.Stat(); (fileInfo.Mode() & os.ModeCharDevice) != 0 {
+		return true
+	}
+	return false
+}
```
vladaionescu, line 11: BTW, I used something like this here: https://github.com/earthly/earthly/blob/main/builder/solver_monitor.go#L81

vladaionescu (approved):

PR #635, opened 2020-12-15
Automatically use https when no ssh keys exist

alexcb:


vladaionescu (approved):
Can you try to remove all the git-instead-of env vars from our gh actions and circle ci setup?

That should serve as a good test.

PR #636, opened 2020-12-15
Warn if using earth instead of earthly

alexcb:

### DIFF
--- a/cmd/earthly/main.go
+++ b/cmd/earthly/main.go
@@ -792,7 +792,31 @@ func (app *earthlyApp) before(context *cli.Context) error {
 	return nil
 }
 
+func (app *earthlyApp) warnIfEarth() {
+	if len(os.Args) == 0 {
+		return
+	}
+	binPath := os.Args[0] // can't use os.Executable() here; because it will give us earthly if executed via the earth symlink
+
+	baseName := path.Base(binPath)
+	if baseName == "earth" {
+		app.console.Warnf("Warning: the earth binary has been renamed to earthly; the earth command is currently symlinked, but is deprecated and will one day be removed.")
+	}
+
+	absPath, err := filepath.Abs(binPath)
+	if err != nil {
+		return
+	}
+	earthlyPath := path.Join(path.Dir(absPath), "earthly")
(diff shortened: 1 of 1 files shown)


cmd/earthly/main.go
```diff
@@ -792,7 +792,21 @@ func (app *earthlyApp) before(context *cli.Context) error {
...
 
+func (app *earthlyApp) warnIfEarth() {
+	if len(os.Args) == 0 {
+		return
+	}
+	binPath := os.Args[0] // can't use os.Executable() here; because it will give us earthly if executed via the earth symlink
+
+	baseName := path.Base(binPath)
+	if baseName == "earth" {
+		app.console.Warnf("Warning: the earth binary has been renamed to earthly; the earth command is currently symlinked, but is deprecated and will one day be removed.")
```
vladaionescu, line 803: We should also say... `Feel free to rm <path-to-earth>`, if we detect that `earthly` exists.

vladaionescu (approved):

cmd/earthly/main.go
alexcb: that's a good idea, I just pushed up a new version to do that.

vladaionescu: Forgot to add files maybe?

alexcb: I don't think so, I did a rebase+force push.

vladaionescu: Oh I see https://github.com/earthly/earthly/pull/637

vladaionescu: nvm, I got wires crossed :-)

vladaionescu (approved):

vladaionescu:
I think there is an issue:

```
~/workspace/earthly main ❯ ./build/linux/amd64/earthly --ci --push -P ./examples/cpp+docker
Once you are ready to switch over to earthly, you can `rm /home/vlad/workspace/earthly/build/linux/amd64/earthly`
```

(I ran earthly, not earth)

alexcb:
Thanks for that fix, good catch.

PR #639, opened 2020-12-16
Idea: Wrap CLI help

agbell:
Earthly CLI  help is really wide and wraps strangely.  Here I am trying to wrap it at around 150 characters wide.  This is still kind of wide but does look better.  
Before:
Screen Shot 2020-12-16 at 11 36 35 AM

After:
Screen Shot 2020-12-16 at 11 36 55 AM

Also, these are the first lines of go I have written so I might need some guidance.

### DIFF
--- a/cmd/earthly/main.go
+++ b/cmd/earthly/main.go
@@ -295,13 +295,13 @@ func newEarthlyApp(ctx context.Context, console conslogging.ConsoleLogger) *eart
 		&cli.BoolFlag{
 			Name:        "ci",
 			EnvVars:     []string{"EARTHLY_CI"},
-			Usage:       "Execute in CI mode (implies --use-inline-cache --save-inline-cache --no-output) *experimental*",
+			Usage:       wrap([]string{"Execute in CI mode (implies --use-inline-cache --save-inline-cache --no-output)", "*experimental*"}),
 			Destination: &app.ci,
 		},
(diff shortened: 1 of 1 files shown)


alexcb:
I wonder if we need this much control over where to wrap it? Just an idea, but what if we had a function like `func wrap(s string, width int) string` which would split on space and automatically wrap it once the line was longer than width? that way we could detect the terminal width and change the wrapping automatically?

agbell:
Yeah, good question.  I found it helpful because the default value and ENV value add to the line length in the terminal, so you have to counteract those, but if we can wrap the whole string including those it would work.  I just wasn't sure how to get ahold of the full string.

alexcb:
That's a good point regarding the ENV name and default value being added; Maybe the function could also take a third argument that estimates how many chars will be appended to the help text?

If you wanted to programatically detect the terminal widths, you could do something like this:
```
import "github.com/creack/pty"

size, err := pty.GetsizeFull(os.Stdin)
size.X <-- will contain the terminal width
```
The interactive debugger already uses pty, so it won't even increase the number of dependencies.

alexcb (approved):
There's currently an open PR in urfav/cli https://github.com/urfave/cli/pull/1119 that will make wrapping easier; let's merge our PR as-is, then come back to it once the auto-wrapping is added to the cli library.

PR #646, opened 2020-12-17
Load secret data out of a file

alexcb:
- if the secret value starts with a '@', then load the contents of the
file it points to.


vladaionescu:
Wondering if it might be better to spell it out as `--secret-file` rather than use a special character. I feel it's more readable that way.

BTW, we should also be able to mount as a file within the build itself. Different PR maybe?

alexcb:
`--secret-file` is probably a bit more user-friendly; I was using the `@` notion that curl uses, but it's still a bit less-known.

vladaionescu (approved):

PR #650, opened 2020-12-17
Fix docker compose + interactive issue

vladaionescu:


buildkitd/dockerd-wrapper.sh
```diff
@@ -46,6 +46,24 @@ execute() {
...
 start_dockerd() {
+    # Use a specific IP range to avoid collision with host dockerd (we need to also connect to host
+    # docker containers for the debugger).
+    mkdir -p /etc/docker
+    cat <<'EOF' >/etc/docker/daemon.json
+{
+    "default-address-pools" : [
+        {
+            "base" : "172.21.0.0/16",
+            "size" : 24
```
alexcb, line 57: any idea what the size is for? is it related to the CIDR size of 16? or completely unrelated?

alexcb (approved):
woohoo!

buildkitd/dockerd-wrapper.sh
vladaionescu: No idea - tried finding info on it but got nothing. I just see all examples with size 24

PR #654, opened 2020-12-18
Support for mounting secrets as files

alexcb:
Implements a new `--secret-file` flag within the RUN command.
```
    RUN --secret-file=/root/this-my-secret=+secrets/SECRET1 cat /root/this-my-secret)
```

### DIFF
--- a/earthfile2llb/converter.go
+++ b/earthfile2llb/converter.go
@@ -131,7 +131,7 @@ func (c *Converter) fromTarget(ctx context.Context, targetName string, platform 
 		c.mts.Final.LocalDirs[dirKey] = dirValue
 	}
 	for _, kv := range saveImage.Image.Config.Env {
-		k, v := variables.ParseKeyValue(kv)
+		k, v, _ := variables.ParseKeyValue(kv)
 		c.varCollection.AddActive(k, variables.NewConstantEnvVar(v), true, false)
 	}
 	c.mts.Final.MainImage = saveImage.Image.Clone()
@@ -286,7 +286,7 @@ func (c *Converter) CopyClassical(ctx context.Context, srcs []string, dest strin
 }
 
 // Run applies the earthly RUN command.
-func (c *Converter) Run(ctx context.Context, args []string, mounts []string, secretKeyValues []string, privileged bool, withEntrypoint bool, withDocker bool, isWithShell bool, pushFlag bool, withSSH bool) error {
+func (c *Converter) Run(ctx context.Context, args, mounts, secretKeyValues []string, privileged, withEntrypoint, withDocker, isWithShell, pushFlag, withSSH bool) error {
 	c.nonSaveCommand()
 	if withDocker {
 		return errors.New("RUN --with-docker is obsolete. Please use WITH DOCKER ... RUN ... END instead")
@@ -615,7 +615,7 @@ func (c *Converter) buildTarget(ctx context.Context, fullTargetName string, plat
 	return mts, nil
 }
 
-func (c *Converter) internalRun(ctx context.Context, args []string, secretKeyValues []string, isWithShell bool, shellWrap shellWrapFun, pushFlag bool, withSSH bool, commandStr string, opts ...llb.RunOption) error {
+func (c *Converter) internalRun(ctx context.Context, args, secretKeyValues []string, isWithShell bool, shellWrap shellWrapFun, pushFlag, withSSH bool, commandStr string, opts ...llb.RunOption) error {
 	finalOpts := opts
 	var extraEnvVars []string
 	// Secrets.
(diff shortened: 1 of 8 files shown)


variables/util.go
```diff
@@ -16,6 +16,38 @@ func ParseKeyValue(env string) (string, string) {
 	return parts[0], v
 }
 
+// ParseEscapedKeyValue pases a key-value type into its parts
+// keys and values containing '=', and '\' must be escaped using '\=', and '\\' respectively
+func ParseEscapedKeyValue(s string) (string, string, bool) {
```
alexcb, line 21: I wonder if it would be safe to simply replace `ParseKeyValue` with this new method, but I wonder if it'll break any backwards compatibility issues where some a value might contain an `=` which would have to be escaped.

earthfile2llb/converter.go
```diff
@@ -643,6 +643,29 @@ func (c *Converter) internalRun(ctx context.Context, args []string, secretKeyVal
 			return fmt.Errorf("Secret definition %s not supported. Must start with +secrets/ or be an empty string", secretKeyValue)
 		}
 	}
+	// Secret Files.
+	for _, secretFileDef := range secretFiles {
+		secretPath, secretID, ok := variables.ParseEscapedKeyValue(secretFileDef)
+		if !ok {
+			return fmt.Errorf("Invalid secret-file definition %s", secretFileDef)
```
vladaionescu, line 650: Nit: We should use lower-case for errors. Mind changing the other `fmt.Errorf`'s in this function too?

earthfile2llb/listener.go
```diff
@@ -281,6 +281,8 @@ func (l *listener) ExitRunStmt(c *parser.RunStmtContext) {
 	withSSH := fs.Bool("ssh", false, "Make available the SSH agent of the host")
 	secrets := new(StringSliceFlag)
 	fs.Var(secrets, "secret", "Make available a secret")
+	secretFiles := new(StringSliceFlag)
+	fs.Var(secretFiles, "secret-file", "mount a secret as a file")
```
vladaionescu, line 285:
```suggestion
	fs.Var(secretFiles, "secret-file", "Mount a secret as a file")
```

vladaionescu, line 285: Can you also expand build args in a for loop that looks like this one?

```
for i, s := range secrets.Args {
  secrets.Args[i] = l.expandArgs(s, true)
}
```

earthfile2llb/converter.go
```diff
@@ -643,6 +643,29 @@ func (c *Converter) internalRun(ctx context.Context, args []string, secretKeyVal
 			return fmt.Errorf("Secret definition %s not supported. Must start with +secrets/ or be an empty string", secretKeyValue)
 		}
 	}
+	// Secret Files.
+	for _, secretFileDef := range secretFiles {
+		secretPath, secretID, ok := variables.ParseEscapedKeyValue(secretFileDef)
```
vladaionescu, line 648: I *think* that we don't need a special implementation for this. In the previous implementation anything after the first `=` is part of the value. So `key=value=something` parses to key=`key` and value=`value=something`. Wouldn't that be enough?

variables/util_test.go
```diff
@@ -0,0 +1,28 @@
...
+	var tests = []struct {
+		kv string
+		k  string
+		v  string
+		ok bool
+	}{
+		{"key=val", "key", "val", true},
+		{"with space=val with space", "with space", "val with space", true},
+		{"dudewheresmyvalue", "", "", false},
+		{`\==\\`, "=", `\`, true},
```
vladaionescu, line 19: We should add `key=value=something` too as a valid parse. I think we rely on something like this somewhere in our code (forgot where).

vladaionescu (commented):
Wonder if we should use a syntax closer to Dockerfile's https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/syntax.md#run---mounttypesecret (`--mount=type=secret`).

I assume this also works with cloud secrets, right? BTW, is it possible to `earthly secrets set` from file or stdin somehow?

earthfile2llb/converter.go
alexcb: tackled separately in https://github.com/earthly/earthly/pull/663

earthfile2llb/listener.go
alexcb: good catch!

alexcb:
> Wonder if we should use a syntax closer to Dockerfile's https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/syntax.md#run---mounttypesecret (--mount=type=secret).

If we did do something like that what would it look like?

`RUN --secret=SECRET2=+secrets/SECRET2` vs `RUN --secret=file=/root/my-secret=+secrets/SECRET2`? but then the parser would see `RUN --secret=file=/root/my-secret` and assume we want a secret named `file` set with the value `/root/my-secret`? (or actually it'll just display an error saying it needs to start with `+secrets/`?


> I assume this also works with cloud secrets, right?

It should, let me try adding in a test for this to be sure.

>  BTW, is it possible to earthly secrets set from file or stdin somehow?

    earthly secrets set --file /tmp/releases /user/my_key

but there's no support for stdin (yet); however the ice-cream operator can help:

    earthly secrets set --file <(echo -n "hello") /user/my_key

vladaionescu:
> If we did do something like that what would it look like?

I was thinking something more like this...

For file mount (this would be the closest to the Dockerfile syntax, and also similar to all our other mount types):

```
RUN --mount=type=secret,id=+secrets/foo/bar,target=/path/where/to/mount
```

For env var the same way we have it today:

```
RUN --secret=x=+secrets/foo/bar
```

> earthly secrets set --file /tmp/releases /user/my_key

:cool:

vladaionescu (approved):

PR #660, opened 2020-12-18
Ignore installID on CI

alexcb:
- containerized CI systems will always generate a new install-ID, we
should ignore them.

### DIFF
--- a/analytics/analytics.go
+++ b/analytics/analytics.go
@@ -161,7 +161,9 @@ func CollectAnalytics(ctx context.Context, earthlyServer string, displayErrors b
 	installID, overrideInstallID := os.LookupEnv("EARTHLY_INSTALL_ID")
 	repoHash := getRepoHash()
 	if !overrideInstallID {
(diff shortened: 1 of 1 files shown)


vladaionescu (approved):
> containerized CI systems will always generate a new install-ID, we should ignore them.

Not quite... self-hosted runners typically reuse the environment.

But perhaps this is a decent compromise.

alexcb:
@vladaionescu we could change to ignore certain types of CIs instead if you think that's better? like `github-actions`, `drone`, `circle-ci` for instance?

alexcb:
alternatively, if we're willing to cut out `segment` completely, there's no harm leaving it as-is and closing this PR.

vladaionescu:
> @vladaionescu we could change to ignore certain types of CIs instead if you think that's better? like github-actions, drone, circle-ci for instance?

It'll never be perfect, because even on GH Actions you can still have a self-hosted runner. I think the change is good the way it  is. Don't think we're losing much data.

PR #664, opened 2020-12-18
Another remote cache test

vladaionescu:


alexcb (approved):
interesting choice of sample files to copy in :)

vladaionescu:
> interesting choice of sample files to copy in :)

It's the JS example, which was crashing before :D

Issue #536, opened 2020-11-22
Support multi-platform builds


LuaxY:
Is it possible to support multi-platform (e.g. amd64, arm64, arm v7) image build with `docker buildx` feature?
With docker command, by using compatible builder, we can now provide `--platform` flag to specify target platform.

`docker [buildx] build --platform linux/amd64,linux/arm64,linux/arm/v7 .`

https://docs.docker.com/buildx/working-with-buildx/

agbell:
Hi @LuaxY, I believe this should be possible and should be a feature we add.  Thank you for submitting it.

vladaionescu:
Experimental support for multi-platform builds is available in `v0.4.3`. Here is a simple example: https://github.com/earthly/earthly/blob/main/examples/multiplatform/Earthfile .

If you try this feature out, I'd be very curious about your experience with it - either here or on our Slack.

LuaxY:
Tested on my project, work perfectly, big thanks!!

Question about local artifacts, example:

`main.go`
```go
package main

import "fmt"

func main() {
	fmt.Println("test")
}
```

`Earthfile`
```Dockerfile
FROM golang:1.15-alpine
RUN apk --update --no-cache add git
WORKDIR /mailtrap

all:
  BUILD \
    --platform=linux/amd64 \
    --platform=linux/arm64 \
    --platform=linux/arm/v7 \
    --platform=linux/arm/v6 \
    +build

build:
  COPY main.go main.go
  RUN go build -o build/test main.go
  SAVE ARTIFACT build/test AS LOCAL build/test
```

This will build binary for each platform but also override the local artifcat

```
+build | Artifact test:master+build/test as local build/test
+build | Artifact test:master+build/test as local build/test
+build | Artifact test:master+build/test as local build/test
+build | Artifact test:master+build/test as local build/test
```

The final binary is the last platform stage to finish the `SAVE ARTIFACT`

```
$ file build/test 
build/test: ELF 32-bit LSB executable, ARM, EABI5 version 1 (SYSV), statically linked, [...]
```

This is the intended behavior? or there is a way to control the artifact name depending the platform?

vladaionescu:
Great!

There are four new built-in args available, which you can use for this:

* `TARGETPLATFORM`
* `TARGETOS`
* `TARGETARCH`
* `TARGETVARIANT`

You need to predeclare them before using them:

```
ARG TARGETPLATFORM
RUN echo "The platform is $TARGETPLATFORM"
```

[...]

PR #667, opened 2020-12-19
Support for multi-platform builds

vladaionescu:
Re https://github.com/earthly/earthly/issues/536

Integration tests to follow in additional PRs.

### DIFF
--- a/cmd/earthly/main.go
+++ b/cmd/earthly/main.go
@@ -26,6 +26,7 @@ import (
 	"text/tabwriter"
 	"time"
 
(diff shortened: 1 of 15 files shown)


cmd/earthly/main.go
```diff
@@ -26,6 +26,7 @@ import (
 	"text/tabwriter"
 	"time"
 
+	"github.com/containerd/containerd/platforms"
```
alexcb, line 29: nit: can we move this to the non-earthly portion of the imports?

alexcb (approved):

PR #668, opened 2020-12-21
git repo hash should not depend on protocol

alexcb:
- remove protocol and user from git repo before hashing to make the
repo_hash the same for users that use ssh and CIs that use https.

### DIFF
--- a/buildcontext/detectgit.go
+++ b/buildcontext/detectgit.go
@@ -57,7 +57,7 @@ func Metadata(ctx context.Context, dir string) (*GitMetadata, error) {
 	}
 	var gitURL string
 	if remoteURL != "" {
-		gitURL, err = parseGitRemoteURL(remoteURL)
+		gitURL, err = ParseGitRemoteURL(remoteURL)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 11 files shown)


buildcontext/detectgit.go
```diff
@@ -128,8 +128,8 @@ func detectIsGitDir(ctx context.Context, dir string) error {
 	return nil
 }
 
-// parseGitRemoteURL converts a gitURL like user@host.com:path/to.git or https://host.com/path/to.git to host.com/path/to
-func parseGitRemoteURL(gitURL string) (string, error) {
+// ParseGitRemoteURL converts a gitURL like user@host.com:path/to.git or https://host.com/path/to.git to host.com/path/to
+func ParseGitRemoteURL(gitURL string) (string, error) {
```
vladaionescu, line 132: Seems like this no longer belongs to this package since it's being used in other contexts too. Perhaps the entire file `detectgit.go` can be moved to something like `gitutil` package?

vladaionescu (approved):

buildcontext/detectgit.go
alexcb: good point, want to take another look after the move?

vladaionescu (approved):
Looks good :+1:

PR #677, opened 2020-12-29
Fix tracking of CI user IDs

vladaionescu:
Previous logic was never reaching `installID = fmt.Sprintf("%x", sha256.Sum256([]byte(ciName+repoHash)))` as `ciName` had to be `"false"`, but `ci` had to be `true`.

Or perhaps I'm misunderstanding this - not sure how this has been apparently working so far.


alexcb (approved):
That's a good catch.

vladaionescu:
Any idea / theory as to how it was seemingly working so far? Not sure if I'm breaking something here :D

vladaionescu:
Just checked the data and it is indeed incorrect. This should fix it.

PR #683, opened 2020-12-31
Add --if-exists to COPY and SAVE ARTIFACT

### LINKED ISSUE #588: Proposal: SAVE ARTIFACT --if-exists
### Proposal

Allow saving an artifact only if file exists.

```
SAVE ARTIFACT --if-exists ./my-file AS LOCAL ./path/my-file
```

### Use-case

* Terraform backups

### CC

@dchw

dchw:
It was super easy to add it to `COPY`; and seemed applicable because you can COPY artifacts from upstream targets (see the `CopyArtifact` function...)

I chose to implement it by exploiting the "Allow Empty Wildcard" option for the LLB copy; as that logic is exactly the logic I also want to follow; just with an exact file name. This changes the filename into a wildcard-type string (that is sufficiently constrained enough to only match the file we want), and then enables the "Allow Empty Wildcard" behavior for the optional artifacts.

Admittedly a hack; but it was certainly less invasive than evaluating where we were at in a `WITH DOCKER` style setup for each artifact, and some `COPY`s. 

Resolves #588.

### DIFF
--- a/.buildkite/macOS.yml
+++ b/.buildkite/macOS.yml
@@ -8,3 +8,4 @@ steps:
       EARTHLY_BUILD_ARGS: "DOCKERHUB_USER_SECRET=+secrets/earthly-technologies/dockerhub/user,DOCKERHUB_TOKEN_SECRET=+secrets/earthly-technologies/dockerhub/token"
     agents:
       os: macOS
+      arch: amd64
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -317,7 +317,7 @@ func (b *Builder) convertAndBuild(ctx context.Context, target domain.Target, opt
 	if opt.NoOutput {
 		// Nothing.
 	} else if opt.OnlyArtifact != nil {
-		err := b.saveArtifactLocally(ctx, *opt.OnlyArtifact, outDir, opt.OnlyArtifactDestPath, mts.Final.Salt, opt)
+		err := b.saveArtifactLocally(ctx, *opt.OnlyArtifact, outDir, opt.OnlyArtifactDestPath, mts.Final.Salt, opt, false)
 		if err != nil {
 			return nil, err
 		}
@@ -358,7 +358,7 @@ func (b *Builder) convertAndBuild(ctx context.Context, target domain.Target, opt
 						Target:   sts.Target,
 						Artifact: saveLocal.ArtifactPath,
 					}
-					err := b.saveArtifactLocally(ctx, artifact, artifactDir, saveLocal.DestPath, sts.Salt, opt)
+					err := b.saveArtifactLocally(ctx, artifact, artifactDir, saveLocal.DestPath, sts.Salt, opt, saveLocal.IfExists)
 					if err != nil {
 						return nil, err
 					}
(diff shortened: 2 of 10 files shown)


builder/builder.go
```diff
@@ -466,6 +466,9 @@ func (b *Builder) saveArtifactLocally(ctx context.Context, artifact domain.Artif
 	if err != nil {
 		return errors.Wrapf(err, "glob")
 	} else if !artifact.Target.IsRemote() && len(fromGlobMatches) <= 0 {
+		if ifExists {
```
dchw, line 469: Kept the name `ifExists` everywhere all the way through in an attempt to draw a direct line from the parameter. However; I feel like `optional` might be a better name from the listener on down... open for suggestions if this is offensive.

llbutil/copyop.go
```diff
@@ -26,14 +27,22 @@ func CopyOp(srcState llb.State, srcs []string, destState llb.State, dest string,
 		baseCopyOpts = append(baseCopyOpts, llb.WithCreatedTime(*defaultTs()))
 	}
 	for _, src := range srcs {
+		if ifExists {
+			// If the copy came in as optional (ifExists), then we need to trigger the
+			// underlying wildcard matching and allow empty wildcards. The matching uses
+			// the filepath.Match syntax, so by simply creating a wildcard where the
+			// first letter needs to match the current first letter gets us the single
+			// match; and no error if it is missing.
```
dchw, line 35: Admittedly a hack. But its clean and airtight after much spelunking and experimenting.

.buildkite/macOS.yml
```diff
@@ -8,3 +8,4 @@ steps:
       EARTHLY_BUILD_ARGS: "DOCKERHUB_USER_SECRET=+secrets/earthly-technologies/dockerhub/user,DOCKERHUB_TOKEN_SECRET=+secrets/earthly-technologies/dockerhub/token"
     agents:
       os: macOS
+      arch: amd64
```
dchw, line 11: Had to since the mac m1 build isnt 100 percent right now anyways... we way want to consider splitting into smoketests per os/arch and let the long test fall where it may

llbutil/copyop.go
dchw: See here for the matching details: https://golang.org/pkg/path/filepath/#Match

llbutil/copyop.go
```diff
@@ -26,14 +27,22 @@ func CopyOp(srcState llb.State, srcs []string, destState llb.State, dest string,
...
+			src = fmt.Sprintf("[%s]%s", string(src[0]), string(src[1:]))
```
vladaionescu, line 36: Ha!

earthfile2llb/listener.go
```diff
@@ -395,7 +397,7 @@ func (l *listener) ExitSaveArtifact(c *parser.SaveArtifactContext) {
 		l.err = fmt.Errorf("no arguments provided to the SAVE ARTIFACT command")
 		return
 	}
-	if fs.NArg() > 5 {
+	if fs.NArg() > 6 {
```
vladaionescu, line 400: This change seems incorrect? The extra flag does not contribute to `NArg()`

dchw: You are correct. I think this is an artifact from before I pushed EOF into the NL...

dchw: With `?` :
With question mark

Without `?`:
without question mark

vladaionescu: Wrong PR? :-)

dchw: Oh man.  Yep. Cleaned.

Anyways, yes. I didnt read docs for Narg and assumed it was total args; and it added up to 5 with keepOwn and keepTs... thats what I get for assuming

alexcb (approved):

vladaionescu (approved):

Issue #685, opened 2021-01-03
`WITH DOCKER` with multiple --load from same target doesn't load images properly


ohadravid:
Hi!

I'm facing an issue when using `WITH DOCKER` with multiple images which are loaded from a target with the same name.
 
Running ` earthly -V -P +test` (using `earthly version v0.4.3`) for the following `Earthfile` (in an empty folder):

```

docker:
    FROM ubuntu


test:
    FROM earthly/dind:alpine
    WITH DOCKER \
            --load a:latest=+docker \
            --load b:latest=+docker
        RUN docker images
    END
```

results in:
```
...
+test | --> WITH DOCKER (install deps)
+test | /usr/local/bin/dockerd
+test | Completed in 302.1184ms
+test | --> WITH DOCKER RUN docker images
+test | Loading images...
+test | Loaded image: a:latest
+test | Loaded image: a:latest
+test | ...done
+test | REPOSITORY   TAG       IMAGE ID       CREATED          SIZE
+test | a            latest    abe76754ebd5   51 minutes ago   72.9MB
+test | Completed in 4.4538029s
...
```

where the expected output should show both `a` and `b` images.

My workaround for now is this:
```

docker:
    FROM ubuntu

a:
    FROM +docker

b:
    FROM +docker

test:
    FROM earthly/dind:alpine
    WITH DOCKER \
            --load a:latest=+a \
            --load b:latest=+b 
        RUN docker images
    END
```

which seems to work as expected.

Thanks for this great tool!

PR #687, opened 2021-01-05
Allow --load with same target, but with different tag names

dchw:
Resolves https://github.com/earthly/earthly/issues/685.

This allows the same upstream target to be used multiple times in a `WITH DOCKER --load`, with different docker tag names. It also oversteps its bounds a little bit and formalizes some of the cache we had set up.


states/solveCache.go
```diff
@@ -0,0 +1,60 @@
...
+)
+
+// SolveCache is a formal version of the cache we keep mapping targets to their LLB state.
+type SolveCache map[string]llb.State
+
+// StateKeyFunc is a function that can build a key for the SolveCache. These keys seem to be
+// Highly convention based, and used elsewhere too (LocalFolders?), so this is a step at
+// formalizing that convention, since we sometimes need one key, and sometimes another.
+// It may give us some toeholds to help with some refactoring later.
+type StateKeyFunc func() string
```
dchw, line 17: Its late; but the whole idea here was to make the key a type. An alternative here would be to "render" the data into the key, and use this as something like `type StateKey string`. I think thats better actually, how do I request a fix of my own PR? ;)

earthfile2llb/withdockerrun.go
```diff
@@ -341,14 +341,14 @@ func (wdr *withDockerRun) solveImage(ctx context.Context, mts *states.MultiTarge
...
+		keyFunc(),
 		llb.SessionID(sessionID),
 		llb.Platform(llbutil.DefaultPlatform()),
 		llb.WithCustomNamef("[internal] docker tar context %s %s", opName, sessionID),
 	)
 	wdr.tarLoads = append(wdr.tarLoads, tarContext)
-	wdr.c.mts.Final.LocalDirs[solveID] = outDir
-	wdr.c.opt.SolveCache[solveID] = tarContext
+	wdr.c.mts.Final.LocalDirs[keyFunc()] = outDir
+	wdr.c.opt.SolveCache.Set(keyFunc, tarContext)
```
dchw, line 351: More or less the problem was what the target hash is not enough here to make assumptions on. We need to also include the name its being assigned.

states/solveCache.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+// Delete removes a LLB state from a given solve cache, using the KeyFunc to derive the key
+func (sc *SolveCache) Delete(kf StateKeyFunc, state llb.State) {
+	delete((*sc), kf())
+}
+
+// KeyFromHashAndTag builds a state key from a given target state and a docker tag.
+// This is useful when you want to reference the same image but with a different name.
+func KeyFromHashAndTag(target *SingleTarget, dockerTag string) (StateKeyFunc, error) {
```
dchw, line 43: See aformementioned idea about a special string type.

dchw (commented):
I had a better idea on something, but leaving this up for comments. I will make the fix I suggested tomorrow.

alexcb (approved):

states/solveCache.go
```diff
@@ -0,0 +1,61 @@
...
+
+// KeyFromState is a simple wrapper to get a key from a given state using the hash of its target.
+func KeyFromState(target *SingleTarget) (StateKey, error) {
+	hash, err := target.TargetInput.Hash()
+	if err != nil {
+		return StateKey(""), errors.Wrap(err, "target input hash")
+	}
+
+	return StateKey(hash), nil
+}
```
vladaionescu, line 61: Nit: Name this file `solvecache.go` for consistency?

examples/tests/single-target-load-multi-name.earth
```diff
@@ -0,0 +1,22 @@
...
+            --pull ubuntu:latest       # Test that we can load other images easily
+        RUN docker images && \
+            docker run a:latest && \
+            echo "-----" && \
+            docker run b:latest && \
+            echo "-----" && \
+            docker run c:latest uname -a && \
+            echo "-----" && \
+            docker run ubuntu:latest uname-a
+    END
```
vladaionescu, line 22: This file is not actually used, right?

vladaionescu (approved):
Nice find!

states/solveCache.go
```diff
@@ -0,0 +1,61 @@
...
+
+	"github.com/moby/buildkit/client/llb"
+	"github.com/pkg/errors"
+)
+
+// SolveCache is a formal version of the cache we keep mapping targets to their LLB state.
+type SolveCache map[StateKey]llb.State
+
+// StateKey is a type for a key in SolveCache. These keys seem to be highly convention based,
+// and used elsewhere too (LocalFolders?). so this is a step atformalizing that convention,
```
vladaionescu, line 14: Those are actually cache folders rather than pure local folders.

vladaionescu (approved):

examples/tests/single-target-load-multi-name.earth
dchw: No. I thought I removed that when I moved it into the with docker tests. Ill remove it.

states/solveCache.go
dchw: Yep; can do. I <3 consistency

states/solveCache.go
dchw: :+1:

PR #690, opened 2021-01-05
Handle mixed files and subdirectories on wildcard

### LINKED ISSUE #689: SAVE ARTIFACT ./* AS LOCAL not capturing all files
## Repro steps
1. `git clone https://github.com/jazzdan/earthly-tons-of-artifacts.git`
2. `cd earthly-tons-of-artifacts`
3. `earthly +docker-setup`

## What I expected to happen
The Earthfile copies everything from `lib` in to the image (`COPY lib/* .`) then saves everything in the current working directory in to the out directory. `out` should be exactly the same as `lib` after `earthly` runs.

## What actually happens
`out` contains a subset of files from `lib`:

```
├── out
│   ├── gdkx-autocleanups.h
│   ├── gdkx11applaunchcontext.h
│   ├── gdkx11cursor.h
│   ├── gdkx11device-core.h
│   ├── gdkx11device-xi2.h
│   ├── gdkx11device.h
│   ├── gdkx11devicemanager-core.h
│   ├── gdkx11devicemanager-xi2.h
│   ├── gdkx11devicemanager.h
│   ├── gdkx11display.h
│   ├── gdkx11displaymanager.h
│   ├── gdkx11dnd.h
│   ├── gdkx11glcontext.h
│   ├── gdkx11keys.h
│   ├── gdkx11monitor.h
│   ├── gdkx11property.h
│   ├── gdkx11screen.h
│   ├── gdkx11selection.h
│   ├── gdkx11utils.h
│   ├── gdkx11visual.h
│   └── gdkx11window.h
```

dchw:
Resolves #689

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -515,10 +515,10 @@ func (b *Builder) saveArtifactLocally(ctx context.Context, artifact domain.Artif
 				}
 			} else {
 				// Remove pre-existing dest dir.
-				// err = os.RemoveAll(to)
-				// if err != nil {
-				// 	return errors.Wrapf(err, "rm -rf %s", to)
-				// }
+				err = os.RemoveAll(to)
+				if err != nil {
+					return errors.Wrapf(err, "rm -rf %s", to)
+				}
 			}
 		}
 
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -485,7 +485,7 @@ func (b *Builder) saveArtifactLocally(ctx context.Context, artifact domain.Artif
 			to = path.Join(artifact.Target.LocalPath, to)
 		} else {
 		}
-		if !srcIsDir && destIsDir {
+		if destIsDir {
 			// Place within dest dir.
 			to = path.Join(to, path.Base(from))
 		}
@@ -515,10 +515,10 @@ func (b *Builder) saveArtifactLocally(ctx context.Context, artifact domain.Artif
 				}
 			} else {
 				// Remove pre-existing dest dir.
-				err = os.RemoveAll(to)
-				if err != nil {
-					return errors.Wrapf(err, "rm -rf %s", to)
-				}
+				// err = os.RemoveAll(to)
+				// if err != nil {
+				// 	return errors.Wrapf(err, "rm -rf %s", to)
+				// }
 			}
 		}
 
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 .antlr/
 .env
 .vscode
+.tmp-earthly-out*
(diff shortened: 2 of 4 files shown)


builder/builder.go
```diff
@@ -515,10 +515,10 @@ func (b *Builder) saveArtifactLocally(ctx context.Context, artifact domain.Artif
 				}
 			} else {
 				// Remove pre-existing dest dir.
-				err = os.RemoveAll(to)
-				if err != nil {
-					return errors.Wrapf(err, "rm -rf %s", to)
-				}
+				// err = os.RemoveAll(to)
```
alexcb, line 518: Do you suspect we'll revisit this commented out code, or can this else case be removed?

dchw: It can be removed. I wanted to run all the tests, etc... wasn't quite ready for a review yet :D

dchw: I  agree. Commented out code should be deleted, thats what source control is for :D

alexcb: hah, whoops, I think I just clicked on it without realising it was a draft :)

builder/builder.go
```diff
@@ -515,10 +515,10 @@ func (b *Builder) saveArtifactLocally(ctx context.Context, artifact domain.Artif
...
+				// if err != nil {
+				// 	return errors.Wrapf(err, "rm -rf %s", to)
+				// }
```
vladaionescu, line 521: This doesn't look right.

The core idea is that if the artifact is a directory it should come out locally as that same directory, and not as a merged version of it+whatever already exists. Reason: If a build is being repeated and the new output no longer includes a certain file, then that file should now be removed.

vladaionescu (changes requested):
It might be useful to also have an integration test here.

builder/builder.go
dchw: Its not; I make my PR's from inside VSCode, and starting as a draft isn't an option :(

builder/builder.go
dchw: I wasnt quite ready for a review yet :D Yall are quick right now?

I question the need for cleanup; but maybe this means that a (`--no-cleanup`/`--merge`) option would be nice. For sure there have been times I wanted that; say comingling earthly-made builds with local made builds.

dchw: ALso this was here just to see what the CI would think since I still havent set up locally to run tests with the new QEmu dependency

dchw:
The integration test I added fails on older Earthly, by producing only two files. The new one includes the subdirectories.

alexcb (approved):
nice test :)

vladaionescu (approved):
Seems correct to me.

Issue #703, opened 2021-01-11
Weird AS LOCAL behavior with directories


jazzdan:
Working from this [simple sample repo](https://github.com/jazzdan/earthly-as-local-weirdness), with an Earthfile that looks like this:

```
everythings-flat:
    COPY qux .
    COPY bar .
    SAVE ARTIFACT ./* AS LOCAL out/standalone/

wat:
    COPY qux .
    COPY bar .
    SAVE ARTIFACT . AS LOCAL out/standalone/
```

When I run `earthly +everythings-flat` I would expect to get a multi level file tree in `out/standalone` that mirrors what is in the repo:

```
❯ tree .
.
├── Earthfile
├── bar
│   ├── baz
│   └── goo
│       └── gone
└── qux
    └── foo
```
. Instead what I get is this:

```
❯ earthly +everythings-flat && tree out/standalone
           buildkitd | Found buildkit daemon as docker container (earthly-buildkitd)
No ssh auth socket detected or zero keys loaded; falling back to https for auto auth values
             context | --> local context .
             context | [██████████] transferring .: ... 100%
   +everythings-flat | *cached* --> COPY qux .
   +everythings-flat | *cached* --> COPY bar .
   +everythings-flat | *cached* --> SAVE ARTIFACT ./* +everythings-flat/* AS LOCAL out/standalone/
              output | --> exporting outputs
              output | [██████████] copying files ... 100%
              output | [██████████] sending tarballs ... 100%
=========================== SUCCESS ===========================
   +everythings-flat | Artifact github.com/jazzdan/earthly-as-local-weirdness:main+everythings-flat/baz as local out/standalone/baz
   +everythings-flat | Artifact github.com/jazzdan/earthly-as-local-weirdness:main+everythings-flat/foo as local out/standalone/foo
   +everythings-flat | Artifact github.com/jazzdan/earthly-as-local-weirdness:main+everythings-flat/gone as local out/standalone/gone
out/standalone
├── baz
├── foo
└── gone
```

Only the leaf node files get copied over, none of the directory structure.

If I change the `SAVE ARTIFACT ./*` to a `SAVE ARTIFACT .`, as in `wat` I get:

```
❯ earthly +wat && tree out/standalone
           buildkitd | Found buildkit daemon as docker container (earthly-buildkitd)
No ssh auth socket detected or zero keys loaded; falling back to https for auto auth values
             context | --> local context .
             context | [██████████] transferring .: ... 100%
[...]

PR #704, opened 2021-01-12
Fix SAVE AS LOCAL issues

vladaionescu:
There were two issues here:

* Seems that for whatever reason, we were using the non-dir copy when saving artifacts in `SeparateArtifactsState`. This must have been due to copy-paste error of some sort, as the copy to `ArtifactsState` was correctly using the dir copy. This was causing copies to copy the contents and not the files themselves.
* Another, separate, issue was that when doing `SAVE ARTIFACT .` and `.` is `/`, then there is no wrapping directory name to use. So the wrapping directory would end up being `index-X`, which was never meant to be output to the user. ~So I reverted the previous fix (https://github.com/earthly/earthly/pull/690) which was using the wrapping dir name in some cases and have fixed it in a different manner (using wildcard detection).~ EDIT: In the end, we decided that this edge case is not worth pursuing as it creates inconsistencies, so I made it an error when trying to save the root directory as an artifact. Can't think of any use-case where this is absolutely necessary.

Fixes https://github.com/earthly/earthly/issues/703


hbenl:
I tested this with the following Earthfile:
```
FROM debian:buster-slim
WORKDIR /build

test:
  RUN mkdir foo && mkdir bar && touch foo/f.txt && touch bar/b.txt
  SAVE ARTIFACT ./* AS LOCAL out/
```

Here are my results with earth 0.4.4 and different `SAVE ARTIFACT` lines:
```
SAVE ARTIFACT ./* AS LOCAL out/
out
├── b.txt
└── f.txt
```
```
SAVE ARTIFACT . AS LOCAL out/
out
└── build
    ├── bar
    │   └── b.txt
    └── foo
        └── f.txt
```
```
SAVE ARTIFACT foo bar AS LOCAL out/
out
└── bar
    └── f.txt
```
```
SAVE ARTIFACT foo/f.txt bar AS LOCAL out/
out
└── bar
```
```
SAVE ARTIFACT foo bar/b.txt AS LOCAL out/
out
└── b.txt
    └── f.txt
```

It looks to me like there could be more than one bug, so please test all these scenarios with this PR.

builder/builder.go
```diff
@@ -471,7 +471,7 @@ func (b *Builder) saveArtifactLocally(ctx context.Context, artifact domain.Artif
 		}
 		return fmt.Errorf("cannot save artifact %s, since it does not exist", artifact.StringCanonical())
 	}
-	isWildcard := (len(fromGlobMatches) > 1)
+	isWildcard := strings.ContainsAny(fromPattern, `*?[`)
```
dchw, line 474: Are we shooting for compatibility with `filepath.Match`? https://golang.org/pkg/path/filepath/#Match

vladaionescu: Yes.. it looks like this is the kind of check filepath uses underneath.

examples/tests/.gitignore
```diff
@@ -0,0 +1,2 @@
+out*
+expected
```
vladaionescu, line 2: This doesn't seem necessary, right @dchw?

vladaionescu:
Thanks @hbenl - we've added a comprehensive test suite here as well in this PR: https://github.com/earthly/earthly/pull/704/files#diff-785e1a52262a239492022c86129a68340467a6a0a10cef599400bd6fab39d183

vladaionescu:
Oh BTW @hbenl - some of your examples make the assumption that `SAVE ARTIFACT` can take multiple sources, if I understand correctly. It cannot.. The second parameter is an optional destination path within the artifact environment. See also [reference docs](https://docs.earthly.dev/earthfile#save-artifact).

examples/tests/.gitignore
dchw: At least as of my last commit; the expected listing was exported as a separate artifact as just a bare file named `expected`.

vladaionescu: Right... but that wasn't actually exported on the host, right? It was exported in the environment of the Earthfile that runs the test.

dchw: Looking at how you tweaked the tests to save each, we should change it to `expected*`.

The file in general is here if you run them locally (outside the main tests `Earthfile`) to prevent any accidents committing the test results.

dchw: I can live with removing it and just being careful; if it bothers you.

dchw (approved):

examples/tests/.gitignore
vladaionescu: :+1:

vladaionescu:
Yay!! Thanks for the help with the extensive testing @dchw!

Issue #705, opened 2021-01-12
Inconsistent COPY --dir behaviour


hbenl:
In some situations, `COPY --dir` seems to ignore the `--dir` option or applies it only to some of the source arguments.
To reproduce, create the following file structure:
```
.
├── Earthfile
└── in
    ├── bar
    │   └── b.txt
    └── foo
        └── f.txt
```
with the following Earthfile:
```
FROM debian:buster-slim
WORKDIR /build

test:
  COPY --dir in copied
  RUN apt update && apt install tree && tree
```

Here's the result with different `COPY` lines (and sometimes an additional `RUN` line):
* this ignores the `--dir` argument entirely:
```
COPY --dir in copied
               +test | .
               +test | `-- copied
               +test |     |-- bar
               +test |     |   `-- b.txt
               +test |     `-- foo
               +test |         `-- f.txt
```
* but this works:
```
COPY --dir in copied/
              +test | .
               +test | `-- copied
               +test |     `-- in
               +test |         |-- bar
               +test |         |   `-- b.txt
               +test |         `-- foo
               +test |             `-- f.txt
```
* it also works if the target directory already exists:
```
RUN mkdir copied
COPY --dir in copied 
               +test | .
               +test | `-- copied
               +test |     `-- in
               +test |         |-- bar
               +test |         |   `-- b.txt
               +test |         `-- foo
               +test |             `-- f.txt
```
* this applies `--dir` to the second argument only:
```
COPY --dir in/foo in/bar copied
               +test | .
               +test | `-- copied
               +test |     |-- bar
               +test |     |   `-- b.txt
               +test |     `-- f.txt
```
* again, this can be fixed by adding a trailing `/`:
```
COPY --dir in/foo in/bar copied/
               +test | .
               +test | `-- copied
               +test |     |-- bar
               +test |     |   `-- b.txt
               +test |     `-- foo
               +test |         `-- f.txt
```
* or by creating the directory before copying:
```
RUN mkdir copied
COPY --dir in/foo in/bar copied 
               +test | .
               +test | `-- copied
               +test |     |-- bar
               +test |     |   `-- b.txt
               +test |     `-- foo
               +test |         `-- f.txt
```

vladaionescu:
The intention for `COPY --dir` is that it behaves similarly to a `cp -r`, with the key distinction that `COPY --dir` also creates the parent destination dirs if necessary.

The trailing `/` is meant to signal Earthly that the destination is a directory where the contents should be placed in.

[...]

PR #706, opened 2021-01-12
Fix inconsistencies related to COPY --dir

vladaionescu:
Fixes https://github.com/earthly/earthly/issues/705

### DIFF
--- a/earthfile2llb/listener.go
+++ b/earthfile2llb/listener.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"flag"
 	"fmt"
+	"path/filepath"
 	"regexp"
 	"strings"
 	"time"
@@ -280,6 +281,9 @@ func (l *listener) ExitCopyStmt(c *parser.CopyStmtContext) {
 		return
 	}
 	if allArtifacts {
+		if dest == "" || dest == "." || len(srcs) > 1 {
+			dest += string(filepath.Separator)
+		}
 		for _, src := range srcs {
 			err = l.converter.CopyArtifact(l.ctx, src, dest, platform, buildArgs.Args, *isDirCopy, *keepTs, *keepOwn, *chown, *ifExists)
 			if err != nil {
--- a/examples/tests/copy.earth
+++ b/examples/tests/copy.earth
@@ -3,24 +3,24 @@ FROM alpine:3.11
 WORKDIR /test
 
 all:
-    BUILD +copy1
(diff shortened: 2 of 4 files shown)


earthfile2llb/listener.go
```diff
@@ -280,6 +281,9 @@ func (l *listener) ExitCopyStmt(c *parser.CopyStmtContext) {
 		return
 	}
 	if allArtifacts {
+		if dest == "" || dest == "." || len(srcs) > 1 {
+			dest += string(filepath.Separator)
```
dchw, line 285: Thankyou for using filepath. One less place to touch when we go to Windows.

examples/tests/copy.earth
```diff
@@ -0,0 +1,167 @@
...
+    BUILD +copy3
+    BUILD +copy4
+    BUILD +copy5
+    BUILD +copy6
+    BUILD +copy-art1
+    BUILD +copy-art2
+    BUILD +copy-art3
+    BUILD +copy-art4
+    BUILD +copy-art5
+    BUILD +copy-art6
```
dchw, line 17: I mean its fine; but can we name them after what they are testing?

examples/tests/copy.earth
```diff
@@ -0,0 +1,167 @@
...
+    RUN find copied | sort | tee ./actual
+    RUN echo "copied
+copied/in
+copied/in/root
+copied/in/sub
+copied/in/sub/1
+copied/in/sub/1/file
+copied/in/sub/2
+copied/in/sub/2/file
+copied/in/sub/file" >./expected
```
dchw, line 47: If you use `-printf '%P\n' as a find option it omits the leading directory and clarifies it some IMO`

earthfile2llb/listener.go
vladaionescu: I'm not really convinced that it's the best use though :-) . Because `filepath` is host platform dependent. Whereas inside the Earthfile, we perhaps want to be guest-platform dependent. So there's still stuff to tweak.

dchw (approved):

examples/tests/copy.earth
vladaionescu: Seems that `-printf` is not available in the `find` available in `alpine` for some reason :-(

examples/tests/copy.earth
vladaionescu: Done

PR #709, opened 2021-01-13
Create test for private GIT CLONE

alexcb:


alexcb:
This has been done in https://github.com/earthly/earthly/pull/964

PR #718, opened 2021-01-15
Add --remote-cache compatibility matrix

dchw:
I manually tested each with `--remote-cache` and with `--ci`

### DIFF
--- a/docs/guides/shared-cache.md
+++ b/docs/guides/shared-cache.md
@@ -24,6 +24,20 @@ Earthly makes available two types of shared caches:
 
 For a summary of the differences see [comparison between inline and explicit cache](#comparison-between-inline-and-explicit-cache).
 
+### Compatibility with major registry providers
+
+Not all registries support the needed manifest formats to allow the usage of each kind of cache. Here is a compatibility matrix for many popular registries:
+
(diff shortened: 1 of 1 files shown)


alexcb (approved):

vladaionescu (approved):
Great stuff!

I would perhaps move this higher up in the guide. Perhaps just above "Types of shared cache"?

docs/guides/shared-cache.md
```diff
@@ -133,6 +133,21 @@ If a project has multiple CI pipelines or `earthly` invocations, it is recommend
 It is currently not possible to push both inline and implicit caches currently.
 {% endhint %}
 
+### Compatibility with major registry providers
+
+Not all registries support the needed manifest formats to allow the usage of `--remote-cache`. Here is a compatibility matrix for many popular registries:
```
vladaionescu, line 138: Oh - also I think both inline cache and explicit cache are affected. So the wording should be adjusted to make it clear that it applies to inline cache too.

vladaionescu (changes requested):

docs/guides/shared-cache.md
dchw: Looks like Inline and Remote are different enough to warrant a different column. Ill make that happen and then move it further up.

agbell (approved):

PR #720, opened 2021-01-20
Add integration tests for all registries we have guides for

dchw:
Pushes and Pulls a unique image to and from:
 * GCR
 * Google Artifact Registry
 * ACR
 * ECR


examples/tests/cloud-push-pull/Earthfile
```diff
@@ -0,0 +1,136 @@
...
+        tar -xvzf google-cloud-sdk-324.0.0-linux-x86_64.tar.gz && \
+        ./google-cloud-sdk/install.sh -q && \
+        ls -la /test/google-cloud-sdk
+    
+    ENV PATH $PATH:/test/google-cloud-sdk/bin
+
+    RUN --secret=GCP_KEY=+secrets/earthly-technologies/gcp/ci-cd-key \
+        echo $GCP_KEY > key.json
+
+    RUN gcloud auth activate-service-account --key-file /test/key.json
```
dchw, line 24: The GCP is under my name; if you need or want access yell? Vlad might defacto have access?

examples/tests/cloud-push-pull/Earthfile
```diff
@@ -0,0 +1,136 @@
...
+
+    RUN --privileged \
+    --entrypoint \
+    --mount=type=tmpfs,target=/tmp/earthly \
+    -- -P +pull
+
+amazon-base:
+    ENV GLIBC_VER=2.31-r0
+
+    # install glibc compatibility for alpine
```
dchw, line 79: Ugh I hat that I have to do this, but the base image is Alpine-based. 100% a copy-paste job from someone on Stack Overflow. It works, not gonna question it.

examples/tests/cloud-push-pull/azure-container-registry.earth
```diff
@@ -0,0 +1,12 @@
...
+push:
+    RUN date | md5sum | head -c 32 > motd
+    ENTRYPOINT cat motd
+
+    SAVE IMAGE --push earthlyintegrationtest.azurecr.io/test:latest
+
+pull:
+    WITH DOCKER --pull earthlyintegrationtest.azurecr.io/test:latest
+        RUN docker run earthlyintegrationtest.azurecr.io/test:latest
+    END
```
vladaionescu, line 12: Theoretically a `FROM earthlyintegrationtest.azurecr.io/test:latest` would involve a pull just as well. Though this works too.

examples/tests/cloud-push-pull/Earthfile
```diff
@@ -0,0 +1,136 @@
...
+        /usr/local/aws-cli/v2/*/dist/awscli/examples \
+    && apk --no-cache del \
+        binutils \
+        curl \
+    && rm glibc-${GLIBC_VER}.apk \
+    && rm glibc-bin-${GLIBC_VER}.apk \
+    && rm -rf /var/cache/apk/*
+
+    RUN apk add go
+    RUN go get -u github.com/awslabs/amazon-ecr-credential-helper/ecr-login/cli/docker-credential-ecr-login
```
dchw, line 106: Also have to install from source since thats really the only way from their installation instructions that works for us here.

examples/tests/cloud-push-pull/azure-container-registry.earth
```diff
@@ -0,0 +1,12 @@
+FROM earthly/dind:alpine-main
+
+push:
+    RUN date | md5sum | head -c 32 > motd
```
dchw, line 4: Should make a unique file; unless we run the same target twice in a second. Force different layers to _actually_ push.

vladaionescu (approved):
Neat!

dchw (commented):
I could go the extra mile here and do some `--no-cache`, and also do some comparison of the random hash... but this seems good enough.

vladaionescu:
I wonder if we should split this into a separate GH actions job? Does this take a long time in CI?

examples/tests/cloud-push-pull/azure-container-registry.earth
dchw: I wanted to actually run it to have the hash I generated print. To ensure its functional.

examples/tests/cloud-push-pull/azure-container-registry.earth
```diff
@@ -0,0 +1,12 @@
+FROM earthly/dind:alpine-main
```
vladaionescu, line 1: I would use the released version of the dind image: `earthly/dind:alpine`.

examples/tests/cloud-push-pull/azure-container-registry.earth
vladaionescu: Yeah, makes sense.

dchw:
> I wonder if we should split this into a separate GH actions job? Does this take a long time in CI?

@vladaionescu not super long; at least no worse than some of our worst examples.

examples/tests/cloud-push-pull/azure-container-registry.earth
dchw: Good catch

dchw:
@vladaionescu Too much logspam running it as part of the main test suite; plus it seemed weirdly unstable on macOS when I ran it alone via Buildkite. Seems fine as a GH action? Mind giving it a run on your end?

vladaionescu:
> plus it seemed weirdly unstable on macOS when I ran it alone via Buildkite. Seems fine as a GH action? Mind giving it a run on your end?

I've ran it 3 times and passed all 3. It seems stable on my macOS. Maybe network issues on mac stadium?

PR #725, opened 2021-01-21
run commands locally with a new LOCALLY statement

alexcb:


earthfile2llb/listener.go
```diff
@@ -158,6 +160,15 @@ func (l *listener) ExitFromStmt(c *parser.FromStmtContext) {
 	for i, ba := range buildArgs.Args {
 		buildArgs.Args[i] = l.expandArgs(ba, true)
 	}
+
+	if imageName == "LOCAL" {
+		imageName = "alpine:3.13"
```
vladaionescu, line 165: I would use an even lighter image here - like busybox.

earthfile2llb/converter.go
```diff
@@ -297,6 +297,16 @@ func (c *Converter) CopyClassical(ctx context.Context, srcs []string, dest strin
 			dest))
 }
 
+func (c *Converter) RunLocal(ctx context.Context, args []string) {
+	finalArgs := withShellAndEnvVars(args, []string{}, true, false)
+	c.mts.Final.MainState = c.mts.Final.MainState.Run(
+		llb.Args(finalArgs),
+		llb.AddMount("/run_on_localhost_hack", llb.Scratch()),
+		llb.IgnoreCache,
+		llb.WithCustomNamef("%sRUN %s", c.vertexPrefix(), strings.Join(args, " ")),
```
vladaionescu, line 306: I wonder if we should mark the output in a special way when it's executing on the local machine.

earthfile2llb/listener.go
vladaionescu, line 165: I would move this hack to the converter - it helps keep the listener focused on syntax only. Perhaps there needs to be a separate call for `FromLocal` in the converter.

earthfile2llb/listener.go
```diff
@@ -158,6 +160,15 @@ func (l *listener) ExitFromStmt(c *parser.FromStmtContext) {
...
+		// TODO trigger local mode
```
vladaionescu, line 166: `TODO` can be removed?

earthfile2llb/converter.go
alexcb: that's a good idea, maybe `LOCAL RUN` or `RUN LOCAL`?

e.g. 
```
alex@mah:~/gh/earthly/earthly$ ./build/linux/amd64/earthly --push +test-local 
golang:1.13-alpine3.11 | --> Load metadata linux/amd64
      busybox:1.32.1 | --> Load metadata linux/amd64
         +test-local | --> FROM busybox:1.32.1
         +test-local | [██████████] resolve docker.io/library/busybox:1.32.1@sha256:c5439d7db88ab5423999530349d327b04279ad3161d7596d2126dfb5b02bfd1f ... 100%
         +test-local | --> LOCAL RUN cat /proc/1/cgroup | grep '^[0-9]\+:cpuset:/$'
         +test-local | 12:cpuset:/
              output | [██████████] sending tarballs ... 100%
=========================== SUCCESS ===========================
         +test-local | 12:cpuset:/
         +test-local | --> LOCAL RUN --push touch /tmp/earthly-test-local
```

vladaionescu: Or even `+test-local *local* | --> RUN cat ...` (this is similar to how we print *failed*). We can then even use a special color (orange?) for local commands to draw more attention.

earthfile2llb/listener.go
alexcb: Good idea on the `FromLocal` function in the converter.

earthfile2llb/converter.go
alexcb: that definitely makes it stand out more.
```
alex@mah:~/junk/local$ /home/alex/gh/earthly/earthly/build/linux/amd64/earthly +both 
       alpine:latest | --> Load metadata linux/amd64
      busybox:1.32.1 | --> Load metadata linux/amd64
              +step1 | --> FROM busybox:1.32.1
              +step2 | --> FROM alpine:latest
              +step1 | [██████████] resolve docker.io/library/alpine:latest@sha256:d9a7354e3845ea8466bb00b22224d9116b183e594527fb5b6c3d30bc01a20378 ... 100%
              +step1 | [██████████] resolve docker.io/library/busybox:1.32.1@sha256:c5439d7db88ab5423999530349d327b04279ad3161d7596d2126dfb5b02bfd1f ... 100%
      +step1 *local* | --> RUN whoami > myuser
      +step1 *local* | --> RUN pwd
      +step1 *local* | /home/alex/junk/local
      +step1 *local* | --> RUN cat /proc/1/cgroup
      +step1 *local* | 12:cpuset:/
      +step1 *local* | 11:perf_event:/
      +step1 *local* | 10:devices:/
      +step1 *local* | 9:blkio:/
      +step1 *local* | 8:freezer:/
      +step1 *local* | 7:cpu,cpuacct:/
      +step1 *local* | 6:net_cls,net_prio:/
      +step1 *local* | 5:rdma:/
      +step1 *local* | 4:memory:/
      +step1 *local* | 3:pids:/
      +step1 *local* | 2:hugetlb:/
      +step1 *local* | 1:name=systemd:/init.scope
      +step1 *local* | 0::/init.scope
              +step2 | *cached* --> RUN cat /proc/1/cgroup
              +step2 | *cached* --> RUN sleep 10
              +step2 | *cached* --> RUN whoami > myuser
              +step2 | *cached* --> RUN cat myuser
              output | [██████████] sending tarballs ... 100%
=========================== SUCCESS ===========================
              +step1 | Did not execute push command RUN --push echo push flag1. Use earthly --push to enable pushing
              +step2 | Did not execute push command RUN --push echo push flag2. Use earthly --push to enable pushing
```

maybe we could leave out the orange feature for another PR (it'll also be a good first issue for anyone wanting to contribute).

alexcb:
This PR makes use of my buildkit hacks which I've opened up a PR under https://github.com/earthly/buildkit/pull/24

Earthfile
```diff
@@ -285,5 +285,12 @@ examples:
     BUILD ./examples/multiplatform-cross-compile+build-all-platforms
     BUILD github.com/earthly/hello-world:main+hello
 
+test-local:
+    FROM LOCAL
+    # If run inside a container, one would expect something like 
+    # 12:cpuset:/docker/e4b6b1698e80c6a2d8ccdfcd689a1ce5828587ada1337ddcb0b9a80caf0087a4/buildkit/83zo724g6vr1ipssy0dic9kpa
+    RUN cat /proc/1/cgroup | grep '^[0-9]\+:cpuset:/$'
+    RUN --push touch /tmp/earthly-test-local
+
```
vladaionescu, line 294: I would put this under `examples/tests/local` to avoid collecting tests here.

earthfile2llb/converter.go
```diff
@@ -291,12 +299,37 @@ func (c *Converter) CopyClassical(ctx context.Context, srcs []string, dest strin
...
+	runStr := fmt.Sprintf("RUN %s%s", strIf(pushFlag, "--push "), strings.Join(args, " "))
+	finalArgs := withShellAndEnvVars(args, []string{}, true, false)
+	opts := []llb.RunOption{
+		llb.Args(finalArgs),
+		llb.AddMount(localhost.RunOnLocalHostMagicStr, llb.Scratch()), // hack to tell buildkit to run this locally
+		llb.IgnoreCache,
+		llb.WithCustomNamef("%s%s", c.vertexPrefix(true), runStr),
+	}
+
+	if pushFlag {
```
vladaionescu, line 318: Maybe we just don't support push under `FROM LOCAL`? Wonder what the implications for allowing this...

Earthfile
alexcb: sounds good, i'll also move the test-fail test into a separate directory too.

earthfile2llb/converter.go
alexcb: Adding support will allow someone to have a bit more control over pushes or not, but they could also just have two separate local targets to support having one perform a push and the other not.

I don't feel strongly attached to it, we can remove it if you think it's a bit too confusing.

vladaionescu: Perhaps @dchw can look at this and see if it conflicts with his run-after-push work.

dchw: Nope this is fine. Almost exactly the same as in `RUN`. This is adaptable fairly easily. Were good here.

vladaionescu (approved):

earthfile2llb/parser/EarthParser.g4
```diff
@@ -44,6 +45,8 @@ fromStmt: FROM (WS stmtWords)?;
 
 fromDockerfileStmt: FROM_DOCKERFILE (WS stmtWords)?;
 
+locallyStmt: LOCALLY WS*;
```
vladaionescu, line 48: We should include `(WS stmtWords)?` here.

I know it doesn't take arguments, but the errors that we produce from `listener` are usually more informative, in case someone does end up adding arguments.

vladaionescu (approved):

earthfile2llb/parser/EarthParser.g4
alexcb: I was wondering if that was needed or not, i'll add it in.

PR #728, opened 2021-01-22
Allow commands to run after --push

dchw:
Current progress. I know there are things to clean up...

Whats left (that I know about):

- [ ] Fix Phase Ordering; especially when got multiple child `BUILD`s
- [ ] Refactor unrolled loops
- [ ] Formalize the "commandStr" for unpushed/unpostpushed items
- [ ] Phase Images
- [ ] Do not let `WITH DOCKER` and friends pull images when the commands are post-push
- [ ] Clean up and add integration tests


dchw:
Too many edge cases here to allow everything; and there are too many places to touch.

Instead, I am going to go with simply allowing `SAVE ARTIFACT`, and fix `SAVE IMAGE --push` to include the state of all `RUN --push` so far. This should get us most of the benefit without the massive refactor and complications.

PR #729, opened 2021-01-22
Add documentation for LOCALLY directive

alexcb:

### DIFF
--- a/docs/earthfile/earthfile.md
+++ b/docs/earthfile/earthfile.md
@@ -106,6 +106,29 @@ Sets a value override of `<value>` for the Dockerfile build arg identified by `<
 
 In a multi-stage Dockerfile, sets the target to be used for the build. This option is similar to the `docker build --target <target-name>` option.
 
+## LOCALLY (**experimental**)
+
+#### Synopsis
+
+* `LOCALLY`
+
+#### Description
+
+{% hint style='danger' %}
+##### Important
+
+This feature is currently in **Experimental** stage
+
+* The feature may break, be changed drastically with no warning, or be removed altogether in future versions of Earthly.
+* Check the [GitHub tracking issue](https://github.com/earthly/earthly/issues/580) for any known problems.
+* Give us feedback on [Slack](https://earthly.dev/slack) in the `#locally` channel.
+{% endhint %}
+
+The `LOCALLY` command instructs earthly to execute the following `RUN` commands locally on the host system.
+This feature should be used with caution as locally run commands have no guarantee they will behave the same on different systems.
+
+Only `RUN` commands are supported under a `LOCALLY` defined target; futhermore only `RUN`'s `--push` flag is supported.
+
 ## RUN
 
 #### Synopsis


docs/earthfile/earthfile.md
```diff
@@ -106,6 +106,16 @@ Sets a value override of `<value>` for the Dockerfile build arg identified by `<
...
 
+## LOCALLY (**experimental**)
+
+#### Synopsis
+
+* `LOCALLY`
+
+#### Description
+
+The `LOCALLY` command instructs earthly to execute the following `RUN` commands locally on the host system. This feature should be used with caution as locally run commands have no guarantee they will behave the same on different systems; where possible it's recommended to use the `FROM` directive instead.
```
vladaionescu, line 117:
```suggestion
The `LOCALLY` command instructs earthly to execute the following `RUN` commands locally on the host system. This feature should be used with caution as locally run commands have no guarantee they will behave the same on different systems.
```

Feels like the final `where` makes the paragraph too complex.

vladaionescu, line 117: We should also say that `LOCALLY` "initializes a new build environment" - the same way FROM does.

Perhaps
> The `LOCALLY` initializes a new build environment and instructs earthly to execute the following `RUN` commands locally...

vladaionescu: We should also say that currently only `RUN` commands are allowed after `LOCALLY`.

docs/earthfile/earthfile.md
```diff
@@ -106,6 +106,16 @@ Sets a value override of `<value>` for the Dockerfile build arg identified by `<
 
 In a multi-stage Dockerfile, sets the target to be used for the build. This option is similar to the `docker build --target <target-name>` option.
 
+## LOCALLY (**experimental**)
```
vladaionescu, line 109: I wonder if we should include a notice like this one: https://github.com/earthly/earthly/blob/main/docs/guides/shared-cache.md#important given that this is a particularly big feature.

docs/earthfile/earthfile.md
alexcb: It seems a bit odd saying it initialises a new environment, because it doesn't initializes anything in the context of creating a new image or snapshot. 

Maybe something like The `LOCALLY` defines a new build target in which `RUN` commands are executed locally on the host system.

vladaionescu: Yeah.. that's true actually. Not sure how best to say that it drops whatever build environment you have before (eg any `FROM` above `LOCALLY`).

Perhaps your original text is the best then. I wouldn't say "defines a new build target" as that's what `target:` does.

docs/earthfile/earthfile.md
alexcb: good plan, I just setup a new slack channel for this.

docs/earthfile/earthfile.md
alexcb: ok, lets keep it as-is for now and perhaps change it later.

vladaionescu (approved):

PR #732, opened 2021-01-25
Add tests for solver monitor's parser method

alexcb:
- these tests were written to better understand how the vertex prefix parser works


vladaionescu (approved):
Thanks!

I have a mental TODO to use the vertex metadata map more and not abuse use the vertex name as much. I'll get to it sometime.

Issue #586, opened 2020-12-05
Proposal: Allow save artifact commands after push commands


vladaionescu:
### Background

**Outputting only at the end is limiting**. The principle of “first perform all the build steps, and only after they complete successfully, perform outputting + pushing” is limiting for certain use cases. Examples:
  * `RUN --push` and then `SAVE ARTIFACT` (eg a terraform backup)
  * A case where an artifact publishing tool auto-increments version number on publish, but then that number needs to be used in other places too

### Proposal

Currently the following is an error in Earthly:

```
RUN --push ...
RUN ...
```

The proposal is to allow this.

### How it works

* If `earth --push` flag is provided, everything executes as expected.
* If no push flag is provided to `earth`, such a build would fail from the get-go, explaining why it would not be able to complete the build.

dchw:
With the addition of allowing `SAVE ARTIFACT` after `--push`, allowing `RUN --push` to affect the state of later `SAVE IMAGE --push`, and `RUN --no-cache`, it seems like almost all use cases have been satisfied.

PR #735, opened 2021-01-27
Allow SAVE ARTIFACT after a `--push`

dchw:


builder/builder.go
```diff
@@ -68,9 +68,10 @@ type BuildOpt struct {
 
 // Builder executes Earthly builds.
 type Builder struct {
-	s        *solver
-	opt      Opt
-	resolver *buildcontext.Resolver
+	s         *solver
+	opt       Opt
+	resolver  *buildcontext.Resolver
+	builtMain bool
```
dchw, line 74: Same concept as phases; just need to know one thing instead of a list index... we may need phases in future

builder/builder.go
```diff
@@ -124,6 +125,9 @@ func (b *Builder) convertAndBuild(ctx context.Context, target domain.Target, opt
 	destPathWhitelist := make(map[string]bool)
 	manifestLists := make(map[string][]manifest) // parent image -> child images
 	var mts *states.MultiTarget
+	depIndex := 0
+	imageIndex := 0
+	dirIndex := 0
```
dchw, line 130: Indexes need to increase monotonically across all artifacts to avoid overwriting issues, so keep them out side and close on them

builder/builder.go
```diff
@@ -381,8 +401,22 @@ func (b *Builder) convertAndBuild(ctx context.Context, target domain.Target, opt
...
+	return sts.MainState
+}
+
+func (b *Builder) targetPhaseArtifacts(sts *states.SingleTarget) []states.SaveLocal {
+	if b.builtMain {
+		return sts.RunPush.SaveLocals
+	}
+	return sts.SaveLocals
+}
+
```
dchw, line 417: Similar things to phases to enable use of the gwClient. However, its not quite the same as before where the state tracks it. Kept it simpler using the existing RunPush struct, and phases only happen here in builder where they are cared about now.

dchw: Open to pushing this down and cleaning up some more; perhaps as a separate PR though? Its just touching a lot to do that

earthfile2llb/converter.go
```diff
@@ -430,19 +430,35 @@ func (c *Converter) SaveArtifact(ctx context.Context, saveFrom string, saveTo st
...
+					"%sSAVE ARTIFACT %s%s %s AS LOCAL %s",
+					c.vertexPrefix(false), strIf(ifExists, "--if-exists "), saveFrom, artifact.String(), saveAsLocalTo))
+		} else {
+			separateArtifactsState = llbutil.CopyOp(
+				c.mts.Final.MainState, []string{saveFrom}, separateArtifactsState,
+				saveToAdjusted, true, true, keepTs, "root:root", ifExists,
+				llb.WithCustomNamef(
+					"%sSAVE ARTIFACT %s%s %s AS LOCAL %s",
+					c.vertexPrefix(false), strIf(ifExists, "--if-exists "), saveFrom, artifact.String(), saveAsLocalTo))
+		}
```
dchw, line 447: State to copy from depends on if we are pushing

earthfile2llb/listener.go
```diff
@@ -462,7 +458,7 @@ func (l *listener) ExitSaveArtifact(c *parser.SaveArtifactContext) {
 	saveFrom := l.expandArgs(fs.Args()[0], false)
 	saveTo = l.expandArgs(saveTo, false)
 	saveAsLocalTo = l.expandArgs(saveAsLocalTo, false)
-	err = l.converter.SaveArtifact(l.ctx, saveFrom, saveTo, saveAsLocalTo, *keepTs, *keepOwn, *ifExists)
+	err = l.converter.SaveArtifact(l.ctx, saveFrom, saveTo, saveAsLocalTo, *keepTs, *keepOwn, *ifExists, l.pushOnlyAllowed)
```
dchw, line 461: `pushOnlyAllowed` is here to signify we have encountered a push. Could rename this var; but didn't because im not globally allowing everything like before.

examples/tests/Earthfile
```diff
@@ -509,3 +509,17 @@ file-copying:
...
+    RUN --privileged \
+        --entrypoint \
+        --mount=type=tmpfs,target=/tmp/earthly \
+        -- +test
+
+    # test that cant copy saved after push
+    RUN --privileged \
+        --entrypoint \
+        --mount=type=tmpfs,target=/tmp/earthly \
+        -- +copy-test 2>&1 | perl -pe 'BEGIN {$status=1} END {exit $status} $status=0 if /not found/;'
```
dchw, line 525: Can't copy artifacts that were saved in a different targets --push

builder/builder.go
```diff
@@ -381,8 +401,22 @@ func (b *Builder) convertAndBuild(ctx context.Context, target domain.Target, opt
...
 func (b *Builder) stateToRef(ctx context.Context, gwClient gwclient.Client, state llb.State, platform *specs.Platform) (gwclient.Reference, error) {
-	if b.opt.NoCache {
+	if b.opt.NoCache || b.builtMain {
```
vladaionescu, line 419: This doesn't look right. I think it should be

```suggestion
	if b.opt.NoCache && !b.builtMain {
```

Meaning that for the push phase, we do not add the ignore cache flag here (instead we rely on the converter having added this flag to individual commands).

If we add the ignore cache flag here for the push phase, then everything the push phase depends on (including the main phase) will be executed with no cache - so the main phase ends up being executed twice.

builder/builder.go
vladaionescu: Not sure what you have in mind by "cleaning up some more"?

This looks pretty good BTW.

vladaionescu (commented):
This looks pretty good - left an important comment, though.

I assume the output has two `===== SUCCESS ======` lines, right? We should print out a more obvious distinction for the two phases. Maybe `===== Build phase SUCCESS ======` and `===== Push phase SUCCESS ======`. Or something like that - whatever looks pretty enough.

earthfile2llb/listener.go
vladaionescu: If you wanted you could do
```suggestion
	isPushPhase := l.pushOnlyAllowed
	err = l.converter.SaveArtifact(l.ctx, saveFrom, saveTo, saveAsLocalTo, *keepTs, *keepOwn, *ifExists, isPushPhase)
```

builder/builder.go
dchw: Cleaning up meaning pushing this down into the states package. Think of it as phases-lite; so the builder doesn't have to be so grabby with SingleTargets fields

vladaionescu: I think the way it is in this PR is simple enough. We could do that if things get more complex.

earthfile2llb/listener.go
dchw: Ehhhhhhhh that feels like extra ceremony. Not worth it IMO; I would rather rename it to `pushEncountered` across all of converter if I went renaming; since that preserves the business logic meaning well enough

vladaionescu: 🤷 

I like ceremonies :-)

builder/builder.go
dchw: What you initially suggested is what I hade before; but I noticed that solving the states for the separate artifacts was showing up as cached. I used to have some `ls` sentinels in here to sho when things were being re-run, but I must have pulled them out before swapping to this.

I was trying to get `SAVE ARTIFACT` inside `--push` to _also_ not cache when running with `--no-cache`. Ill reqork this some to get both working

dchw:
@vladaionescu I agree on more distinctive success lines. Though, that is also the current behavior.

Sticking the name in the line feels fine; All the alternatives I can think of would require a little more work.

vladaionescu:
Yeah, let's stick the name in. I think the current behavior is inconsistent: only prints success once, but can print both success and then fail.

dchw:
@vladaionescu Updated to your logic.

The one downside to this is that unchanged artifacts after `--push` still use cache even when specifying `--no-cache` on the CLI.

dchw:
How do you feel about solving artifacts state with something like this:

```go
func (b *Builder) artifactStateToRef(ctx context.Context, gwClient gwclient.Client, state llb.State, platform *specs.Platform) (gwclient.Reference, error) {
	if b.opt.NoCache || b.builtMain{
		state = state.SetMarshalDefaults(llb.IgnoreCache)
	}
	return llbutil.StateToRef(ctx, gwClient, state, platform, b.opt.CacheImports)
}
```

vladaionescu:
Sounds good - I think we had that bug before too.

vladaionescu (approved):

dchw:
This is part of the actual solution we plan to ship in place of #728. Relates to #586.

PR #738, opened 2021-01-28
Fix build

vladaionescu:
* fix dind build for arm
* arm/v7 dind on Ubuntu no longer supported. Seems that it cannot be installed via apt-get:
  ```
  +dind-ubuntu *failed* | Package docker-ce is not available, but is referred to by another package.
  +dind-ubuntu *failed* | This may mean that the package is missing, has been obsoleted, or
  +dind-ubuntu *failed* | is only available from another source

  +dind-ubuntu *failed* | E: Package 'docker-ce' has no installation candidate
  +dind-ubuntu *failed* | E: Unable to locate package docker-ce-cli
  +dind-ubuntu *failed* | E: Unable to locate package containerd.io
  +dind-ubuntu *failed* | E: Couldn't find any package by glob 'containerd.io'
  +dind-ubuntu *failed* | E: Couldn't find any package by regex 'containerd.io'
  ```
* reorganize GitHub actions: spread build of +all across multiple jobs
* don't push anything else other than +prerelease

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,7 +40,7 @@ jobs:
         if: ${{ failure() }}
 
   misc-tests:
-    name: Misc tests
+    name: misc tests
     runs-on: ubuntu-latest
     env:
       FORCE_COLOR: 1
@@ -118,7 +118,7 @@ jobs:
         if: ${{ failure() }}
 
   cloud-integration:
-    name: Cloud Integration
+    name: cloud-integration
     runs-on: ubuntu-latest
     env:
       FORCE_COLOR: 1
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -263,7 +263,16 @@ jobs:
...
+    needs:
+      - tests
+      - misc-tests
+      - examples
+      - cloud-integration
+      - secrets-integration
+      - private-repo-test
+      - secrets
+      - test-local
+      - all
```
alexcb, line 275: nice reformatting change

alexcb (approved):

PR #739, opened 2021-01-28
Add RUN `--no-cache`

### LINKED ISSUE #585: Proposal: RUN --no-cache
Introduce a new flag for the `RUN` command which instructs the Earthly to never use cache for it.

```Dockerfile
RUN --no-cache ...
```

Use-cases:

* Temporarily add this flag in the Earthfile to debug flaky tests
* Otherwise, use in situations where the command needs to actually run every time (eg refresh some credentials)

dchw:
Documentation coming in a later PR to go into `next`; assuming this implementation is acceptable.

Resolves #585

### DIFF
--- a/cmd/earthly/main.go
+++ b/cmd/earthly/main.go
@@ -1065,8 +1065,7 @@ func (app *earthlyApp) insertZSHCompleteEntry() error {
 func (app *earthlyApp) run(ctx context.Context, args []string) int {
 	err := app.cliApp.RunContext(ctx, args)
 
-	rpcRegex := regexp.MustCompile(`rpc error: code = .+ desc = .+:\s`)
-
+	rpcRegex := regexp.MustCompile(`(?U)rpc error: code = .+ desc = .+:\s`)
 	if err != nil {
 		if strings.Contains(err.Error(), "security.insecure is not allowed") {
 			app.console.Warnf("Error: --allow-privileged (-P) flag is required\n")
--- a/earthfile2llb/listener.go
+++ b/earthfile2llb/listener.go
@@ -427,10 +427,6 @@ func (l *listener) ExitSaveArtifact(c *parser.SaveArtifactContext) {
 	if l.shouldSkip() {
 		return
 	}
-	if l.pushOnlyAllowed {
-		l.err = fmt.Errorf("no non-push commands allowed after a --push: %s", c.GetText())
-		return
-	}
 	fs := flag.NewFlagSet("SAVE ARTIFACT", flag.ContinueOnError)
 	keepTs := fs.Bool("keep-ts", false, "Keep created time file timestamps")
 	keepOwn := fs.Bool("keep-own", false, "Keep owner info")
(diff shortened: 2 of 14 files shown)


cmd/earthly/main.go
```diff
@@ -1065,8 +1065,7 @@ func (app *earthlyApp) insertZSHCompleteEntry() error {
 func (app *earthlyApp) run(ctx context.Context, args []string) int {
 	err := app.cliApp.RunContext(ctx, args)
 
-	rpcRegex := regexp.MustCompile(`rpc error: code = .+ desc = .+:\s`)
-
+	rpcRegex := regexp.MustCompile(`(?U)rpc error: code = .+ desc = .+:\s`)
```
dchw, line 1068: Ancillary fix: Root errors with `:` were getting trimmed. This makes the regex ungreedy, so it will only prune the `rpc error` portion.

earthfile2llb/listener.go
```diff
@@ -330,6 +330,7 @@ func (l *listener) ExitRunStmt(c *parser.RunStmtContext) {
 		"Include the entrypoint of the image when running the command")
 	withDocker := fs.Bool("with-docker", false, "Deprecated")
 	withSSH := fs.Bool("ssh", false, "Make available the SSH agent of the host")
+	noCache := fs.Bool("no-cache", false, "Always run this specific item, ignoring cache. Like --push, but without --push on the CLI")
```
vladaionescu, line 333: I wouldn't mention push, as it might be confusing for someone who doesn't fully understand --push.

```suggestion
	noCache := fs.Bool("no-cache", false, "Always run this specific item, ignoring cache")
```

earthfile2llb/listener.go
```diff
@@ -370,6 +371,10 @@ func (l *listener) ExitRunStmt(c *parser.RunStmtContext) {
 			l.err = fmt.Errorf("the --privileged flag has no effect when used with the LOCALLY directive: %s", c.GetText())
 			return
 		}
+		if *noCache {
```
dchw, line 374: Didn't make any special effort here. Of course its not cached? Any bad or invalid assumptions I made and need to check?

earthfile2llb/withdockerrun.go
```diff
@@ -164,7 +164,7 @@ func (wdr *withDockerRun) Run(ctx context.Context, args []string, opt WithDocker
 		return errors.Wrap(err, "compute dind id")
 	}
 	shellWrap := makeWithDockerdWrapFun(dindID, tarPaths, opt)
-	return wdr.c.internalRun(ctx, finalArgs, opt.Secrets, opt.WithShell, shellWrap, false, false, runStr, runOpts...)
+	return wdr.c.internalRun(ctx, finalArgs, opt.Secrets, opt.WithShell, shellWrap, false, false, false, runStr, runOpts...)
```
dchw, line 167: `WITH DOCKER` just doesnt do it. I need to test this still; now that I type this...

examples/tests/Earthfile
```diff
@@ -509,3 +510,22 @@ file-copying:
...
+run-no-cache:
+    COPY run-no-cache.earth ./Earthfile
+
+    # Run twice to allow the second one to attempt to cache things
+    RUN --privileged \
+        --entrypoint \
+        -- +test
+    RUN --privileged \
+        --entrypoint \
+        -- +test 2>&1 | perl -pe 'BEGIN {$status=0} END {exit $status} $status=1 if /\*cached\* --> .* motd2/;'  # Fail if we cached any of the motd2 lines, which are after the --no-cache
```
dchw, line 523: Note statuses flipped here. Its easier to test that anything I dont want cached isnt, rather than everything I want is.

earthfile2llb/listener.go
alexcb: This seems like a good decision.

PR #743, opened 2021-01-29
Add test to make sure repo hash stays the same

alexcb:
- this is to prevent us from changing the hashing function
which would make it impossible to track repos


vladaionescu (approved):

alexcb:
I had to change this test to be run directly under GHA because it requires the `.git` directory to be present.

vladaionescu (approved):

PR #751, opened 2021-01-30
Turn non-constant build args into constant ones on the spot

vladaionescu:
Note to reviewer: it's easiest to review this commit-by-commit.

This enables the use of variable build args in non-RUN expressions. For example, something like the following is now possible:

```
ARG src=$(cat ./what-to-copy.txt)
COPY $src ./
```

It also simplifies the TargetInput-related logic as we don't have to account for the provenience tree of build args.


alexcb (approved):
Nice change!

PR #753, opened 2021-02-01
Eliminate unnecessary double build and double push of save images

vladaionescu:


builder/builder.go
```diff
@@ -175,7 +175,7 @@ func (b *Builder) convertAndBuild(ctx context.Context, target domain.Target, opt
 		}
 		res := gwclient.NewResult()
 		if !b.builtMain {
-			ref, err := b.stateToRef(ctx, gwClient, b.targetPhaseState(mts.Final), mts.Final.Platform)
+			ref, err := b.stateToRef(ctx, gwClient, mts.Final.MainState, mts.Final.Platform)
```
vladaionescu, line 178: This change is a no-op - just made this change for clarity as it only applies to main state.

builder/builder.go
```diff
@@ -285,7 +285,11 @@ func (b *Builder) convertAndBuild(ctx context.Context, target domain.Target, opt
 					}
 				}
 			}
-			if !sts.Target.IsRemote() && !opt.NoOutput && !opt.OnlyFinalTargetImages && opt.OnlyArtifact == nil {
+			performSaveLocals := (!sts.Target.IsRemote() &&
+				!opt.NoOutput &&
+				!opt.OnlyFinalTargetImages &&
+				opt.OnlyArtifact == nil)
+			if performSaveLocals {
```
vladaionescu, line 292: This change is a no-op - just slightly easier to read.

builder/builder.go
```diff
@@ -348,9 +352,18 @@ func (b *Builder) convertAndBuild(ctx context.Context, target domain.Target, opt
...
+			if sts.RunPush.Initialized {
+				hasRunPush = true
+				break
+			}
+		}
+		if hasRunPush {
+			err = b.s.buildMainMulti(ctx, bf, onImage, onArtifact, onFinalArtifact, "--push")
+			if err != nil {
+				return nil, errors.Wrapf(err, "build push")
+			}
```
vladaionescu, line 366: This should be a no-op (I think) - we're avoiding executing run-pushes if no run push was defined anywhere.

builder/builder.go
```diff
@@ -208,7 +208,7 @@ func (b *Builder) convertAndBuild(ctx context.Context, target domain.Target, opt
 				shouldPush := opt.Push && saveImage.Push && !sts.Target.IsRemote() && saveImage.DockerTag != ""
 				shouldExport := !opt.NoOutput && opt.OnlyArtifact == nil && !(opt.OnlyFinalTargetImages && sts != mts.Final) && saveImage.DockerTag != ""
 				useCacheHint := saveImage.CacheHint && b.opt.CacheExport != ""
-				if !shouldPush && !shouldExport && !useCacheHint {
+				if (!shouldPush && !shouldExport && !useCacheHint) || b.builtMain {
```
vladaionescu, line 211: This line is the meat of the PR

dchw: This works until I add my PR, #754

Does this also fix that?

dchw (approved):
I'm on board as a release-patch to maintain current behavior

PR #758, opened 2021-02-04
Support for copying artifacts from LOCALLY target to containerized target

alexcb:
This is still very much so a draft and is not ready for review.

This requires https://github.com/earthly/buildkit/pull/26 be merged first (then we can backout my buildkit/Earthfile change)


buildkitd/Earthfile
```diff
@@ -1,6 +1,6 @@
 
 buildkitd:
-    ARG BUILDKIT_BASE_IMAGE=github.com/earthly/buildkit:earthly-main+build
+    ARG BUILDKIT_BASE_IMAGE=github.com/alexcb/buildkit:0ae61da79f8edcfa1a2a230c722de3ad6df187e5+build
```
alexcb, line 3: TODO remove this once earthly-main gets updated.

vladaionescu (approved):

PR #759, opened 2021-02-04
Improvements for Apple Silicon

vladaionescu:
* Make `./earthly` pull the right buildkit image
* `docker run` the right platform for buildkit
* Include platform in version output
* Track platform in analytics

With this, `earthly -P +test` works on arm64, including `WITH DOCKER`. However, `earthly -P --platform=linux/amd64 +test` does not work when ran on an M1 mac, because `WITH DOCKER` qemu emulation is not functional yet.

### DIFF
--- a/buildkitd/buildkitd.go
+++ b/buildkitd/buildkitd.go
@@ -224,12 +224,12 @@ func Start(ctx context.Context, image string, settings Settings, reset bool) err
 			"-p", fmt.Sprintf("127.0.0.1:%d:8373", settings.DebuggerPort))
 	}
 
-	isExp, err := isExperimentalDocker(ctx)
+	supPlat, err := supportsPlatform(ctx)
 	if err != nil {
-		fmt.Printf("Warning: Could not detect if Docker has the experimental flag enabled: %s\n", err.Error())
+		fmt.Printf("Warning: Could not detect if Docker supports the --platform flag: %s\n", err.Error())
 		// Keep going.
 	}
-	if isExp {
+	if supPlat {
 		args = append(args, platformFlag())
 	}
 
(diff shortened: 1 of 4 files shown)


alexcb (approved):

buildkitd/buildkitd.go
```diff
@@ -403,3 +408,21 @@ func isRootlessDocker(ctx context.Context) (bool, error) {
...
+	output, _ := cmd.CombinedOutput()
+	return bytes.Contains(output, []byte("Unable to find image"))
+}
+
+func platformFlag() string {
+	arch := runtime.GOARCH
+	if runtime.GOARCH == "arm" {
+		arch = "arm/v7"
+	}
+	return fmt.Sprintf("--platform=linux/%s", arch)
```
dchw, line 427: Thank you for not shelling out. Is `arm/v7` a safe assumption?

earthly
```diff
@@ -13,19 +13,21 @@ last_check_state_path=/tmp/last-earthly-prerelease-check
...
     earth_bin_path=/earthly-linux-amd64
+    bk_platform=linux/amd64
     if [ "$(uname)" == "Darwin" ]; then
         if [ "$(uname -m)" == "arm64" ]; then
             earth_bin_path=/earthly-darwin-arm64
+            bk_platform=linux/arm64
         else
             earth_bin_path=/earthly-darwin-amd64
         fi
     fi
```
dchw, line 28: [But what about Windows ARM64?](https://en.wikipedia.org/wiki/Trollface#/media/File:Trollface_non-free.png)

dchw (approved):
Looks good to me. Anything else we need to mark up on BuildKite to get CI regular on this? I can investigate

buildkitd/buildkitd.go
vladaionescu: I saw that Docker's libraries do the same - if they see arm, they assume v7. v8 shows up as arm64 so we're safe there.

vladaionescu:
> Anything else we need to mark up on BuildKite to get CI regular on this?

Not much - I was planning on playing with BuildKite today actually.

dchw:
> > Anything else we need to mark up on BuildKite to get CI regular on this?
> 
> Not much - I was planning on playing with BuildKite today actually.

Awesome. I already have `arch` tags set up for that; the Mac is tagged as `arch=M1`.

PR #761, opened 2021-02-04
Adapt buildkite scripts for M1 too

vladaionescu:


.buildkite/tests.sh
```diff
@@ -20,23 +42,19 @@ while ! docker ps; do
...
-bash -c "! ./build/$EARTH_OS/amd64/earthly --ci +test-fail"
-
-echo "Build examples1"
-./build/"$EARTH_OS"/amd64/earthly --ci -P +examples1
+bash -c "! $earthly --ci ./examples/tests/fail+test-fail"
 
-echo "Build examples2"
-./build/"$EARTH_OS"/amd64/earthly --ci -P +examples2
+echo "Build examples"
+"$earthly" --ci -P +examples
```
alexcb, line 60: I sure hope the path to the earthly binary doesn't contain any spaces; but good to know it'll still work if it did :grinning:

alexcb (approved):

PR #769, opened 2021-02-06
Initial implementation of an Earthly AST and the interpreter

vladaionescu:
Note to reviewer: This may be easier to review commit-by-commit.

* The AST can be printed to stdout using `earthly debug ast <path-to-earthfile>`.
* The new interpreter is not enabled by default. It needs to be enabled via `--enable-ast`. There is some leftover duplication as a result of this feature flagging (eg two listeners, two earthfile2llb), but this should go away relatively soon.
* I've added some GitHub actions tests for examples and +test with the new interpreter.
* I've also enabled the new feature flag for the buildkite tests, `./earthly` and `+earthly-docker` (used in some integration tests).

Other changes:
* ineffassign wasn't working - I've updated it
* Incidentally many errors are much better now as they show file name and line:column numbers too: 
  Screenshot from 2021-02-05 22-00-28


builder/builder.go
```diff
@@ -158,19 +159,37 @@ func (b *Builder) convertAndBuild(ctx context.Context, target domain.Target, opt
...
-				DockerBuilderFun:     b.MakeImageAsTarBuilderFun(),
-				CleanCollection:      b.opt.CleanCollection,
-				Platform:             opt.Platform,
-				VarCollection:        b.opt.VarCollection,
-				BuildContextProvider: b.opt.BuildContextProvider,
-				CacheImports:         b.opt.CacheImports,
-				UseInlineCache:       b.opt.UseInlineCache,
-				UseFakeDep:           b.opt.UseFakeDep,
-			})
+			if b.opt.EnableAst {
```
alexcb, line 162: It took me a while to see they only difference between the if and the else is just Earthfile2LLB2 vs Earthfile2LLB; where as the convert options are identical. Do we expect that the ConvertOpt values will ever change between the two different calls? If not is is possible to move them outside of the if/else

vladaionescu: Yeah, I should do that - I left this a bit messy.

Will fix.

ast/spec/earthfile.go
```diff
@@ -0,0 +1,77 @@
...
+type IfStatement struct {
+	Expression     []string        `json:"expression"`
+	IfBody         Block           `json:"ifBody"`
+	ElseIf         []ElseIf        `json:"elseIf"`
+	HasElse        bool            `json:"hasElse"`
+	ElseBody       *Block          `json:"elseBody,omitempty"`
+	SourceLocation *SourceLocation `json:"sourceLocation,omitempty"`
+}
+
+// ElseIf is the AST representation of an else if clause.
```
alexcb, line 55: Do we need to represent an ElseIF in the AST? I think it's good to support it in the parser, but by the time the interpreter receives it would it simplify the interpreter code if it was simple an IFBody and ElseBody, then each ElseBody would container another IfStatement nested under it.

alexcb (approved):
looks great

ast/spec/earthfile.go
vladaionescu: I was debating this in my head. It's not absolutely necessary. I figured it looks more like an AST if there is less interpreting in the listener. But perhaps that is not strictly the goal here - but rather to have some lower-level representation of the operations.

ast/spec/earthfile.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+// SourceLocation is an optional reference to the original source code location.
+type SourceLocation struct {
+	File        string `json:"file,omitempty"`
+	StartLine   int    `json:"startLine"`
+	StartColumn int    `json:"startColumn"`
+	EndLine     int    `json:"endLine"`
+	EndColumn   int    `json:"endColumn"`
+}
```
dchw, line 77: Do we want a version here; in case we want to make incompatible changes? Either per-node or whole-AST?

vladaionescu: Perhaps. I would add it only when we need it, to keep things simple for now. We don't yet store it or use it between different versions of Earthly.

builder/builder.go
vladaionescu: Fixed now - I merged the two earthfile2llb's to clean things up.

ast/spec/earthfile.go
alexcb: I was just thinking out loud here. If we keep it, does it make it easier to display clear error messages?

dchw (approved):
> Perhaps. I would add it only when we need it, to keep things simple for now. We don't yet store it or use it between different versions of Earthly. 

Fair; just thinking out loud

ast/spec/earthfile.go
vladaionescu: Maybe... it definitely maps more closely to the original code. But it's not a big difference.

It may be more significant that for us it's easier to read the AST for debug purposes, as there's less JSON nesting.

dchw:
Just playing with the implementation and what we could do with it:

```
❯ diff <(jq -S . orig.json) <(jq -S . new.json)
943a944,973
>       "name": "earthly-windows-amd64",
>       "recipe": [
>         {
>           "command": {
>             "args": [
>               "--build-arg",
>               "GOOS=darwin",
>               "--build-arg",
>               "GOARCH=arm64",
>               "--build-arg",
>               "VARIANT=",
>               "--build-arg",
>               "GO_EXTRA_LDFLAGS=",
>               "+earthly/*",
>               "./"
>             ],
>             "name": "COPY"
>           }
>         },
>         {
>           "command": {
>             "args": [
>               "./*"
>             ],
>             "name": "SAVE ARTIFACT"
>           }
>         }
>       ]
>     },
>     {
```

vladaionescu:
> Just playing with the implementation and what we could do with it:

Pretty cool! I was thinking to use a few guinea pig Earthfiles for such testing purposes. If the resulting AST drifts over time, it needs to be checked in - and so it becomes part of the code review.

ast/spec/earthfile.go
vladaionescu: I ended up removing it - we can add it back in the future if we change our minds.

Issue #772, opened 2021-02-08
Starting any containers during build fails


raphinesse:
First time user here. I was very excited when I read about earthly, but unfortunately nothing is working for me so far.

If I run `earthly +foo` using the following `Earthfile`:

```Dockerfile
FROM alpine

foo:
    RUN touch foo
```

I get this output:
```
           buildkitd | Found buildkit daemon as docker container (earthly-buildkitd)
              alpine | --> Load metadata linux/amd64
               +base | --> FROM alpine
               +base | [          ] resolve docker.io/library/alpine@sha256:08d6ca16c60fe7490c03d10dc339d9fd8ea67c6466dea8d558526b1330a85930 ... 0%                [██████████] resolve docker.io/library/alpine@sha256:08d6ca16c60fe7490c03d10dc339d9fd8ea67c6466dea8d558526b1330a85930 ... 100%
               +base | [██████████] sha256:4c0d98bf9879488e0407f897d9dd4bf758555a78e39675e72b5124ccf12c2580 ... 100%
               +base | [██████████] extracting sha256:4c0d98bf9879488e0407f897d9dd4bf758555a78e39675e72b5124ccf12c2580 ... 100%
                +foo | --> RUN touch foo
                +foo | container_linux.go:367: starting container process caused: process_linux.go:340: applying cgroup configuration for process caused: mkdir /sys/fs/cgroup/blkio/buildkit: no such file or directory
                +foo | ERROR: Command exited with non-zero code: RUN touch foo
Repeating the output of the command that caused the failure
================================ FAILURE [main] ================================
                +foo *failed* | --> RUN touch foo
                +foo *failed* | container_linux.go:367: starting container process caused: process_linux.go:340: applying cgroup configuration for process caused: mkdir /sys/fs/cgroup/blkio/buildkit: no such file or directory
                +foo *failed* | ERROR: Command exited with non-zero code: RUN touch foo
Error: exit code: 1
```
Changing the base image or the commands to be run does not seem to change anything.

Building and running the corresponding `Dockerfile` works just fine:
```Dockerfile
FROM alpine
RUN touch foo
```

### Environment

- Host OS: Ubuntu 20.04
- earthly version v0.5.0
- Docker version 19.03.11, build dd360c7

vladaionescu:
Hi @raphinesse, I'm not seeing anything obvious that may be causing this.

Can you provide the output of

```
docker logs earthly-buildkitd
```
```
docker volume inspect earthly-cache
```
```
[...]

PR #773, opened 2021-02-08
Better buildkit status reporting

vladaionescu:
* Provides `docker pull` progress update to the user when buildkit image is not available locally (user feedback: https://github.com/earthly/earthly/issues/772)
* Prints the buildkit logs when the error is "transport is closing" (save a round-trip during issue reporting)


buildkitd/buildkitd.go
```diff
@@ -200,6 +200,13 @@ func Start(ctx context.Context, image string, settings Settings, reset bool) err
 	if err != nil {
 		return err
 	}
+	// Pulling is not strictly needed, but it helps display some progress status to the user in
+	// case the image is not available locally.
+	err = MaybePull(ctx, image)
+	if err != nil {
+		fmt.Printf("Error: %v\n", err.Error())
```
dchw, line 207: Is it worth telling the user too that it still might work?

dchw (approved):

PR #774, opened 2021-02-09
Add note about WITH DOCKER and multiplatform

vladaionescu:


alexcb (approved):
I really like the "don't do this", but instead "do this" example.

PR #775, opened 2021-02-09
Refactor grammar and listener to output AST for IF

vladaionescu:
* Removes the old listener
* Refactors the grammar structure for `WITH` and `IF` for `listener` simplicity
* Creates AST for IF and IF-ELSE
* Reworked the way INDENT and DEDENT tokens work (needed in order to support IF in the base target)
* Significantly improved error messages for both lexing and parsing errors


ast/parser/EarthParser.g4
```diff
@@ -39,8 +39,34 @@ commandStmt:
...
+withCommand:
+	dockerCommand;
+
+dockerCommand: DOCKER (WS stmtWords)?;
+
+// ifStmt ---------------------------------------------------------------------
+
+ifStmt: ifClause (NL+ WS? elseIfClause)* (NL+ WS? elseClause)? NL+ WS? END;
+
+ifClause: IF WS ifExpr (NL+ WS? ifBlock)?;
```
alexcb, line 59: I'm not sure if i'm reading this correctly, but does this mean an ifBlock is optional? Does this mean one can do this:

```
IF test "$a" = "$b"
END
```
without any RUN or copy or any other statements in the code block?

alexcb: The nice thing about this, is if someone wanted to comment out some lines one could do:

```
IF test "$a" = "$b"
#    RUN something a user wants to temporarily comment out during development
END
```

Not sure exactly how I feel about this if that's the intended case; seems like it could work.

vladaionescu: Yes, exactly. I was wondering if we should allow this - I decided to allow given that all programming languages allow you to do this.

alexcb (approved):
looks great.

PR #776, opened 2021-02-10
Display hint for platform related errors

alexcb:
Here's an example of the error-fix hint:
```
$ ./build/linux/amd64/earthly -i --platform arm64 github.com/earthly/hello-world:main+hello
           buildkitd | ...Done
      busybox:1.32.0 | --> Load metadata linux/arm64
g/e/hello-world:main+base | platform=linux/arm64
g/e/hello-world:main+base | --> FROM busybox:1.32.0
g/e/hello-world:main+base | [██████████] resolve docker.io/library/busybox:1.32.0@sha256:bde48e1751173b709090c2539fdf12d6ba64e88ec7a4301591227ce925f3c678 ... 100%
g/e/hello-world:main+hello | platform=linux/arm64
g/e/hello-world:main+hello | --> RUN echo 'Hello, world!'
g/e/hello-world:main+hello | .buildkit_qemu_emulator: /usr/bin/earth_debugger: Invalid ELF image for this architecture
g/e/hello-world:main+hello | ERROR: Command exited with non-zero code: RUN echo 'Hello, world!'
================================ FAILURE [main] ================================
g/e/hello-world:main+hello *failed* | platform=linux/arm64
g/e/hello-world:main+hello *failed* | --> RUN echo 'Hello, world!'
g/e/hello-world:main+hello *failed* | .buildkit_qemu_emulator: /usr/bin/earth_debugger: Invalid ELF image for this architecture
g/e/hello-world:main+hello *failed* | ERROR: Command exited with non-zero code: RUN echo 'Hello, world!'
Error: build target: build main: bkClient.Build: failed to solve: rpc error: code = Unknown desc = executor failed running [/dev/.buildkit_qemu_emulator /bin/sh -c  /usr/bin/earth_debugger /bin/sh -c 'echo '"'"'Hello, world!'"'"'']: exit code: 255

[whitespace added to this PR comment to show the new addition to the warning message below]

Are you using --platform to target a different architecture? You may have to manually install Qemu.
For more information see https://docs.earthly.dev/guides/multi-platform
```

### DIFF
--- a/builder/solver_monitor.go
+++ b/builder/solver_monitor.go
@@ -234,7 +234,7 @@ func newSolverMonitor(console conslogging.ConsoleLogger, verbose bool) *solverMo
 	}
 }
 
-func (sm *solverMonitor) monitorProgress(ctx context.Context, ch chan *client.SolveStatus, phaseText string) error {
+func (sm *solverMonitor) monitorProgress(ctx context.Context, ch chan *client.SolveStatus, phaseText string) (string, error) {
 	sm.mu.Lock()
 	sm.ongoing = true
 	sm.mu.Unlock()
@@ -311,12 +311,14 @@ Loop:
 				}
 				err := sm.printOutput(vm, logLine.Data)
 				if err != nil {
-					return err
+					return "", err
 				}
 			}
 		}
 	}
+	failedVertexOutput := ""
 	if errVertex != nil {
+		failedVertexOutput = string(errVertex.tailOutput.Bytes())
 		sm.reprintFailure(errVertex, phaseText)
 	}
 	sm.mu.Lock()
@@ -328,7 +330,7 @@ Loop:
 	sm.ongoing = false
 	sm.mu.Unlock()
 	sm.PrintTiming()
-	return nil
+	return failedVertexOutput, nil
 }
 
 func (sm *solverMonitor) printOutput(vm *vertexMonitor, data []byte) error {
--- a/cmd/earthly/main.go
+++ b/cmd/earthly/main.go
@@ -1092,11 +1092,17 @@ func (app *earthlyApp) insertZSHCompleteEntry() error {
 }
 
 func (app *earthlyApp) run(ctx context.Context, args []string) int {
-	err := app.cliApp.RunContext(ctx, args)
-
 	rpcRegex := regexp.MustCompile(`(?U)rpc error: code = .+ desc = `)
+	err := app.cliApp.RunContext(ctx, args)
 	if err != nil {
 		ie, isInterpereterError := earthfile2llb.GetInterpreterError(err)
+
+		var failedOutput string
+		var buildErr *builder.BuildError
+		if errors.As(err, &buildErr) {
+			failedOutput = buildErr.VertexLog()
+		}
+
 		if strings.Contains(err.Error(), "security.insecure is not allowed") {
 			app.console.Warnf("Error: --allow-privileged (-P) flag is required\n")
 		} else if strings.Contains(err.Error(), "failed to fetch remote") {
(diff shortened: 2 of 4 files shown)


builder/solver_monitor.go
```diff
@@ -317,6 +317,9 @@ Loop:
 		}
 	}
 	if errVertex != nil {
+		v := ctx.Value("failed_output")
+		ptr, _ := v.(*string)
+		*ptr = string(errVertex.tailOutput.Bytes())
```
vladaionescu, line 322: This is too hacky.... :-)

Here's an alternative: Save it as a field in solver monitor and have a getter fun for it in builder. On failure, `buildAction` in `main.go` saves the output in a field as part of `app` (eg `lastFailureOutput`), which can be accessed from `run`.

cmd/earthly/main.go
```diff
@@ -1105,6 +1106,11 @@ func (app *earthlyApp) run(ctx context.Context, args []string) int {
 				"Check your git auth settings.\n" +
 					"Did you ssh-add today? Need to configure ~/.earthly/config.yml?\n" +
 					"For more information see https://docs.earthly.dev/guides/auth\n")
+		} else if strings.Contains(failedOutput, "Invalid ELF image for this architecture") {
+			app.console.Warnf("Error: %v\n", err)
+			app.console.Printf(
+				"Are you using --platform to target a different architecture? You may have to manually install Qemu.\n" +
```
vladaionescu, line 1112:
```suggestion
				"Are you using --platform to target a different architecture? You may have to manually install QEMU.\n" +
```

builder/solver_monitor.go
alexcb: haha, why thank you :)

I was considering adding it to the solvermonitor; do you think it makes sense to then propigate it all the way back to where `app.cliApp.RunContext` is called? We're going to need to reference the builder object which gets created under  `actionBuild`. would it make sense to store an instance of the builder object under the `earthlyApp` struct?

Another somewhat (maybe less) hacky idea would be to have the builder return a `BuilderError` type, which we could create which would include both the underlying error as well as another field with the errVertex output.

I'm definitely up for suggestions.

vladaionescu: I was thinking that after an err is returned from builder in `actionBuild`, we can retrieve the output through a getter in builder that has access to `sm`. Then I was thinking of putting that output in a var in `app` that we can access from `run`.

A `BuilderError` that embeds the output too could be very elegant. One thing to be careful though is that sometimes errors are passed to buildkit and then buildkit gives it back to us - but it's in string form - the type of error is lost (it's wrapped in rpc stuff - we only get the string contents). So the error needs to be created in the right place in order for it to get to the caller without going through rpc.

alexcb: I had to do a combination of the two due to buildkit bubbling up the original error as a string; I created a `BuildError` (so I don't get linting issues with Builder.Builder* stuttering.

builder/builder.go
```diff
@@ -366,7 +366,7 @@ func (b *Builder) convertAndBuild(ctx context.Context, target domain.Target, opt
 		if hasRunPush {
 			err = b.s.buildMainMulti(ctx, bf, onImage, onArtifact, onFinalArtifact, "--push")
 			if err != nil {
-				return nil, errors.Wrapf(err, "build push")
+				return nil, errors.Wrapf(NewBuildError(err, b.s.vertexFailureOutput), "build push")
```
vladaionescu, line 369: Could `buildMainMulti` return this error itself? Maybe a defer could do a wrapping if err != nil.

vladaionescu (approved):
Looks good!

builder/builder.go
alexcb: Good point, I moved it there instead.

vladaionescu (approved):

PR #778, opened 2021-02-10
Add command to set config file values via CLI

dchw:
Still needs some integration tests, and a couple fixes; but is ready for some eyes and questions..

Sample inputs/results:

```
earthly config --dry-run global.run_path my/path

global:
    run_path: my/path

```

```
earthly config --dry-run  global.buildkit_additional_args "[userns, --root]"

global:
    buildkit_additional_args: [userns, --root]

```

```
earthly config --dry-run git."git.sr.ht".pattern my-pattern

git:
    git.sr.ht:
        pattern: my-pattern

```

```
earthly config --dry-run git {git.sr.ht: {pattern: my-pattern}}

git:
    git.sr.ht:
        pattern: my-pattern
```


cmd/earthly/main.go
```diff
@@ -805,6 +806,19 @@ func newEarthlyApp(ctx context.Context, console conslogging.ConsoleLogger) *eart
...
+			Hidden: true, // Experimental.
+			Action: app.actionConfig,
+			Flags: []cli.Flag{
+				&cli.BoolFlag{
+					Name:        "dry-run",
+					Usage:       "Print the changed config file to the consile instead of writing it out",
+					Destination: &app.configDryRun,
+				},
+			},
+		},
```
dchw, line 821: Maybe we need a command to just print your config? But that command is also named `cat ~/.earthly/config.yml`...

config/config.go
```diff
@@ -170,3 +173,192 @@ func defaultRunPath() string {
...
+		// Empty file, or a simple comment results in a null document.
+		// Not handled well, so manufacture somewhat acceptable document
+		fullDoc := string(config) + "\n---"
+		yaml.Unmarshal([]byte(fullDoc), base)
+		base.Content = []*yaml.Node{{Kind: yaml.MappingNode}}
+	}
+
+	pathParts := splitPath(path)
+
+	err := validatePath(reflect.TypeOf(Config{}), pathParts)
```
dchw, line 193: Uses the `yaml` tags on our existing structs to validate. We can just add config to the structs like normal and it should magically also be writeable.

config/config.go
```diff
@@ -170,3 +173,192 @@ func defaultRunPath() string {
...
+	if err != nil {
+		return []byte{}, errors.Wrap(err, "path is not valid")
+	}
+
+	yamlValue, err := valueToYaml(value)
```
dchw, line 198: We parse it as YAML because we needed to support sequences for the `buildkit_additional_args` feature. Turns out this also ends up letting us support complex objects on the other end as inline YAML. I think this is neat.

config/config.go
```diff
@@ -170,3 +173,192 @@ func defaultRunPath() string {
...
+	fixStyling = func(node *yaml.Node) {
+		node.Style = 0
+
+		for _, n := range node.Content {
+			fixStyling(n)
+		}
+	}
+	fixStyling(valueNode)
+
+	return valueNode.Content[0], nil
```
dchw, line 267: The value is wrapped in a `DocumentNode`; and we dont need that node. We just want the juicy _actual_ values. If it parsed correctly, we will always have this value. I expect an error from unmarshal before this fails.

config/config.go
```diff
@@ -170,3 +173,192 @@ func defaultRunPath() string {
...
+			path = setYamlValue(c, path, value)
+		}
+
+	case yaml.MappingNode:
+		for i := 0; i < len(node.Content); i += 2 {
+			// Keys/Values are inline. Count by twos to get it right.
+			key := node.Content[i]
+			val := node.Content[i+1]
+
+			if len(path) > 0 && key.Value == path[0] {
```
dchw, line 323: Having consumed the path, either by construction or search is the base case for this recursion

config/config.go
```diff
@@ -170,3 +173,192 @@ func defaultRunPath() string {
...
+				if len(path) == 0 {
+					node.Content[i+1] = value
+					return []string{}
+				}
+
+				path = setYamlValue(val, path, value)
+			}
+		}
+
+	default: // Sequence, Scalar nodes get skipped
```
dchw, line 335: As well as parse errors; I think I cleaned up all empty/zero documents getting in here?

config/config.go
```diff
@@ -170,3 +173,192 @@ func defaultRunPath() string {
...
+	} else if err != nil {
+		return []byte{}, errors.Wrapf(err, "failed to read from %s", configPath)
+	}
+
+	return yamlData, nil
+}
+
+// WriteConfigFile writes the config file to disk with preset permission 0644
+func WriteConfigFile(configPath string, data []byte) error {
+	return ioutil.WriteFile(configPath, data, 0644)
```
dchw, line 363: I just pulled these perms out of the air. If these are bad let me know

earthfile2llb/withdockerrun.go
```diff
@@ -21,7 +21,7 @@ import (
 	gwclient "github.com/moby/buildkit/frontend/gateway/client"
 	specs "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
-	"gopkg.in/yaml.v2"
+	"gopkg.in/yaml.v3"
```
dchw, line 24: The tests still pass... YOLO

cmd/earthly/main.go
vladaionescu: I think it would be nice nevertheless - similar to how you can inspect git config. But probably not as high in priority as setting values.

vladaionescu (approved):
Seems good to me. Some integration tests would be great - but that can be in separate PR since this is a hidden feature.

scripts/tests/self-hosted-private-repo.sh
```diff
@@ -73,13 +73,7 @@ cd ~
...
 
 # test that earthly has access to it
-cat <<EOF >> ~/.earthly/config.yml
-git:
-  myserver:
-    pattern: 'myserver/([^/]+)'
-    substitute: 'ssh://root@$ip:2222/root/my/really/weird/path/\$1.git'
-    auth: ssh
-EOF
+$earthly config git {myserver: {pattern: 'myserver/([^/]+)', substitute: 'ssh://root@$ip:2222/root/my/really/weird/path/\$1.git', auth: ssh}}
```
vladaionescu, line 76:
```suggestion
"$earthly" config git {myserver: {pattern: 'myserver/([^/]+)', substitute: 'ssh://root@$ip:2222/root/my/really/weird/path/\$1.git', auth: ssh}}
```

PR #780, opened 2021-02-11
Add some json-based AST tests

vladaionescu:
This takes all the tests we have in `./examples/tests/*.earth` plus the `WITH DOCKER` ones and checks that the AST json is the same as the one we had before. It should guard against uninteded AST drift.

Side note: a `FOR` would be great here.


alexcb (approved):
some sort of for or UDF would make that nicer, but the test coverage looks great.

PR #783, opened 2021-02-17
test earthly exports work

alexcb:
This test is for validating the exporting from earthly to docker works.

It's written in a way that a user can run `./scripts/tests/export.sh` locally to run the test, or allow CI to invoke the script.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -257,6 +257,27 @@ jobs:
         run: docker logs earthly-buildkitd
         if: ${{ failure() }}
 
+  export-test:
+    runs-on: ubuntu-latest
+    env:
+      FORCE_COLOR: 1
+      EARTHLY_TOKEN: "${{ secrets.EARTHLY_TOKEN }}"
+      EARTHLY_INSTALL_ID: "earthly-githubactions"
+    steps:
+      - uses: actions/checkout@v2
+      - name: Download released earthly
+        run: "sudo /bin/sh -c 'wget https://github.com/earthly/earthly/releases/latest/download/earthly-linux-amd64 -O /usr/local/bin/earthly && chmod +x /usr/local/bin/earthly'"
+      - name: Docker Login (non fork only)
+        run: docker login --username "${{ secrets.DOCKERHUB_USERNAME }}" --password "${{ secrets.DOCKERHUB_TOKEN }}"
+        if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name == github.repository
+      - name: Build latest earthly using released earthly
+        run: earthly --use-inline-cache +for-linux
+      - name: run export tests
+        run: env earthly=./build/linux/amd64/earthly scripts/tests/export.sh
+      - name: Buildkit logs (runs on failure)
+        run: docker logs earthly-buildkitd
+        if: ${{ failure() }}
+
   all-buildkitd:
     name: +all-buildkitd
     runs-on: ubuntu-latest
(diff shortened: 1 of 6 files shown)


scripts/tests/export.sh
```diff
@@ -0,0 +1,70 @@
...
+earthly=${earthly:=earthly}
+earthly=$(realpath $earthly)
+echo "running tests with $earthly"
+
+# prevent the self-update of earthly from running (this ensures no bogus data is printed to stdout,
+# which would mess with the secrets data being fetched)
+date +%s > /tmp/last-earthly-prerelease-check
+
+# ensure earthly login works (and print out who gets logged in)
+$earthly account login
```
vladaionescu, line 13: I think this is needed for shellcheck

```suggestion
"$earthly" account login
```

scripts/tests/export.sh
```diff
@@ -0,0 +1,70 @@
...
+test3:
+    FROM busybox:latest
+    RUN true # hack needed to make this work
+    SAVE IMAGE earthly-export-test-3:test
+EOF
+
+$earthly prune --reset
+$earthly +test3
+
+docker run --rm earthly-export-test-3:test | grep "running default cmd"
```
vladaionescu, line 70: Should we add this script to the shellcheck list?

.github/workflows/ci.yml
```diff
@@ -257,6 +257,29 @@ jobs:
         run: docker logs earthly-buildkitd
         if: ${{ failure() }}
 
+  export-test:
+    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name == github.repository
```
vladaionescu, line 261: This test seems excluded for fork PRs - is this necessary?

scripts/tests/export.sh
alexcb: that's a good point, any concerns about me adding `./scripts/tests/*.sh` to the Earthfile so we don't forget to add any future scripts?

scripts/tests/export.sh
alexcb: it's actually not complaining about this; but it definitely breaks if I'm working out of a directory that has a space in it; i'll fix this in all of our scripts.

.github/workflows/ci.yml
alexcb: i'll remove it; just a copy and past from one of our excluded tests

scripts/tests/export.sh
vladaionescu: SGTM

vladaionescu (approved):

.github/workflows/ci.yml
```diff
@@ -257,6 +257,27 @@ jobs:
...
+      - uses: actions/checkout@v2
+      - name: Download released earthly
+        run: "sudo /bin/sh -c 'wget https://github.com/earthly/earthly/releases/latest/download/earthly-linux-amd64 -O /usr/local/bin/earthly && chmod +x /usr/local/bin/earthly'"
+      - name: Docker Login (non fork only)
+        run: docker login --username "${{ secrets.DOCKERHUB_USERNAME }}" --password "${{ secrets.DOCKERHUB_TOKEN }}"
+        if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name == github.repository
+      - name: Build latest earthly using released earthly
+        run: earthly --use-inline-cache +for-linux
+      - name: run export tests
+        run: env earthly=./build/linux/amd64/earthly scripts/tests/export.sh
```
vladaionescu, line 276: BTW - relatively small tests (<5m to execute) could also go under Misc tests.

vladaionescu (approved):

Issue #789, opened 2021-02-18
Then linux-arm64 release binary is actually the darwin-arm64 binary


bhansconnect:
The title pretty much describes the issue.

On the [release page](https://github.com/earthly/earthly/releases/tag/v0.5.1), the darwin-arm64 and the linux-arm64 binaries are the same. They both look to be compiled for Mac and not for linux.

vladaionescu:
Just checked - this is confirmed.

vladaionescu:
I removed `earthly-linux-arm64` from the release for the time being.

vladaionescu:
This is now fixed in the latest release https://github.com/earthly/earthly/releases/tag/v0.5.2

PR #790, opened 2021-02-18
Fix for build arg issue when not overriding default value

vladaionescu:
This was causing our earthly-linux-arm64 binary to be the same as earthly-darwin-arm64.

Fixes https://github.com/earthly/earthly/issues/789

### DIFF
--- a/earthfile2llb/converter.go
+++ b/earthfile2llb/converter.go
@@ -622,7 +622,7 @@ func (c *Converter) Arg(ctx context.Context, argKey string, defaultArgValue stri
 		return err
 	}
 	c.mts.Final.TargetInput = c.mts.Final.TargetInput.WithBuildArgInput(
-		effective.BuildArgInput(argKey, effective.ConstantValue()))
+		effective.BuildArgInput(argKey, defaultArgValue))
 	return nil
 }
 
--- a/examples/tests/Earthfile
+++ b/examples/tests/Earthfile
@@ -58,6 +58,7 @@ ga:
     BUILD +run-no-cache
     BUILD +save-artifact-after-push
     BUILD +push-build
+    BUILD +build-arg-repeat
     BUILD ./autocompletion+test-all
     BUILD ./with-docker+all
     BUILD ./with-docker-compose+all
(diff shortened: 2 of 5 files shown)


earthfile2llb/converter.go
```diff
@@ -622,7 +622,7 @@ func (c *Converter) Arg(ctx context.Context, argKey string, defaultArgValue stri
 		return err
 	}
 	c.mts.Final.TargetInput = c.mts.Final.TargetInput.WithBuildArgInput(
-		effective.BuildArgInput(argKey, effective.ConstantValue()))
+		effective.BuildArgInput(argKey, defaultArgValue))
```
vladaionescu, line 625: This is the bug fix itself.

examples/tests/Earthfile
```diff
@@ -564,4 +565,26 @@ push-build:
...
+    RUN --privileged \
+        --entrypoint \
+        --mount=type=tmpfs,target=/tmp/earthly \
+        -- +build-all-2
+    RUN test -f ./output/out-other-1
+    RUN test -f ./output/out-default-1
+    RUN cat ./output/out-other-1 | grep "A=other"
+    RUN cat ./output/out-other-1 | grep "B=1"
+    RUN cat ./output/out-default-1 | grep "A=default"
+    RUN cat ./output/out-default-1 | grep "B=1"
```
vladaionescu, line 590: The test to validate the fix (fails before the fix, passes after).

release/Earthfile
```diff
@@ -52,6 +53,14 @@ release-github:
...
         test -f ./release/earthly-linux-arm7 && \
         test -f ./release/earthly-linux-arm64
+    RUN file ./release/earthly-linux-amd64 | grep "x86-64"
+    RUN file ./release/earthly-linux-amd64 | grep "ELF 64-bit"
+    RUN file ./release/earthly-darwin-amd64 | grep "Mach-O 64-bit x86_64"
+    RUN file ./release/earthly-darwin-arm64 | grep "Mach-O 64-bit arm64"
+    RUN file ./release/earthly-linux-arm7 | grep "ARM, EABI5 version 1"
+    RUN file ./release/earthly-linux-arm7 | grep "ELF 32-bit"
+    RUN file ./release/earthly-linux-arm64 | grep "aarch64"
+    RUN file ./release/earthly-linux-arm64 | grep "ELF 64-bit"
```
vladaionescu, line 63: Double-check during the release that we got it right.

alexcb: I like this check :+1:

alexcb (approved):

Issue #791, opened 2021-02-19
Docker-out-of-Docker Support?


naramore:
If I install the docker client in a container, and pass the host docker socket to it at run-time, then the docker container has access to the host's docker engine. But when I attempt to run earthly from within the docker container, it always fails because it can't mount against the `RunPath`.

After investigating the code a bit, I found this:
https://github.com/earthly/earthly/blob/70b632433ae662848416e00386c8491382db6c0a/config/config.go#L75

and the `RunPath` is not configurable.

So the issue is that earthly is executing from within the container which is where it 'looks up' it's `RunPath`, but when it attempts to mount against that `RunPath`, unless the container and host have that same path, it fails.

Would it be possible to make the RunPath optionally configurable? Or perhaps some other configuration option to better support running earthly with the [Docker-out-of-Docker](https://tdongsi.github.io/blog/2017/04/23/docker-out-of-docker/) setup?

This would make running earthly in a [VSCode Remote:Container](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers) setup much nicer.

Thanks for this awesome project!

vladaionescu:
Hi @naramore - it seems that `RunPath` is actually not really used so we will remove it in the next release. It used to be necessary to make the `earthly -i` mode working. It now works through a different mechanism.

In any case, until the next release, I *think* that you can configure the path via the earthly config. In v0.5.2 you can do `earthly config global.run_path /some/other/path`. Really all that is necessary is for that path to exist on the host that the Docker daemon is running.

I should also say that running Earthly's buildkit as Docker-out-of-Docker is not really supported, but also it should just work. Let us know if you run into other issues!

PR #792, opened 2021-02-19
Get rid of old RunPath config and mount

vladaionescu:
This is no longer used AFAICT.

Fixes https://github.com/earthly/earthly/issues/791


alexcb (approved):
looks good, I can't find any reference to `/run/earthly` elsewhere.

Issue #779, opened 2021-02-10
Proposal: Flow control structures (IF and FOR)


vladaionescu:
# IF / IF ELSE

## Use-cases

* Ability to switch between FROM <something> and LOCALLY based on a build arg. This allows the user to reuse CI setup (FROM alpine) in their development workflow (LOCALLY). A common set of instructions are used (eg RUN go download, RUN go build etc), but the execution can be switched between container and local via an IF.
* With the new UDC proposal, it it harder to rely on bash if’s alone. The decision of whether to run a UDC or not cannot be controlled via a bash if. In this context first-class support for IF is necessary.
* This kind of optional use of secrets is clunky to use: https://github.com/earthly/earthly/blob/main/examples/tests/Earthfile#L8-L15. In order to turn the use of a secret on, the entire path of the secret needs to be specified as part of a build arg (like this https://github.com/earthly/earthly/blob/main/.github/workflows/ci.yml#L72). It would be much nicer if we could use an IF with a much simpler build arg like DO_DOCKER_LOGIN=true.

## Proposal

```
IF <run expression>
	RUN ...
	COPY ...
	EXPOSE ...
END
```

```
IF <run expression>
	RUN ...
	COPY ...
ELSE
	COPY ...
	RUN ...
END
```

```
IF <run expression>
	RUN ...
	COPY ...
ELSE IF <run expression>
	COPY ...
	RUN ...
END
```

## Example 1 (optional docker login in our integration tests)

```
ARG DO_DOCKER_LOGIN=false
IF $DO_DOCKER_LOGIN
  ARG DOCKERHUB_USER_SECRET=+secrets/earthly-technologies/dockerhub/user
  ARG DOCKERHUB_TOKEN_SECRET=+secrets/earthly-technologies/dockerhub/token
  RUN --secret USERNAME=$DOCKERHUB_USER_SECRET \
    --secret TOKEN=$DOCKERHUB_TOKEN_SECRET \
    docker login --username="$USERNAME" --password="$TOKEN"
END
```

## Example 2 (ability to run build locally)

```
ARG LOCAL=false
IF $LOCAL
	LOCALLY
ELSE
	FROM +code
END
RUN go build ...
```

## Example 3 (use bash comparisons)

```
FROM +something
IF \
	[ "$var" = "abc" ] && \
	[ -f ./some-file ] && \
	curl ...| \
	grep ....
	FROM +something-else
END
```

## Notes
* Don’t interpret true / false. Just run and look at exit code.
[...]

PR #793, opened 2021-02-19
Support for IF

vladaionescu:
Re https://github.com/earthly/earthly/issues/779


earthfile2llb/converter.go
```diff
@@ -352,6 +353,60 @@ func (c *Converter) RunLocal(ctx context.Context, args []string, pushFlag bool)
...
+	var opts []llb.RunOption
+	mountRunOpts, err := parseMounts(mounts, c.mts.Final.Target, c.mts.Final.TargetInput, c.cacheContext)
+	if err != nil {
+		return 0, errors.Wrap(err, "parse mounts")
+	}
+	opts = append(opts, mountRunOpts...)
+	if privileged {
+		opts = append(opts, llb.Security(llb.SecurityModeInsecure))
+	}
+	runStr := fmt.Sprintf(
```
alexcb, line 379: How will this work if we're in a LOCALLY target? the RunLocal doesn't use internalRun.

examples/tests/if.earth
```diff
@@ -0,0 +1,131 @@
+FROM alpine:3.13
```
alexcb, line 1: I think it would be good to also try this test with LOCALLY targets.

alexcb (approved):
Looks good, but I think we'll have to do some additional work to support this in LOCALLY targets.

examples/tests/if.earth
vladaionescu: True... I'll need to do that in a future PR.

vladaionescu: https://github.com/earthly/earthly/pull/794

Issue #796, opened 2021-02-21
SSL/TLS trust issues: unable to add custom certificates or disable verification


yonkeltron:
Hi, I'm working in a corporate environment where I need to specify the certificates which should be used for verification of remote connections. Unfortunately, right now I can't pull from internal registries since I get the following error:

```
...snip...
     r/r/ubi8:latest | --> Load metadata linux/amd64
     r/r/ubi8:latest | WARN: (Load metadata linux/amd64) failed to do request: Head https://REGISTRY/PATH: x509: certificate signed by unknown authority
Error: failed to do request: Head https://REGISTRY/PATH: x509: certificate signed by unknown authority
```

Since the registry is contacted every time regardless of local image presence (outlined further here in #345), it appears that I have a blocker to adopting Earthly.

How can I provide more information to troubleshooting this? I would be happy to try some things out, as well. I really appreciate all of your time and efforts.

Thanks,
+Jonathan

agbell:
Hi @yonkeltron,
One thing you could try to get around this issue, for now, is to extend the earthly/buildkit image to add in your certificate. There are [instructions for this ](https://blog.earthly.dev/mitmproxy/) here under "Extending an Alpine Image to Add a Certificate Authority". After creating that image, start it up and use the `--buildkit-host=` flag on earthly to specify its usage.  

BuildKit also has config options for this found [here](https://github.com/moby/buildkit/blob/master/docs/buildkitd.toml.md) and you could extend the image or possibly volume mount in a version of /etc/buildkitd.toml that sets things as insecure.

I think we need to provide a better way to expose this though.

vladaionescu:
We should implement proper support for this. This could be an Earthly config item. Earthly can then automatically configure buildkit and also possibly mount any necessary cert files into the buildkit container.

vladaionescu:
Hi @yonkeltron - the next release of Earthly will have support for this via the configuration shown in https://github.com/earthly/earthly/pull/797. Let me know if this works for you once you give it a try.

> Usage example:
> 
> ```yaml
> global:
>   buildkit_additional_args: ["-v", "<absolute-path-to-ca-file>:/etc/config/add.ca"]
>   buildkit_additional_config: |
>     [registry."<registry-hostname>"]
>       ca=["/etc/config/add.ca"]
> ```

Note that you will also be able to use an insecure registry

```yaml
global:
  buildkit_additional_config: |
    [registry."<registry-hostname>"]
      http = true
      insecure = true
```

[...]

PR #797, opened 2021-02-22
Add ability to specify additional buildkit config and also to mount certificates into buildkit

vladaionescu:
Re https://github.com/earthly/earthly/issues/796

This adds the ability to use a private registry with self-signed certificates via a combination of the two config items `global.buildkit_additional_args` and `global.buildkit_additional_config`.

Usage example:

```yaml
global:
  buildkit_additional_args: ["-v", "<absolute-path-to-ca-file>:/etc/config/add.ca"]
  buildkit_additional_config: |
    [registry."<registry-hostname>"]
      ca=["/etc/config/add.ca"]
```

Replace `<absolute-path-to-ca-file>` and `<registry-hostname>` with your relevant settings.


alexcb (approved):

dchw (approved):
LGTM.

I understand the mount, any reason to keep the registry settings out of the rest of our custom git registry config?

vladaionescu:
This is actually for docker registries not git.

Planning to make this a bit more natively supported in the future - but this should be a pretty good start. It also allows for possibly other buildkit configuration to be specified, in case the user wants to add in a quick hack that requires buildkit config changes.

dchw:
> This is actually for docker registries not git.

Well I misread the original issue; my bad 😬. As I said I am good here; we may also end up needing this kind of thing there eventually though...

PR #799, opened 2021-02-22
Docs for self-signed certs in registries

vladaionescu:
Re https://github.com/earthly/earthly/issues/796


dchw (approved):

dchw:
Looks like we can also mount keypairs if needed too and feels clear enough.

PR #801, opened 2021-02-22
Initial implementation for strict flag

dchw:
Small prereq that can standalone as I add in support for `RUN --interactive`

Still needs an integration test though

### DIFF
--- a/earthfile2llb/converter.go
+++ b/earthfile2llb/converter.go
@@ -271,6 +271,10 @@ func (c *Converter) FromDockerfile(ctx context.Context, contextPath string, dfPa
 
 // Locally applies the earthly Locally command.
 func (c *Converter) Locally(ctx context.Context, platform *specs.Platform) error {
+	if !c.opt.AllowLocally {
+		return errors.New("LOCALLY cannot be used when --strict is specified")
+	}
+
 	return c.fromClassical(ctx, "scratch", platform, true)
 }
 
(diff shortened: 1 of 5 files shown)


earthfile2llb/earthfile2llb.go
```diff
@@ -53,6 +53,8 @@ type ConvertOpt struct {
 	UseInlineCache bool
 	// UseFakeDep is an internal feature flag for fake dep.
 	UseFakeDep bool
+	// AllowLocally is an internal fature flag for controlling if LOCALLY directives can be used.
```
vladaionescu, line 56:
```suggestion
	// AllowLocally is an internal feature flag for controlling if LOCALLY directives can be used.
```

earthfile2llb/converter.go
```diff
@@ -271,6 +271,10 @@ func (c *Converter) FromDockerfile(ctx context.Context, contextPath string, dfPa
 
 // Locally applies the earthly Locally command.
 func (c *Converter) Locally(ctx context.Context, platform *specs.Platform) error {
+	if !c.opt.AllowLocally {
+		return errors.New("LOCALLY cannot be used when --strict is specified")
+	}
+
```
dchw, line 277: This is in converter, because it (to me) helps interpreter stay clean. Interpreter is for language, Converter is where CLI options + interpreted language options are combined, and so also is where the relevant error lives.

vladaionescu (approved):
:+1:

PR #804, opened 2021-02-24
Correctly exclude .tmp-earthly-out

vladaionescu:
The implicit exclude of `.tmp-earthly-out` was not working because the temp dir contained some numbers at the end and the exclude pattern was simply `.tmp-earthly-out/`. With this PR, the temp dirs are now stored under `./.tmp-earthly-out/tmpXXXXXX` instead of `./.tmp-earthly-outXXXXXX`.

The wrapping `.tmp-earthly-out` dir is removed only if it's empty.

It may be possible in the future to also remove any old leftover temp dirs within `.tmp-earthly-out` using some file lock-based logic to detect whether the temp dir is still being used by some process (a parallel earthly run).

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -655,11 +655,11 @@ func (b *Builder) tempEarthlyOutDir() (string, error) {
 	if err != nil {
 		return "", errors.Wrap(err, "mk temp dir for artifacts")
 	}
-	// Note that because this removes the parent dir, it will cause the cleanup of any
-	// previous run tmp dir(s) as well. This is intentional so that there is GCing of any
-	// leftovers from ctrl+C'd runs.
 	b.opt.CleanCollection.Add(func() error {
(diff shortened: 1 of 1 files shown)


builder/builder.go
```diff
@@ -645,3 +644,22 @@ func (b *Builder) saveArtifactLocally(ctx context.Context, artifact domain.Artif
 	}
 	return nil
 }
+
+func (b *Builder) tempEarthlyOutDir() (string, error) {
+	tmpParentDir := filepath.Join(".", ".tmp-earthly-out")
```
alexcb, line 649: Is the filepath.Join needed, or can it just be this:
```suggestion
	tmpParentDir := ".tmp-earthly-out"
```

vladaionescu: You're right

alexcb (approved):

PR #805, opened 2021-02-24
Copy to localhost during solve

alexcb:
implements copying artifacts from containerized targets to `LOCALLY` targets. This functionality is different from `SAVE ARTIFACT AS LOCAL`, in that it is performed during the build rather than at the end of the build during the export stage.

This depends on https://github.com/earthly/buildkit/pull/30

### DIFF
--- a/earthfile2llb/interpreter.go
+++ b/earthfile2llb/interpreter.go
@@ -268,6 +268,10 @@ func (i *Interpreter) handleRun(ctx context.Context, cmd spec.Command) error {
 			return Errorf(cmd.SourceLocation, "secrets need to be implemented for the LOCALLY directive")
 		}
 
+		if i.withDocker != nil {
+			return fmt.Errorf("WITH docker not currently supported under a LOCALLY target")
+		}
+
 		err = i.converter.RunLocal(ctx, fs.Args(), *pushFlag)
 		if err != nil {
 			return WrapError(err, cmd.SourceLocation, "apply RUN")
--- a/earthfile2llb/interpreter.go
+++ b/earthfile2llb/interpreter.go
@@ -269,7 +269,7 @@ func (i *Interpreter) handleRun(ctx context.Context, cmd spec.Command) error {
 		}
 
 		if i.withDocker != nil {
-			return fmt.Errorf("WITH docker not currently supported under a LOCALLY target")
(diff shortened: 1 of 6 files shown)


earthfile2llb/interpreter.go
```diff
@@ -268,11 +268,15 @@ func (i *Interpreter) handleRun(ctx context.Context, cmd spec.Command) error {
 			return Errorf(cmd.SourceLocation, "secrets need to be implemented for the LOCALLY directive")
 		}
 
-		err = i.converter.RunLocal(ctx, fs.Args(), *pushFlag)
-		if err != nil {
-			return WrapError(err, cmd.SourceLocation, "apply RUN")
+		if i.withDocker == nil {
```
vladaionescu, line 271: Nit: It's easier to read if this condition is reversed - the more rare cases being nested deeper.

earthfile2llb/interpreter.go
```diff
@@ -416,9 +420,16 @@ func (i *Interpreter) handleCopy(ctx context.Context, cmd spec.Command) error {
...
+					return WrapError(err, cmd.SourceLocation, "copy artifact locally")
+				}
+			} else {
+				err = i.converter.CopyArtifact(ctx, src, dest, platform, buildArgs.Args, *isDirCopy, *keepTs, *keepOwn, *chown, *ifExists)
+				if err != nil {
+					return WrapError(err, cmd.SourceLocation, "copy artifact")
+				}
 			}
 		}
 	} else {
```
vladaionescu, line 435: This else case should be handled too somehow. Maybe not in this PR, but at the very least we should return an error when attempting

```
LOCALLY
COPY ./some-file ./
```

earthfile2llb/interpreter.go
alexcb: I have a follow up PR which will expand on handling the withDocker logic which is I suspect will be more complicated than the without logic. I'll swap it for now and might swap back if it helps with readability later.

earthfile2llb/interpreter.go
alexcb: that's a good point, I overlooked this; i'll change it to an error for this PR and replicate+fix later

vladaionescu (approved):

Issue #786, opened 2021-02-18
Cargo-Chef with Earthly caching issue


bhansconnect:
[Cargo-Chef](https://crates.io/crates/cargo-chef) is a tool for being able to cache rust build dependencies. Since, sadly, rust does not have a nice way to tell it to go and download all dependencies. Chef is made with docker containers in mind. Its general process is:
1) prepare - Generate a recipe in a container with the full source
2) cache - Load the recipe into another container and use it to build dependencies
3) build - Load the build dependencies from the cacher and complete your build

If you modify your code without changing a dependency, step 1 will run again, but produce the same recipe. As such, step 2 will notice the recipe is the same and simply used the cached image with dependencies in it. Finally, step 3 will run again.
In the case you do modify a dependency, the recipe will change and step 2 will rerun.

[Naively copying](https://github.com/bhansconnect/earthly-rust-chef/blob/main/Earthfile%20-%20Suggested) the equivalent Dockerfile into an Earthfile does not work as expected. With every single change, Earthly will rerun step 2, even if it generates the same recipe in step 1.
[A workaround](https://github.com/bhansconnect/earthly-rust-chef/blob/main/Earthfile#L16) to at least make this partially work is to make the prepare script only depend on the `Cargo.toml` files and to generate fake `main.rs` and `lib.rs` files. This is not terrible, but it is quite tedious for sizable projects. Also, all of the copy and run operations to move these different directories take quite a long time to generate. Lastly, if a change is made to a `Cargo.toml` file that does not modify the recipe, it will still invalidate all caching.

For a repo with a sample project I set up: https://github.com/bhansconnect/earthly-rust-chef

It would be really nice if copying a new version of an artifact that has the same sha would cause caching to be kept.
Otherwise, it would be nice if there was some sort of copy operation that went faster by grouping more things together.
I believe it is the cost of caching every individual copy operation that makes this slow.

vladaionescu:
Thanks for the detailed explanation and sample project. Super useful! We'll work out a solution for this.

bhansconnect:
Just an extra note: For the chain of copies being slow for me, I think I may be hitting this bug related to zfs and caching layers: https://github.com/moby/moby/issues/31247

vladaionescu:
I don't think that that zfs bug applies - underneath Earthly uses a docker volume for the cache and that volume is provided as ext4 by Docker.

agbell:
[...]

PR #809, opened 2021-02-24
Add early cutoff cache optimization example

agbell:
This example demonstrates not repeating unnecessary subsequent steps after file changes.  This is called 'Early Cut Off Optimization' in the 'Build systems a la cart' paper.  

I also believe this is the behavior being referenced in #786 .


vladaionescu (approved):
Very nice addition

PR #812, opened 2021-02-25
Create config path if missing

dchw:
Looser permissions, plus create directory separately.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -356,5 +356,5 @@ func WriteConfigFile(configPath string, data []byte) error {
 		return err
 	}
 
-	return ioutil.WriteFile(configPath, data, 0755)
+	return ioutil.WriteFile(configPath, data, 0644)
 }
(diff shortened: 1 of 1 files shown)


alexcb (approved):

config/config.go
```diff
@@ -350,5 +351,10 @@ func ReadConfigFile(configPath string, contextSet bool) ([]byte, error) {
 
 // WriteConfigFile writes the config file to disk with preset permission 0644
 func WriteConfigFile(configPath string, data []byte) error {
-	return ioutil.WriteFile(configPath, data, 0644)
+	err := os.MkdirAll(path.Dir(configPath), 0755)
+	if err != nil {
+		return err
+	}
+
+	return ioutil.WriteFile(configPath, data, 0755)
```
alexcb, line 359: seems a bit odd to mark this config file as executable, I think setting it to `0644` is more correct.

dchw:
Not even sure if this is needed since it seems to work in 0.5.2, was just
throwing up something to see if it helped you

config/config.go
dchw: Probably

Issue #813, opened 2021-02-25
Different behavior from local and remote targets


diegobernardes:
The configure target is used to set up the repository setting some Git configs. Both executions happened at the same filesystem path.

Executing locally:
```
➜ earthly --no-cache +configure
           buildkitd | Found buildkit daemon as docker container (earthly-buildkitd)
   python:3.9-buster | --> Load metadata linux/amd64
          +configure | --> FROM python:3.9-buster
             context | --> local context .
          +configure | [██████████] resolve docker.io/library/python:3.9-buster@sha256:88e7df21c03536caeca064dd04a6cac7868b4fdf9e4d849e679e43718eea922c ... 100%
             context | [██████████] transferring .: ... 100%
          +configure | --> COPY --dir .git .git
          +configure | --> RUN git config pull.rebase true && git config remote.origin.prune true && git config branch.main.mergeoptions "--ff-only" && git clone https://github.com/awslabs/git-secrets.git && cd git-secrets && make install && git secrets --install -f && git secrets --register-aws
          +configure | Cloning into 'git-secrets'...
          +configure | tput: No value for $TERM and no -T specified
          +configure | tput: No value for $TERM and no -T specified
          +configure | ✓ Installed commit-msg hook to .git/hooks/commit-msg
          +configure | tput: No value for $TERM and no -T specified
          +configure | tput: No value for $TERM and no -T specified
          +configure | ✓ Installed pre-commit hook to .git/hooks/pre-commit
          +configure | tput: No value for $TERM and no -T specified
          +configure | tput: No value for $TERM and no -T specified
          +configure | ✓ Installed prepare-commit-msg hook to .git/hooks/prepare-commit-msg
          +configure | OK
          +configure | --> SAVE ARTIFACT .git/hooks/* +configure/* AS LOCAL .git/hooks/
          +configure | --> SAVE ARTIFACT .git/config +configure/config AS LOCAL .git/
              output | --> exporting outputs
              output | [██████████] copying files ... 100%
              output | [██████████] sending tarballs ... 100%
================================ SUCCESS [main] ================================
          +configure | Artifact github.com/org/repo:main+configure/config as local .git/config
          +configure | Artifact github.com/org/repo:main+configure/applypatch-msg.sample as local .git/hooks/applypatch-msg.sample
          +configure | Artifact github.com/org/repo:main+configure/commit-msg.sample as local .git/hooks/commit-msg.sample
[...]

PR #815, opened 2021-02-25
Keep git dir in git build context

vladaionescu:
Let me know if you can think of any downsides if we do this.

Fixes https://github.com/earthly/earthly/issues/813

### DIFF
--- a/buildcontext/git.go
+++ b/buildcontext/git.go
@@ -209,6 +209,7 @@ func (gr *gitResolver) resolveGitProject(ctx context.Context, gwClient gwclient.
 
 	gitOpts = []llb.GitOption{
 		llb.WithCustomNamef("[context %s] git context %s", gitURL, target.StringCanonical()),
+		llb.KeepGitDir(),
 	}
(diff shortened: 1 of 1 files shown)


alexcb:
I could see this backfiring for users who structure their build in a way that depends on `.git`, only to discover if they distributed a source-code release as a `.tar.gz`, the building out of that source code will fail since the `.git` directory won't exist.

By allowing the `.git` to be copied into the build we'll ultimately giving the power to the user to decide weather or not the `.git` contents should be referenced or not. 

Perhaps a different solution would be to allow a `--strict` mode which skips the `.git` contents from being included. That would make it problematic for users to issue commands like `git rev-parse HEAD` and would force them to use the earthly env variables instead. That could work for their own build scripts, but what if they wanted to use a vendored tool which required a .git; it seems like a lot to ask users to ditch those tools in favour of (re)writing them to use our env variables (vendor lockin).

Ultimately I think it's fine to include the `.git` even under the `--strict` mode.

alexcb (approved):

Issue #807, opened 2021-02-24
Confusing error when missing FROM


ohadravid:
For this Earthfile:
```

hello:
    RUN mkdir /foo
    COPY bar /foo
```

Version: `earthly version v0.5.1 linux/amd64`

```
$ earthly +hello
           buildkitd | Found buildkit daemon as docker container (earthly-buildkitd)
No ssh auth socket detected or zero keys loaded; falling back to https for auto auth values
             context | --> local context .
              +hello | --> RUN mkdir /foo
              +hello | WARN: (RUN mkdir /foo) invalid mountinput dest:"/run/secrets/earthly_debugger_settings" mountType:SECRET secretOpt:<ID:"earthly_debugger_settings" mode:292 >
             context | [          ] transferring .: ... 0%
             context | WARN: Canceled
Error: failed to load cache key: invalid mountinput dest:"/run/secrets/earthly_debugger_settings" mountType:SECRET secretOpt:<ID:"earthly_debugger_settings" mode:292 >
```

The fix is easy: just add the missing FROM, but the error is very unintuitive.

Thanks! 😄

vladaionescu:
It's probably best if we don't allow any command other than `FROM`, `FROM DOCKERFILE` or `LOCALLY` as the first thing being interpreted.  Then we can display a meaningful error if the user forgets.

This is also relevant for the IF proposal (#779), where the user may be tempted to start off with an IF (but it doesn't work).

PR #817, opened 2021-02-25
Detect first command and error out if it's not a FROM-like command

vladaionescu:
... plus some exceptions, like `BUILD` and `ARG`.

Fixes https://github.com/earthly/earthly/issues/807


alexcb:
Does this PR mean the following Earthfile will no longer be valid?

```
FROM busybox:latest

foo:
    RUN busybox | head
```

vladaionescu:
No, it is valid. Because all targets have an implicit `FROM +base`. It's valid as long as base also has a `FROM` and the from chain eventually leads to a classical `FROM`.

alexcb (approved):

PR #818, opened 2021-02-25
Example for creating rpm and yum repo

alexcb:
- an example for generating a rpm from scratch
- testing the rpm can be installed
- an example for creating a yum repo
- testing the yum repo can be served over http and a centos client can
install the earthly rpm via yum

### DIFF
--- a/examples/yum-repo/Earthfile
+++ b/examples/yum-repo/Earthfile
@@ -3,22 +3,28 @@ deps:
     RUN yum install -y createrepo rpm-build wget
 
 rpm:
+    ARG EARTHLY_VERSION
     FROM +deps
     WORKDIR /work
-    RUN wget -q https://github.com/earthly/earthly/releases/download/v0.5.3/earthly-linux-amd64 -O /usr/local/bin/earthly && chmod +x /usr/local/bin/earthly
-    COPY buildrpm.sh .
-    RUN ./buildrpm.sh
-    SAVE ARTIFACT /root/rpmbuild/RPMS/x86_64/earthly-1.0.0-1.x86_64.rpm AS LOCAL earthly-1.0.0-1.x86_64.rpm
+    RUN echo "$EARTHLY_VERSION" | grep '^[0-9]\+.[0-9]\+.[0-9]\+$' # ensure version is of the form 1.2.3
+    RUN wget -q "https://github.com/earthly/earthly/releases/download/v${EARTHLY_VERSION}/earthly-linux-amd64" -O /usr/local/bin/earthly && chmod +x /usr/local/bin/earthly
+    COPY earthly.spec .
+    RUN sed -i "s/__earthly_version__/$EARTHLY_VERSION/" earthly.spec
+    RUN cat earthly.spec
+    RUN rpmbuild --target x86_64 -bb earthly.spec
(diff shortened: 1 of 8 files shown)


alexcb:
Here's a proof of concept for creating an earthly rpm as well as generating a self-hosted yum repo which can be served on any static webhost. Best to start it with `earthly -P +test`

examples/yum-repo/buildrpm.sh
```diff
@@ -0,0 +1,38 @@
+#/bin/sh
+set -e
+
+mkdir -p rpmbuild/{BUILD,RPMS,SOURCES,SPECS,SRPMS}
+
+cat > earthly.spec << EOF
+Summary: Build automation tool for the container era
+Name: earthly
+Version: 1.0.0
```
alexcb, line 9: this will have to be fixed

examples/yum-repo/Earthfile
```diff
@@ -0,0 +1,55 @@
...
+
+yum-repo-tester:
+    FROM +deps
+    RUN yum install -y curl
+
+    RUN echo "[earthly-test]" > /etc/yum.repos.d/earthly-test.repo
+    RUN echo "name=Earthly Test Repository" >> /etc/yum.repos.d/earthly-test.repo
+    RUN echo "baseurl=http://127.0.0.1:8000/CentOS/8/os/x86_64/" >> /etc/yum.repos.d/earthly-test.repo
+    RUN echo "enabled=1" >> /etc/yum.repos.d/earthly-test.repo
+    RUN echo "gpgcheck=0" >> /etc/yum.repos.d/earthly-test.repo # TODO setup gpg signing
```
dchw, line 39: Proof positive we need heredoc

dchw:
Super cool. PPAs are even easier I assume!

examples/yum-repo/buildrpm.sh
```diff
@@ -0,0 +1,38 @@
...
+/usr/bin/earthly
+
+%changelog
+* Thu Feb 25 2021 alex <alex@earthly.dev>
+- initial poc
+
+%clean
+echo maybe-rm-rf $RPM_BUILD_ROOT
+
+EOF
```
vladaionescu, line 35: Would it be better if we had an actual distinct `earthly.spec` file?

You could use [envsubst](https://linux.die.net/man/1/envsubst) to template $RPM_BUILD_ROOT.

examples/yum-repo/Earthfile
vladaionescu, line 39: Note that this should work

```suggestion
    RUN echo "[earthly-test]
name=Earthly Test Repository
baseurl=http://127.0.0.1:8000/CentOS/8/os/x86_64/
enabled=1
gpgcheck=0" > /etc/yum.repos.d/earthly-test.repo # TODO setup gpg signing
```

vladaionescu (commented):
Nice

examples/yum-repo/Earthfile
alexcb: wonderful, I was hoping I could do something like this, thanks.

examples/yum-repo/buildrpm.sh
alexcb: that's a much better idea, i'll do that in place of my shell script.

alexcb:
This has been added in via https://github.com/earthly/earthly/pull/973 and https://github.com/earthly/earthly/pull/976

PR #824, opened 2021-02-26
Fix local-remote issue from release

vladaionescu:
Revert most of https://github.com/earthly/earthly/pull/810 as it was the issue here.

Add some better logic about what platform to use during export.

### DIFF
--- a/ast/parser/Earthfile
+++ b/ast/parser/Earthfile
@@ -8,7 +8,8 @@ RUN apk add --update --no-cache \
 
 # Install antlr.
 WORKDIR /usr/local/lib
-RUN curl -O https://www.antlr.org/download/antlr-4.9.1-complete.jar
+ARG ANTLR_VERSION=4.9.1
+RUN curl -O https://www.antlr.org/download/antlr-${ANTLR_VERSION}-complete.jar
 WORKDIR /earthly
 
 parser:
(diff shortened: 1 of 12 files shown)


ast/parser/Earthfile
```diff
@@ -8,16 +8,15 @@ RUN apk add --update --no-cache \
 
 # Install antlr.
 WORKDIR /usr/local/lib
-ARG ANTLR_VERSION=4.9.1
```
alexcb, line 11: Is this change intentional? now if we ever want to change the version we need to update it in two locations instead of just this one place. Alternatively maybe we can have curl output to just `antlr-complete.jar` so we only need to hardcode the version in one place.

llbutil/platform.go
```diff
@@ -64,12 +65,33 @@ func PlatformToString(p *specs.Platform) string {
...
+	if p1 == nil {
+		return p2, nil
 	}
-	return override
+	if p2 == nil {
+		return p1, nil
+	}
+	plat1 := platforms.Normalize(*p1)
+	plat2 := platforms.Normalize(*p2)
+	if plat1.OS != plat2.OS {
```
alexcb, line 79: nice additional error checking here :)

alexcb (commented):
Code looks good, just not sure about the `4.9.1` hardcoded value appearing twice.

ast/parser/Earthfile
vladaionescu: This was part of the revert. Anyway, fixed now.

alexcb (approved):

llbutil/platform.go
vladaionescu: Also part of the revert :-)

dchw (approved):

llbutil/platform.go
alexcb: hah, nevermind me :)

PR #826, opened 2021-02-26
Allow platform contradictions - the right way this time

vladaionescu:
This is a redo of https://github.com/earthly/earthly/pull/810

The issue we were seeing during the release was that there was a confusion between the default platform (`linux/amd64`) and `platform=nil`. Because they point to the same thing, the mts Visited cache would kick in and the platform information would not be carried over sometimes.


alexcb (approved):

vladaionescu:
This had to be reverted due to some issues discovered during the release process. Will redo this later.

mariusvniekerk:
Has this made its way back in?

vladaionescu:
Not yet.

PR #831, opened 2021-03-02
Detect detached head and bucket as CI for analytics purposes

vladaionescu:


alexcb (approved):
much more descriptive than "unknown" :+1:

PR #832, opened 2021-03-03
disable analytics for buildkite builds

alexcb:


vladaionescu (approved):

alexcb:
@vladaionescu it looks like the pined version on the m1 doesn't support the `config` subcommand. Is it ok if I update https://github.com/earthly/earthly/blob/disable-analytics-for-buildkit-builds/.buildkite/tests.sh#L17 to a newer version?

alexcb:
I upgraded earthly on the m1 to a more recent version which now supports the `config` command.

PR #833, opened 2021-03-03
RUN --interactive

dchw:
## Executive Summary:

Allows interactive sessions at the ends of builds; with the option to persist those changes into the result or not.

### General:
```
test:
    FROM alpine:3.13
    RUN --interactive type=save ash
```

### In `WITH DOCKER`:
```
test:
    FROM alpine:3.13
    WITH DOCKER --pull ubuntu:latest
        RUN --interactive type=ephemeral docker run -it ubuntu:latest /bin/sh -c 'apt update && apt install -y elixir && iex'
    END
```

### In a `--push`:
```
test:
    FROM alpine:3.13
    RUN --push --interactive type=save ash
```

Of note, ephemeral sessions are evaluated at the end; much like Images and Artifacts. This is to avoid logging interruptions, and avoid other parallelization issues (see the new `InteractiveSession` struct and its similarities to `Save(Image/Artifact)`. Saved sessions are evaluated inline and do not suffer from this issue.

## Invalid, but tempting things you may attempt:

* Use within a `LOCALLY`. Just open an new shell, why would you do this?
* Multiple `--interactive` sessions within a single invocation
  * Includes creating an `IF`-like statement by hiding one in a `--push`
* Any statements after an `--interactive` 
  * Includes `FROM `-ing a target with an `--interactive`
* Usage without a `type` arg. This is because I can't "undo" parsing the next token as part of the interactive sessions arguments; which lead to weird syntax (`--interactive --`) or accidentally stealing the first part of the command.
  * Alternative 1: Two args, `--interactive-save` / `--interactive-ephemeral` that bake this  into the arg itself
  * Alternative 2: Two args, but one only modifies the first. `--interactive` and `--omit` to instruct it to ignore the state
  * Alternative 3: Interactive becomes a `RUN`-like full command; and has options `--save` or `--omit` (default)
* Usage when `--strict` is present or implied.

## Later Ideas:

* Remove the single, and last restrictions. I tried to code this so it would more-or-less be code removal only to provide this.
* Use the `NewContainer` buildkit API. This would require an equal, but different set of hacks.
* Find a way to let builds run in the background without interrupting the flow of the interactive session.
* Distribute an interactive Zork container.

## Things Left To Do:

- [ ] Documentation
- [ ] Some kind of integration test (this could get weird)


earthfile2llb/converter.go
```diff
@@ -147,6 +147,9 @@ func (c *Converter) fromTarget(ctx context.Context, targetName string, platform
 	if err != nil {
 		return errors.Wrapf(err, "apply build %s", depTarget.String())
 	}
+	if mts.Final.RanInteractive {
+		return errors.New("Cannot FROM a target ending with an --interactive")
+	}
```
dchw, line 152: This check was needed because we wouldn't otherwise fail on this case:

```
test:
    FROM alpine:3.13
    RUN apk add bash
    RUN --interactive type=ephemeral bash

from-interactive:
    FROM +test
```

vladaionescu (approved):
This looks great!

The only comment I have is regarding that `type` switch. It feels very wordy for the user to remember. I would go with your first alternative, but have one as "the default".

> Alternative 1: Two args, --interactive-save / --interactive-ephemeral that bake this into the arg itself

I would do  `--interactive` (this is the ephemeral one) and `--interactive-keep` (save has a somewhat different meaning in other places of Earthly).

dchw:
> I would do `--interactive` (this is the ephemeral one) and `--interactive-keep` (save has a somewhat different meaning in other places of Earthly).

I did this. Plus allowed `SAVE IMAGE` after an `--interactive` command to make it worth actually having the keep. Will be sure to note this in the (forthcoming) docs.

vladaionescu (approved):

PR #840, opened 2021-03-04
With docker locally

alexcb:
the code needs some loving, but here's an initial not-ready-for-review draft of running with docker locally.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -256,6 +256,10 @@ jobs:
         run: "ls /tmp/earthly-test-local"
       - name: Run general local tests (TODO this is re-testing the +test-local target)
         run: "./build/linux/amd64/earthly --use-inline-cache --save-inline-cache --no-output ./examples/tests/local+all"
+      - name: Buildkit logs (runs on failure)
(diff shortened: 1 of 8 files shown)


vladaionescu (commented):
Looks pretty good to me so far. No compose support, but it's a start.

Wondering if we should reuse more of the other `WITH DOCKER` implementation. But then again, I can also see this as being different enough.

alexcb:
> Looks pretty good to me so far. No compose support, but it's a start.

I was able to get docker compose working in this test: https://github.com/earthly/earthly/pull/840/files#diff-4141aec9ede776e8e9f60ce8be4517a0431dbc996c9fdd9977ff34a4c8489026R13

what's sort of interesting is there's no explicit COPY directive to pull in the docker-compose.yml file since it's already in the current directory on the local machine.

> Wondering if we should reuse more of the other `WITH DOCKER` implementation. But then again, I can also see this as being different enough.

Yeah I was wondering a bit about that too, there's some code that was copy/pasted from there, but a fair amount of it changed.

vladaionescu (approved):

PR #843, opened 2021-03-04
Improve analytics

vladaionescu:
* Better detection for Jenkins

### DIFF
--- a/analytics/analytics.go
+++ b/analytics/analytics.go
@@ -28,12 +28,12 @@ func detectCI() (string, bool) {
 		"GITHUB_WORKFLOW": "github-actions",
 		"CIRCLECI":        "circle-ci",
 		"JENKINS_HOME":    "jenkins",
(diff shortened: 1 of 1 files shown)


analytics/analytics.go
```diff
@@ -28,6 +28,7 @@ func detectCI() (string, bool) {
 		"GITHUB_WORKFLOW": "github-actions",
 		"CIRCLECI":        "circle-ci",
 		"JENKINS_HOME":    "jenkins",
+		"AGENT_WORKDIR":   "jenkins", // https://github.com/jenkinsci/docker-agent/blob/master/11/alpine/Dockerfile#L35
```
alexcb, line 31: Can we move this to the bottom of the list, just in case it overlaps with a different CI system? I'm worried about this conflicting with a different CI system down the road.

vladaionescu: Yeah, good idea.

alexcb (approved):

PR #857, opened 2021-03-10
Disallow multiple targets with the same name

dchw:
This is more or less a small validation framework we can grow. Example output:

image


alexcb:
This seems like a nice to have feature; is it possible to add a test for this?

earthfile2llb/validator.go
```diff
@@ -0,0 +1,55 @@
+package earthfile2llb
```
vladaionescu, line 1: Could this file go under the `ast` package and have public functions?

earthfile2llb/earthfile2llb.go
```diff
@@ -115,6 +115,9 @@ func Earthfile2LLB(ctx context.Context, target domain.Target, opt ConvertOpt) (m
 	if err != nil {
 		return nil, err
 	}
+	if err := validateAst(bc.Earthfile); err != nil {
+		return nil, err
+	}
```
vladaionescu, line 120: I would put this check in the build context resolver, right after the parse call. That way  `bc` always has a validated `bc.Earthfile`.

earthfile2llb/validator.go
dchw: Yeah; I had started here hoping to do a little more but decided it wasnt worth it. Ill move it to finish formalizing that

earthfile2llb/earthfile2llb.go
dchw: How about putting it in the `Parse` method itself? Then its always valid after parsing, and we don't have to make validation public?

dchw:
> This seems like a nice to have feature; is it possible to add a test for this?

Not just nice to have; we are allowing duplicate target names within a file right now :(

earthfile2llb/earthfile2llb.go
vladaionescu: That might be better

dchw:
@alexcb Test added.

alexcb (approved):

PR #864, opened 2021-03-16
Minimum Windows Compilation

dchw:
Still a WIP. This is the minimum set of changes to at least get it to compile and run trivial Earth targets. Not yet self-hosting.


dchw:
Currently investigating why it is trying to use Windows paths when copying from another target (within `FROM DOCKERFILE` while trying to self-host). These paths seems to be changing within the buildkit client itself... going to take more sleuthing.

debugger/terminal/terminal_windows.go
```diff
@@ -0,0 +1,13 @@
...
+package terminal
+
+import (
+	"context"
+
+	"github.com/pkg/errors"
+)
+
+func ConnectTerm(ctx context.Context, addr string) error {
+	return errors.New("Debugger not supported on Windows yet")
```
vladaionescu, line 12:
```suggestion
	return errors.New("debugger not supported on Windows yet")
```

Nit: Error strings should start with lower-case

vladaionescu (commented):
Looks good so far. We could even merge this as is.

dchw:
> Looks good so far. We could even merge this as is.

Lets do it then. Piece by piece!

vladaionescu (approved):

PR #865, opened 2021-03-16
test-case for broken escape handling

alexcb:
I'm struggling with figuring out a fix in the parser, but here's a case to reproduce it.


examples/tests/quotes.earth
```diff
@@ -17,9 +18,18 @@ test-env-quoting:
...
 
-test-env-quoting2:
-    ENV BACKSLASH "\\"
-    RUN test "$BACKSLASH" = "\\"
+test-env-unquoted-backslash:
+    RUN touch /before-backslash
+    RUN echo \\
+    RUN touch /after-backslash
+    RUN ls /before-backslash
+    RUN ls /after-backslash
```
alexcb, line 26: this right here is failing; since the `RUN echo \\` line is turning into `RUN echo \    RUN touch /after-backslash`

Issue #581, opened 2020-12-05
Proposal: User-defined commands (UDCs)


vladaionescu:
### Proposal

This proposal includes three key parts:
  * The transition from the syntax `--build-arg XYZ=value` to `--xyz=value`
  * ~The extension of ARGs to allow other types besides the default string type~ EDIT: This part has been abandoned for now.
  * User-defined commands (UDCs). A concept somewhat similar to Earthly targets - they can be referenced across directories and projects and they are made up of a series of Earthly commands. However the key difference is that Earthly targets come with a specific build context (the dir where `COPY ./file ./dest` copies files from) and also Earthly targets reset the build environment as a result of a `FROM` command. User-defined commands on the other hand inherit both the build context and the build environment from the caller.

### Example

```Dockerfile
INSTALL_DEPS:
  COPY go.mod go.sum ./
  RUN go mod download
  SAVE ARTIFACT go.mod AS LOCAL go.mod
  SAVE ARTIFACT go.sum AS LOCAL go.sum
```

Usage example

```Dockerfile
IMPORT github.com/bla/bla+INSTALL_DEPS

code:
  +INSTALL_DEPS
  COPY src ./src
  RUN go build …
```

OR (without the IMPORT)

```
code:
  github.com/bla/bla+INSTALL_DEPS
  COPY src ./src
  RUN go build …
```

An example with args

```Dockerfile
COMPILE_GO:
  ARG copy=./
  ARG out
  ARG GOCACHE=/go-cache
  COPY $copy ./src
  RUN --mount=type=cache,target=$GOCACHE \
    go build -o $out src/*.go

command1:
  +COMPILE_GO --copy=./cmd/command1 --out=./build/command1

command2:
  +COMPILE_GO --copy=./cmd/command2 --out=./build/command2
```

### Notes

* User-defined commands (UDCs) are distinguished from targets by the fact that they are all upper case. Targets may not have upper case characters.
* A UDC executes in the context in which it is called. FROM is not allowed.
* The build context is whatever build context the caller has (COPY commands work as if they are inlined in the caller)
* Arguments are defined the same way as build args, but are lower-case and can be specified like command-line args eg `ARG copy` can be passed as `+<command> --copy=....`
* In UDCs args are not transitive - they have to be passed explicitly

### ARG types

```Dockerfile
MY_COPY:
  ARG --bool recursive
  ARG --positional --string src
  ARG --positional --string dest=./
[...]

PR #867, opened 2021-03-17
Implementation of new style args and support for args in UDCs

vladaionescu:
Re #581 

Add support for new style args:

```
FROM +something --arg=foo
BUILD +something --arg=foo
COPY (+something/artifact --arg=foo) ./dest
```

Add support for `ARG`s in UDCs.

```
DO +MY_UDC --arg1=foo --arg2=bar

MY_UDC:
  COMMAND
  ARG arg1=default
  ARG arg2
  ...
```

Args in UDCs have their own, isolated scope.

The old arg style still works with non-UDC commands, for backwards compatibility. They will be deprecated in the future.

```
FROM --build-arg arg=foo +something
```


examples/tests/command.earth
```diff
@@ -29,3 +68,29 @@ COMMANDCEPTION:
 ANOTHER_COMMAND:
     COMMAND
     DO +TOUCH
+
+COMMAND_ARGS:
+    COMMAND
+    ARG arg1=default
+    RUN touch "./$arg1"
+
+RECURSIVE:
```
alexcb, line 77: amazing!

alexcb (approved):

examples/tests/command.earth
agbell: This is great 🥇

PR #868, opened 2021-03-17
Implementation of IMPORT

vladaionescu:
Re #581 

Syntax is:

```
IMPORT github.com/foo/bar:v1.2.3
IMPORT github.com/foo/buz:main AS zulu

...

FROM bar+target
BUILD zulu+something
```


variables/collection.go
```diff
@@ -166,6 +174,11 @@ func (c *Collection) ExitFrame() {
 	c.effectiveCache = nil
 }
 
+// IsStackAtBase returns whether the stack has size 1.
+func (c *Collection) IsStackAtBase() bool {
+	return len(c.stack) == 1
```
alexcb, line 179: nit: should we check for a condition like

    if len(c.stack) == 0 { panic("shouldnt happen") }

then again if it's empty that might cause some nil exceptions elsewhere. :shrug:

alexcb (approved):

variables/collection.go
vladaionescu: Yes - we actually have such a check in `popStackFrame`

PR #869, opened 2021-03-18
Accept new style args on the command line too

vladaionescu:


alexcb:
I see a lot of `fmt.Errorf` to `errors.Errorf`, should we add adopt errors.Errorf in all places and add a test that greps for fmt.Errorf and errors out during our linting process?

vladaionescu:
> I see a lot of fmt.Errorf to errors.Errorf, should we add adopt errors.Errorf in all places and add a test that greps for fmt.Errorf and errors out during our linting process?

Perhaps - I figured we'd go with some consistency. At least one file at a time. Could be a separate PR if we did this for the whole repo though.

alexcb (approved):

PR #870, opened 2021-03-18
Add docs for the latest commands: IF, DO, COMMAND, IMPORT

vladaionescu:
Preview: https://app.gitbook.com/@earthly/s/earthly-staging/~/revisions/-MW6acp3S4kvi1vZtWUT/v/vlad%2Fdocs-new-commands/guides/udc

### DIFF
--- a/docs/earthfile/earthfile.md
+++ b/docs/earthfile/earthfile.md
@@ -755,7 +755,7 @@ ELSE
 END
 ```
 
-The reason this is invalid is because the `IF` condition does not have a build environment initialized, to execute in.
+The reason this is invalid is because the `IF` condition is actually running the `/usr/bin/[` executable to test if the condition is true or false, and therefore requires a valid build environment has been initialized.
 
 Here is how this might be fixed.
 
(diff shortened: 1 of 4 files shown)


docs/earthfile/earthfile.md
```diff
@@ -684,6 +690,197 @@ Specifies the platform for any referenced `--load` and `--pull` images.
...
+# NOT A VALID EARTHFILE.
+ARG base=alpine
+IF [ "$base" = "alpine" ]
+    FROM alpine:3.13
+ELSE
+    FROM ubuntu:20.04
+END
+```
+
+The reason this is invalid is because the `IF` condition does not have a build environment initialized, to execute in.
```
alexcb, line 758: I wonder if we should explicitly call out `/usr/bin/[` to explain what's actually happening; not all users realise that `[` is actually a program.
```suggestion
The reason this is invalid is because the `IF` condition is actually running the `/usr/bin/[` executable to test if the condition is true or false, and therefore requires a valid build environment has been initialized.
```

alexcb (approved):

PR #871, opened 2021-03-18
generate railroad diagrams for lexer and parser

alexcb:
- the generate-railroad-diagrams target will save two html files locally
which can be used to visualize the lexer and parser antlr code.

here's an example of a screenshot of a generated html page containing the railroad diagram.

image


vladaionescu (approved):
Wow

PR #875, opened 2021-03-20
Use a UDC for tests

vladaionescu:


examples/tests/Earthfile
```diff
@@ -525,71 +371,45 @@ file-copying:
...
+        --post_command="2>&1 | perl -pe 'BEGIN {\\\$status=1} END {exit \\\$status} \\\$status=0 if /not found/;'"
 
 push-build:
-    COPY push-build.earth ./Earthfile
-    RUN --privileged \
-        --entrypoint \
-        --mount=type=tmpfs,target=/tmp/earthly \
-        -- --push +test 2>&1 > output
-    
+    DO +RUN_EARTHLY --earthfile=push-build.earth --target=+test --post_command="2>&1 > output"
```
dchw, line 398:
```suggestion
    DO +RUN_EARTHLY --earthfile=push-build.earth --target=+test --extra_args="--push" --post_command="2>&1 > output"
```

builder/solver_monitor.go
```diff
@@ -519,8 +519,8 @@ func (sm *solverMonitor) reprintFailure(errVertex *vertexMonitor, phaseText stri
 	errVertex.printError()
 }
 
-var vertexRegexp = regexp.MustCompile("^\\[([^\\]]*)\\] (.*)$")
-var targetAndSaltRegexp = regexp.MustCompile("^([^\\(]*)(\\(([^\\)]*)\\))? (.*)$")
+var vertexRegexp = regexp.MustCompile("(?s)^\\[([^\\]]*)\\] (.*)$")
```
alexcb, line 522: what does `(?s)` match? I'm had a quick search for `(?` under https://github.com/google/re2/wiki/Syntax and couldn't figure it out.

vladaionescu: It's a regex flag to include `\n` in `.` matches. I had an issue with the multi-line script not showing up in the output because it was classified as internal, because one of these two regexes didn't match.

See `(?flags)` here https://golang.org/pkg/regexp/syntax/ under `Grouping`.

alexcb: `s              let . match \n (default false)`; ah ha! thanks for pointing that out. makes more sense now.

alexcb (approved):
looks like a good use of the COMMAND to reduce the amount of code; the only bit that looks a bit odd (at first) is `--post_command="| grep 'hello world'"`, but it makes sense.

PR #883, opened 2021-03-23
Corey/retry metadata

dchw:
Just want the CI to run to test this retry test I made

### DIFF
--- a/earthfile2llb/converter.go
+++ b/earthfile2llb/converter.go
@@ -18,6 +18,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/opencontainers/go-digest"
+
 	"github.com/earthly/earthly/buildcontext"
 	"github.com/earthly/earthly/debugger/common"
 	"github.com/earthly/earthly/domain"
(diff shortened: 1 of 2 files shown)


dchw:
Ha! Retry does work here. This Mac build would have failed: https://buildkite.com/earthly-technologies/earthly-mac-ci/builds/2151#825d7858-0390-4475-92e0-67ba2cfb1bea

Looks like there is some retry deep down inside the resolver, but only triggers on certain errors.

Probably better to pull in a legit retry library here? Or something with gentle backoff...

dchw:
Even does this in Circle! Would have saved this run: https://app.circleci.com/pipelines/github/earthly/earthly/3975/workflows/52cce0e9-4520-43c1-83c1-1a0372a72fc1/jobs/3987

dchw:
The fact that this build succeeded _first try_ (well, except for the M1) I think shows that this does, indeed fix it. IMO there is room for other stuff inside to get retried (cache? Image push/pulls?)

dchw:
Again on Circle: https://app.circleci.com/pipelines/github/earthly/earthly/3975/workflows/99296c5d-614b-442a-b897-7dc33bddc2d6/jobs/3988

vladaionescu (commented):
I think we should do something like this.

It might be good to add the TLS error in buildkit retry whitelist. Wonder if we should be doing a retry on our side too in addition.

vladaionescu:
I ended up doing this: https://github.com/moby/buildkit/pull/2036 for now. I cherry-picked this in our `buildkit` `earthly-main` branch.

We'll keep an eye on it and if flaky stuff keeps happening, we should also do something like this PR on our side too. In the meantime, I'll close.

(BTW, this library seemed pretty good, if we ever need to come back to this: https://github.com/cenkalti/backoff)

PR #887, opened 2021-03-23
Test limiting parallelism

vladaionescu:
For https://github.com/earthly/buildkit/pull/34


dchw:
Didn't see why this was closed. An interesting idea that requires no external config is some multiple of `GOMAXPROCS` or the runtime equivalent.

vladaionescu:
The change is only needed on the buildkit side. This PR was opened for running the CI only. The relevant change has been applied to `earthly-main`.

PR #892, opened 2021-03-25
Various random improvements

vladaionescu:
* Print stack traces when `--verbose` is used
* `fmt.Errorf` -> `errors.Errorf` s.t. it includes stack trace
* Random staticcheck fixes
* Limit the amount of time a connection is attempted for when waiting for buildkitd to start up

### DIFF
--- a/conslogging/conslogging.go
+++ b/conslogging/conslogging.go
@@ -255,7 +255,7 @@ func (cl ConsoleLogger) PrintBytes(data []byte) {
 	}
 	if len(output) > 0 {
 		c.Fprintf(cl.outW, "%s", string(output))
-		// output = output[:0] // needed if output is used futher in the future
+		// output = output[:0] // needed if output is used further in the future
 	}
 }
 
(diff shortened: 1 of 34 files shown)


conslogging/conslogging.go
```diff
@@ -256,7 +255,7 @@ func (cl ConsoleLogger) PrintBytes(data []byte) {
 	}
 	if len(output) > 0 {
 		c.Fprintf(cl.outW, "%s", string(output))
-		output = output[:0]
+		// output = output[:0] // not needed
```
dchw, line 258: Nit: if its not needed we should delete it

dchw (approved):
LGTM

conslogging/conslogging.go
vladaionescu: I'd leave it in case we need to add more code after that line that uses output somehow. Then it is needed and it might not be obvious.

dchw: Nit on the nit: Maybe note in the comment when it would be needed since its not obvious

PR #897, opened 2021-03-29
fix mapping of custom projects to self-hosted git repos

alexcb:
- fix our handling of non-standard (e.g. self-hosted) git repos; this bug only affected accessing subdirectories of self-hosted git repos.
- extend our self hosted private repo test to reference commands that are stored remotely but from a local Earthfile
    
### DIFF
--- a/scripts/tests/self-hosted-private-repo.sh
+++ b/scripts/tests/self-hosted-private-repo.sh
@@ -1,6 +1,20 @@
 #!/usr/bin/env bash
 set -eu
 
+if [ -z ${GITHUB_ACTIONS+x} ]; then
+    echo "this script should only be run from GHA; if run locally it will modify your ssh settings"
+    exit 1
+fi
+
+# setup a trap to display logs on errors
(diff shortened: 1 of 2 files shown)


scripts/tests/self-hosted-private-repo.sh
```diff
@@ -68,18 +73,69 @@ git branch -M trunk
...
+
+# next test that we can reference commands in the weird repo;
+# create a local Earthfile (that wont be saved to git)
+cat <<EOF >> Earthfile
+
+IMPORT myserver/project/weirdcommands:trunk
+
+FROM alpine:latest
+
+testweirdtouch:
```
dchw, line 131: Got to be the funniest target name we have made yet :D

alexcb: oh no, now that you've pointed it out I've had bloodhound gang flashbacks.

vladaionescu (approved):
:cool:

PR #906, opened 2021-04-01
Add code to allow overriding MTU from earhly YAML.

dchw:
Simple as `earthly config global.cni_mtu 12345`.

### DIFF
--- a/cmd/earthly/main.go
+++ b/cmd/earthly/main.go
@@ -903,6 +903,12 @@ func (app *earthlyApp) before(context *cli.Context) error {
 	app.buildkitdSettings.AdditionalArgs = app.cfg.Global.BuildkitAdditionalArgs
 	app.buildkitdSettings.AdditionalConfig = app.cfg.Global.BuildkitAdditionalConfig
 
+	// ensure the MTU is something allowable in IPv4, cap enforced by type. Zero is autodetect.
+	if app.buildkitdSettings.CniMtu != 0 && app.buildkitdSettings.CniMtu < 68 {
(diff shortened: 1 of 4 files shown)


cmd/earthly/main.go
```diff
@@ -903,6 +903,12 @@ func (app *earthlyApp) before(context *cli.Context) error {
 	app.buildkitdSettings.AdditionalArgs = app.cfg.Global.BuildkitAdditionalArgs
 	app.buildkitdSettings.AdditionalConfig = app.cfg.Global.BuildkitAdditionalConfig
 
+	// ensure the MTU is something allowable in IPv4, cap enforced by type. Zero is autodetect.
+	if app.buildkitdSettings.CniMtu != 0 && app.buildkitdSettings.CniMtu < 68 {
+		return errors.New("Invalid overridden MTU size")
```
vladaionescu, line 908: Nit: lower-case errors

```suggestion
		return errors.New("invalid overridden MTU size")
```

vladaionescu (approved):

cmd/earthly/main.go
dchw: I get that idiomatic Go wants lower case errors, but it feels wrong to present sentences that start with lower case to the user.

vladaionescu: It's actually presented as `Error: invalid overridden MTU size` to the user

dchw: Still feels wrong to me. But I am willing to follow convention.

alexcb (approved):

PR #910, opened 2021-04-05
Earthly installation improvements

vladaionescu:
* Nicer Buildkit pull message
* Perform buildkit pull and start as part of bootstrap
* Don't classify detached head as CI due to possible false positives

### DIFF
--- a/buildkitd/buildkitd.go
+++ b/buildkitd/buildkitd.go
@@ -33,8 +33,6 @@ var (
 	ErrBuildkitStartFailure = errors.New("buildkitd failed to start (in time)")
 )
 
-const reportPull = false // flip to print pull messages
-
 // Address is the address at which the daemon is available.
 var Address = fmt.Sprintf("docker-container://%s", ContainerName)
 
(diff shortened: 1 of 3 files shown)


buildkitd/buildkitd.go
```diff
@@ -423,12 +435,16 @@ func MaybePull(ctx context.Context, image string) error {
 	}
 	args = append(args, image)
 	cmd = exec.CommandContext(ctx, "docker", args...)
-	cmd.Stdout = os.Stdout
-	cmd.Stderr = os.Stderr
+	console.
+		WithPrefix("buildkitd-pull").
```
dchw, line 439: Nit: the message also says it is pulling. We don't do separate prefixes for when other images are being pulled or pushed, why would we here?

buildkitd/buildkitd.go
```diff
@@ -212,12 +214,16 @@ func Start(ctx context.Context, image string, settings Settings, reset bool) err
 	if err != nil {
 		return err
 	}
-	// Pulling is not strictly needed, but it helps display some progress status to the user in
-	// case the image is not available locally.
-	err = MaybePull(ctx, image)
-	if err != nil {
-		fmt.Printf("Error: %s. Attempting to start buildkitd anyway...\n", err.Error())
-		// Keep going - it might still work.
+	if reportPull {
```
dchw, line 217: Why are we gating this right now?

buildkitd/buildkitd.go
vladaionescu: It's mainly because of the `...Done` message. There's one for pulling and one for starting.

buildkitd/buildkitd.go
vladaionescu: Not sure if this is causing user drops - I would turn it off for now as an extra precaution.

dchw (approved):

dchw (approved):

PR #911, opened 2021-04-05
Add `earthly bootstrap --ci` flag

vladaionescu:
* Marks the installation as a CI installation via the Earthly config
* Skips bash/zsh autocompletion installation as part of bootstrap


dchw (approved):

vladaionescu:
Probably not really that useful...

PR #912, opened 2021-04-06
Switch uuid package

alexcb:


analytics/analytics.go
```diff
@@ -19,8 +19,8 @@ import (
 	"github.com/earthly/earthly/gitutil"
 	"github.com/earthly/earthly/syncutil"
 
-	uuid "github.com/nu7hatch/gouuid"
 	"github.com/pkg/errors"
+	uuid "gopkg.in/go-on/go.uuid.v1"
```
vladaionescu, line 23: BTW we use this package in other parts of our code: https://github.com/google/uuid - would the Google one have the same functionality?

alexcb: that's a good point, and `NewUUID` returns a v1. I'll switch to that one.

vladaionescu (approved):

PR #915, opened 2021-04-06
Write to ash_history too

alexcb:
- the debugger should produce an .ash_history file incase the user
enters ash instead of bash.

### DIFF
--- a/cmd/debugger/main.go
+++ b/cmd/debugger/main.go
@@ -17,6 +17,7 @@ import (
 	"github.com/alessio/shellescape"
 	"github.com/creack/pty"
 	"github.com/fatih/color"
+	multierror "github.com/hashicorp/go-multierror"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
@@ -66,6 +67,7 @@ func handleWinChangeData(ptmx *os.File, data []byte) error {
 }
 
 func populateShellHistory(cmd string) error {
+	var result error
 	for _, f := range []string{
 		"/root/.ash_history",
 		"/root/.bash_history",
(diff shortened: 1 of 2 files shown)


cmd/debugger/main.go
```diff
@@ -66,12 +66,18 @@ func handleWinChangeData(ptmx *os.File, data []byte) error {
...
-	if err != nil {
-		return err
+	for _, f := range []string{
+		"/root/.ash_history",
+		"/root/.bash_history",
+	} {
+
+		f, err := os.Create(f)
+		if err != nil {
+			return err
```
vladaionescu, line 76: Wonder if we should `continue`?

cmd/debugger/main.go
```diff
@@ -66,12 +66,18 @@ func handleWinChangeData(ptmx *os.File, data []byte) error {
 }
 
 func populateShellHistory(cmd string) error {
-	f, err := os.Create("/root/.bash_history")
-	if err != nil {
-		return err
+	for _, f := range []string{
+		"/root/.ash_history",
```
vladaionescu, line 70: Do we need to also `MkdirAll("/root")`?

vladaionescu (approved):

cmd/debugger/main.go
alexcb: I don't think so, if that were the case the image would be very broken I suspect.

cmd/debugger/main.go
alexcb: good point, how about I switch it over to using multierrors?

vladaionescu: sounds good

vladaionescu (approved):

PR #916, opened 2021-04-06
new command to run Dockerfile directly

alexcb:
- behind the scenes we 1) convert the Dockerfile into an Earthfile, 2)
run an interactive build on the converted Earthfile, then 3) delete the
generated Earthfile.

### DIFF
--- a/cmd/earthly/main.go
+++ b/cmd/earthly/main.go
@@ -2174,6 +2174,10 @@ func (app *earthlyApp) actionDocker(c *cli.Context) error {
 		return err
 	}
 
+	fmt.Fprintf(os.Stderr, "Warning: earthly does not support all dockerfile commands and is highly experimental as a result, use with caution.\n")
+
+	app.imageMode = false
+	app.artifactMode = false
 	app.interactiveDebugging = true
 	flagArgs := []string{}
 	nonFlagArgs := []string{"+build"}
--- a/cmd/earthly/main.go
+++ b/cmd/earthly/main.go
@@ -524,6 +524,26 @@ func newEarthlyApp(ctx context.Context, console conslogging.ConsoleLogger) *eart
 			},
 		},
 		{
+			Name:        "docker",
+			Usage:       "Build a Dockerfile without converting to an Earthfile",
+			Description: "Builds a dockerfile",
+			Hidden:      true, // Experimental.
+			Action:      app.actionDocker,
+			Flags: []cli.Flag{
+				&cli.StringFlag{
+					Name:        "dockerfile",
+					Usage:       "Path to dockerfile input, or - for stdin",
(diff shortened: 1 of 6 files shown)


cmd/earthly/main.go
```diff
@@ -2139,8 +2159,36 @@ func (app *earthlyApp) actionPrune(c *cli.Context) error {
...
+}
+
 func (app *earthlyApp) actionDocker2Earthly(c *cli.Context) error {
-	return docker2earthly.Docker2Earthly(app.dockerfilePath, app.earthfilePath, app.earthfileFinalImage)
+	app.commandName = "docker2earthly"
+	err := docker2earthly.Docker2Earthly(app.dockerfilePath, app.earthfilePath, app.earthfileFinalImage)
+	if err != nil {
+		return err
+	}
+	fmt.Fprintf(os.Stderr, "An Earthfile has been generated; to run it use: earthly +build; then run with docker run -ti %s\n", app.earthfileFinalImage)
```
vladaionescu, line 2190: Should we standardize on using `app.console` for messages to the user? Perhaps console should just always print to stderr.

cmd/earthly/main.go
```diff
@@ -2194,13 +2242,18 @@ func (app *earthlyApp) actionBuild(c *cli.Context) error {
...
 	if err != nil {
 		return errors.Wrapf(err, "parse args %s", strings.Join(c.Args().Slice(), " "))
 	}
+
+	return app.actionBuildImp(c, flagArgs, nonFlagArgs)
+}
+func (app *earthlyApp) actionBuildImp(c *cli.Context, flagArgs, nonFlagArgs []string) error {
+	var target domain.Target
+	var artifact domain.Artifact
+	destPath := "./"
```
vladaionescu, line 2256: I think it's possible to run `earthly --artifact docker2earthly` and cause artifact mode to be enabled, which wouldn't make sense in this case.

Not sure what's the right way to prevent this. In any case - that can be a separate PR.

vladaionescu (commented):
What's the main motivation for this functionality?

It feels like a brittle way to run Dockerfiles, with no advantages over just running `docker build`.

alexcb:
> It feels like a brittle way to run Dockerfiles, with no advantages over just running `docker build`.

The only advantage is that if an error occurs it'll drop you into an interactive shell. I have it default-enabled, but now that I think of it maybe it should be an option, so then people will discover the feature more easily while searching google for "how to debug dockerfile interactively".

vladaionescu (approved):

cmd/earthly/main.go
vladaionescu: Makes sense - can't remember if it does already. In any case, it should be easy to switch in the `conslogging` package.

PR #926, opened 2021-04-12
Disallow privileged and local execution from remote Earthfiles

alexcb:
- We should avoid running local and priviledged commands that are
referenced from remote Earthfiles.
- We will design a way to configure trusted repos in a follow up PR.

### DIFF
--- a/earthfile2llb/converter.go
+++ b/earthfile2llb/converter.go
@@ -79,7 +79,7 @@ func NewConverter(ctx context.Context, target domain.Target, bc *buildcontext.Da
 }
 
 // From applies the earthly FROM command.
-func (c *Converter) From(ctx context.Context, imageName string, platform *specs.Platform, buildArgs []string) error {
+func (c *Converter) From(ctx context.Context, imageName string, platform *specs.Platform, allowPrivileged bool, buildArgs []string) error {
 	err := c.checkAllowed("FROM")
 	if err != nil {
 		return err
@@ -93,7 +93,7 @@ func (c *Converter) From(ctx context.Context, imageName string, platform *specs.
 	c.setPlatform(platform)
 	if strings.Contains(imageName, "+") {
 		// Target-based FROM.
-		return c.fromTarget(ctx, imageName, platform, buildArgs)
+		return c.fromTarget(ctx, imageName, platform, allowPrivileged, buildArgs)
 	}
 
 	// Docker image based FROM.
@@ -126,12 +126,12 @@ func (c *Converter) fromClassical(ctx context.Context, imageName string, platfor
 	return nil
 }
 
-func (c *Converter) fromTarget(ctx context.Context, targetName string, platform *specs.Platform, buildArgs []string) error {
+func (c *Converter) fromTarget(ctx context.Context, targetName string, platform *specs.Platform, allowPrivileged bool, buildArgs []string) error {
 	depTarget, err := domain.ParseTarget(targetName)
 	if err != nil {
 		return errors.Wrapf(err, "parse target name %s", targetName)
 	}
-	mts, err := c.buildTarget(ctx, depTarget.String(), platform, buildArgs, false, true)
+	mts, err := c.buildTarget(ctx, depTarget.String(), platform, allowPrivileged, buildArgs, false, true)
 	if err != nil {
 		return errors.Wrapf(err, "apply build %s", depTarget.String())
 	}
@@ -184,7 +184,7 @@ func (c *Converter) FromDockerfile(ctx context.Context, contextPath string, dfPa
 		// The Dockerfile and build context are from a target's artifact.
 		// TODO: The build args are used for both the artifact and the Dockerfile. This could be
 		//       confusing to the user.
-		mts, err := c.buildTarget(ctx, contextArtifact.Target.String(), platform, buildArgs, false, false)
+		mts, err := c.buildTarget(ctx, contextArtifact.Target.String(), platform, false, buildArgs, false, false)
 		if err != nil {
 			return err
 		}
@@ -297,7 +297,7 @@ func (c *Converter) Locally(ctx context.Context, workdirPath string, platform *s
 }
 
 // CopyArtifactLocal applies the earthly COPY artifact command which are invoked under a LOCALLY target.
-func (c *Converter) CopyArtifactLocal(ctx context.Context, artifactName string, dest string, platform *specs.Platform, buildArgs []string, isDir bool) error {
+func (c *Converter) CopyArtifactLocal(ctx context.Context, artifactName string, dest string, platform *specs.Platform, allowPrivileged bool, buildArgs []string, isDir bool) error {
 	err := c.checkAllowed("COPY")
 	if err != nil {
 		return err
@@ -307,7 +307,7 @@ func (c *Converter) CopyArtifactLocal(ctx context.Context, artifactName string, 
 	if err != nil {
 		return errors.Wrapf(err, "parse artifact name %s", artifactName)
 	}
-	mts, err := c.buildTarget(ctx, artifact.Target.String(), platform, buildArgs, false, false)
+	mts, err := c.buildTarget(ctx, artifact.Target.String(), platform, allowPrivileged, buildArgs, false, false)
 	if err != nil {
 		return errors.Wrapf(err, "apply build %s", artifact.Target.String())
 	}
@@ -341,7 +341,7 @@ func (c *Converter) CopyArtifactLocal(ctx context.Context, artifactName string, 
 }
 
 // CopyArtifact applies the earthly COPY artifact command.
-func (c *Converter) CopyArtifact(ctx context.Context, artifactName string, dest string, platform *specs.Platform, buildArgs []string, isDir bool, keepTs bool, keepOwn bool, chown string, ifExists bool) error {
+func (c *Converter) CopyArtifact(ctx context.Context, artifactName string, dest string, platform *specs.Platform, allowPrivileged bool, buildArgs []string, isDir bool, keepTs bool, keepOwn bool, chown string, ifExists bool) error {
 	err := c.checkAllowed("COPY")
 	if err != nil {
 		return err
@@ -351,7 +351,7 @@ func (c *Converter) CopyArtifact(ctx context.Context, artifactName string, dest 
 	if err != nil {
 		return errors.Wrapf(err, "parse artifact name %s", artifactName)
 	}
-	mts, err := c.buildTarget(ctx, artifact.Target.String(), platform, buildArgs, false, false)
+	mts, err := c.buildTarget(ctx, artifact.Target.String(), platform, allowPrivileged, buildArgs, false, false)
 	if err != nil {
 		return errors.Wrapf(err, "apply build %s", artifact.Target.String())
 	}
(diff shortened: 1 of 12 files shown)


earthfile2llb/interpreter.go
```diff
@@ -493,6 +497,10 @@ func (i *Interpreter) handleRun(ctx context.Context, cmd spec.Command) error {
 		return nil
 	}
 
+	if i.target.IsRemote() && *privileged {
+		return i.errorf(cmd.SourceLocation, "Permission denied: unwilling to run privileged command from remote Earthfile")
+	}
+
```
vladaionescu, line 503: This looks good, but I wouldn't make this change right now as it might break people's builds and have no way to get them back. First we need a setting to allow this behavior: "i know what I'm doing" kind of thing.

Perhaps we can do this in a later PR and only focus on `LOCALLY` for now? `LOCALLY` should be fine to change in this PR because it's an experimental feature.. so we made no promises to maintain API / compatiblity.

alexcb: I should have left this as a Draft PR; I'm happy to leave it open until we have a complete story for listing trusted repos

earthfile2llb/interpreter.go
```diff
@@ -445,6 +445,10 @@ func (i *Interpreter) handleRun(ctx context.Context, cmd spec.Command) error {
 	// Note: Not expanding args for the run itself, as that will be take care of by the shell.
 
 	if i.local {
+		if i.target.IsRemote() {
+			return i.errorf(cmd.SourceLocation, "Permission denied: unwilling to run locally command from remote Earthfile")
+		}
```
vladaionescu, line 450: It might be better if this is placed in `handleLocally`. Then it'll cover any command, not just `RUN`.

earthfile2llb/interpreter.go
vladaionescu: Sounds good

earthfile2llb/interpreter.go
alexcb: oh that's a spectacular point! thanks for catching that.

earthfile2llb/earthfile2llb.go
```diff
@@ -63,6 +64,10 @@ type ConvertOpt struct {
 	// ie if there are any non-SAVE commands after the first SAVE command,
 	// or if the target is invoked via BUILD command (not COPY nor FROM).
 	HasDangling bool
+	// Console is for logging
+	Console conslogging.ConsoleLogger
```
vladaionescu, line 68: Thanks for passing this through - it was long overdue :D

earthfile2llb/interpreter.go
```diff
@@ -393,14 +399,50 @@ func (i *Interpreter) handleFrom(ctx context.Context, cmd spec.Command) error {
...
+
+	return i.getAllowPrivileged(depTarget, allowPrivileged)
+}
+
+func (i *Interpreter) getAllowPrivileged(depTarget domain.Target, allowPrivileged bool) (bool, error) {
+	if depTarget.IsRemote() {
+		return i.allowPrivileged && allowPrivileged, nil
+	}
+	if allowPrivileged {
+		i.console.Warnf("allow-privileged has no effect when refencing a local or relative target")
```
vladaionescu, line 432:
```suggestion
		i.console.Warnf("allow-privileged has no effect when referencing a local or relative target\n")
```

vladaionescu: I would maybe just use `Printf` so it doesn't show up as red?

vladaionescu (commented):
Looks good so far

alexcb:
This has one test commented out which I'm hoping to fix in a follow up PR, to deal with this non-related bug:

```
FROM alpine:latest
foo:
    LOCALLY
    RUN whoami
bar:
    FROM +foo
    RUN echo should this also be locally?
```
where `earthly +bar` fails.

earthfile2llb/converter.go
```diff
@@ -769,18 +769,18 @@ func (c *Converter) SaveImage(ctx context.Context, imageNames []string, pushImag
...
 	}
 	c.nonSaveCommand()
-	_, err = c.buildTarget(ctx, fullTargetName, platform, buildArgs, true, false)
+	_, err = c.buildTarget(ctx, fullTargetName, platform, allowPrivileged, buildArgs, true, false)
 	return err
 }
 
 // BuildAsync applies the earthly BUILD command asynchronously.
-func (c *Converter) BuildAsync(ctx context.Context, fullTargetName string, platform *specs.Platform, buildArgs []string) chan error {
+func (c *Converter) BuildAsync(ctx context.Context, fullTargetName string, platform *specs.Platform, allowPrivileged bool, buildArgs []string) chan error {
```
vladaionescu, line 783: Forgot to use `allowPrivileged` here?

earthfile2llb/converter.go
```diff
@@ -1129,11 +1129,12 @@ func (c *Converter) prepBuildTarget(ctx context.Context, fullTargetName string,
 	return target, opt, propagateBuildArgs, nil
 }
 
-func (c *Converter) buildTarget(ctx context.Context, fullTargetName string, platform *specs.Platform, buildArgs []string, isDangling bool, isFrom bool) (*states.MultiTarget, error) {
+func (c *Converter) buildTarget(ctx context.Context, fullTargetName string, platform *specs.Platform, allowPrivileged bool, buildArgs []string, isDangling bool, isFrom bool) (*states.MultiTarget, error) {
 	target, opt, propagateBuildArgs, err := c.prepBuildTarget(ctx, fullTargetName, platform, buildArgs, isDangling, isFrom)
 	if err != nil {
 		return nil, err
 	}
+	opt.AllowPrivileged = allowPrivileged
```
vladaionescu, line 1137: This should go in `prepBuildTarget` instead. This will cover `BuildAsync` too.

earthfile2llb/earthfile2llb.go
```diff
@@ -63,6 +64,10 @@ type ConvertOpt struct {
...
+
+	AllowPrivileged bool
```
vladaionescu, line 70: Mind adding a comment for this one?

earthfile2llb/interpreter.go
```diff
@@ -600,6 +650,7 @@ func (i *Interpreter) handleCopy(ctx context.Context, cmd spec.Command) error {
 	keepTs := fs.Bool("keep-ts", false, "Keep created time file timestamps")
 	keepOwn := fs.Bool("keep-own", false, "Keep owner info")
 	ifExists := fs.Bool("if-exists", false, "Do not fail if the artifact does not exist")
+	allowPrivilegedFlag := fs.Bool("allow-privileged", false, "allow targets to assume privileged mode")
```
vladaionescu, line 653:
```suggestion
	allowPrivilegedFlag := fs.Bool("allow-privileged", false, "Allow targets to assume privileged mode")
```

earthfile2llb/withdockerrun.go
```diff
@@ -287,7 +287,7 @@ func (wdr *withDockerRun) load(ctx context.Context, opt DockerLoadOpt) error {
 	if err != nil {
 		return errors.Wrapf(err, "parse target %s", opt.Target)
 	}
-	mts, err := wdr.c.buildTarget(ctx, depTarget.String(), opt.Platform, opt.BuildArgs, false, false)
+	mts, err := wdr.c.buildTarget(ctx, depTarget.String(), opt.Platform, true, opt.BuildArgs, false, false)
```
vladaionescu, line 290: This doesn't seem like an obvious choice. For example, if putting together a set of images from a bunch of remote repos, usually you don't need the privileged mode just for building the image. I would make this required for the user to specify. e.g. `WITH DOCKER --allow-privileged --load=+target`.

earthfile2llb/withdockerrunlocal.go
```diff
@@ -43,7 +43,7 @@ func (wdrl *withDockerRunLocal) load(ctx context.Context, opt DockerLoadOpt) (st
 	if err != nil {
 		return "", errors.Wrapf(err, "parse target %s", opt.Target)
 	}
-	mts, err := wdrl.c.buildTarget(ctx, depTarget.String(), opt.Platform, opt.BuildArgs, false, false)
+	mts, err := wdrl.c.buildTarget(ctx, depTarget.String(), opt.Platform, true, opt.BuildArgs, false, false)
```
vladaionescu, line 46: Similarly here.

vladaionescu (commented):
Looks great!

I found a couple of remaining edge cases.

earthfile2llb/converter.go
alexcb: Good catch, that matches up nicely with the unused allowPriviledged in BuildAsync.

earthfile2llb/withdockerrun.go
alexcb: I was thinking that the only way this code would be hit would be after it went through https://github.com/earthly/earthly/blob/main/earthfile2llb/interpreter.go#L431-L433 which would fail if privileged wasn't available.

Now that you mention it, this does seem wrong. What about a case where there's two different remote targets to load, but you only want to give one privileged access? e.g.

```
WITH DOCKER --load=(github.com/earthly/foo+target --allow-privileged ) --load=github.com/other/repo+bar
```

maybe we need to add a allowPrivileged bool to DockerLoadOpt too?

vladaionescu: This would overlap with the new style build args unfortunately.

I wouldn't worry about it, because there are workarounds if the user wants finer granularity:

```
foo-target:
  FROM --allow-privileged github.com/earthly/foo+target

...

WITH DOCKER \
        --load=+foo-target \
        --load=github.com/other/repo+bar
```

Or

```
IMPORT --allow-privileged github.com/earthly/foo

...

WITH DOCKER \
        --load=foo+target \
        --load=github.com/other/repo+bar
```

alexcb: Those are very reasonable workaround, and good catch on this, I've added an AllowPrivileged bool to the DockerLoadOpt struct now.

earthfile2llb/withdockerrunlocal.go
alexcb: fixed :)

earthfile2llb/interpreter.go
```diff
@@ -801,6 +858,7 @@ func (i *Interpreter) handleBuild(ctx context.Context, cmd spec.Command, async b
 	fs.Var(platformsStr, "platform", "The platform to build")
 	buildArgs := new(StringSliceFlag)
 	fs.Var(buildArgs, "build-arg", "A build arg override passed on to a referenced Earthly target")
+	allowPrivilegedFlag := fs.Bool("allow-privileged", false, "allow targets to assume privileged mode")
```
vladaionescu, line 861:
```suggestion
	allowPrivilegedFlag := fs.Bool("allow-privileged", false, "Allow targets to assume privileged mode")
```

earthfile2llb/interpreter.go
```diff
@@ -600,6 +650,7 @@ func (i *Interpreter) handleCopy(ctx context.Context, cmd spec.Command) error {
 	keepTs := fs.Bool("keep-ts", false, "Keep created time file timestamps")
 	keepOwn := fs.Bool("keep-own", false, "Keep owner info")
 	ifExists := fs.Bool("if-exists", false, "Do not fail if the artifact does not exist")
+	allowPrivilegedFlag := fs.Bool("allow-privileged", false, "Allow targets to assume privileged mode")
```
vladaionescu, line 653: I wonder if this works as a workaround for not supporting short-form flags?

```suggestion
	allowPrivilegedFlag1 := fs.Bool("allow-privileged", false, "Allow targets to assume privileged mode")
	allowPrivilegedFlag2 := fs.Bool("P", false, "Allow targets to assume privileged mode")
	allowPrivilegedFlag := allowPrivilegedFlag1 || allowPrivilegedFlag2
```

vladaionescu (approved):
One last thought

earthfile2llb/interpreter.go
alexcb: I was thinking about that, but figure it's best to just swap out the lib. It would be a bit funny to allow `-allow-privileged` and `--P` if we did this change.

vladaionescu: Yeah, I agree we should swap out the lib eventually. I was thinking more like a stop-gap solution - despite the weird `--P` 😄

PR #935, opened 2021-04-19
Allow priviledged flag for imports

alexcb:


domain/importtracker.go
```diff
@@ -50,12 +60,17 @@ func (ir *ImportTracker) Add(importStr string, as string, global bool) error {
...
+	allowPrivileged := currentlyPrivileged
 	if parsedImport.IsImportReference() {
 		return errors.Errorf("IMPORT %s not supported", importStr)
 	} else if parsedImport.IsRemote() {
 		path = parsedImport.GetGitURL()
+		allowPrivileged = allowPrivileged && allowPrivilegedFlag
 	} else if parsedImport.IsLocalExternal() {
 		path = parsedImport.GetLocalPath()
+		if allowPrivilegedFlag {
+			ir.console.Printf("the --allow-privileged flag has no effect when referencing a local or relative target\n")
```
vladaionescu, line 72: Nit: Relative targets are also local (internal vs external is the naming I've used before for same earthfile vs different earthfile)

```suggestion
			ir.console.Printf("the --allow-privileged flag has no effect when referencing a local target\n")
```

domain/importtracker.go
```diff
@@ -82,34 +97,40 @@ func (ir *ImportTracker) Add(importStr string, as string, global bool) error {
...
+			fullPath:        importStr,
+			allowPrivileged: allowPrivileged,
+		}
 	}
 	return nil
 }
 
 // Deref resolves the import (if any) and returns a reference with the full path.
-func (ir *ImportTracker) Deref(ref Reference) (Reference, error) {
+func (ir *ImportTracker) Deref(ref Reference) (Reference, bool, bool, error) {
```
vladaionescu, line 118: Might be worth naming the outputs, in this case, to help with readability.

earthfile2llb/converter.go
```diff
@@ -1016,17 +1016,17 @@ func (c *Converter) Healthcheck(ctx context.Context, isNone bool, cmdArgs []stri
...
 	err := c.checkAllowed("IMPORT")
 	if err != nil {
 		return err
 	}
-	return c.varCollection.Imports().Add(importStr, as, isGlobal)
+	return c.varCollection.Imports().Add(importStr, as, isGlobal, currentlyPrivileged, allowPrivilegedFlag)
 }
 
-// ResolveReference resolves a reference's build context given the current state: relativity to the Earthfile, imports etc.
+// ResolveReference resolves a reference's build context given the current state: relativity to the Earthfile, imports etc. NOTE: this function is used to support UDFs (user defined functions), and does not make use of the allowPrivileged values returned by Deref.
```
vladaionescu, line 1027:
```suggestion
// ResolveReference resolves a reference's build context given the current state: relativity to the Earthfile, imports etc. NOTE: this function is used to support UDCs (user defined commands), and does not make use of the allowPrivileged values returned by Deref.
```

earthfile2llb/converter.go
```diff
@@ -1016,17 +1016,17 @@ func (c *Converter) Healthcheck(ctx context.Context, isNone bool, cmdArgs []stri
...
 func (c *Converter) ResolveReference(ctx context.Context, ref domain.Reference) (*buildcontext.Data, error) {
-	derefed, err := c.varCollection.Imports().Deref(ref)
+	derefed, _, _, err := c.varCollection.Imports().Deref(ref)
```
vladaionescu, line 1029: Wouldn't this information be needed for UDCs too?

domain/importtracker.go
alexcb: good point :)

earthfile2llb/converter.go
alexcb: very good point!

earthfile2llb/converter.go
alexcb: I was on the fence about this one, because even without it, it could be abused:

```
# github.com/scary/repo
UNSAFE:
    COMMAND
    RUN rm -rf /
    
# local Earthfile
foo:
    LOCALLY
    DO github.com/scary/repo+UNSAFE
```

since the unsafe command doesn't require priviledged mode, it could be pulled into a locally target and could do a fair amount of damage; I was worried that adding a `--allow-priviledged` flag could give users an unsafe reassurance that it would protect them in this case. If we did extend it for this, we could have it simply block `RUN --priviledged` calls, or calls to other LOCALLY targets.

maybe it's still good to extend the proection here for users who want to use UDCs within a containerized target?

vladaionescu: I would do it still, because even in your example, the user had to specify LOCALLY to get into trouble first

alexcb: I just submitted a follow up commit with an implementation for this.

vladaionescu (approved):

PR #937, opened 2021-04-20
Update tutorial based on user feedback

vladaionescu:
This fixes a number of user feedback items regarding the tutorial, including:

* missing `go.sum` and `package-lock.json` files
* awkward `// package.json` comment at the beginning of the file
* user having trouble to copy-paste examples (there's now a one-liner that user can run from terminal to copy material over)
* java example wasn't working (missing build.gradle file)
* `index.html` file was used from `dist`, which is an output dir
* outputting `go.sum` too early in the tutorial, confusing the user
* merging pages 1 and 2, as the user wasn't sure how to verify their Earthfile while they were on page 1 and were feeling uneasy (answer was on page 2, but before the page turn, user was scratching head)
* diagram explaining the structure of the target reference was presented to early in the tutorial

In addition, each part is now tested individually for each language as part of a new set of tests.


docs/basics/part-1-a-simple-earthfile.md
```diff
@@ -94,9 +121,9 @@ docker:
 
 The code of the app might look like this
 
-```java
-// src/main/java/hello/HelloWorld.java
+`./src/main/java/hello/HelloWorld.java`
```
alexcb, line 124: nice change.

alexcb (approved):
looks great!

It might be a long-shot, but is there any way we can dedup the contents of the examples/tutorial code with the docs? or somehow use grep to ensure the contents of all the files under the examples/tutorial appear under the docs to protect us against code-drift? (not necessarily a suggestion for this PR, but down the road?)

vladaionescu:
> It might be a long-shot, but is there any way we can dedup the contents of the examples/tutorial code with the docs? or somehow use grep to ensure the contents of all the files under the examples/tutorial appear under the docs to protect us against code-drift? (not necessarily a suggestion for this PR, but down the road?)

I was meaning to do this but couldn't find an easy way. If there's a unix tool that easily finds the contents of one file within another file, it would be very easy to add such a check.

alexcb:
> If there's a unix tool

I had a few tries with a combination of find and grep -Z but couldn't come up with a working example. Let's punt on it then.

PR #939, opened 2021-04-20
Tutorial: Links to tutorial files. Remove note box which causes copy-paste issues.

vladaionescu:


alexcb (approved):

vladaionescu:
Seems that the info box is not the cause of the copy-paste bug. Doing only the links in this PR.

PR #940, opened 2021-04-22
move interpreter flags to separate file

alexcb:
- move interpreter flags to separate file for clearer visibility of flags
- switch from go's flag library to go-flags which supports both short and long names

### DIFF
--- a/earthfile2llb/flags.go
+++ b/earthfile2llb/flags.go
@@ -1,0 +1,120 @@
+package earthfile2llb
+
+import (
+	"fmt"
+	"os"
+	"time"
+
+	"github.com/pkg/errors"
+
+	flags "github.com/jessevdk/go-flags"
+)
+
+type ifOpts struct {
+	Privileged bool     `long:"privileged" description:"Enable privileged mode"`
+	WithSSH    bool     `long:"ssh" description:"Make available the SSH agent of the host"`
+	NoCache    bool     `long:"no-cache" description:"Always run this specific item, ignoring cache"`
+	Secrets    []string `long:"secret" description:"Make available a secret"`
+	Mounts     []string `long:"mount" description:"Mount a file or directory"`
+}
+
+type runOpts struct {
+	Push            bool     `long:"push" description:"Execute this command only if the build succeeds and also if earthly is invoked in push mode"`
+	Privileged      bool     `long:"privileged" description:"Enable privileged mode"`
+	WithEntrypoint  bool     `long:"entrypoint" description:"Include the entrypoint of the image when running the command"`
+	WithDocker      bool     `long:"with-docker" description:"Deprecated"`
+	WithSSH         bool     `long:"ssh" description:"Make available the SSH agent of the host"`
+	NoCache         bool     `long:"no-cache" description:"Always run this specific item, ignoring cache"`
+	Interactive     bool     `long:"interactive" description:"Run this command with an interactive session, without saving changes"`
+	InteractiveKeep bool     `long:"interactive-keep" description:"Run this command with an interactive session, saving changes"`
+	Secrets         []string `long:"secret" description:"Make available a secret"`
+	Mounts          []string `long:"mount" description:"Mount a file or directory"`
+}
+
+type fromOpts struct {
+	AllowPrivileged bool     `long:"allow-privileged" description:"Allow commands under remote targets to enable privileged mode"`
+	BuildArgs       []string `long:"build-arg" description:"A build arg override passed on to a referenced Earthly target"`
+	Platform        string   `long:"platform" description:"The platform to use"`
+}
+
+type fromDockerfileOpts struct {
+	BuildArgs []string `long:"build-arg" description:"A build arg override passed on to a referenced Earthly target and also to the Dockerfile build"`
+	Platform  string   `long:"platform" description:"The platform to use"`
+	Target    string   `long:"target" description:"The Dockerfile target to inherit from"`
+	Path      string   `short:"f" description:"Not supported"`
+}
+
+type copyOpts struct {
+	From            string   `long:"from" description:"Not supported"`
+	IsDirCopy       bool     `long:"dir" description:"Copy entire directories, not just the contents"`
+	Chown           string   `long:"chown" description:"Apply a specific group and/or owner to the copied files and directories"`
+	KeepTs          bool     `long:"keep-ts" description:"Keep created time file timestamps"`
+	KeepOwn         bool     `long:"keep-own" description:"Keep owner info"`
+	IfExists        bool     `long:"if-exists" description:"Do not fail if the artifact does not exist"`
+	AllowPrivileged bool     `long:"allow-privileged" description:"Allow targets to assume privileged mode"`
+	Platform        string   `long:"platform" description:"The platform to use"`
+	BuildArgs       []string `long:"build-arg" description:"A build arg override passed on to a referenced Earthly target"`
+}
+
+type saveArtifactOpts struct {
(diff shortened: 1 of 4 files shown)


alexcb:
If this draft looks good, I'll extend it to support all of our parsing in interpreter.go

earthfile2llb/flags.go
```diff
@@ -0,0 +1,22 @@
...
+	Privileged      bool     `short:"P" long:"privileged" description:"Enable privileged mode"`
+	WithEntrypoint  bool     `long:"entrypoint" description:"Include the entrypoint of the image when running the command"`
+	WithDocker      bool     `long:"with-docker" description:"Deprecated"`
+	WithSSH         bool     `long:"ssh" description:"Make available the SSH agent of the host"`
+	NoCache         bool     `long:"no-cache" description:"Always run this specific item, ignoring cache"`
+	Interactive     bool     `long:"interactive" description:"Run this command with an interactive session, without saving changes"`
+	InteractiveKeep bool     `long:"interactive-keep" description:"Run this command with an interactive session, saving changes"`
+	Secrets         []string `long:"secret" description:"Make available a secret"`
+	Mounts          []string `long:"mount" description:"Mount a file or directory"`
+}
```
vladaionescu, line 18: This looks really nice and clean!

earthfile2llb/flags.go
```diff
@@ -0,0 +1,22 @@
+package earthfile2llb
+
+import (
+	flags "github.com/jessevdk/go-flags"
+)
+
+type runOpts struct {
+	Push            bool     `long:"push" description:"Execute this command only if the build succeeds and also if earthly is invoked in push mode"`
+	Privileged      bool     `short:"P" long:"privileged" description:"Enable privileged mode"`
```
vladaionescu, line 9: The short version would not correspond to the Dockerfile equivalent. Although I can see how this matches to the short version of `--allow-privileged`.

I'm starting to think that the long version of flags in scripts is better because it helps with readability. Giving the user two options means that there will be a mixture of Earthfiles out there some using long version, some using short version and some using a mix, which makes Earthfiles less readable overall.

What do you think of not providing a short version of any flag in Earthfiles (even `--allow-privileged`)? Just long versions.

vladaionescu (commented):
I think this is a great refactor - makes things easier to read for sure.

Not sure about the short form of flags - see comment below. (I know we had talked before otherwise)

earthfile2llb/flags.go
alexcb: I'm honestly fine with that, `--allow-privileged` really stands out to a user reading the Earthfile, where as `-P` might not jump out as much as a security concern.

We could still make this switch and just keep the long version, then possibly come back to it if we get a lot of requests for a shorter version. The only real downside is users (myself included) would have to learn how to consistently spell privileged correctly.

vladaionescu: Sounds like a plan. I agree that the library switch is worthwhile nevertheless.

alexcb:
Changing libraries here will change what happens when an unknown flag is encountered; currently earthly prints this:

```
$ ./earthly -P ~/broken/allow-priv+test
...
       alpine:latest | --> Load metadata linux/amd64
flag provided but not defined: -P
Usage of RUN:
  -entrypoint
    	Include the entrypoint of the image when running the command
  -interactive
    	Run this command with an interactive session, without saving changes
  -interactive-keep
    	Run this command with an interactive session, saving changes
  -mount value
    	Mount a file or directory
  -no-cache
    	Always run this specific item, ignoring cache
  -privileged
    	Enable privileged mode
  -push
    	Execute this command only if the build succeeds and also if earthly is invoked in push mode
  -secret value
    	Make available a secret
  -ssh
    	Make available the SSH agent of the host
  -with-docker
    	Deprecated
Error: /home/alex/broken/allow-priv/Earthfile line 5:4 invalid RUN arguments [-P cat /proc/self/status | grep CapEff | grep 0000003fffffffff]: flag provided but not defined: -P
in		/home/alex/broken/allow-priv+test
```

However with changing the library, the new one doesn't print the available flags by default, but I can catch the error and call printhelp before bubbling up the error which would then look like this:

```
$ ./build/linux/amd64/earthly ~/test/allow-priv+test
...
       alpine:latest | --> Load metadata linux/amd64
unknown flag `P'

RUN [options] args:
  --push                Execute this command only if the build succeeds and also if earthly is invoked in push mode
  --privileged          Enable privileged mode
  --entrypoint          Include the entrypoint of the image when running the command
  --with-docker         Deprecated
  --ssh                 Make available the SSH agent of the host
  --no-cache            Always run this specific item, ignoring cache
  --interactive         Run this command with an interactive session, without saving changes
  --interactive-keep    Run this command with an interactive session, saving changes
  --secret=             Make available a secret
  --mount=              Mount a file or directory
Error: /home/alex/broken/allow-priv/Earthfile line 5:4 invalid RUN arguments [-P cat /proc/self/status | grep CapEff | grep 0000003fffffffff]: unknown flag `P'
in		/home/alex/broken/allow-priv+test

```

earthfile2llb/flags.go
```diff
@@ -0,0 +1,120 @@
...
+	Mounts          []string `long:"mount" description:"Mount a file or directory"`
+}
+
+type fromOpts struct {
+	AllowPrivileged bool     `long:"allow-privileged" description:"Allow commands under remote targets to enable privileged mode"`
+	BuildArgs       []string `long:"build-arg" description:"A build arg override passed on to a referenced Earthly target"`
+	Platform        string   `long:"platform" description:"The platform to use"`
+}
+
+type fromDockerOpts struct {
```
vladaionescu, line 40:
```suggestion
type fromDockerfileOpts struct {
```

earthfile2llb/interpreter.go
```diff
@@ -369,38 +359,34 @@ func (i *Interpreter) handleFrom(ctx context.Context, cmd spec.Command) error {
...
 		return i.pushOnlyErr(cmd.SourceLocation)
 	}
-	fs := flag.NewFlagSet("FROM", flag.ContinueOnError)
-	allowPrivilegedFlag := fs.Bool("allow-privileged", false, "Enable privileged mode")
-	buildArgs := new(StringSliceFlag)
-	fs.Var(buildArgs, "build-arg", "A build arg override passed on to a referenced Earthly target")
-	platformStr := fs.String("platform", "", "The platform to use")
-	err := fs.Parse(getArgsCopy(cmd))
+	opts := fromOpts{}
+	args, err := parseArgs("RUN", &opts, getArgsCopy(cmd))
```
vladaionescu, line 363: Should this be `FROM`?

```suggestion
	args, err := parseArgs("FROM", &opts, getArgsCopy(cmd))
```

earthfile2llb/interpreter.go
```diff
@@ -369,38 +359,34 @@ func (i *Interpreter) handleFrom(ctx context.Context, cmd spec.Command) error {
...
 	if err != nil {
-		return i.wrapError(err, cmd.SourceLocation, "invalid FROM arguments")
+		return i.wrapError(err, cmd.SourceLocation, "invalid RUN arguments %v", cmd.Args)
```
vladaionescu, line 365:
```suggestion
		return i.wrapError(err, cmd.SourceLocation, "invalid FROM arguments %v", cmd.Args)
```

vladaionescu (commented):
This PR is a breath of fresh air 😄 🎉

earthfile2llb/interpreter.go
alexcb: good catch!

vladaionescu (approved):

PR #943, opened 2021-04-23
`sudo earthly bootstrap` should use the same `install_id` as the calling user

vladaionescu:
~Uses the magical incantation `eval echo ~$SUDO_USER` to figure out the sudo user's home dir and use that dir for `install_id`.~

TIL https://golang.org/pkg/os/user/#Lookup

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -31,7 +31,7 @@ deps:
 code:
     FROM +deps
     COPY --platform=linux/amd64 ./ast/parser+parser/*.go ./ast/parser/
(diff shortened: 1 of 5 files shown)


alexcb:
I think it's a good idea, but I think we're missing a part where we ensure the created `~/.earthly` directory is chown'ed to the `$SUDO_USER`.

vladaionescu:
Done.

The `install_id` file is still owned by root, but is readable by the user. This should be good enough.

alexcb (approved):
looks great.

PR #944, opened 2021-04-23
docs for allow-priviledged flag

alexcb:

### DIFF
--- a/docs/earthfile/earthfile.md
+++ b/docs/earthfile/earthfile.md
@@ -44,7 +44,7 @@ Each recipe contains a series of commands, which are defined below. For an intro
 #### Synopsis
 
 * `FROM <image-name>`
-* `FROM [--build-arg <key>=<value>] [--platform <platform>] <target-ref>`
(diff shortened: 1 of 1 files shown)


alexcb:
@vladaionescu how did you manage to get `https://docs.earthly.dev/docs/earthfile#privileged` to point to the first one? in my case referencing `#allow-privileged` is pointing to the second anchor.

docs/earthfile/earthfile.md
```diff
@@ -1018,6 +1066,13 @@ If an `IMPORT` is defined in the `base` target of the Earthfile, then it becomes
 
 For more information see the [target, artifact and command references guide](../guides/target-ref.md).
 
+#### Options
+
+##### `--allow-privileged`
+
+Same as [`FROM --allow-privileged`](#allow-privileged).
```
vladaionescu, line 1073: I would expand a little more here, given that this is the `IMPORT` command. We should say that it allows any mention of the alias to use privileged.

alexcb: That's a good call, how does it sound now?

vladaionescu (approved):
:+1:

PR #949, opened 2021-04-26
Update Buildkitd Settings

dchw:
This commit is a slice of some of the changes I have made to the Buildkit settings struct. More will follow, I just made the whole ball of mud too big :(

This makes the settings struct the place for all the buildkit arguments, plus removes the homegrown hashing we were doing and replaces it with `hashstructure`. This swap clarifies which settings contribute to restarting buildkit as well.

Additionally, it pushes some details that leaked into the main `earthly` command into buildkit.


buildkitd/settings.go
```diff
@@ -1,57 +1,47 @@
...
-	AdditionalConfig string   `json:"additionalConfig"`
-	CniMtu           uint16   `json:"cnuMtu"`
+	CacheSizeMb      int
+	GitURLInsteadOf  string
+	Debug            bool
+	DebuggerPort     int
+	AdditionalArgs   []string
+	AdditionalConfig string
+	CniMtu           uint16
+	Timeout          time.Duration `hash:"ignore"`
```
alexcb, line 21: I like the `hash:"ignore"`

alexcb (approved):

vladaionescu (approved):
Neat!

buildkitd/settings.go
dchw: Yes, me too. It makes life better. Fun fact: one of our deps already used this, so no real harm!

Issue #619, opened 2020-12-14
Support FROM DOCKERFILE -f


vladaionescu:
@rulatir requests support for -f in https://github.com/earthly/earthly/issues/615#issuecomment-743977844

* Should be able to specify the Dockerfile to use.
* `-f` should work the same as the `FROM DOCKERFILE` path: it can either be a path within the build context or be an artifact path.

n-rodriguez:
hi there! any news?

vladaionescu:
I'll add this to the roadmap as this has been requested quite a bit.

vladaionescu:
This will go out with the next release.

The syntax is

```
FROM DOCKERFILE -f <local-dockerfile-path-relative-to-current-earthfile> <local-build-context-path-relative-to-current-earthfile>
```

or

```
FROM DOCKERFILE -f +target/path/to/some.Dockerfile +another-target/path/to/build/context
```

Or any combination of the above.

vladaionescu:
This has now been released.

n-rodriguez:
> This has now been released.

awesome!

PR #951, opened 2021-04-27
Docs for `FROM DOCKERFILE -f`

vladaionescu:
Re https://github.com/earthly/earthly/issues/619

### DIFF
--- a/docs/earthfile/earthfile.md
+++ b/docs/earthfile/earthfile.md
@@ -731,13 +731,13 @@ Sets the number of retries before a container is considered `unhealthy`. Default
 
 #### Synopsis
 
-* `FROM DOCKERFILE [--build-arg <key>=<value>] [--platform <platform>] [--target <target-name>] <context-path>`
+* `FROM DOCKERFILE [options...] <context-path>`
 
 #### Description
 
 The `FROM DOCKERFILE` command initializes a new build environment, inheriting from an existing Dockerfile. This allows the use of Dockerfiles in Earthly builds.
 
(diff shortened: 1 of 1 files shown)


docs/earthfile/earthfile.md
```diff
@@ -751,6 +751,10 @@ This feature is currently in **Beta** and it has the following limitations:
 
 #### Options
 
+##### `-f <dockerfile-path>`
+
+Specify an alternative Dockerfile to use. The `<dockerfile-path>` can be either a path on the host system, relative to the current Earthfile, or an artifact reference pointing to the `Dockerfile`.
```
dchw, line 756:
```suggestion
Specify an alternative Dockerfile to use. The `<dockerfile-path>` can be either a path on the host system, relative to the current Earthfile, or an artifact reference pointing to a `Dockerfile`.
```

dchw: So a Dockerfile as an artifact must be named +some-target/Dockerfile? At least, thats how I initially parsed this sentence. Assuming that i s not true, I would probably phrase it like this

dchw: Side note: I went to add a link to what an "artifact reference" is. The closest I found is the description in `COPY`. This seems like a fundamental term I should find a better definition for when I search "artifact reference" over our docs.

dchw (approved):
Small nit if I understood right

docs/earthfile/earthfile.md
vladaionescu: It doesn't have to be named `Dockerfile` - good catch.

Ah right.. it's in `target-ref.md`. I'll add a link to that too.

PR #952, opened 2021-04-27
More Buildkit Cleaning

dchw:
Pull in more details that have leaked into the main `earthly` command. Additionally, use the same retry logic when a provided buildkit address is used.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -94,7 +94,7 @@ jobs:
       - name: Execute earthly docker command
         run: (cd examples/tests/docker && ../../.././build/linux/amd64/earthly docker --tag examples-test-docker:latest && diff <(docker run --rm examples-test-docker:latest) <(echo "hello dockerfile") )
       - name: Execute test similar to homebrew test in https://github.com/Homebrew/homebrew-core/blob/master/Formula/earthly.rb
-        run: "./build/linux/amd64/earthly --buildkit-host 127.0.0.1 ./examples/tests/with-docker+all 2>&1 | grep 'connect provided buildkit: timeout'"
+        run: "./build/linux/amd64/earthly --buildkit-host 127.0.0.1 ./examples/tests/with-docker+all 2>&1 | grep 'buildkitd failed to start'"
       - name: Execute private image test (non fork only)
         run: ./build/linux/amd64/earthly --ci ./examples/tests+private-image-test
         if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name == github.repository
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -94,7 +94,7 @@ jobs:
       - name: Execute earthly docker command
         run: (cd examples/tests/docker && ../../.././build/linux/amd64/earthly docker --tag examples-test-docker:latest && diff <(docker run --rm examples-test-docker:latest) <(echo "hello dockerfile") )
       - name: Execute test similar to homebrew test in https://github.com/Homebrew/homebrew-core/blob/master/Formula/earthly.rb
-        run: ./build/linux/amd64/earthly --buildkit-host 127.0.0.1 ./examples/tests/with-docker+all 2>&1 | grep 'Error while dialing invalid address 127.0.0.1'
+        run: "./build/linux/amd64/earthly --buildkit-host 127.0.0.1 ./examples/tests/with-docker+all 2>&1 | grep 'connect provided buildkit: timeout'"
```
dchw, line 97: Is there an equivalent change I will need to make for our homebrew stuff? I am unfamiliar with our packaging there because its automated away so nicely.

alexcb: This is just a silly test that homebrew wanted to validate the earthly binary could be executed. We explained that earthly requires buildkit to run (which isn't available when this homebrew test is run). This was the work-around the homebrew maintainers agreed upon.

Will this PR change the error message that's printed when earthly runs and can't connect to buildkit? (e.g. if you change 127.0.0.1 to a different IP that times out, does it produce the same message as what's expected here?) If this gets broken then we'll need to change this https://github.com/Homebrew/homebrew-core/blob/master/Formula/earthly.rb#L46-L47

buildkitd/buildkitd.go
```diff
@@ -57,6 +72,11 @@ func NewClient(ctx context.Context, console conslogging.ConsoleLogger, image str
 
 // ResetCache restarts the buildkitd daemon with the reset command.
 func ResetCache(ctx context.Context, console conslogging.ConsoleLogger, image string, settings Settings) error {
+	// Prune by resetting container.
+	if settings.BuildkitHost != "" {
+		return errors.New("Cannot reset cache of a provided buildkit-host setting")
```
vladaionescu, line 77: Nit: consistency

```suggestion
		return errors.New("cannot reset cache of a provided buildkit-host setting")
```

vladaionescu (commented):
Looks good.

Alex's comment is the only thing that needs attention.

.github/workflows/ci.yml
dchw: Any externally provided host now prints a different message, because it shares the same retry logic now. I did this because in our all-in-one earthly image, its easier to start buildkit and earthly concurrently and let earthly just use it when it becomes avaliable. It also helps with things like transient failures in a distributed environment.

alexcb: I'm fine if you want to change the error message, we'll just have to remember to update the homebrew test when we release the next version.

dchw: I changed the message we check for to `buildkitd failed to start`, and added a note to the manual change we need to make during the next release.

alexcb (approved):

vladaionescu (approved):

PR #954, opened 2021-04-27
Add a release step to check broken links in our docs

vladaionescu:


alexcb (approved):
I just gave it a try, much faster than I expected.

PR #960, opened 2021-04-28
Attempt git ssh or fallback to https v2

alexcb:


buildcontext/gitlookup.go
```diff
@@ -155,6 +188,47 @@ func (gl *GitLookup) getGitMatcher(path string) (string, *gitMatcher, error) {
 	return "", nil, ErrNoMatch
 }
 
+// detectProtocol will update the gitMatcher protocol if it is set to auto
+func (gl *GitLookup) detectProtocol(host string) (protocol string, err error) {
```
dchw, line 192: This might be a nit; but the only thing I would do in addition here is maybe enumerate the protocols to make it clearer down the line. Like:

```
type gitProtocol string

const (
	auto gitProtocol = "auto"
	ssh gitProtocol = "ssh"
)
```

This would ensure that git protocols stay on the rails? At the very least I would move them into a constants list and name them, the type just helps them only get used where they should be used.

dchw (approved):

buildcontext/gitlookup.go
alexcb: that's a fair point, i'll make a follow up PR for that. (I have some other git-related changes I'm currently working on which would cause a bit more rebasing issues).

vladaionescu:
Any chance that we could also write a test for this?

alexcb:
A lot of it is already covered under https://github.com/earthly/earthly/blob/main/scripts/tests/secrets-integration.sh @vladaionescu , but I can create a new similar test that additionally checks for a case where the ssh agent is running and has an unauthorized key loaded.

Issue #962, opened 2021-04-30
Create test for auto SSH


alexcb:
create a new test similar to https://github.com/earthly/earthly/blob/main/scripts/tests/secrets-integration.sh that covers more edge cases.

e.g. the ssh agent is running and has an unauthorized key loaded.

_Originally posted by @alexcb in https://github.com/earthly/earthly/issues/960#issuecomment-830186148_

alexcb:
A number of tests for authenticating with (and without) ssh have been created under https://github.com/earthly/earthly/tree/main/scripts/tests/auth

PR #963, opened 2021-04-30
test that unauthorized keys dont affect public access

alexcb:
- expand test to cover case where an unauthorized key is also loaded


alexcb:
fixes #962

vladaionescu (approved):

Issue #708, opened 2021-01-13
GIT CLONE command not working with private repos


alexcb:
Consider an Earthfile like:

```
WORKDIR /environ

private:
    GIT CLONE \
        --branch=master \
        gitlab.com/alexcb/privproj .

public:
    GIT CLONE \
        --branch=main \
        github.com/earthly/earthly .
```

I can run `earthly +public` without any issues; however when I try to run it against my private repo with `earthly +private`, it fails with:

```
$ earthly-v0.4.4 --verbose --debug +private
           buildkitd | Found buildkit daemon as docker container (earthly-buildkitd)
            +private | gitlab.com/alexcb/privproj
            +private | --> GIT CLONE (--branch master) gitlab.com/alexcb/privproj
               +base | --> WORKDIR /environ
               +base | Completed in 31.858µs
            +private | fatal: could not read Username for 'https://gitlab.com': terminal prompts disabled
            +private | WARN: (GIT CLONE (--branch master) gitlab.com/alexcb/privproj) failed to fetch remote https://gitlab.com/alexcb/privproj: exit status 128
            +private | Completed in 798.444207ms
Summary of timing information
Note that the times do not include the expansion of commands like BUILD, FROM, COPY (artifact).
            +private | (gitlab.com/alexcb/privproj) 798.444207ms
               +base | () 31.858µs
===============================================================
Total       	798.476065ms
Total (real)	929.851393ms
Error: build target: build main: bkClient.Build: failed to solve: rpc error: code = Unknown desc = failed to load cache key: failed to fetch remote https://gitlab.com/alexcb/privproj: exit status 128
Check your git auth settings.
Did you ssh-add today? Need to configure ~/.earthly/config.yml?
For more information see https://docs.earthly.dev/guides/auth
```

This is with  `~/.earthly/config.yml`:

```
global:
  cache_size_mb: 40000
git:
  gitlab.com:
    auth: https
    user: alexcb
    password: "secret"
```

however the auth works when I reference the git repo on the command line via: `earthly gitlab.com/alexcb/privproj+docker`

vladaionescu:
I see two issues here:
* `gitlab.com/alexcb/privproj` doesn't look like a valid git URL
* I don't think we should be using the git auth configured in Earthly for the `GIT CLONE` command. But perhaps there are good reasons to support this.

alexcb:
I changed my earthfile to:

```
WORKDIR /environ

private:
    GIT CLONE \
        --branch=master \
[...]

Issue #928, opened 2021-04-15
failed to fetch remote git@github.com:earthly/hello-world.git: exit status 128


alexcb:
earthly will attempt to use git ssh keys when accessing github if it 1) detects SSH_AUTH_SOCK is set, and 2) one or more ssh keys are loaded. However this will fail if none of the loaded keys are able to auth with github.

We should have it fall back to http when ssh fails instead of using our current heuristic.

steps to reproduce:

```
alex@mah:~$ earthly github.com/earthly/hello-world:main+hello
           buildkitd | Found buildkit daemon as docker container (earthly-buildkitd)
No ssh auth socket detected or zero keys loaded; falling back to https for auto auth values
      busybox:1.32.0 | --> Load metadata linux/amd64
g/e/hello-world:main+base | --> FROM busybox:1.32.0
g/e/hello-world:main+base | [██████████] resolve docker.io/library/busybox:1.32.0@sha256:bde48e1751173b709090c2539fdf12d6ba64e88ec7a4301591227ce925f3c678 ... 100%
g/e/hello-world:main+hello | *cached* --> RUN echo 'Hello, world!'
              output | --> exporting outputs
              output | [██████████] sending tarballs ... 100%
=================================== SUCCESS ====================================
alex@mah:~$ earthly --no-cache github.com/earthly/hello-world:main+hello
           buildkitd | Found buildkit daemon as docker container (earthly-buildkitd)
No ssh auth socket detected or zero keys loaded; falling back to https for auto auth values
      busybox:1.32.0 | --> Load metadata linux/amd64
g/e/hello-world:main+base | --> FROM busybox:1.32.0
g/e/hello-world:main+base | [██████████] resolve docker.io/library/busybox:1.32.0@sha256:bde48e1751173b709090c2539fdf12d6ba64e88ec7a4301591227ce925f3c678 ... 100%
g/e/hello-world:main+hello | --> RUN echo 'Hello, world!'
g/e/hello-world:main+hello | Hello, world!
              output | --> exporting outputs
              output | [██████████] sending tarballs ... 100%
=================================== SUCCESS ====================================
alex@mah:~$ echo $SSH_AUTH_SOCK
/tmp/ssh-pNVodb61zjAI/agent.2687262

alex@mah:~$ ssh-keygen 
Generating public/private rsa key pair.
Enter file in which to save the key (/home/alex/.ssh/id_rsa): /tmp/bogus_rsa
Enter passphrase (empty for no passphrase): 
[...]

PR #964, opened 2021-04-30
Git clone should use configured auth

alexcb:
- make use of ~/.netrc credentials when no config is set under `~/.earthly/config.yml`
- make use of auth credentials when performing a GIT CLONE command within an Earthfile.

This fixes https://github.com/earthly/earthly/issues/708 and https://github.com/earthly/earthly/issues/928


buildcontext/gitlookup.go
```diff
@@ -290,6 +324,67 @@ func (gl *GitLookup) GetCloneURL(path string) (string, string, string, error) {
 	return gitURL, subPath, keyScan, nil
 }
 
+// ConvertCloneURL takes a url such as https://github.com/user/repo.git or git@github.com:user/repo.git
+// and makes use of configured git credentials and protocol preferences to convert it into the appropriate
+// https or ssh protocol.
+// it also returns a keyScan
+func (gl *GitLookup) ConvertCloneURL(inURL string) (string, string, error) {
```
alexcb, line 331: TODO: consolidate the duplicate code from GetCloneURL and ConvertCloneURL

vladaionescu (approved):

PR #965, opened 2021-05-03
Refactor auth integration tests

alexcb:
- split up different components of the auth tests into seperate files.

- each file under scripts/test/auth/test-*.sh can be automatically added
  to the github workflow by running:

    earthly ./scripts/tests/auth+generate-github-tasks


Earthfile
```diff
@@ -47,15 +47,31 @@ update-buildkit:
...
+lint-scripts-misc:
+    FROM +lint-scripts-base
     COPY ./earthly ./scripts/install-all-versions.sh ./buildkitd/entrypoint.sh ./earthly-buildkitd-wrapper.sh \
         ./buildkitd/dockerd-wrapper.sh ./buildkitd/docker-auto-install.sh \
         ./release/envcredhelper.sh ./.buildkite/*.sh \
         ./scripts/tests/*.sh \
-        ./shell_scripts/
-    RUN shellcheck shell_scripts/*
+        ./
+    RUN shellcheck *.sh
```
dchw, line 62: Shellcheck is not in the ARM Alpine repos. The M1 failure did fail on shellcheck. We would need to move that out of the path or force x64 images there

scripts/tests/auth/test-private-repo-with-https-config-file.sh
```diff
@@ -0,0 +1,18 @@
...
+source "$(dirname "$0")/setup.sh"
+
+# setup earthly config
+cat << EOF > /tmp/earthconfig.https
+git:
+  github.com:
+    auth: https
+    user: cinnamonthecat
+    password: "$GITHUB_PASSWORD"
+EOF
```
dchw, line 13: Why not use `earthly config` here?

dchw (approved):

dchw:
Approved, assuming M1 tests pass :D

scripts/tests/auth/test-private-repo-with-https-config-file.sh
alexcb: This test is to cover the case where a user creates the config on their own.

There's a different test under `test-private-repo-with-https-config-command.sh`  that's more or less the same as this one, but uses `earthly config` instead.

PR #969, opened 2021-05-04
Include GHA workflow generation script for auth tests

alexcb:
- includes a script that regenerates github action integration tests
for auth related tests that can't run via earthly in earthly.

### DIFF
--- a/scripts/tests/auth/Earthfile
+++ b/scripts/tests/auth/Earthfile
@@ -1,0 +1,19 @@
+FROM alpine:latest
+
+generate-github-tasks:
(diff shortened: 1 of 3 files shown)


dchw (approved):

scripts/tests/auth/Earthfile
```diff
@@ -0,0 +1,19 @@
+FROM alpine:latest
+
+generate-github-tasks:
+    LOCALLY # this was used since COPY ../../../.github/workflows/ci.yml didn't work (perhaps because it matches ".git")
```
vladaionescu, line 4: It doesn't work because it's not part of the build context...

PR #972, opened 2021-05-07
Display which secret was not found in error message

alexcb:
Previously if one referenced a secret which was not available, earthly
would fail and simply return "Error: not found"; this changes the
behavious to display which secret was not found.

For example, consider an Earthfile like:

    foo:
        FROM alpine:latest
        RUN --secret password=+secrets/password env

If you invoke earthly +foo (without passing a value for the password secret), now it will display
an error such as:

    Error: unable to lookup secret password: not found

Where as previously it just displayed:

    Error: not found

### DIFF
--- a/examples/tests/Earthfile
+++ b/examples/tests/Earthfile
@@ -119,6 +119,11 @@ secrets-test:
         --earthfile=secrets.earth \
         --extra_args="--secret SECRET1 --secret SECRET2=bar --secret-file SECRET3=/my-secret-file" \
         --target=+test
+    DO +RUN_EARTHLY \
+        --earthfile=secrets.earth \
+        --extra_args="--secret SECRET1 --secret SECRET2=bar" \
+        --target=+test \
+        --post_command="2>&1 | perl -pe 'BEGIN {\\\$status=1} END {exit \\\$status} \\\$status=0 if /unable to lookup secret SECRET3: not found/;'"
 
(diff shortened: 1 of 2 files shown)


earthfile2llb/earthfile2llb.go
```diff
@@ -77,6 +77,10 @@ type ConvertOpt struct {
 	// Parallelism is a semaphore controlling the maximum parallelism.
 	Parallelism *semaphore.Weighted
 
+	// AvailableSecrets is a map to indicate of secrets are available within the attachables
+	// this is used to include better error messages when a secret is missing
+	AvailableSecrets map[string]interface{}
```
dchw, line 82: Why `interface{}` over `struct{}`, especially since you populate it with empty struct?

dchw (approved):

earthfile2llb/earthfile2llb.go
alexcb: good point, i'll change that.

alexcb:
@dchw my initial draft PR wasn't checking the correct secret keys, nor did it work with our hosted secrets store, it turns out displaying which secret wasn't found is easier (however we loose the bit that display the Earthfile that was referencing it). If we did want to also display the location we would have had to include it in the converter.go code instead (but that still wouldn't work with hosted secrets)).

PR #973, opened 2021-05-11
Earthfile for generating deb repo

alexcb:
- This provides targets for releasing new deb packages to our deb
repository.


release/apt-repo/Earthfile
```diff
@@ -0,0 +1,175 @@
+FROM alpine:3.13
+
+deps:
+    FROM ubuntu:20.10
+    RUN apt-get update && apt-get install -y dpkg-dev wget dpkg-sig
+
+deb:
+    ARG EARTHLY_VERSION
+    ARG EARTHLY_PLATFORM
+    ARG PKG_NAME=earthly_$EARTHLY_VERSION-1_$EARTHLY_PLATFORM
```
alexcb, line 10: I would like to change `EARTHLY_VERSION` to `RELEASE_TAG` which requires it be of the form `v1.2.3`; however I couldn't figure out how to programatically cut the first char to make `PKG_NAME` get evaluated as earthly_1.2.3-1_amd64. Is it possible to do something like `ENV PKG_NAME=earthly_$(echo "$EARTHLY_VERSION" | cut -c 2-)-1_$EARTHLY_PLATFORM`?

vladaionescu: Something like this should work:
```
ARG VERSION=$(echo "$EARTHLY_VERSION" | cut -c 2-)
ENV PKG_NAME=earthly_${VERSION}-1_${EARTHLY_PLATFORM}
```

vladaionescu (approved):
Looks good AFAIK - might be worth for @dchw to also take a look at this.

PR #974, opened 2021-05-11
Use TCP, add TLS

dchw:
Use TCP as the protocol for communicating with buildkit, to allow better future compatibility with remote and windows users.

Also adds some code to make it so bootstrap can change ownership of the `.earthly` directory.

Of note, this does _not_ have code to bootstrap if no bootstrap was detected before running. I do think I have some cleanup to do there still for existing user to make this fade into the background.

Any comments and testing is welcome and needed.


buildkitd/certificates.go
```diff
@@ -0,0 +1,173 @@
...
+	CertBytes []byte
+	Cert      *x509.Certificate
+}
+
+const (
+	buildkit = "buildkit"
+	earthly  = "earthly"
+)
+
+//GenerateCertificates creates and saves a CA and certificates for both sides of an mTLS TCP connection.
```
vladaionescu, line 32: Nit
```suggestion
// GenerateCertificates creates and saves a CA and certificates for both sides of an mTLS TCP connection.
```

cmd/earthly/main.go
```diff
@@ -411,10 +420,60 @@ func newEarthlyApp(ctx context.Context, console conslogging.ConsoleLogger) *eart
...
+			Value:       "tcp://127.0.0.1:8372",
 			EnvVars:     []string{"EARTHLY_BUILDKIT_HOST"},
 			Usage:       wrap("The URL to use for connecting to a buildkit host. ", "If empty, earthly will attempt to start a buildkitd instance via docker run"),
 			Destination: &app.buildkitHost,
 		},
+		&cli.IntFlag{
+			Name:        "buildkit-port",
+			Value:       config.DefaultBuildkitPort,
+			EnvVars:     []string{"EARTHLY_BUILDKIT_PORT"},
+			Usage:       wrap("The URL to use for connecting to a buildkit host. ", "If empty, earthly will attempt to start a buildkitd instance via docker run"),
```
vladaionescu, line 432: This text doesn't look right as this is an `int` flag, not a URL.

cmd/earthly/main.go
```diff
@@ -411,10 +420,60 @@ func newEarthlyApp(ctx context.Context, console conslogging.ConsoleLogger) *eart
...
+			Destination: &app.buildkitPort,
+		},
+		&cli.IntFlag{
+			Name:        "debugger-port",
+			Value:       config.DefaultDebuggerPort,
+			EnvVars:     []string{"EARTHLY_DEBUG_PORT"},
+			Usage:       wrap("The URL to use for connecting to a debugging host. ", "If empty, earthly will attempt to start a buildkitd instance via docker run"),
```
vladaionescu, line 439: This one too.

vladaionescu (commented):
Looks good overall!

Can this functionality be feature flagged? I feel like this is a pretty fundamental change and it would be good to have a way to turn it on / off easily for testing performance first.

dchw:
> Can this functionality be feature flagged? I feel like this is a pretty fundamental change and it would be good to have a way to turn it on / off easily for testing performance first.

It it turns out to be as simple as a different address / masking the grpc section... maybe.

I don't believe this will have a noticeable impact on performance. It should all remain locally on your machine... the old way was using `docker exec` at the end of the day, and the same protos...

cmd/earthly/main.go
dchw: Yeah it was originally a URL as I was working on this.

cmd/earthly/main.go
```diff
@@ -411,10 +420,60 @@ func newEarthlyApp(ctx context.Context, console conslogging.ConsoleLogger) *eart
...
 			EnvVars:     []string{"EARTHLY_BUILDKIT_HOST"},
 			Usage:       wrap("The URL to use for connecting to a buildkit host. ", "If empty, earthly will attempt to start a buildkitd instance via docker run"),
 			Destination: &app.buildkitHost,
 		},
+		&cli.IntFlag{
+			Name:        "buildkit-port",
+			Value:       config.DefaultBuildkitPort,
+			EnvVars:     []string{"EARTHLY_BUILDKIT_PORT"},
+			Usage:       wrap("The URL to use for connecting to a buildkit host. ", "If empty, earthly will attempt to start a buildkitd instance via docker run"),
+			Destination: &app.buildkitPort,
```
vladaionescu, line 433: Could this port just be inferred from `--buildkit-host`?

vladaionescu:
> It it turns out to be as simple as a different address / masking the grpc section... probably can feature flag it. Ill do that next.

I think the `docker-container://` helper also needs to be imported.

> I don't believe this will have a noticeable impact on performance. It should all remain locally on your machine... the old way was using docker exec at the end of the day, and the same protos...

Makes sense. I'm thinking if large files can cause extra CPU due to encryption in underpowered situations... or some other edge case we're not foreseeing.

We could either keep the flag off for a release or two and only enable it only for ourselves (kinda like we're doing with the parallel converter) + the official image. Or if we think it's safe enough, we can have it enabled straight away for everyone and dogfood it only until the next release.

cmd/earthly/main.go
vladaionescu: Actually, I can see why these are separate settings - the port is the buildkit listening port and the URL is what Earthly uses to connect to it. It's confusing at first, but I can't think of a way to improve this.

vladaionescu: Maybe call it "local port" rather than just "port"? `--buildkit-local-port` and `EARTHLY_BUILDKIT_LOCAL_PORT`

dchw: Right now, we do. But I also added a deprecation notice if you include the port here. The remote settings are now three parts:

* Root host
* Buildkit port (uses root host)
* Debugger port (uses root host)

I did not want to allow the hosts to diverge between these two parts.

dchw: Local port would be a misnomer; since it could be used to specify remote ports too (assuming the root URL was not considered local"). I do agree that naming could be improved though.

dchw:
I've split some parts off this. Closing to re-open a new, better one.

PR #977, opened 2021-05-12
expand secret-file path

alexcb:
Typically the shell will expand paths such as `~/foo`; however when given
as an argument to `--secret-file secret=~/foo`, the shell did not expand
this path. This addresses this issue.

### DIFF
--- a/util/fileutil/homedir.go
+++ b/util/fileutil/homedir.go
@@ -1,0 +1,36 @@
+package fileutil
+
+import (
+	"os"
+	"os/user"
+	"runtime"
+)
+
+// HomeDir returns the home directory of the current user, an additional sudoUser
+// is returned if the user is currently running as root
+func HomeDir() (homeDir string, sudoUser *user.User, err error) {
+	if runtime.GOOS == "windows" {
+		homeDir, err := os.UserHomeDir()
+		return homeDir, nil, err
+	}
+	// See if SUDO_USER exists. Use that user's home dir.
+	sudoUserName, ok := os.LookupEnv("SUDO_USER")
+	if ok {
+		sudoUser, err := user.Lookup(sudoUserName)
+		if err == nil && sudoUser.HomeDir != "" {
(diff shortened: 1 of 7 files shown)


util/fileutil/homedir.go
```diff
@@ -0,0 +1,36 @@
...
+
+// HomeDir returns the home directory of the current user, an additional sudoUser
+// is returned if the user is currently running as root
+func HomeDir() (homeDir string, sudoUser *user.User, err error) {
+	if runtime.GOOS == "windows" {
+		homeDir, err := os.UserHomeDir()
+		return homeDir, nil, err
+	}
+	// See if SUDO_USER exists. Use that user's home dir.
+	sudoUserName, ok := os.LookupEnv("SUDO_USER")
```
vladaionescu, line 17: I think @dchw had a nice function for fetching the current user or sudo user in his PR. Might be worth merging them somehow.

util/fileutil/homedirs.go
```diff
@@ -0,0 +1,67 @@
...
+	"path"
+	"path/filepath"
+	"runtime"
+	"strings"
+
+	"github.com/pkg/errors"
+)
+
+// GetUserHomeDirs returns a map of all users and their homedirs
+func GetUserHomeDirs() (map[string]string, error) {
```
vladaionescu, line 18: I feel that this is too much logic for what is normally needed... but hey... I guess it's now already written.

util/fileutil/homedir.go
alexcb: This code came from the cliutil.DetectHomeDir() function; If we can fold more code into using this function that sounds like a good plan.

util/fileutil/homedirs.go
alexcb: this code was moved here from our `autocompletion.getUsers()`, no new code was written in this PR for this function.

vladaionescu (approved):

PR #979, opened 2021-05-12
Add config docs.

dchw:
I put the docs above `account` because its lesser used, and it was long enough so as to bury this command. It also felt like a more natural positioning?

### DIFF
--- a/docs/earthly-command/earthly-command.md
+++ b/docs/earthly-command/earthly-command.md
@@ -296,11 +296,11 @@ Enables modifying, adding, and updating values in `~/.earthly/config.yml`. It do
 
 #### Options
 
-#### `--help`
+##### `--help`
 
 Prints help text, along with some examples.
 
-#### `[key] --help`
+##### `[key] --help`
 
 Prints help for the specific key, including what it is used for and what kind of value it needs to be.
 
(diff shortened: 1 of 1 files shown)


docs/earthly-command/earthly-command.md
```diff
@@ -282,6 +282,69 @@ Instructs earthly to issue a "prune all" command to the buildkit daemon.
...
+
+#### Synopsis
+
+```
+earthly [options] config [key] [value]
+```
+
+#### Description
+
+Enables modifying, adding, and updating values in `~/.earthly/config.yml`. It does its best to preserve formatting and existing comments. You can also set complex values using YAML literals.
```
alexcb, line 295: I wonder if we can make this more concise like this?
```suggestion
Manipulates values in `~/.earthly/config.yml`. The `value` parameter can be a string or YAML literals.
```

alexcb (approved):

docs/earthly-command/earthly-command.md
```diff
@@ -282,6 +282,69 @@ Instructs earthly to issue a "prune all" command to the buildkit daemon.
...
+
+#### Options
+
+#### `--help`
```
vladaionescu, line 299: The options should have 5 `#`s

docs/earthly-command/earthly-command.md
dchw: Its not just strings, its literally only YAML literals. Which include bare numbers and boolean values too. I can make the first word manipulates.

dchw: I shortened it some?

vladaionescu (approved):

PR #981, opened 2021-05-13
fix panic when netrc file exists but doesnt contain host

alexcb:
- fixes a panic that occurs when a ~/.netrc file exists but is either
empty, invalid, or doesn't contain login information for the host that
is being looked up.

### DIFF
--- a/buildcontext/gitlookup.go
+++ b/buildcontext/gitlookup.go
@@ -237,6 +237,8 @@ func (gl *GitLookup) detectProtocol(host string) (protocol string, err error) {
 	return
 }
 
+var errNoRCHostEntry = fmt.Errorf("no netrc host entry")
+
 func (gl *GitLookup) lookupNetRCCredential(host string) (login, password string, err error) {
 	homeDir, _, err := cliutil.DetectHomeDir()
 	if err != nil {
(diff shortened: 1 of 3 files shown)


buildcontext/gitlookup.go
```diff
@@ -246,6 +248,10 @@ func (gl *GitLookup) lookupNetRCCredential(host string) (login, password string,
 	if err != nil {
 		return "", "", errors.Wrap(err, "failed to parse .netrc")
 	}
+	machine := n.Machine(host)
+	if machine == nil {
+		return "", "", errors.Wrapf(errNoRCHostEntry, "failed to lookup netrc entry for %s", host)
```
vladaionescu, line 253: Would this case cause it to fall back to other auth mechanisms?

alexcb: it falls back to unauthenticated https, as the error messages are actually ignored: https://github.com/earthly/earthly/blob/fix-panic-when-netrc-entry-missing/buildcontext/gitlookup.go#L304

alexcb: and this is only when the auth is set to "auto" (the default), first it tries connecting via ssh, then netrc, then un-authed.

vladaionescu: great

dchw (approved):

PR #983, opened 2021-05-13
buildkitd was not building on arm64 without calling apk fix

alexcb:


vladaionescu (approved):
That's weird

alexcb:
I just tried running this again with `earthly +all-buildkitd`, and it's working without this PR. I suspect there was an upstream issue in the apk repos.

PR #989, opened 2021-05-19
Bootstrap if simple heuristics say we haven't run it yet.

dchw:
Pulled the logic out of the TCP pull request.

Note that this does not separate the shell completions from bootstrap itself yet. It seemed like we could skate by and see if we wanted to do that later?

I was on board with logging to stderr; but given the number of places I had to redirect I am concerned; but not enough to change course. Just want to think more deeply one more time before we merge something like this.


conslogging/conslogging.go
```diff
@@ -61,7 +61,7 @@ type ConsoleLogger struct {
 // Current returns the current console.
 func Current(colorMode ColorMode, prefixPadding int) ConsoleLogger {
 	return ConsoleLogger{
-		outW:           os.Stdout,
+		outW:           os.Stderr, // So logs dont sully any intended outputs of commands.
```
dchw, line 64: As discussed in standup this morning

examples/tests/Earthfile
```diff
@@ -431,7 +431,7 @@ save-artifact-after-push:
         --post_command="2>&1 | perl -pe 'BEGIN {\\\$status=1} END {exit \\\$status} \\\$status=0 if /not found/;'"
 
 push-build:
-    DO +RUN_EARTHLY --earthfile=push-build.earth --target=+test --extra_args="--push" --post_command="2>&1 > output"
+    DO +RUN_EARTHLY --earthfile=push-build.earth --target=+test --extra_args="--push" --post_command="> output 2>&1"
```
dchw, line 434: Flipping this order means we capture stderr properly

util/cliutil/earthlydir.go
```diff
@@ -68,3 +68,23 @@ func DetectHomeDir() (homeDir string, sudoUser *user.User, err error) {
...
+	if err != nil {
+		return false
+	}
+
+	earthlyDir := filepath.Join(homeDir, ".earthly")
+	if !fileutil.DirExists(earthlyDir) {
+		return false
+	}
+
+	installID := filepath.Join(homeDir, ".earthly", "install_id")
```
vladaionescu, line 84: Will this heuristic work even if analytics reporting is turned off via config?

vladaionescu (approved):
:+1:

Perhaps these two should be moved to stderr too?

* https://github.com/earthly/earthly/blob/main/builder/docker.go#L72
* https://github.com/earthly/earthly/blob/main/builder/docker.go#L85

(This one is already done https://github.com/earthly/earthly/blob/main/buildkitd/buildkitd.go#L485)

vladaionescu:
Oh.. did we want to move auto-complete installation under a new flag too? `earthly bootstrap --with-autocomplete`?

Otherwise the bootstrapping might not work due to missing `sudo`.

util/cliutil/earthlydir.go
dchw: Probably not. I actually think we should remove this as a check since its not even directly controlled by bootstrap.

dchw:
> Oh.. did we want to move auto-complete installation under a new flag too? `earthly bootstrap --with-autocomplete`?
> 
> Otherwise the bootstrapping might not work due to missing `sudo`.

I was trying to see if we could get away without it, and although CI does work, it does fail in bootstrap a lot and really increases noise. Ill add that flag.

dchw:
>     * https://github.com/earthly/earthly/blob/main/builder/docker.go#L72
> 
>     * https://github.com/earthly/earthly/blob/main/builder/docker.go#L85

I can dig some; what are these hooked up to?

vladaionescu:
> I can dig some; what are these hooked up to?

Not sure what you mean. They're hooked up to stdout - they show up during the output phase, when Earthly exports images to the local Docker.

PR #990, opened 2021-05-19
Autobootstrap Cleanup

dchw:
Should address most comments from https://github.com/earthly/earthly/pull/989 that was unintentionally automerged.


vladaionescu (approved):
:+1:

PR #993, opened 2021-05-19
Bootstrap ensures permissions of everything within it. 

dchw:
Only works with `sudo`.

Also other cleanups:

* Only print messages for autocompletion if we install the autocompletion
* Different method for detecting homedir

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,6 +48,9 @@ jobs:
       EARTHLY_INSTALL_ID: "earthly-githubactions"
     steps:
       - uses: actions/checkout@v2
+      - uses: actions/setup-go@v1
+        with:
+          go-version: 1.16
       - name: Download released earthly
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -74,6 +77,9 @@ jobs:
       EARTHLY_INSTALL_ID: "earthly-githubactions"
     steps:
       - uses: actions/checkout@v2
+      - uses: actions/setup-go@v1
+        with:
+          go-version: 1.16
```
vladaionescu, line 82: Hmm... What's this for?

vladaionescu (approved):

.github/workflows/ci.yml
dchw: The ubuntu-latest runner only has Go 1.15. I need this action in cases where we use go run since the function I used to recurse through .earthly is new to 1.16.

vladaionescu: Ah right. I had forgotten that we use `go run` in some places.

Issue #896, opened 2021-03-29
Proposal: Change what is being output and pushed


vladaionescu:
This proposal gives the user more control over what is being output and what is being pushed, and also it helps prevent pushes when they are not intended in monorepo setups.

## Background

Today, the following are produced locally at the end of a successful build:

* All `SAVE ARTIFACT ... AS LOCAL ...`, with local Earthfiles
* All images with tag names (both local and remote Earthfiles)

The following is NOT produced:

* `SAVE ARTIFACT ... AS LOCAL ...` produced in remote targets

This strategy is meant to optimize for the situations where you actually need things locally, for local development iteration.

If an artifact of a remote target is needed, the user can issue a separate earthly command to output it.

`earthly --artifact github.com/foo/bar+buz/artifact ./local/dest`

### Situations where this strategy falls short

Sometimes images are produced when they are not needed, thus slowing down the build, while other times the user expects remote artifacts to be output, but aren’t.

Examples where images are produced but are not needed:
* A common base image target is mentioned in a `FROM` to create a microservice. The user is interested only in producing the microservice image, not the base image too.
* A `COPY` artifact where the artifact happens to be generated by some logic that happens to include `SAVE IMAGE` (a common base image target, or a microservice that was used in the generation process).

Examples of situations where remote artifacts are expected locally:
* ? (@dchw had a use-case IIRC?)

### Pushes background

Today, pushing (when enabled) takes place for all local targets.

This can create situations where a dependency is accidentally pushed when that’s not intended if the user uses a monorepo setup, where each sub-project has its own separate build.

## Proposal

We change the logic s.t. output is only produced for the targets invoked either directly (mentioned directly on the command-line) or via calls to `BUILD`. `FROM` and `COPY` (and others e.g. `WITH DOCKER --load`) do not cause images to be output locally.

So chains like these will produce outputs:

* `SAVE`
* `BUILD -> SAVE`
* `BUILD -> BUILD -> SAVE`
* `BUILD -> BUILD -> BUILD -> SAVE`
etc

While chains like these will NOT produce outputs:

* `FROM -> SAVE`
* `COPY -> SAVE`
* `FROM -> BUILD -> SAVE`
* `BUILD -> FROM -> SAVE`
* `BUILD -> BUILD -> COPY -> SAVE`
etc

This ensures that outputs are produced only when they are intended. If the user wants to both output an image and use `FROM` on it, they can issue

```
BUILD +target
FROM +target
```

[...]

PR #995, opened 2021-05-20
Only save artifacts when they are directly referenced

alexcb:
Re https://github.com/earthly/earthly/issues/896

Artifacts will only be saved locally when their target is directly
referenced, or referenced by a BUILD.

This will cause targets that are referenced from a FROM, or COPY from
*not* saving artifacts as a side-effect.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -159,6 +159,8 @@ func (b *Builder) convertAndBuild(ctx context.Context, target domain.Target, opt
 
 	sharedLocalStateCache := earthfile2llb.NewSharedLocalStateCache()
 
+	featureFlagOverrides := b.opt.FeatureFlagOverrides
+
 	destPathWhitelist := make(map[string]bool)
 	manifestLists := make(map[string][]manifest) // parent image -> child images
 	var mts *states.MultiTarget
@@ -191,9 +193,9 @@ func (b *Builder) convertAndBuild(ctx context.Context, target domain.Target, opt
 				Parallelism:          b.opt.Parallelism,
 				Console:              b.opt.Console,
 				GitLookup:            b.opt.GitLookup,
-				FeatureFlagOverrides: b.opt.FeatureFlagOverrides,
+				FeatureFlagOverrides: featureFlagOverrides,
 				LocalStateCache:      sharedLocalStateCache,
-			})
+			}, true)
 			if err != nil {
 				return nil, err
 			}
@@ -339,10 +341,10 @@ func (b *Builder) convertAndBuild(ctx context.Context, target domain.Target, opt
 					}
 				}
 			}
-			performSaveLocals := (!sts.Target.IsRemote() &&
-				!opt.NoOutput &&
+			performSaveLocals := (!opt.NoOutput &&
 				!opt.OnlyFinalTargetImages &&
 				opt.OnlyArtifact == nil)
+
 			if performSaveLocals {
 				for _, saveLocal := range b.targetPhaseArtifacts(sts) {
 					ref, err := b.artifactStateToRef(childCtx, gwClient, sts.SeparateArtifactsState[saveLocal.Index], sts.Platform)
@@ -499,7 +501,6 @@ func (b *Builder) convertAndBuild(ctx context.Context, target domain.Target, opt
 		dirIndex := 0
 		for _, sts := range mts.All() {
 			console := b.opt.Console.WithPrefixAndSalt(sts.Target.String(), sts.ID)
-
 			for _, saveImage := range sts.SaveImages {
 				shouldPush := opt.Push && saveImage.Push && !sts.Target.IsRemote() && saveImage.DockerTag != ""
 				shouldExport := !opt.NoOutput && saveImage.DockerTag != ""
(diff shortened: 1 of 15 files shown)


vladaionescu:
Remember to feature-flag this ;-)

earthfile2llb/converter.go
```diff
@@ -1122,7 +1122,7 @@ func (c *Converter) ExpandArgs(word string) string {
 	return c.varCollection.Expand(word)
 }
 
-func (c *Converter) prepBuildTarget(ctx context.Context, fullTargetName string, platform *specs.Platform, allowPrivileged bool, buildArgs []string, isDangling bool, isFrom bool) (domain.Target, ConvertOpt, bool, error) {
+func (c *Converter) prepBuildTarget(ctx context.Context, fullTargetName string, platform *specs.Platform, allowPrivileged, doSaves bool, buildArgs []string, isDangling bool, isFrom bool) (domain.Target, ConvertOpt, bool, error) {
```
vladaionescu, line 1125: I would merge `doSaves` and `isFrom` into a single parameter `cmdName` and use that.

```go
opt.doSaves = (cmdName == 'BUILD' && c.opt.doSaves)
```

vladaionescu (commented):
I feel like the logic is split too much between the interpreter and the converter making it hard to follow.

I think you can keep everything only in converter and simply check in `buildTarget` / `prepBuildTarget` whether the command is `BUILD` or not, in order to pass the `opt.doSaves` onwards.

Then the decision to not do anything on saves can be done in the `converter.SaveArtifact` and`converter.SaveImage`.

This should keep all the logic in the converter and make the interpreter decoupled.

earthfile2llb/converter.go
alexcb: good plan, that made it much cleaner.

vladaionescu:
This is still being worked on, right? This still needs the changes in `builder.go`.

earthfile2llb/earthfile2llb.go
```diff
@@ -105,18 +111,19 @@ func Earthfile2LLB(ctx context.Context, target domain.Target, opt ConvertOpt) (m
...
 	err = features.ApplyFlagOverrides(ftrs, opt.FeatureFlagOverrides)
 	if err != nil {
 		return nil, errors.Wrapf(err, "failed to apply version feature overrides")
 	}
+	opt.Features = ftrs
 
-	if ftrs.ReferencedSaveOnly {
-		fmt.Printf("TODO feature-flip referenced save artifact as local feature in a future PR.\n")
+	if !ftrs.ReferencedSaveOnly {
+		// 0.5 behaviour is to save all artifacts and images in all cases (if they are directly referenced, or indirectly triggered)
```
vladaionescu, line 125: Not really.. in 0.5 artifacts are to be saved if they are part of local targets, whereas images are always saved.

builder/builder.go
```diff
@@ -511,46 +526,44 @@ func (b *Builder) convertAndBuild(ctx context.Context, target domain.Target, opt
...
-					artifact := domain.Artifact{
-						Target:   sts.Target,
-						Artifact: saveLocal.ArtifactPath,
-					}
-					err := b.saveArtifactLocally(ctx, artifact, artifactDir, saveLocal.DestPath, sts.ID, opt, saveLocal.IfExists)
-					if err != nil {
-						return nil, err
-					}
-					dirIndex++
+			for _, saveLocal := range sts.SaveLocals {
```
vladaionescu, line 529: Shouldn't the `!sts.Target.IsRemote()` still be checked somewhere for 0.5 behavior?

earthfile2llb/earthfile2llb.go
alexcb: Good catch, I will expand the tests to make sure the save images still occur.

builder/builder.go
alexcb: I needed to remove it to support the case of performing a build such as:

    earthly github.com/earthly/earthly:main+deps

which would then save the `go.mod` and `go.sum` files locally; however if the target is remote, we perform this extra check earlier up:

```
	if target.IsRemote() {
		if len(featureFlagOverrides) != 0 {
			featureFlagOverrides += ","
		}
		featureFlagOverrides += "referenced-save-only"
	}
```

which will prevent the remote target from saving anything else that was indirectly referenced.

I was thinking that in this case it's fair to assume that if the user is directly referencing a remote target, then that user wants any directly referenced files to be saved.

There might be an issue where a user gives us a local Earthfile that then references something that's remote, and none of these Earthfiles have any VERSION flags, then those remote targets might also be saved -- let me double check this case (and write a test for it if one doesn't exist).

earthfile2llb/earthfile2llb.go
alexcb: @vladaionescu I just looked over https://github.com/earthly/earthly/issues/896 again and it says:

> We change the logic s.t. output is only produced for the targets invoked either directly (mentioned directly on the command-line) or via calls to BUILD. FROM and COPY (and others e.g. WITH DOCKER --load) do not cause images to be output locally.

Is this still correct? I have this code only locally saving artifacts or images when they are referenced via `FROM` or `BUILD`.

builder/builder.go
alexcb: In addition this also had to be added to https://github.com/earthly/earthly/blob/selective-push/earthfile2llb/converter.go#L1199

vladaionescu (approved):

PR #999, opened 2021-05-21
One-time section, include docs update.

dchw:
Add a section to hold all the one-time release items I am generating, and include a one-time item for updating `bootstrap` invocations on the website.

### DIFF
--- a/cmd/earthly/main.go
+++ b/cmd/earthly/main.go
@@ -1381,6 +1381,8 @@ func (app *earthlyApp) actionBootstrap(c *cli.Context) error {
 			}
 			fmt.Print(compEntry)
 			return nil
+		case "":
+			break
 		default:
 			return errors.Errorf("unhandled source %q", app.homebrewSource)
 		}
--- a/cmd/earthly/main.go
+++ b/cmd/earthly/main.go
@@ -1359,34 +1359,35 @@ func symlinkEarthlyToEarth() error {
 
 func (app *earthlyApp) actionBootstrap(c *cli.Context) error {
 	app.commandName = "bootstrap"
+	defer cliutil.EnsurePermissions()
 
 	var err error
 	console := app.console.WithPrefix("bootstrap")
 
-	if app.bootstrapWithAutocomplete || app.homebrewSource != "" {
-		switch app.homebrewSource {
-		case "bash":
-			compEntry, err := bashCompleteEntry()
-			if err != nil {
-				fmt.Fprintf(os.Stderr, "Warning: failed to enable bash-completion: %s\n", err)
-				return nil // zsh-completion isn't available, silently fail.
-			}
-			fmt.Print(compEntry)
-			return nil
-		case "zsh":
-			compEntry, err := zshCompleteEntry()
-			if err != nil {
-				fmt.Fprintf(os.Stderr, "Warning: failed to bootstrap zsh-completion: %s\n", err)
-				return nil // zsh-completion isn't available, silently fail.
-			}
-			fmt.Print(compEntry)
-			return nil
-		case "":
-			break
-		default:
-			return errors.Errorf("unhandled source %q", app.homebrewSource)
+	switch app.homebrewSource {
+	case "bash":
+		compEntry, err := bashCompleteEntry()
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Warning: failed to enable bash-completion: %s\n", err)
+			return nil // zsh-completion isn't available, silently fail.
 		}
+		fmt.Print(compEntry)
+		return nil
+	case "zsh":
+		compEntry, err := zshCompleteEntry()
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Warning: failed to bootstrap zsh-completion: %s\n", err)
+			return nil // zsh-completion isn't available, silently fail.
+		}
+		fmt.Print(compEntry)
+		return nil
+	case "":
+		break
+	default:
+		return errors.Errorf("unhandled source %q", app.homebrewSource)
+	}
 
+	if app.bootstrapWithAutocomplete {
 		// Because this requires sudo, it should warn and not fail the rest of it.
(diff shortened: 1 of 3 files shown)


release/README.md
```diff
@@ -56,6 +56,10 @@
 * Copy the release notes you have written before and paste them in the Earthly Community slack channel `#announcements`, together with a link to the release's GitHub page. If you have Slack markdown editing activated, you can copy the markdown version of the text.
 * Ask Adam to tweet about the release.
 
+### One-Time (clear this section when done during release)
+
+* Update the website with the documentation for bootstrap, include `--with-autocomplete` as part of the installation.
```
vladaionescu, line 61: I think it's fine to not include this flag for the mac installation since the autocomplete is done through homebrew. We could remove sudo for mac then.

dchw: You helped me find a bug that would have broken autocomplete on macOS again. That is included here and an addendum added to this point.

vladaionescu (approved):

cmd/earthly/main.go
```diff
@@ -1363,7 +1363,7 @@ func (app *earthlyApp) actionBootstrap(c *cli.Context) error {
 	var err error
 	console := app.console.WithPrefix("bootstrap")
 
-	if app.bootstrapWithAutocomplete {
+	if app.bootstrapWithAutocomplete || app.homebrewSource != "" {
```
vladaionescu, line 1366: I think these shuold be two separate if blocks actually.

vladaionescu (changes requested):

cmd/earthly/main.go
vladaionescu: Basically, the switch should be completely independent from `bootstrapWithAutocomplete`.

And `bootstrapWithAutocomplete` should only wrap the stuff that comes after the switch.

alexcb: Currently homebrew calls:

    "#{bin}/earthly", "bootstrap", "--source", "bash"

(see https://github.com/Homebrew/homebrew-core/blob/master/Formula/earthly.rb#L33 )

which should print out something like this:

    complete -o nospace -C '/home/alex/bin/earthly-v0.5.8' earthly

(or a different path depending on the system).

If this block was broken up into a seperate part, maybe we can remove the need to change how homebrew generates this output?

dchw: Yep. The switch above actually returns if it was a `--source` call. I had just skimmed it and though it was getting the proper data for autocompletion.

I gated the switch separate from the other stuff, and then also needed to move the permissions insurance to a defer. See the new fix and if this gets us there?

examples/tests/bootstrap/test-bootstrap.sh
```diff
@@ -136,4 +136,23 @@ if [[ $(stat --format '%G' "$HOME/.earthly/config.yml") != "$GRP" ]]; then
   echo "earthly config is not owned by the users group"
   stat "$HOME/.earthly/config.yml"
   exit 1
-fi
\ No newline at end of file
+fi
+
+echo "=== Test 6: Homebrew Source ==="
```
alexcb, line 141: nice test addition!

examples/tests/bootstrap/test-bootstrap.sh
```diff
@@ -136,4 +136,23 @@ if [[ $(stat --format '%G' "$HOME/.earthly/config.yml") != "$GRP" ]]; then
...
+
+bash=$("$earthly" bootstrap --source bash)
```
alexcb, line 143: What about expanding this test to ensure it still works when docker isn't present? e.g.

    DOCKER_HOST="bootstrap doesn't have docker" earthly bootstrap --source bash

vladaionescu (approved):

examples/tests/bootstrap/test-bootstrap.sh
dchw: That should be a separate test with the `--no-buildkit` flag, as thats what would fail if docker is not present.

alexcb: my worry is that https://github.com/Homebrew/homebrew-core/blob/master/Formula/earthly.rb#L33-L36 will fail if docker isn't available, which would prevent the brew package from being built.

dchw: I can add a test around that flag I mentioned earlier. We've already done releases with the buildkit docker stuff in there?

alexcb: I don't think we have:

```
alex@mah:~/gh/earthly/earthly$ DOCKER_HOST="bootstrap doesn't have docker" earthly-v0.5.13 bootstrap --source bash
complete -o nospace -C '/home/alex/bin/earthly-v0.5.13' earthly
alex@mah:~/gh/earthly/earthly$ DOCKER_HOST="bootstrap doesn't have docker" earthly bootstrap --source bash
           buildkitd | Is docker installed and running? Are you part of the docker group?
Error: bootstrap new buildkitd client: docker not available
```

dchw: Its been in there since April 5 - PR #910. We released successfully 5x since then?

dchw: I'll add a test for the `--no-buildkit` flag and add a note to update the recipe on next release.

alexcb: > add a note to update the recipe on next release

Are you talking about the homebrew forumula?

If yes, I don't understand why we can't maintain backwards compatibility and have earthly continue to work the way homebrew expects it to.

dchw: Reolved. Added `DOCKER_HOST ` test (and others) for `--source`, and an individual test for `--no-buildkit`. We will break it out later to avoid this kind of confusion again.

alexcb (approved):

Issue #991, opened 2021-05-19
Proposal: VERSION entry


vladaionescu:
## Background

One of the key reasons for using Earthly is the consistency in which builds are executed. While this is important, Earthly itself is under constant development and is evolving. By that definition, Earthly will evolve over time in the way it executes builds, and thus to some extent they will no longer be as consistent with the way a previous version of Earthly would produce results.

We need a way to pin the way an Earthfile is interpreted, to help maintain consistency, while allowing Earthly itself to continue to evolve.

## Proposal

The proposal is to introduce a new `VERSION` entry at the top of every Earthfile. This version declaration must be **the very first command in the Earthfile**. This command would tell Earthly what version of Earthly the Earthfile is written for.

Example

```Dockerfile
VERSION 0.7

FROM alpine

build:
  RUN echo abracadabra
```

The version number would only specify `<major>.<minor>` and would always leave out the `<patch>` part of the [semver](https://semver.org/). Earthly releases anyway maintain compatibility between patch version increments. Minor version increments take place when there is an incompatibility introduced.

If a version of Earthly previous to 0.7 encounters this declaration, it will return an error and not execute the build. If a version of Earthly greater (or equal) than 0.7 encounters this declaration, it will downgrade its capabilities to support the behavior the way it was specified in version 0.7.

### Importing cross-versions.

Referencing other Earthfiles of different versions is allowed.

A 0.6 Earthfile can import a 0.7 Earthfile and vice-versa. The parser and interpreter need to be smart enough to support this.

### Support lifecycle

If a version is too old, it will not be supported.

We aim to support at least 4 minor versions behind and also at least a whole year since the last release of that minor version. This should be a nice balance between keeping Earthly development nimble enough, while not breaking people's builds abruptly.

## Rollout

At the time of writing the latest version of Earthly is `v0.5.13`. We plan to introduce `VERSION` as an optional declaration for the next 4 minor versions. If `VERSION` is not declared, it is assumed to be `0.5` so as to maintain backwards compatibility. From version 0.9 of Earthly onwards, `VERSION` will become mandatory.

| Earthly Version      | Behavior |
| ----------- | ----------- |
| `v0.5.13`      | `VERSION` not supported - not a recognized command |
| `v0.5.14+`      | `VERSION` is optional. If `VERSION` declared is different from `0.5`, return an error |
[...]

PR #1001, opened 2021-05-21
VERSION keyword for Earthfile

alexcb:
- currently the VERSION is optional, but it will allow us to feature-flip
options at a future point.

Part of https://github.com/earthly/earthly/issues/991


ast/tests/build-arg.ast.json
```diff
@@ -1,4 +1,8 @@
 {
+  "Version": {
+    "major": 0,
+    "minor": 0
+  },
```
vladaionescu, line 5: I think it's better if this is left out when it's not set. It's less error prone when the JSON AST matches exactly to what is actually visible in text form.

ast/spec/earthfile.go
```diff
@@ -2,6 +2,7 @@ package spec
 
 // Earthfile is the AST representation of an Earthfile.
 type Earthfile struct {
+	Version        Version
```
vladaionescu, line 5: `json:"version"` (lower-case, to match the other fields)

vladaionescu, line 5: Should probably also be optional (`*Version` and `omitempty`)

ast/version.go
```diff
@@ -0,0 +1,108 @@
...
+				EndColumn:   lineWidth,
+			}
+		}
+
+		return &version, nil
+	}
+
+	// No version was found
+	return nil, nil
+}
```
vladaionescu, line 108: ~This could use some tests - it's nicely isolated.~

NVM - seems like you have some nice integration tests.

earthfile2llb/earthfile2llb.go
```diff
@@ -97,6 +99,16 @@ func Earthfile2LLB(ctx context.Context, target domain.Target, opt ConvertOpt) (m
 	if err != nil {
 		return nil, errors.Wrapf(err, "resolve build context for target %s", target.String())
 	}
+
+	ftrs, err := features.GetFeatures(bc.Earthfile.Version)
+	if err != nil {
+		return nil, errors.Wrapf(err, "resolve build context for target %s", target.String())
```
vladaionescu, line 105: (You'll need to also define a `bc.Earthfile.Version.String()` or similar)

```suggestion
		return nil, errors.Wrapf(err, "resolve feature set for version %s for target %s", bc.Earthfile.Version.String(), target.String())
```

earthfile2llb/earthfile2llb.go
```diff
@@ -97,6 +99,16 @@ func Earthfile2LLB(ctx context.Context, target domain.Target, opt ConvertOpt) (m
...
+	}
+
+	if ftrs.ReferencedSaveOnly {
+		fmt.Printf("TODO feature-flip referenced save artifact as local feature in a future PR.\n")
```
vladaionescu, line 109: Somewhat unrelated, but this will apply to both save artifact and save image, right?

examples/tests/version/Earthfile
```diff
@@ -0,0 +1,46 @@
+ARG DOCKERHUB_AUTH=true
+FROM ../../..+earthly-integration-test-base --DOCKERHUB_AUTH=$DOCKERHUB_AUTH
+
+IMPORT .. AS tests
```
vladaionescu, line 4: 🆒

examples/tests/version/Earthfile
```diff
@@ -0,0 +1,46 @@
...
+    BUILD +test-single-line
+    BUILD +test-single-line-with-args
+    BUILD +test-multi-line
+    BUILD +test-multi-line-with-comment
+    BUILD +test-multi-line-with-comment2
+    BUILD +test-multi-line-with-comment3
+    BUILD +test-multi-line-with-comment4
+    BUILD +test-multi-line-with-args
+    BUILD +test-multi-line-with-args2
+    BUILD +test-multi-line-with-newline
```
vladaionescu, line 46: Some more edge-cases:
* A comment before version, possibly with whitespace in-between
* Whitespace, then `VERSION`
* `VERSION`, then EOF (no new line)

features/features.go
```diff
@@ -0,0 +1,63 @@
...
+
+	if version == nil {
+		return &ftrs, nil
+	}
+
+	if version.Args == nil {
+		return nil, errUnexpectedArgs
+	}
+
+	parsedArgs, err := flagutil.ParseArgs("VERSION", &ftrs, version.Args)
```
vladaionescu, line 36: Could this be made to ignore feature flags that we don't recognize? Maybe only show a warning in such cases?

features/features.go
```diff
@@ -0,0 +1,63 @@
...
+		return nil, errors.Wrapf(err, "failed to parse major version %q", majorAndMinor[0])
+	}
+	minor, err := strconv.Atoi(majorAndMinor[1])
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to parse minor version %q", majorAndMinor[1])
+	}
+
+	// TODO depending on versions here, flip on/off features.
+	_ = major
+	_ = minor
```
vladaionescu, line 60: Might be useful to carry a string form for this, to display in error messages if needed?

vladaionescu (commented):
Cool stuff

earthfile2llb/earthfile2llb.go
alexcb: That was my intention, a single bool to flip both of those on and off.

examples/tests/version/Earthfile
alexcb: Good ideas.

Having an empty Earthfile is problematic, as I don't think I can run it without a `+test` target to run; however, I discovered I can `IMPORT` an empty Earthfile simply containing a `VERSION 0.5`, and that's passing.

features/features.go
alexcb: I'm unsure if issuing a warning is a good idea here, 

What about a case where a user wants to make use of a certain feature but that feature doesn't exist? for example:

    VERSION --enable-foo 0.5

If we ran this on an earlier version of earthly, then it would just print a warning and then likely fail later on when foo isn't enabled.

Can you think of an example where this ignore unknown flags behaviour would help a user?

earthfile2llb/earthfile2llb.go
alexcb: Currently the structs in the ast/spec package only have JSON tags and no methods (e.g. see https://github.com/earthly/earthly/blob/version/ast/spec/earthfile.go ). A quick idea is to just use `%v` and `bc.Earthfile.Version.Args`, or we could just marshal the str via `json.marshal`?

examples/tests/version/Earthfile
vladaionescu: You could try to run `+base`

features/features.go
vladaionescu: Not a strong opinion. I was thinking of a case where multiple people on the team are using varying versions of Earthly and only some have the version with that feature flag. If the feature flag errors out on a version of Earthly that doesn't support it, then the users wouldn't be able to use any target from that Earthfile.

I'm not convinced either that this behavior is useful. So maybe it's better to return an error if we don't recognize it.

examples/tests/version/Earthfile
alexcb: of course! I forget that can be run from the command line :+1:

features/features.go
alexcb: Maybe that's where `--strict` could come into play? I'm in favour of leaving it as-is until a use-case comes up requiring it.

vladaionescu: Yeah, less is more. As-is is best.

features/features.go
alexcb: I added in `Version()` and `String()` methods on the `Features` struct to make it easier to get a string version for debugging/errors.

vladaionescu (approved):

PR #1002, opened 2021-05-21
Try using direct aarch64 shellcheck download?

dchw:
Th M1 continues to fail in CI when it comes to Shellcheck. This should enable Shellcheck natively on the M1, eliminating the emulation layer as a source of problems.

This is finding some lint issues, but appears to make it further more consistently than before.

### DIFF
--- a/scripts/tests/auth/test-self-hosted.sh
+++ b/scripts/tests/auth/test-self-hosted.sh
@@ -110,10 +110,10 @@ rm -rf odd-project
 # test that earthly has access to it
 "$earthly" config git "{myserver: {pattern: 'myserver/([^/]+)', substitute: 'ssh://root@$ip:2222/root/my/really/weird/path/\$1.git', auth: ssh}}"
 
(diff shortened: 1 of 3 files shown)


scripts/tests/auth/test-self-hosted.sh
```diff
@@ -110,10 +110,10 @@ rm -rf odd-project
 # test that earthly has access to it
 "$earthly" config git "{myserver: {pattern: 'myserver/([^/]+)', substitute: 'ssh://root@$ip:2222/root/my/really/weird/path/\$1.git', auth: ssh}}"
 
-echo === Test remote build under repo root ===
+echo "=== Test remote build under repo root ==="
```
dchw, line 113: This is most of the lint issues the shellcheck stage is finding on the M1. Wonder if its due to the one in Alpine's repo being old? Maybe migrate the other platforms over to using a manual download installation?

vladaionescu (approved):

PR #1003, opened 2021-05-24
TCP (without TLS)

dchw:
This enables using the TCP mode behind a feature flag. This does not include TLS, which will also be behind a different feature flag.

Should make this more reviewable and incremental.


buildkitd/buildkitd.go
```diff
@@ -265,8 +269,16 @@ func Start(ctx context.Context, console conslogging.ConsoleLogger, image string,
 		// Debugger only supported in top-most earthly.
 		// TODO: Main reason for this is port clash. This could be improved in the future,
 		//       if needed.
-		args = append(args,
-			"-p", fmt.Sprintf("127.0.0.1:%d:8373", settings.DebuggerPort))
+
+		// These are controlled by us and should have been validated already
```
vladaionescu, line 273: In that case, panic on `err` perhaps?

dchw:
Converting to draft until I add on the correct entrypoint for the experimental docker image too.

cmd/earthly/main.go
```diff
@@ -412,10 +417,27 @@ func newEarthlyApp(ctx context.Context, console conslogging.ConsoleLogger) *eart
...
 			Destination: &app.buildkitHost,
 		},
+		&cli.IntFlag{
+			Name:        "buildkit-port",
+			Value:       config.DefaultBuildkitPort,
+			EnvVars:     []string{"EARTHLY_BUILDKIT_PORT"},
+			Usage:       wrap("The port to use for connecting to a buildkit host. ", "Used in combination with buildkit-host to produce the URL to dial."),
+			Destination: &app.buildkitPort,
+			Hidden:      true,
+		},
```
vladaionescu, line 432: I feel like these are hard to follow because the host and port exist in two separate fields. It's a bit unclear what happens in some edge cases like if the user wanted to connect to a host without a port number.

I would do the configuration a little differently:

- `buildkit-host` remains like it is today: a full URL.
- `buildkit-port` turns into something like `buildkit-listening-port` and is used for configuring the buildkit daemon that we manage. We can choose a default port for this. Say `8372`, like you chose here.

The different possible cases:
* When `buildkit-host` is of the form `docker-container://` and it matches the container name we expect, again, we manage the daemon ourselves, but we use the old connectivity. (This should be the default)
* When `buildkit-host`'s port matches `tcp://127.0.0.1:<buildkit-listening-port>`, we manage the daemon ourselves and use TCP.
* When `buildkit-host` is anything else, we don't manage the daemon ourselves - we assume that the user did what was necessary to bring this it up separately.

So to turn on the feature, you would simply do

```
earthly --buildkit-host=127.0.0.1:8372
```

I think that this is a little simpler because the full address is not comprised of multiple flags and you can repoint earthly to a different buildkit with a single flag override.

vladaionescu, line 432: Since this flag may trigger a buildkit restart, it's probably better if it's placed in the config, rather than as a command-line flag.

vladaionescu (commented):
Looks good overall - but I had a comment about the configuration.

What do you think?

buildkitd/buildkitd.go
dchw: If you are ok with `panic` I am

cmd/earthly/main.go
dchw: The tl;dr:
* Config is structured this way because I wanted to align how we also deal with the address we connect to for the debugger. I wanted to make it consistent, and also impossible for the user to accidentally let these diverge. Letting the hosts vary implies that you can debug from a different host than what buildkitd is running on, and I thought we weren't into that.
* I would rather keep a simple boolean for enabling TCP than requiring an address. The ports, and even the host root, are entirely optional (as they would be if we make TCP the default, which I hope we do). The options are there if you want to customize (which would likely only be for remote buildkits, or local port clash. It makes the flagged feature behave more like it would if it were productionized).

I think theres a middle ground here. Perhaps this is palatable:
* Drop the k8s-like port + host config style in favor of two addresses. Warn (but don't stop?) if the hosts diverge.
Keep the --use-tcp option, and have it simply swap the default address... more or less like it does now, and would likely in the future. Its essentially short for --buildkit-host tcp://127.0.0.1:8372 , which is the 80% case.
* Old behavior reigns if --use-tcp is not specified. Pinned town more completely:
  * Old buildkit host setting is honored, we just simply try to connect to it.
  * A new debug host setting would be ignored completely in favor of the old port setting. Port setting is depricated when the feature flag is on.
  * --use-tcp turns into a config-file only setting.

dchw:
Moved out of draft so the entrypoint can be evaluated on its own merits, and to keep this incremental.

cmd/earthly/main.go
```diff
@@ -412,10 +415,26 @@ func newEarthlyApp(ctx context.Context, console conslogging.ConsoleLogger) *eart
...
 			Name:        "buildkit-host",
+			Value:       buildkitd.DockerAddress,
 			EnvVars:     []string{"EARTHLY_BUILDKIT_HOST"},
 			Usage:       wrap("The URL to use for connecting to a buildkit host. ", "If empty, earthly will attempt to start a buildkitd instance via docker run"),
 			Destination: &app.buildkitHost,
 		},
+		&cli.StringFlag{
+			Name:        "debugger-host",
+			EnvVars:     []string{"EARTHLY_DEBUGGER_HOST"},
+			Usage:       wrap("The URL to use for connecting to a debugger host. ", "If empty, earthly use the default debugger port, combined with the desired buildkit host."),
```
vladaionescu, line 426: Nit: The `debugger-port` usage says host + port forms the URL, while this one says it's itself the URL. This should be clarified.

cmd/earthly/main.go
```diff
@@ -966,6 +1008,66 @@ func (app *earthlyApp) before(context *cli.Context) error {
...
+	if err != nil {
+		return nil, errors.Errorf("%s is not a valid URL", addr)
+	}
+
+	if parsed.Scheme != "tcp" {
+		return nil, errors.Errorf("%s is not a valid scheme. Only tcp is allowed at this time.", parsed.Scheme)
+	}
+
+	if parsed.Port() == "" {
+		return nil, errors.Errorf("%s does not contain a port number.", addr)
```
vladaionescu, line 1065: This feels like it shouldn't be an error. URLs without a port are still valid? Or maybe this is because it's `tcp://` and you can't infer a port unless it's `http://` or `https://`. Yeah, I think I gave myself the explanation :D

cmd/earthly/main.go
```diff
@@ -1032,6 +1134,10 @@ func (app *earthlyApp) processDeprecatedCommandOptions(context *cli.Context, cfg
 		app.buildkitdSettings.CacheSizeMb = cfg.Global.BuildkitCacheSizeMb
 	}
 
+	if context.IsSet("debugger-port") && app.cfg.Global.BuildkitScheme == "tcp" {
+		app.console.Warnf("Warning: specifying the port using the --buildkit-host setting is deprecated. Set it in ~/.earthly/config.yml under the global buildkit_port setting.\n")
```
vladaionescu, line 1138: Should this be like this?

```suggestion
		app.console.Warnf("Warning: specifying the port using the --debugger-port setting is deprecated. Set it in ~/.earthly/config.yml under the global debugger_port setting.\n")
```

config/config.go
```diff
@@ -26,14 +34,17 @@ type GlobalConfig struct {
...
 	BuildkitAdditionalArgs   []string `yaml:"buildkit_additional_args"   help:"Additional args to pass to buildkit when it starts. Useful for custom/self-signed certs, or user namespace complications."`
 	BuildkitAdditionalConfig string   `yaml:"buildkit_additional_config" help:"Additional config to use when starting the buildkit container; like using custom/self-signed certificates."`
 	CniMtu                   uint16   `yaml:"cni_mtu"                    help:"Override auto-detection of the default interface MTU, for all containers within buildkit"`
+	BuildkitScheme           string   `yaml:"buildkit_transport"         help:"Change how Earthly communicates with its buildkit daemon. Valid options are: docker-container, tcp. TCP is experimental."`
+	BuildkitHost             string   `yaml:"buildkit_host"              help:"The address of your buildkit, remote or local."`
+	DebuggerHost             string   `yaml:"debugger_host"              help:"The address of your debugger, remote or local."`
 
 	// Obsolete.
-	CachePath string `yaml:"cache_path"`
+	CachePath    string `yaml:"cache_path"    help:" *Depricated* The path to keep Earthly's cache."`
```
vladaionescu, line 46:
```suggestion
	CachePath    string `yaml:"cache_path"    help:" *Deprecated* The path to keep Earthly's cache."`
```

config/config.go
```diff
@@ -26,14 +34,17 @@ type GlobalConfig struct {
...
+	DebuggerPort int    `yaml:"debugger_port" help:" *Depricated* What port should the debugger (and other interactive sessions) use to communicate."`
```
vladaionescu, line 47:
```suggestion
	DebuggerPort int    `yaml:"debugger_port" help:" *Deprecated* What port should the debugger (and other interactive sessions) use to communicate."`
```

vladaionescu (approved):
Looks good! Just a few nits

cmd/earthly/main.go
dchw: This option should have been removed, actually. Keeping it as it was; it should remain in the config file only, I thought it had always been there as I was moving to the dual-URL approach.

vladaionescu (approved):

PR #1009, opened 2021-05-26
Instrument frequency of earthfile commands

alexcb:

### DIFF
--- a/analytics/analytics.go
+++ b/analytics/analytics.go
@@ -113,7 +113,11 @@ func isGitDir() bool {
 }
 
 func getRepoHash() string {
-	repo := getRepo()
+	return RepoHashFromCloneURL(getRepo())
+}
+
+// RepoHashFromCloneURL returns the repoHash of a ref
+func RepoHashFromCloneURL(repo string) string {
 	if repo == "unknown" || repo == "" {
 		return repo
 	}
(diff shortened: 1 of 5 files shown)


analytics/analytics.go
```diff
@@ -25,6 +25,9 @@ import (
 	"github.com/pkg/errors"
 )
 
+var counts = make(map[string]map[string]int)
+var countsMutex = &sync.Mutex{}
```
vladaionescu, line 29: I would put this in a struct and make that struct a global singleton. (In case we need to ever make this a non-global it's easier)

vladaionescu, line 29: Nit
```suggestion
var countsMutex sync.Mutex
```

analytics/analytics.go
```diff
@@ -202,6 +206,18 @@ func saveData(server string, data *EarthlyAnalytics) error {
...
+	countsMutex.Lock()
+	defer countsMutex.Unlock()
+	m, ok := counts[subsystem]
+	if !ok {
+		m = map[string]int{}
+		counts[subsystem] = m
+	}
+	m[key]++
+}
+
```
dchw, line 220: Why not `sync.Map`? https://golang.org/pkg/sync/#Map

alexcb: I wasn't familiar with sync.Map. I just watched [GopherCon 2017 - Lightning Talk: Bryan C Mills - An overview of sync.Map
](https://www.youtube.com/watch?v=C1EtfDnsdDs) which explains a bit about it. I'm not sure it'll be good for our use case since we're doing a lot of writes to the same key (i.e. incrementing the value of a specific key).

The go docs mention this:
```
The Map type is specialized. Most code should use a plain Go map instead, with separate locking or coordination, for better type safety and to make it easier to maintain other invariants along with the map content.

The Map type is optimized for two common use cases: (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow, or (2) when multiple goroutines read, write, and overwrite entries for disjoint sets of keys.
```

so I think it's best to stick with the current mutex+lock approach.

Furthermore, I'm unsure of sync.Map could (easily) handle being serialized to json.

analytics/analytics.go
alexcb: Sounds like a valid point, I moved these into a struct under a seperate analytics/counters.go file.

analytics/counters.go
```diff
@@ -0,0 +1,35 @@
...
+		c.counts[subsystem] = m
+	}
+	m[key]++
+}
+
+// getMap locks the Counter mutex, and returns the underlying counters map,
+// and a function that must be called when the map is no longer being used.
+func (c *Counters) getMap() (map[string]map[string]int, func()) {
+	c.mutex.Lock()
+	return c.counts, c.mutex.Unlock
```
vladaionescu, line 34: We could also deep-clone to keep the unlocking in the same function. Not a strong opinion though.

vladaionescu (approved):

Issue #500, opened 2020-11-11
SAVE IMAGE is slow, even when there's no work to be done


jazzdan:
I was observing that for highly optimized builds the slowest part can be saving images. For example, [here](https://github.com/jazzdan/earthly-save-image)’s a repo where `earth +all` takes 12s if everything is cached (like I run `earth +all` twice). Yet if I comment the `SAVE IMAGE`  lines out the total time drops to 2s. This implies that `SAVE IMAGE` is doing a lot of work, even when nothing has changed.

Is there anything I can do speed up SAVE IMAGE in instances like this? I’m surprised that SAVE IMAGE does anything if the image hasn’t changed, is it possible for it do some more sophisticated content negotiation with the layers? 

After [talking with](https://earthlycommunity.slack.com/archives/C01DL2928RM/p1605112044036800) @agbell in Slack I hypothesized that it might not be possible for Earthly to what images/layers the host has. This is all conjecture on my part, but:

If Earth is running in a container then it doesn’t know the state of the registry on the host machine, and what layers it has. It’s only option is to export the entire image to the host, which on a Mac could be slow because containers on a mac are actually running in a VM.

Maybe if Earth could mount/be aware of the host docker registry it could just do docker push?
This reminds me of similar problems that are being solved in the Kubernetes local cluster space https://github.com/kubernetes/enhancements/tree/master/keps/sig-cluster-lifecycle/generic/1755-communicating-a-local-registry

vladaionescu:
Hi @jazzdan - you are spot-on. The output of the build is passed to docker as a tar file, so there are inherent inefficiencies resulting from that. I think indeed there may be ways in which Earthly could talk to host docker while it's building and decide to skip generating tars (or even individual layers) if the image (or parts of it) is already available.

jazzdan:
Switching to smaller base images in development helped a lot, but it'd be nice to not have to do that, especially for development. :)

jazzdan:
This is becoming a bigger problem for us as we save more images from our Earthly build. I've thought of a couple interim solutions:

1) We just shouldn't make building an image a part of our dev loop, images should only be built for deploys and we should use something like Tilt to update our local kubernetes cluster without having to build images.
2) Have Earthly export the Docker context to a temp directory, and do the `docker build` outside of Earthly so that we can take advantage of the more efficient layer caching. This would require building more machinery outside of Earthly though.

vladaionescu:
[...]

PR #1010, opened 2021-05-26
Perform local exports via a buildkit-hosted local registry

vladaionescu:
Part of https://github.com/earthly/earthly/issues/500

Depends on https://github.com/earthly/buildkit/pull/39

This feature is turned off by default. To turn it on you can run

```bash
earthly config global.local_registry_host 'tcp://127.0.0.1:8371'
```

(It seems that somehow docker pull automatically trusts insecure registries on `127.0.0.1:*`)

How this works:
  * Buildkit runs a registry inside of it
  * The registry has a special storage driver built on top of a buildkit content provider. The storage driver is read-only. No pushes allowed from the outside.
  * When a build takes place,
    * the Earthly exporter registers the built images with that same buildkit content provider
    * then it pings the client via an attachable (`pullping`) that it is ready to perform pulls on a set of images
    * the client runs a series of `docker pull` commands to grab what is needed from the buildkit local registry
    * the client signals back to the exporter (via the same `pullping` call) that it has finished pulling
    * the exporter exits, cancelling the context. The context cancel triggers a GC in the content provider, so the images are no longer available to be pulled


alexcb (approved):

buildkitd/buildkitd.go
```diff
@@ -44,35 +44,35 @@ var TCPAddress = "tcp://127.0.0.1:8372"
 
 // TODO: Implement all this properly with the docker client.
 
-// NewClient returns a new buildkitd client.
-func NewClient(ctx context.Context, console conslogging.ConsoleLogger, image string, settings Settings, opts ...client.ClientOpt) (*client.Client, error) {
+// NewClient returns a new buildkitd client, together with a boolean specifying whether the buildkit is local.
```
dchw, line 47: Could we just make `isLocal` public and call it where needed above? Since all but one invocation ignores that parameter.

dchw (approved):
I'm probably nit-picking on some of these. Hence the approval. Though the port thing is the one I feel most strongly about.

buildkitd/buildkitd.go
dchw: Would just make this somewhat cleaner?

buildkitd/buildkitd.go
```diff
@@ -277,16 +277,23 @@ func Start(ctx context.Context, console conslogging.ConsoleLogger, image string,
...
 		args = append(args, "-p", fmt.Sprintf("127.0.0.1:%s:8373", dbURL.Port()))
 
-		if settings.UseTCP {
-			args = append(args, "-p", fmt.Sprintf("127.0.0.1:%s:8372", bkURL.Port()))
+		if settings.LocalRegistryAddress != "" {
+			lrURL, err := url.Parse(settings.LocalRegistryAddress)
+			if err != nil {
+				panic("Local registry address was not a URL when attempting to start buildkit")
+			}
+			args = append(args, "-p", fmt.Sprintf("127.0.0.1:%s:8374", lrURL.Port()))
```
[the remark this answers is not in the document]
vladaionescu: Done.

cmd/earthly/main.go
```diff
@@ -999,12 +1001,12 @@ func (app *earthlyApp) before(context *cli.Context) error {
 	return nil
 }
 
-func (app *earthlyApp) getBuildkitAndDebuggerAddressesOriginal() (string, string, error) {
+func (app *earthlyApp) getAddressesOriginal() (string, string, string, error) {
```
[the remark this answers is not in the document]
vladaionescu: Done.

buildkitd/buildkitd.go
vladaionescu: Done.

PR #1013, opened 2021-05-27
Add TLS support

dchw:
Added the TLS support, Can be controlled independently of the TCP setting.


vladaionescu (approved):
Very nice - I like the decoupling from the TCP connectivity.

PR #1014, opened 2021-05-27
Fix bootstrap --source trying to create earthly dir

vladaionescu:


alexcb (approved):

dchw (approved):

cmd/earthly/main.go
```diff
@@ -2722,7 +2722,9 @@ func processSecrets(secrets, secretFiles []string, dotEnvMap map[string]string)
 func defaultConfigPath() string {
 	earthlyDir, err := cliutil.GetEarthlyDir()
 	if err != nil {
-		panic(err)
+		fmt.Fprintf(os.Stderr, "Error getting earthly dir: %v\n", err.Error())
+		earthlyDir = "/etc/.earthly" // A reasonable fallback.
```
alexcb, line 2726: having it hidden under /etc/ seems a bit odd, maybe remove the dor?
```suggestion
		earthlyDir = "/etc/earthly" // A reasonable fallback.
```

vladaionescu: With the `.` it matches this https://github.com/earthly/earthly/blob/main/util/cliutil/earthlydir.go#L50

dchw (approved):

PR #1019, opened 2021-05-28
Auto install bash completion file on debian

alexcb:
I tried to make use of `dh $@ --with bash-completion` in a rules file however I just couldn't get auto-completion to persist when I logged out and back in. Ultimately I figured we could stick with the approach that homebrew uses to have the earthly binary output the auto-completion contents.

### DIFF
--- a/release/apt-repo/postinst
+++ b/release/apt-repo/postinst
@@ -1,0 +1,32 @@
+#!/usr/bin/env bash
+set -e
+
+# install bash auto completion
+BASH_COMPLETION_DIR="/usr/share/bash-completion/completions"
+if [ -d "$BASH_COMPLETION_DIR" ]
+then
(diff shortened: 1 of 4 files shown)


release/apt-repo/postinst
```diff
@@ -0,0 +1,32 @@
...
+    earthly bootstrap --source zsh > "$ZSH_COMPLETION_DIR/_earthly"
+fi
+
+# skip bootstraping if docker isn't installed or running
+if ! command -v docker &> /dev/null
+then
+    echo "docker was not found; skipping earthly bootstrap"
+    exit
+fi
+if ! docker info 2>/dev/null >/dev/null
```
vladaionescu, line 24: Very nice

vladaionescu (approved):
:+1: :cool:

Issue #1026, opened 2021-05-31
Earthly 0.5.15 fails to run commands


omerbenamram:
Hi,

After updating earthly from  0.5.13 to  0.5.15, every command fails with some variation of this message:
```
Error getting earthly dir: lookup user for homedir: user: unknown userid 116002263
```

Reverting to  0.5.13 fixes it.

Wondering if this is related to https://github.com/earthly/earthly/pull/977?

alexcb:
Thanks for the report @omerbenamram , can you let us know what OS you are using?

omerbenamram:
@alexcb Ubuntu 18.04 :)

alexcb:
@omerbenamram I haven't been able to replicate the issue, but I've tried changing some error handling in the home directory detection code.

Since I can't replicate the issue, would you be able to try out this new code to see if this fixes the issue for you?

You should be able to compile earthly out of my branch via:

    mkdir /tmp/test-earthly-fix && cd /tmp/test-earthly-fix && git clone https://github.com/earthly/earthly.git . && git checkout ignore-errors-when-detecting-home-dir && earthly +for-linux && ls ./build/linux/amd64/earthly

Then if you run `./build/linux/amd64/earthly --version`, you should see the git branch and sha referenced:

    earthly version ignore-errors-when-detecting-home-dir-95fe6666001aeb4031be7c174681cefd012354ca linux/amd64

Thanks.

omerbenamram:
This version indeed runs as expects :)

alexcb:
Thanks for giving it a test :)

I'll work on getting this branch ready to merge into main, it might be tomorrow by the time we get a fix released.

alexcb:
This has been released under https://github.com/earthly/earthly/releases/tag/v0.5.16

PR #1027, opened 2021-05-31
Ignore errors while detecting earthly home dir

alexcb:
Earthly attempts to find the config path under the $SUDO_USER's ~/.earthly path, otherwise if $SUDO_USER isn't set, then under $HOME/.earthly, then the users' home directory (for cases where the HOME env isn't set), then if all of those fail, "/etc". If any of these heuristics fail, errors will be silently ignored.

`GetEarthlyDir()` has been changed to always return a string, without any error as the heuristic will always fallback to `/etc`.

A new function: `GetOrCreateEarthlyDir()` has been created which will call `GetEarthlyDir()` and additionally create the path if it does not exist. This function will continue to return errors which occur during directory creation.

This PR additionally deduplicates some of the path look up code to ensure there's only one mechanism that constructs the ".earthly" path, and also prevents the `before` handler from calling the homebrew specific `--source` handler code which should only ever be executed via `earthly bootstrap --source "bash|zsh"`.

### DIFF
--- a/buildcontext/gitlookup.go
+++ b/buildcontext/gitlookup.go
@@ -240,10 +240,7 @@ func (gl *GitLookup) detectProtocol(host string) (protocol string, err error) {
 var errNoRCHostEntry = fmt.Errorf("no netrc host entry")
 
 func (gl *GitLookup) lookupNetRCCredential(host string) (login, password string, err error) {
-	homeDir, _, err := cliutil.DetectHomeDir()
-	if err != nil {
-		return "", "", errors.Wrap(err, "failed to detect home dir")
-	}
+	homeDir, _ := cliutil.DetectHomeDir()
 	n, err := netrc.Parse(filepath.Join(homeDir, ".netrc"))
 	if err != nil {
 		return "", "", errors.Wrap(err, "failed to parse .netrc")
--- a/util/cliutil/earthlydir.go
+++ b/util/cliutil/earthlydir.go
@@ -84,7 +84,11 @@ func getHomeFromUserCurrent() (string, *user.User, bool) {
 	if u.HomeDir == "" {
 		return "", nil, false
 	}
-	return u.HomeDir, u, true
+
+	// do NOT return the user here, because the user is only
+	// required during the SUDO_USER case; where as this case
+	// the permissions will belong to the current user and won't need changing.
+	return u.HomeDir, nil, true
 }
 
 // DetectHomeDir returns the home directory of the current user, together with
--- a/util/cliutil/earthlydir.go
+++ b/util/cliutil/earthlydir.go
@@ -11,96 +11,109 @@ import (
 )
 
 var earthlyDir string
-var earthlyDirErr error
 var earthlyDirOnce sync.Once
+var earthlyDirSudoUser *user.User
+
+var earthlyDirCreateOnce sync.Once
+var earthlyDirCreateErr error
 
 // GetEarthlyDir returns the .earthly dir. (Usually ~/.earthly).
-func GetEarthlyDir() (string, error) {
+// This function will not attempt to create the directory if missing, for that functionality use to the GetOrCreateEarthlyDir function.
+func GetEarthlyDir() string {
 	earthlyDirOnce.Do(func() {
-		earthlyDir, earthlyDirErr = makeEarthlyDir()
+		earthlyDir, earthlyDirSudoUser = getEarthlyDirAndUser()
 	})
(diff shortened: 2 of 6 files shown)


alexcb:
This PR attempts to fix #1026 (which i haven't been able to replicate).

alexcb:
While I can't replicate the issue, @omerbenamram gave it a test and [confirmed it fixed the issue](https://github.com/earthly/earthly/issues/1026#issuecomment-851693811)

I did a quick sanity test to ensure the home directory heuristic is still working with `SUDO_USER`, and it is picking it up:

```
$ sudo rm -rf /root/.earthly ~/.earthly
[sudo] password for alex: 

$ sudo ./build/linux/amd64/earthly bootstrap
           buildkitd | ...Done
           bootstrap | Bootstrapping successful.

$ sudo ls /root/.earthly
ls: cannot access '/root/.earthly': No such file or directory

$ ls ~/.earthly
install_id
```

buildcontext/gitlookup.go
```diff
@@ -240,10 +240,7 @@ func (gl *GitLookup) detectProtocol(host string) (protocol string, err error) {
 var errNoRCHostEntry = fmt.Errorf("no netrc host entry")
 
 func (gl *GitLookup) lookupNetRCCredential(host string) (login, password string, err error) {
-	homeDir, _, err := cliutil.DetectHomeDir()
-	if err != nil {
-		return "", "", errors.Wrap(err, "failed to detect home dir")
-	}
+	homeDir, _ := cliutil.DetectHomeDir()
```
vladaionescu, line 243: This means that the fallback is `""` right?

I guess that's the best we can do if we tried everything.

util/cliutil/earthlydir.go
```diff
@@ -11,96 +11,109 @@ import (
...
-	if err != nil {
-		return nil, false, err
+// IsBootstrapped provides a tentatively correct guess about the state of our bootstrapping.
+func IsBootstrapped() bool {
+	return fileutil.DirExists(GetEarthlyDir())
+}
+
+// EnsurePermissions changes the permissions of all earthly files to be owned by the user and their group.
+func EnsurePermissions() error {
+	earthlyDir, sudoUser := getEarthlyDirAndUser()
```
vladaionescu, line 114: I *think* that the user is returned even if it's not a `SUDO_USER` case.

I think the logic is ok either way, but perhaps we shouldn't name the variable `sudoUser` just to make it less misleading.

vladaionescu (approved):
Looks good - mostly some nits

util/cliutil/earthlydir.go
```diff
@@ -11,96 +11,109 @@ import (
...
 var earthlyDirOnce sync.Once
+var earthlyDirSudoUser *user.User
+
+var earthlyDirCreateOnce sync.Once
+var earthlyDirCreateErr error
 
 // GetEarthlyDir returns the .earthly dir. (Usually ~/.earthly).
-func GetEarthlyDir() (string, error) {
+// This function will not attempt to create the directory if missing, for that functionality use to the GetOrCreateEarthlyDir function.
+func GetEarthlyDir() string {
```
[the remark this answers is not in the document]
alexcb: The best part about changing the function signature, is the compiler failed on all parts where it was referenced. In nearly all cases it was changed to use `GetOrCreateEarthlyDir()`.

buildcontext/gitlookup.go
alexcb: The fallback is `"/etc"`, which happens under https://github.com/earthly/earthly/pull/1027/files#diff-b8cb87e3ddc4c19120c5a344c081c9f68b80be59f98f1d1bf7b73ae791c4cbeaR104

This function should never return `""`.

vladaionescu: Ah right :+1:

util/cliutil/earthlydir.go
alexcb: that's a good question, I intended on it only ever being the sudo user, but I see we could potentiall return a different user in this function: https://github.com/earthly/earthly/blob/7015379873db0c51b7791b5527961e5f449496ea/util/cliutil/earthlydir.go#L79-L88

I think it's best for us to return nil here as there would be no need to change the permissions for this case.

alexcb: Here's the change for this issue: https://github.com/earthly/earthly/commit/3346674201e92a84c097192bbbc84952fbf9976e

PR #1028, opened 2021-06-01
Add Experimental Windows Build

dchw:
This should wrap up most of the needs for experimental usage. Docs coming ahead of next release.


Earthfile
```diff
@@ -249,12 +249,23 @@ earthly-darwin-arm64:
...
+        +earthly/* ./
+    SAVE ARTIFACT ./*
+
 earthly-all:
     COPY +earthly-linux-amd64/earthly ./earthly-linux-amd64
     COPY +earthly-linux-arm7/earthly ./earthly-linux-arm7
     COPY +earthly-linux-arm64/earthly ./earthly-linux-arm64
     COPY +earthly-darwin-amd64/earthly ./earthly-darwin-amd64
     COPY +earthly-darwin-arm64/earthly ./earthly-darwin-arm64
+    COPY +earthly-windows-amd64/earthly ./earthly-windows-amd64
```
vladaionescu, line 268:
```suggestion
    COPY +earthly-windows-amd64/earthly.exe ./earthly-windows-amd64.exe
```

util/llbutil/copyop.go
```diff
@@ -16,7 +16,7 @@ import (
 func CopyOp(srcState pllb.State, srcs []string, destState pllb.State, dest string, allowWildcard bool, isDir bool, keepTs bool, chown string, ifExists bool, opts ...llb.ConstraintsOpt) pllb.State {
 	destAdjusted := dest
 	if dest == "." || dest == "" || len(srcs) > 1 {
-		destAdjusted += string(filepath.Separator)
+		destAdjusted += string("/") // TODO: needs to be the containers platform, not the earthly hosts platform. For now, this is always Linux.
```
vladaionescu, line 19: :+1:

vladaionescu (commented):
Can you also add tests like these for the release?

https://github.com/earthly/earthly/blob/main/release/Earthfile#L51-L63

dchw:
> Can you also add tests like these for the release?

On it.

vladaionescu (approved):
All looks good.

I think you may have trouble with getting the GHA script to work. But hey.. worth a shot.

It may be worth adding a windows runner to Buildkite instead.

vladaionescu (approved):

dchw:
CI on this is still a WIP. Trying to use the WSL agent to launch this native binary results in a hang, and I have raise the issue here: https://github.com/buildkite/agent/issues/1447.

There are two (potentially) easy ways forward:
* (Re)-push the WSL dependency down. This means using the native Windows Buildkite agent, and invoking `earthly` via WSL directly. IIRC we are doing it this way for speed reasons? Perhaps we would re-learn a bad lesson there.
* Spin up another Windows box, and add the native Windows Buildkite Agent. This would keep any speed gains we got from WSL at the cost of managing another machine.

We cannot:
* Use the Buildkite Elastic stack for Windows. Unless we could get higher utilization by _also_ moving our WSL builds over, I am not sure its worth the even greater complexity and potential cost; especially for our relatively simple needs.
*  Run a native Windows agent alongside the WSL2 agent. This would result in conflict fighting over the `earthly-buildkitd` container name.
  * Though, allowing configuration for this name would certainly allow us to add density fairly easily? :thinking:

dchw:
@vladaionescu Would love to merge this to avoid it becoming stale. CI will continue outside this PR?

PR #1034, opened 2021-06-04
Parsing for `FOR ... IN ...`

vladaionescu:
* Parsing and syntax highlighting for `FOR`
* Also syntax highlighting for `VERSION`

Interpreter part not yet implemented.

Part of #779


alexcb (approved):

alexcb:
after looking through the PR thinking how does this work, I noticed `Interpreter part not yet implemented.` :laughing: 
looks like a good start.

PR #1040, opened 2021-06-07
Shellrepeater client doesn't handle the specifying the network as a URL Scheme

dchw:
This should surface any future errors properly as well.

The TCP is there to help ensure some consistency in how we specify addresses that connect to external services. This also allows us to implement shared validation for these URLs as well.

cc @alexcb

### DIFF
--- a/cmd/earthly/main.go
+++ b/cmd/earthly/main.go
@@ -2629,7 +2629,20 @@ func (app *earthlyApp) actionBuildImp(c *cli.Context, flagArgs, nonFlagArgs []st
 	cleanCollection := cleanup.NewCollection()
 	defer cleanCollection.Close()
 
-	go terminal.ConnectTerm(c.Context, app.debuggerHost)
+	go func() {
+		// Dialing doesnt accept URLs, it accepts an address and a "network". These cannot be handled as URL schemes.
+		// Since Shellrepeater hard-codes TCP, we drop it here and log the error if we fail to connect.
+
+		u, err := url.Parse(app.debuggerHost)
+		if err != nil {
+			panic("debugger host was not a URL")
+		}
+
+		err = terminal.ConnectTerm(c.Context, u.Host)
+		if err != nil {
+			app.console.Warnf("Failed to connect to terminal: %s", err.Error())
+		}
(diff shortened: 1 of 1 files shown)


cmd/earthly/main.go
```diff
@@ -2629,7 +2629,20 @@ func (app *earthlyApp) actionBuildImp(c *cli.Context, flagArgs, nonFlagArgs []st
...
 	defer cleanCollection.Close()
 
-	go terminal.ConnectTerm(c.Context, app.debuggerHost)
+	go func() {
+		// Dialing doesnt accept URLs, it accepts an address and a "network". These cannot be handled as URL schemes.
+		// Since Shellrepeater hard-codes TCP, we drop it here and log the error if we fail to connect.
+
+		u, err := url.Parse(app.debuggerHost)
+		if err != nil {
+			panic("debugger host was not a URL")
```
dchw, line 2638: This `panic` is consistent with what we do if parsing fails when starting buildkit.

alexcb: should we also error out if `u.Scheme != "tcp"`?

cmd/earthly/main.go
```diff
@@ -2629,7 +2629,20 @@ func (app *earthlyApp) actionBuildImp(c *cli.Context, flagArgs, nonFlagArgs []st
...
+		}
+
+		err = terminal.ConnectTerm(c.Context, u.Host)
+		if err != nil {
+			app.console.Warnf("Failed to connect to terminal: %s", err.Error())
```
alexcb, line 2643: :tada: :+1: :1st_place_medal: :ship:

alexcb (approved):

cmd/earthly/main.go
dchw: We already do. See the URL validation we do in `before`: https://github.com/earthly/earthly/blob/b3f056b510d822731710f3e463abbc27ed19fcd7/cmd/earthly/main.go#L1111-L1126

alexcb: awesome! thanks for pointing that out.

PR #1041, opened 2021-06-07
Add verbose calls to conslogging

alexcb:
New Debugf and DebugBytes will print their messages only when
conslogging is running under verbose mode.

### DIFF
--- a/conslogging/conslogging.go
+++ b/conslogging/conslogging.go
@@ -46,6 +46,7 @@ type ConsoleLogger struct {
 	colorMode ColorMode
 	isCached  bool
 	isFailed  bool
+	verbose   bool
 
 	// The following are shared between instances and are protected by the mutex.
 	mu             *sync.Mutex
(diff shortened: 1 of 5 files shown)


conslogging/conslogging.go
```diff
@@ -245,6 +247,20 @@ func (cl ConsoleLogger) PrintBytes(data []byte) {
 	}
 }
 
+// Debugf prints formatted text to the console when verbose flag is set.
+func (cl ConsoleLogger) Debugf(format string, args ...interface{}) {
```
vladaionescu, line 251: Wondering if we should call this something like `Verbosef` or `VerbosePrintf`. The word debug can be overloaded when talking about printing: is it for us trying to debug Earthly, or is it for the user trying to debug their Earthfile.

alexcb: that's a good point. I was initially thinking that perhaps one day we could even extend it to be more like syslog and have levels like: Debug, Info, Warn, Error, then rather than having verbose=true/false, we could have more control over setting the levels, but that doesn't address the `-V` vs -D` flags we currently have.

vladaionescu (approved):

PR #1042, opened 2021-06-07
Document --if-exists options

alexcb:

### DIFF
--- a/docs/earthfile/earthfile.md
+++ b/docs/earthfile/earthfile.md
@@ -338,6 +338,10 @@ Instructs Earthly to not overwrite the file creation timestamps with a constant.
 
 Instructs Earthly to keep file ownership information. This applies only to the *artifact form* and has no effect otherwise.
 
+##### `--if-exists`
+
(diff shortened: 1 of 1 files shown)


docs/earthfile/earthfile.md
```diff
@@ -338,6 +338,10 @@ Instructs Earthly to not overwrite the file creation timestamps with a constant.
 
 Instructs Earthly to keep file ownership information. This applies only to the *artifact form* and has no effect otherwise.
 
+##### `--if-exists`
+
+Only copy source if it exists; if it does not exist, earthly will simply ignore the COPY command and won't treat any missing sources as failures.
```
vladaionescu, line 343: This isn't yet available for the classical version of `COPY`, I believe: https://github.com/earthly/earthly/blob/main/earthfile2llb/converter.go#L411 . It should be easy to support though.

vladaionescu: It should be easy to add though

alexcb: I changed this PR to go into `next`, now that https://github.com/earthly/earthly/pull/1174 has been merged.

vladaionescu (approved):

PR #1045, opened 2021-06-07
Use conslogging for logging debug term messages

alexcb:
Since the debug terminal is called within earthly, it should use the
conslogger along with an "internal-term" prefix. The previous logrus
logger was no correctly logging debug messages due to

    logrus.StandardLogger().Out = ioutil.Discard

being set (to prevent other imported libraries from logging).

### DIFF
--- a/debugger/terminal/terminal_other.go
+++ b/debugger/terminal/terminal_other.go
@@ -11,8 +11,8 @@ import (
 	"sync"
 	"syscall"
 
+	"github.com/earthly/earthly/conslogging"
 	"github.com/earthly/earthly/debugger/common"
-	"github.com/earthly/earthly/logging"
 
 	"github.com/creack/pty"
 	"github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


debugger/terminal/terminal_other.go
```diff
@@ -11,8 +11,8 @@ import (
 	"sync"
 	"syscall"
 
+	"github.com/earthly/earthly/conslogging"
 	"github.com/earthly/earthly/debugger/common"
-	"github.com/earthly/earthly/logging"
```
vladaionescu, line 15: Any chance we can get rid of this package entirely?

vladaionescu (approved):

debugger/terminal/terminal_other.go
alexcb: It's used by the shellrepeater for logging; and is designed for structured-logging for use in servers. Maybe we should rename it to `structuredlogging`?

vladaionescu: Ok no worries. Maybe something shorter, like `slog`? In any case, this is unrelated to this PR.

alexcb: I'll open a new PR for that rename.

PR #1050, opened 2021-06-08
Bless Earthly Docker Image

dchw:
Revamp entrypoint to make it more user-friendly and button it down more. Dogfoods it within our own unit testing. Docs coming once merged so I don't have to keep those up to date too with any changes here.

Sample invocation 1:

`docker run --privileged -t -e NO_DOCKER=1 -v $(pwd):/src earthly/earthly:v0.5.15 +for-linux`

 * `-t` lets Earthly detect that there is a TTY. Better than using `FORCE_COLOR`, when possible.
 * `-e NO_DOCKER=1` Skips the Docker check. Useful for when you don't need to save images, etc.
 * `-v $(pwd):/src` is mounting the current dir into the conventional spot for invoking earthly.

Sample invocation 2:

`docker run --privileged -t -v ~/.docker:/root/.docker -v /var/run/docker.sock:/var/run/docker.sock -e DOCKER_HOST="unix:///var/run/docker.sock" -v $(pwd):/src earthly/earthly:v0.5.15 +for-linux`

* Note that we mount the Docker socket plus specify the `DOCKER_HOST` variable. If the daemon is remote or accessible otherwise, the socket mount would not be needed. Also note mounting `.docker`; this preserves any credentials. Should we support a `docker login` here?


Earthfile
```diff
@@ -284,12 +283,15 @@ earthly-docker:
 earthly-integration-test-base:
     FROM +earthly-docker
     ENV EARTHLY_CONVERSION_PARALLELISM=5
-    RUN earthly config global.disable_analytics true
-    RUN earthly config global.local_registry_host 'tcp://127.0.0.1:8371'
+    ENV GLOBAL_CONFIG={disable_analytics: true, local_registry_host: 'tcp://127.0.0.1:8371'}
+    ENV NO_DOCKER=1
+    ENV SRC_DIR=/test
```
dchw, line 288: The above options are ones from the new image that would also be user-facing.

`GLOBAL_CONFIG` should be self explanatory.
`NO_DOCKER` exists in case you are in an environment where BYOD is not possible, and you want to still build artifacts. It skips the check for connectivity to a live Docker daemon.
`SRC_DIR` lets you override where you will execute earthly from. By default it is `/src`, but here all our tests are in `/test`, which exercises the option.

Earthfile
```diff
@@ -284,12 +283,15 @@ earthly-docker:
...
+    ENV NETWORK_MODE=host
```
dchw, line 289: Moved from the base image config since its our tests that need `host`, not the base image (I've been able to run `+for-linux` under the image without using host networking).

earthly-entrypoint.sh
```diff
@@ -0,0 +1,70 @@
...
+
+  # Light check if docker is functional
+  if ! docker images > /dev/null 2>&1; then
+    echo "Docker appears not to be connected. Please check your DOCKER_HOST variable, and try again."
+    exit 1
+  fi
+fi
+
+# Is container running as privileged? This is currently required.
+if ! captest --text | grep sys_admin > /dev/null; then
```
dchw, line 35: Perhaps theres a better cap to check for? This seemed the most straightforward. Would love to enumerate all the caps we actually need and check for them all...

examples/tests/Earthfile
```diff
@@ -228,8 +228,7 @@ star-test-todo:
         test "$cached_lines" == "6"
 
 dockerfile-test:
-    COPY --dir dockerfile ./
-    WORKDIR /test/dockerfile
+    COPY dockerfile/* ./
```
dchw, line 231: Fix working dir so it blends in with the rest of the tests appropriately.

examples/tests/version/Earthfile
```diff
@@ -3,6 +3,8 @@ FROM ../../..+earthly-integration-test-base --DOCKERHUB_AUTH=$DOCKERHUB_AUTH
 
 IMPORT .. AS tests
 
+WORKDIR /test
```
dchw, line 6: Fix working directory to fall in line with our integration tests `SRC_DIR`.

earthly-entrypoint.sh
```diff
@@ -0,0 +1,68 @@
...
+
+# Apply global configuration
+if [ -n "$GLOBAL_CONFIG" ]; then
+  earthly config global "$GLOBAL_CONFIG"
+fi
+
+# Apply git configuration
+if [ -n "$GIT_CONFIG" ]; then
+  earthly config git "$GIT_CONFIG"
+fi
```
vladaionescu, line 13: I've seen other images allow to mount a config file instead. It might be cleaner and more future-proof as just a file maybe?

earthly-entrypoint.sh
```diff
@@ -0,0 +1,68 @@
...
+
+# We prefer you use EARTHLY_TOKEN directly, but this is here just in case
+if [ -n "$EARTHLY_EMAIL" ] && [ -n "$EARTHLY_PASSWORD" ]; then
+  earthly account login --email "$EARTHLY_EMAIL" --password "$EARTHLY_PASSWORD"
+fi
```
vladaionescu, line 18: Maybe it's better if we don't support the less secure option at all and just require people to create tokens.

earthly-entrypoint.sh
vladaionescu: Maybe we shouldn't require privileged if the user brings their own ~bottle~ buildkit?

earthly-entrypoint.sh
```diff
@@ -0,0 +1,68 @@
...
+  EARTHLY_BUILDKIT_HOST="tcp://$(hostname):8372" # hostname is not recognized as local for this reason
+  export EARTHLY_BUILDKIT_HOST
+else
+  export EARTHLY_BUILDKIT_HOST="$BUILDKIT_HOST"
+fi
+
+echo "Using $EARTHLY_BUILDKIT_HOST as buildkit daemon"
+
+# Use the desired target dir for running a target, saves typing if you use the convention
+BASE_DIR="/src"
```
vladaionescu, line 59: I've seen `/workspace` as being a common name for this.

vladaionescu (approved):
Mostly nits, hence approve

earthly-entrypoint.sh
dchw: :thinking: Good point. Ill move the check inside.

earthly-entrypoint.sh
dchw: I went back-and-forth here. Nothing is stopping you from mounting that as-is; but, as an example within our own testing, the mount option is unavaliable, and this felt nicer than a bunch of `RUN earthly config` lines.

I'll invoke `earthly` with `--config /some/well/known/path`, and document that you can mount it too. These options will stay for when mounting isn't an option.

earthly-entrypoint.sh
dchw: Ill drop this then.

earthly-entrypoint.sh
```diff
@@ -0,0 +1,68 @@
...
+  # Light check if docker is functional
+  if ! docker images > /dev/null 2>&1; then
+    echo "Docker appears not to be connected. Please check your DOCKER_HOST variable, and try again."
+    exit 1
+  fi
+fi
+
+# Is container running as privileged? This is currently required.
+if ! captest --text | grep sys_admin > /dev/null; then
+  echo "Container appears to be running unprivilged. Currently, privileged mode is required."
```
alexcb, line 36:
```suggestion
  echo "Container appears to be running unprivileged. Currently, privileged mode is required."
```

earthly-entrypoint.sh
vladaionescu: Don't see why not - just needs to be documented.

PR #1051, opened 2021-06-08
Log transferring

alexcb:
log files which are sent to buildkit
    
When running in verbose mode, log which files are sent to buildkit including file stats and files which are excluded (ignored).

In regular (non-verbose) mode, the progressbars have been replaced with a message displaying total bytes sent:

image

and under verbose mode, it displays all files:

image

The progress bar had to be removed as it was overwriting the final output of text while running in verbose mode, and to be able to programatically disable it for only verbose mode would have required a fair amount of changes to the buildkit code.

This depends on https://github.com/earthly/buildkit/pull/44


buildkitd/Earthfile
```diff
@@ -11,7 +11,7 @@ buildkitd:
             ARG BUILDKIT_BASE_IMAGE=$BUILDKIT_PROJECT+build
         END
     ELSE
-        ARG BUILDKIT_BASE_IMAGE=github.com/earthly/buildkit:c896a903b73a001a756c8d40330874451dbe38c4+build
+        ARG BUILDKIT_BASE_IMAGE=github.com/earthly/buildkit:907351dab34de3d22aeebea7a03593473ab7a1f0+build
```
alexcb, line 14: I will update this once https://github.com/earthly/buildkit/pull/44 is merged.

vladaionescu:
I wonder if the file stat updates can be made less verbose. For example by printing how many files+dirs were stat'd per tree, instead of listing every single file.

alexcb:
> I wonder if the file stat updates can be made less verbose. For example by printing how many files+dirs were stat'd per tree, instead of listing every single file.

We would need some additional changes to fsutils to differentiate between files and dirs, but we could count the total number of stats per walk.

Another possibility would be to change `-V` to handle multiple `V`s which would increase the verbosity level, so at `-V` we could just skip printing stats, and at `-VV` we would show them.

vladaionescu:
> We would need some additional changes to fsutils to differentiate between files and dirs, but we could count the total number of stats per walk.

Yeah, total number would be great.

I'd avoid multiple levels of verbosity, to keep things simple.

alexcb:
this is a little tidier:

image

vladaionescu:
That looks really good

alexcb:
another small tweak for the text:

image

buildcontext/local.go
```diff
@@ -38,7 +39,7 @@ func (lr *localResolver) resolveLocal(ctx context.Context, ref domain.Reference)
 				// remote not detected.
 				if errors.Is(err, gitutil.ErrNoGitBinary) {
 					// TODO: Log this properly in the console.
-					fmt.Printf("Warning: %s\n", err.Error())
+					lr.console.Warnf("Warning: %s\n", err.Error())
```
vladaionescu, line 42: Remove the TODO comment?

vladaionescu (approved):
Just a nit

buildcontext/local.go
alexcb: hah, good point :)

PR #1053, opened 2021-06-10
Docker image documentation

dchw:
Tried to add CI to share these documentation pages with the appropriate DockerHub repositories.

*Edit:* Re-reading some from the action, I have some doubts if we can actually do this. If we cannot, then I will replace the CI step with a manual step in the release process :(


docs/docker-images/all-in-one.md
```diff
@@ -0,0 +1,76 @@
...
+
+### Simple Usage
+
+```bash
+$ docker run --privileged -t -e NO_DOCKER=1 -v $(pwd):/workspace earthly/earthly:latest +for-linux
+```
+
+Heres a quick breakdown:
+
+- `--privileged` is required when you are using the internal `buildkit`. This is because `buildkit` currently requires is when used with `earthly`.
```
vladaionescu, line 15:
```suggestion
- `--privileged` is required when you are using the internal `buildkit`. This is because `buildkit` currently requires it for overlayfs support.
```

docs/docker-images/all-in-one.md
```diff
@@ -0,0 +1,76 @@
+This image contains `earthly`, `buildkit`, and some extra configuration to enable the two to work together. All that's missing is your source code! This image is mainly intended for use in containerized CI scenarios, or where maintaining a persistent installation of `earthly` isn't possible.
+
+## Quickstart
+
+Want to just get started? Here are a couple sample `docker run` commands that cover the most common use-cases:
+
+### Simple Usage
+
+```bash
+$ docker run --privileged -t -e NO_DOCKER=1 -v $(pwd):/workspace earthly/earthly:latest +for-linux
```
vladaionescu, line 10: Mind removing the `$`? We've not used `$` in other snippets, to allow the user to copy-paste easier.

docs/docker-images/all-in-one.md
```diff
@@ -0,0 +1,76 @@
...
+- `--privileged` is required when you are using the internal `buildkit`. This is because `buildkit` currently requires is when used with `earthly`.
+- `-t` tells Docker to emulate a TTY. This makes the `earthly` log output colorized.
+- `-e NO_DOCKER=1` skips the check for a functional Docker daemon. If you are not exporting any images, you do not need Docker.
+- `-v $(pwd):/workspace` mounts the source code into the conventional location within the docker container. Earthly is executed from this directory when starting the container. Any artifacts saved within this folder remain on your local machine.
+- `+for-linux` is the target to be invoked. All arguments specified after the image tag will be passed to `earthly`.
+
+### More Complicated Usage
+
+```bash
+$ docker run -t --privileged -v $(pwd):/workspace:rw --network=host -v ~/.earthly/config.yml:/etc/.earthly/config.yml -e DOCKER_HOST="tcp://0.0.0.0:2375" earthly/earthly:corey_entrypoint --ci -P +for-linux
```
vladaionescu, line 24: `$`

vladaionescu, line 24: It's more common to use `/var/run/docker.sock` for Docker. Would that be a better example?

docs/docker-images/buildkit-standalone.md
```diff
@@ -0,0 +1,61 @@
...
+| CNI_MTU                             | MTU of first default interface | Set this when we autodetect the MTU incorrectly. The device used for autodetection can be shown by the command  `ip route show \| grep default \| cut -d' ' -f5 \| head -n 1` |
+| EARTHLY_RESET_TMP_DIR               | `false`                        | Cleans out `EARTHLY_TMP_DIR` before running, if set to `true`. Useful when you host-mount an temp dir across runs.                                                            |
+| NETWORK_MODE                        | `cni`                          | Specifies the networking mode of `buildkitd`. Default uses a CNI bridge network, configured with the `CNI_MTU`.                                                               |
+| EARTHLY_TMP_DIR                     | `/tmp/earthly`                 | Specifies the location of `earthly`s temp dir. You can also mount an external volume to this path to preserve the contents across runs.                                       |
```
vladaionescu, line 59: This reminds me - we should recommend people to mount a Docker volume to this dir almost all the time when using buildkitd (buildkit image and earthly image with internal buildkit).

Reason: If the underlying FS does not support overlayfs on top of it then buildkit will switch to the dumb snapshoter (I think it's called native?), which instead of using copy-on-write layers, it just copies the entire file system on every single layer. This makes disk space run out really fast, which is very inconvenient.

If you don't mount anything there, then it's basically overlayfs on top of overlayfs, which is not supported.

For the most part, only `ext4` and Docker volumes are supported. So most of the time, on Linux you could mount a host volume, but on Mac and Windows, the host volume would not work with overlayfs. But Docker volumes are supported in any situation.

docs/docker-images/buildkit-standalone.md
```diff
@@ -0,0 +1,61 @@
...
+
+Assuming you ran this on another machine named `fast-builder`, you could use this remote `buildkitd` by setting `EARTHLY_BUILDKIT_HOST=tcp://fast-builder:8372`, or by specifying the address in your `config.yml`.
+
+## Using This Image
+
+### Requirements
+
+#### Privileged Mode
+
+This image needs to be run as a privileged container. This is because `buildkitd` needs appropriate access to start and run additional containers itself via `runc`.
```
vladaionescu, line 40: I think it is mainly needed for the non-fuse overlayfs, I believe.

docs/docker-images/buildkit-standalone.md
```diff
@@ -0,0 +1,61 @@
...
+
+#### External Usage
+
+To use this image externally, it requires you to forward a port on your machine to the containers port 8372. You will need to ensure that external access ti the machine on the port you chose is possible as well.
```
vladaionescu, line 44:
```suggestion
To use this image externally, it requires you to forward a port on your machine to the containers port 8372. You will need to ensure that external access to the machine on the port you chose is possible as well.
```

docs/docker-images/buildkit-standalone.md
```diff
@@ -0,0 +1,61 @@
...
+
+If you are using this container locally, please note that `EARTHLY_BUILDKIT_HOST` values where the host is `127.0.0.1`, ` ::1/128`, or `localhost` are considered local and will result in Earthly attempting to manage the buildkit container itself. Consider using your hostname, or another alternative name in these cases.
```
vladaionescu, line 46: When I first read this it made me think that EARTHLY_BUILDKIT_HOST is a setting of buildkit given the topic of this page. Maybe clarify this a bit more?

docs/docker-images/buildkit-standalone.md
```diff
@@ -0,0 +1,61 @@
...
+| EARTHLY_ADDITIONAL_BUILDKIT_CONFIG  |                                | Additional `buildkitd` config to append to the generated configuration file.                                                                                                  |
+| BUILDKIT_LOCAL_REGISTRY_LISTEN_PORT | `8371`                         | What port should the internal cache registry listen on?                                                                                                                       |
+| BUILDKIT_TCP_TRANSPORT_ENABLED      |                                | Set to `true` when the `buildkitd` instance is going to be used remotely                                                                                                      |
+| BUILDKIT_TLS_ENABLED                |                                | Set to `true` when the `buildkittd` instance will require mTLS from the clients. You will also need to mount certificates into the right place (`/etc/*.pem`).                |
```
vladaionescu, line 55: I wonder if this should be enabled automatically when using TCP so that it is secure by default. So then we would have instead BUILDKIT_TLS_DISABLED or BUILDKIT_TLS_INSECURE.

docs/docker-images/buildkit-standalone.md
```diff
@@ -0,0 +1,61 @@
+This image contains `buildkit` with some Earthly-specific setup. This is what Earthly will start when using a local daemon. You can also start it up yourself and use it as a remote/shared buildkit daemon.
+
+## Quickstart
```
vladaionescu, line 3: Docker Hub images often have some information about the tags available. We should mention that `latest` points to the latest released version and that `prerelease` points to the latest build on `main`.

For `buildkit`, we've only tested versions of the image that have the exact same version as `earthly` itself. Perhaps this is something we can improve in the future, but for now perhaps we should recommend that people use the exact same version of earthly and buildkit.

vladaionescu (commented):
Looks good so far!

docs/docker-images/all-in-one.md
dchw: Yes; but why would you when you are already running privileged? The earlier example works locally; this was intended to show you could use an external `DOCKER_HOST`, as you might in a CI.

docs/docker-images/all-in-one.md
dchw: Preventing blind copy paste was why I did it ;) I can remove it if you would like. These CLIs will almost always require tweaking.

docs/docker-images/all-in-one.md
dchw: I tweaked the address and wording a bit to make this less host-centric, at the expense of the command not working out of the box for a user.

docs/docker-images/buildkit-standalone.md
dchw: Added a section with less detail.

docs/docker-images/buildkit-standalone.md
dchw: With this image, the code to generate the certs does not exist. Using this standalone is a strictly BYOCerts affair.

agbell (approved):

docs/docker-images/all-in-one.md
vladaionescu: > Yes; but why would you when you are already running privileged?

One reason is that Docker automatically creates `/var/run/docker.sock`, but it doesn't automatically listen on TCP.

I think both are good examples. Perhaps the table mentioning DOCKER_HOST could mention both examples.

docs/docker-images/all-in-one.md
```diff
@@ -0,0 +1,87 @@
...
+| BUILDKIT_HOST                       | `tcp://<hostname>:8372`        | The address of your buildkit host. Use this when you have a remote `buildkitd` you would like to connect to.                                                                                          |             
+| SRC_DIR                             | `/workspace`                   | The working directory for `earthly`. Usually host-mounted.                                                                                                                                            |
+| EARTHLY_ADDITIONAL_BUILDKIT_CONFIG  |                                | Additional `buildkitd` config to append to the generated configuration file.                                                                                                                          |
+| BUILDKIT_LOCAL_REGISTRY_LISTEN_PORT | `8371`                         | What port should the internal cache registry listen on?                                                                                                                                               |
```
vladaionescu, line 79: I think this should be disabled for both images. The reason is that Earthly code assumes the registry is at `127.0.0.1:8371` and doesn't support anything else. If it's enabled, buildkit will try to export via registry and earthly would not be able to receive the images, leading to failures.

docs/docker-images/all-in-one.md
dchw: I guess what I was trying to get at is that if you are already privileged, isn't best practice to run your own daemon? I'll add more details in the table if that works for you rather than adding another quickstart with `-v /var/run/docker.sock:/var/run/docker.sock`?

docs/docker-images/all-in-one.md
dchw: I haven't touched this setting at all; this was just documenting what was there. You want me to actively force it to the default value of 8371?

docs/docker-images/all-in-one.md
vladaionescu: Yeah that makes sense

docs/docker-images/all-in-one.md
vladaionescu: Perhaps don't document its existence just yet. The default should be unset, in which case buildkit will not use a registry for exports.

dchw: It remains unset in here. The defaults in this table arent always from my script; but rather are what they end up defaulting to if unspecified... eventually. I will remove this from the config.

docs/docker-images/all-in-one.md
```diff
@@ -0,0 +1,88 @@
...
+| EARTHLY_RESET_TMP_DIR               | `false`                        | Cleans out `EARTHLY_TMP_DIR` before running, if set to `true`. Useful when you host-mount an temp dir across runs.                                                                                    |
+| NETWORK_MODE                        | `cni`                          | Specifies the networking mode of `buildkitd`. Default uses a CNI bridge network, configured with the `CNI_MTU`.                                                                                       |
+| EARTHLY_TMP_DIR                     | `/tmp/earthly`                 | Specifies the location of `earthly`s temp dir. You can also mount an external volume to this path to preserve the contents across runs.                                                               |
+| CACHE_SIZE_MB                       | `0`                            | How big should the `buildkitd` cache be allowed to get, in MiB? 0 is unbounded.                                                                                                                       |
```
vladaionescu, line 87:
```suggestion
| CACHE_SIZE_MB                       | `0`                            | How big should the `buildkitd` cache be allowed to get, in MiB? A value of 0 sets the cache size to "adaptive", causing buildkit to detect the available size of the system and choose a limit automatically.                                                                                                                       |
```

vladaionescu (approved):
One more detail.

Looks good after that :+1: 🎉

docs/docker-images/buildkit-standalone.md
vladaionescu: It may be best to

1. Change the quick examples to always include a Docker volume mount - not a host mount.
2. Include enough details to make the user understand that they should mount something here almost always.

If people get issues related to disk space or aggressive GCing, it can be very hard for us to figure out that they are using a bad underlying FS. It can lead to a very bad user experience with little signal that this is the issue.

vladaionescu (changes requested):

docs/docker-images/buildkit-standalone.md
```diff
@@ -0,0 +1,73 @@
...
+- `--privileged` is required. This is because `earthly` needs some privileged `buildkit` functionality.
+- `-t` tells Docker to emulate a TTY. This makes the `buildkit` log output colorized.
+- `-v /tmp/earthly:/tmp/earthly:rw` mounts the hosts `/tmp/earthly` into the containers `/tmp/earthly`. This is used as a temporary/working directory for `buildkitd` during builds.
+
+Assuming you are running this on your machine, you could use this `buildkitd` by setting `EARTHLY_BUILDKIT_HOST=docker-container://<container-name>`, or by specifying the appropriate values in `config.yml`.
+
+### Usage (Use As Remote)
+
+```bash
+docker run --privileged -t -v /tmp/earthly:/tmp/earthly:rw -e BUILDKIT_TCP_TRANSPORT_ENABLED=true -p 127.0.0.1:8372:8372 earthly/buildkitd:latest
```
vladaionescu, line 33: For remote usage, the user might want to `-p 8372:8372` to expose externally too, not just to `127.0.0.1`?

dchw:
> Change the quick examples to always include a Docker volume mount - not a host mount.

I can do that.

> Include enough details to make the user understand that they should mount something here almost always.

I thought we were there with the section under Requirements? What else should I be saying there?

vladaionescu (approved):
Awesome! Thanks for bearing with me!

alexcb (approved):

PR #1054, opened 2021-06-11
Create docs on build-args and secrets

alexcb:
here's a quick link to the preview: https://app.gitbook.com/@earthly/s/earthly-staging/~/revisions/-Mbw9Zrr6M34py5BBEBo/v/secret-docs/docs/guides/build-args

### DIFF
--- a/docs/guides/build-args.md
+++ b/docs/guides/build-args.md
@@ -1,0 +1,193 @@
+# Build Arguments and Secrets
+
+## Introduction
+
+One of the core features of earthly is support for build arguments. Build arguments
+can be used to dynamically set environment variables inside the context of [RUN commands](../earthfile.md#run).
+
+Build arguments can be passed between targets or from the command line. They encourage
+writing generic Earthfiles and ultimately promote greater code-reuse.
+
+Additionally, earthly defines secrets which are similar to build arguments, but are exposed as environment
+variables when explicitly allowed.
+
+## A Quick Example
+
+Arguments are declared with the [ARG](../earthfile.md#arg) keyword.
+
+Let's consider a "hello world" example that allows us to change who is being greeted (e.g. hello banana, hello eggplant etc).
+We will create a hello target that accepts the `name` argument:
+
+```Dockerfile
+FROM alpine:latest
+
+hello:
+    ARG name
+    RUN echo "hello $name"
+```
+
+Then we will specify a value for the `name` argument on the command line when we invoke `earthly`:
+
+```bash
+earthly --build-arg name=world +hello
+```
+
+This will output
+
+```
+    buildkitd | Found buildkit daemon as docker container (earthly-buildkitd)
+alpine:latest | --> Load metadata linux/amd64
+         +foo | --> FROM alpine:latest
+         +foo | [██████████] resolve docker.io/library/alpine:latest@sha256:69e70a79f2d41ab5d637de98c1e0b055206ba40a8145e7bddb55ccc04e13cf8f ... 100%
+         +foo | name=world
+         +foo | --> RUN echo "hello $name"
+         +foo | hello world
+       output | --> exporting outputs
+```
+
+If we re-run `earthly --build-arg name=world +hello`, we will see that the echo command is cached (and won't re-display the hello world text):
+
+```
++foo | *cached* --> RUN echo "hello $name"
+```
+
+## Setting Argument Values
+
+Argument values can be set multiple ways:
+
+1. On the command line
+
+   The value can be directly specified on the command line (as shown in the previous example):
+   
+   ```
+   earthly --build-arg name=world +hello
+   ```
+
+2. From an environment variable
+
+   If no value is given for name, then earthly will look for the value in the corresponding
+   environment variable on the localhost:
+   
+   ```bash
+   export name="banana"
+   earthly --build-arg name +hello
+   ```
+
+3. From a `.env` file
+
+   It is also possible to create an `.env` file to contain the build arguments to pass
+   to earthly. First create an `.env` file with:
+   
+   ```
+   name eggplant
+   ```
+   
+   Then simply run earthly:
+   
+   ```bash
+   earthly +hello
+   ```
+
+## Passing Argument values to targets
+
+Build arguments can also be set when calling build targets. If multiple build arguments values are defined for the same argument name,
+earthly will build the target for each value. For example, we can create a new `greetings` target which calls `+hello` multiple times:
+
+```dockerfile
+greetings:
+    BUILD \
+        --build-arg name=world \
+        --build-arg name=banana \
+        --build-arg name=eggplant \
+        +hello
+```
+
+Then when we call `earthly +greetings`, earthly will call `+hello` three times:
+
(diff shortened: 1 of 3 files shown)


agbell (approved):

docs/guides/build-args.md
```diff
@@ -0,0 +1,193 @@
...
+The file will not be saved to the image snapshot.
+{% endhint %}
+
+The value for `+secrets/passwd` must then be supplied to earthly on the command line. Either directly via:
+
+```bash
+earthly --secret passwd=itsasecret +hush
+```
+
+or if the value is omitted, then earthly will attempt to lookup the value from an environment variable on the localhost:
```
vladaionescu, line 154: We should probably mention the option to store secrets in the cloud and link to the relevant guide.

EDIT: I see that it is mentioned below, but in a slightly different context. I would also mention here, without going into details.

docs/guides/build-args.md
```diff
@@ -0,0 +1,193 @@
...
+
+## Storage of local secrets
+
+Earthly stores the contents of command-line-supplied secrets in memory on the localhost. When a `RUN` command that requires a secret is evaluated by BuildKit, the BuildKit
+daemon will request the secret from the earthly command-line process and will temporarily mount the secret inside the runc container that is evaluating the `RUN` command.
+Once the command finishes the secret is unmounted. It will not persist as an environment variable within the saved container snapshot. Secrets will persist in-memory
+until the earthly command exits.
+
+{% hint style='danger' %}
+Earthly and BuildKit communicate over an unencrypted TCP connection *only* when running on the same host; therefore any users on the localhost with network packet inspection
```
vladaionescu, line 187: IIRC it's not a TCP connection currently. We're migrating towards TCP at some point, but it'll be an encrypted connection. @dchw has more details.

docs/guides/cloud-secrets.md
```diff
@@ -87,6 +93,15 @@ You can build it via:
 earthly +build
 ```
 
+{% hint style='info' %}
+### Naming of local and cloud-based secrets
+
+The only difference between the naming of locally-supplied and cloud-based secrets is that cloud secrets will contain
+two or more slashes; where as locally-defined secret names will only start with the `+secrets/` suffix,
```
vladaionescu, line 100: Might be worth including an example of a cloud secret key.

docs/guides/cloud-secrets.md
```diff
@@ -162,6 +177,16 @@ EARTHLY_TOKEN=...
 Which will then force Earthly to use that token when accessing secrets. This is useful for cases
 where running an ssh-agent is impractical.
 
+# Security Details
+
+The Earthly command uses HTTPS to communicate with the cloud secrets server. The server encrypts all
+secrets using OpenPGP's implemtation of AES256 before storing it in a database.
```
vladaionescu, line 183:
```suggestion
secrets using OpenPGP's implementation of AES256 before storing it in a database.
```

vladaionescu, line 183: Perhaps we should also mention that a distinct encryption key is used for each Earthly organization and that we use industry-standard security practices for managing encryption keys in the cloud.

vladaionescu (commented):
Very useful guide!

Some remarks below...

docs/guides/build-args.md
```diff
@@ -0,0 +1,193 @@
...
+   Then simply run earthly:
+   
+   ```bash
+   earthly +hello
+   ```
+
+## Passing Argument values to targets
+
+Build arguments can also be set when calling build targets. If multiple build arguments values are defined for the same argument name,
+earthly will build the target for each value. For example, we can create a new `greetings` target which calls `+hello` multiple times:
```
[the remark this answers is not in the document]
alexcb: that's a good plan, at the very least it would make it easier for people to find if they are searching for `"earthly" "build matrix"`.

docs/guides/build-args.md
alexcb: that's a good point, the default is for socket-based connection: `default: "docker-container://earthly-buildkitd"`; I'll remove this.

docs/guides/cloud-secrets.md
alexcb: I took another look at our server code, and we don't have distinct per-organization encryption keys, but only a global key.

(as an aside, things like account passwords are hashed via bcrypt and can't be decrypted).

I'll add on `industry-standard security practices for managing encryption keys in the cloud` though.

docs/guides/cloud-secrets.md
alexcb: I called out the fact that cloud secrets start with `+secrets/<user or organization/>`; what do you think about that?

docs/guides/build-args.md
alexcb: that's a good point, I added a small link.

docs/guides/cloud-secrets.md
vladaionescu: sgtm

docs/guides/cloud-secrets.md
vladaionescu: We should adjust this on the server side - mind adding an issue so we don't forget? The individual org encryption keys should be stored encrypted with the master key.

pmembrey: Not something for this PR, but I think it's definitely worth considering support for client side encryption. This is beneficial in a few ways, such as the client not having to trust the server, but also then you no longer have to manage toxic assets like credentials for other org's critical / key systems.

For example, I can forsee that many orgs would like to store a github token as a secret so that they can greatly simplify checking out repositories. However, if it's server side only encryption, the client must now explicitly trust that the server (and in turn you) to keep it safe. Anyone with that token can access any of that orgs repos...

One way to do this without huge pain (even as a v1), would be to accept a symmetric key on the client that encrypts the secret with that key before uploading it. The server side stuff won't need to change - it just sends, receives and stores as normal - and on the client side, if the key is set, use that key to decrypt all the secrets that are coming in. Potentially could add a simple tag so the secret so that the client can identify explicitly that it's encrypted, and which key it needs to decrypt it. This also allows for mixed keys and mixing encrypted and non encrypted secrets.

Just some thoughts :)

alexcb: We went back and forth figuring out a way to do this and ultimately ended up with a system that does grant the server the ability to decrypt secrets in order to keep setting up CI tasks.

I don't see any technical issues with allowing users to specify a symmetric key for use strictly on the client side before calling the appropriate [set](https://github.com/earthly/earthly/blob/e8b0e570c5d843afad953ab8caab118d34adc229/secretsclient/client.go#L568) and [get](https://github.com/earthly/earthly/blob/e8b0e570c5d843afad953ab8caab118d34adc229/secretsclient/client.go#L561) functions. But the downside is now the symmetric key must be passed around different members of the team in addition to managing access to the secrets org. Then if you want to grant github actions access to secrets, you'll have to trust it and grant it access via an `EARTHLY_TOKEN` as well as this symmetric key.

vladaionescu (approved):

Issue #971, opened 2021-05-07
Support CI-specific tokens


vladaionescu:
GitHub and GitLab pass `GITHUB_TOKEN` and `CI_JOB_TOKEN `, respectively, to their CI runs, so that the CI can access the repository for basic things such as checking out the code.

It would be great if Earthly could pick these up automatically to configure git auth, in case the CI workflow needs to reference targets in other repos, or even access another branch of the same repo.

This would simplify the use of remote targets significantly in such scenarios.

CC @alexcb

alexcb:
After some experimentation it seems that `GITHUB_TOKEN` only has access for the current repo and does not have access to other repos in the same organisation. Perhaps there's a setting somewhere to grant org-level access?

This can still be useful for cases where a repo wants to reference a different branch and needs to check that out though.

alexcb:
For the case of github, we will have two env variables set by github:

    GITHUB_ACTOR="user" # e.g. "alexcb"
    GITHUB_REPOSITORY="org/repo" # e.g. "earthly/earthly"
    GITHUB_TOKEN="it's a secret"

Then we could add some logic similar to this:

    if $GITHUB_TOKEN != "" && hostToClone == "github.com" && gitPathToClone == $GITHUB_REPOSITORY && protocol == "auto" {
        set protocol="https"
        set user=$GITHUB_ACTOR
        set password=$GITHUB_TOKEN
    }

We would need to be careful to only set the user/password when using the specific $GITHUB_REPOSITORY, as the auth would be rejected when trying to clone other repos, even if public.

alexcb:
Since the `GITHUB_TOKEN` only has access to the current repo and nothing else, this seems like a fairly unique request to support it internall within earthly.

Instead, one will need to issue a call to `earthly config git ...` to setup access prior to calling `earthly`.

Here's an example:

```
jobs:
  example-tests:
    runs-on: ubuntu-latest
    env:
      FORCE_COLOR: 1
      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
    steps:
      - uses: earthly/actions/setup-earthly@v1
        with:
          version: v0.5.10
      - uses: actions/checkout@v2
      - name: configure earthly to use https with the GITHUB_TOKEN
        run: "earthly config git \"{\"github.com\": {\"auth\": \"https\", \"user\": \"$GITHUB_ACTOR\", \"password\": \"$GITHUB_TOKEN\"}}\""
      - name: reference a different branch of the current repo
[...]

PR #1059, opened 2021-06-14
draft: use github token for cloning

alexcb:
please ignore for now.


alexcb:
we decided not to implement this due to the github token only granting access to the git repo and not org. See https://github.com/earthly/earthly/issues/971 for details.

PR #1061, opened 2021-06-15
Re-add DockerHub docs to the CI process.

dchw:
Now that permissions have been sorted, we can re-add this. I have tested the action locally and successfully updated the descriptions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -649,7 +649,7 @@ jobs:
       - name: Build and push +prerelease
         run: ./build/linux/amd64/earthly --ci --push +prerelease
       - name: Update DockerHub description for earthly/earthly
-        uses: peter-evans/dockerhub-description@v2
+        uses: peter-evans/dockerhub-description@616d1b63e806b630b975af3b4fe3304307b20f40
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_PASSWORD }}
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -648,6 +648,22 @@ jobs:
         run: ./build/linux/amd64/earthly config global.local_registry_host 'tcp://127.0.0.1:8371'
       - name: Build and push +prerelease
         run: ./build/linux/amd64/earthly --ci --push +prerelease
+      - name: Update DockerHub description for earthly/earthly
+        uses: peter-evans/dockerhub-description@v2
```
alexcb, line 652: Can we trust Peter?

alexcb: is there a way to pin it to a git sha to avoid it from changing?

alexcb: or alternative, we could fork this action to prevent it from changing.

dchw: This action appears to be widely used; and what I've seen of the code when I was poking around it looked clean. I'll look to see if we can pin it? I don't think there is significant reason to worry though.

alexcb: https://julienrenaux.fr/2019/12/20/github-actions-security-risk/ makes some good points about why we should create a fork (to protect us from the repo being compromised or deleted).

dchw: I would rather just script it than maintain a fork. Ill switch it to a hand-rolled script then? Or is the SHA pinning enough.

Do we have other actions we should be forking too that I could handle at the same time?

alexcb: pinning a sha1 is sufficient to prevent the issue of the repo being compromised.

dchw: Latest commit should have it pinned.

alexcb (approved):

.github/workflows/ci.yml
alexcb: > Do we have other actions we should be forking too that I could handle at the same time?

Aside from the official github actions, we also use `docker/setup-qemu-action@v1` but it doesn't make use of credentials.

PR #1064, opened 2021-06-15
Include version and help flags in autocomplete

alexcb:
The --help and --version flags are added automatically to urfav, and aren't
included in the flags we define for earthly; therefore we must
explicitly add them.

### DIFF
--- a/autocomplete/complete.go
+++ b/autocomplete/complete.go
@@ -208,6 +208,8 @@ func GetPotentials(compLine string, compPoint int, app *cli.App) ([]string, erro
 		flags = getVisibleFlags(cmd.Flags)
 	} else {
 		flags = getVisibleFlags(app.Flags)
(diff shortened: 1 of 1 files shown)


autocomplete/complete.go
```diff
@@ -208,6 +208,8 @@ func GetPotentials(compLine string, compPoint int, app *cli.App) ([]string, erro
 		flags = getVisibleFlags(cmd.Flags)
 	} else {
 		flags = getVisibleFlags(app.Flags)
+		// append flags that urfav/clu automatically include
```
vladaionescu, line 211: cli*

vladaionescu (approved):

autocomplete/complete.go
alexcb: fixed!

PR #1071, opened 2021-06-23
Introduce a --no-follow option for copying

alexcb:
The --no-follow option will instruct buildkit not to follow symlinks.
This is helpful for cases where one wishes to copy an invalid symlink.


alexcb:
This change doesn't fix the issue; we are still hitting the same error:

```
failed to compute cache key: "/test/nonexistant" not found: not found
```

PR #1073, opened 2021-06-24
No follow

alexcb:
This introduces a `--no-follow` flag for both COPY and SAVE ARTIFACT commands, which pass the NoFollow flag to buildkit's CopyOp which will prevent symlinks from being followed. This allows users to save and copy invalid symlinks (which point to non-existent paths).

When using `SAVE ARTIFACT` on files, `--no-follow` is required if the file is an invalid symlink. Furthermore, when that file is copied `--no-follow` must also be used on the `COPY` command.

However, when using `SAVE ARTIFACT` on directories which contain one or more symlink, the `--no-follow` flag has no effect on buildkit's internal workings, and therefore is not required by earthly. The `--no-follow` flag is still however required on the `COPY` command.

### DIFF
--- a/earthfile2llb/converter.go
+++ b/earthfile2llb/converter.go
@@ -227,7 +227,7 @@ func (c *Converter) FromDockerfile(ctx context.Context, contextPath string, dfPa
 		buildContext = llbutil.ScratchWithPlatform()
 		buildContext = llbutil.CopyOp(
 			mts.Final.ArtifactsState, []string{contextArtifact.Artifact},
-			buildContext, "/", true, true, false, "", false,
+			buildContext, "/", true, true, false, "", false, false,
 			llb.WithCustomNamef(
 				"[internal] FROM DOCKERFILE (copy build context from) %s%s",
 				joinWrap(buildArgs, "(", " ", ") "), contextArtifact.String()))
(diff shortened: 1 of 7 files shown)


vladaionescu:
I would maybe call this `--symlink-no-follow` so that it's obvious to the reader what "no follow" refers to.

alexcb:
> so that it's obvious to the reader what "no follow" refers to.

That's a good point, it'll make it easier to search for.

earthfile2llb/converter.go
```diff
@@ -496,7 +497,7 @@ func (c *Converter) RunExitCode(ctx context.Context, commandName string, args, m
 		"%s %s%s%s",
 		commandName, // eg IF
 		strIf(privileged, "--privileged "),
-		strIf(noCache, "--no-cache "),
+		strIf(noCache, "--symlink-no-cache "),
```
vladaionescu, line 500: Probably unintentional? :-)

earthfile2llb/converter.go
```diff
@@ -618,7 +619,7 @@ func (c *Converter) Run(ctx context.Context, args, mounts, secretKeyValues []str
 		strIf(withDocker, "--with-docker "),
 		strIf(withEntrypoint, "--entrypoint "),
 		strIf(pushFlag, "--push "),
-		strIf(noCache, "--no-cache "),
+		strIf(noCache, "--symlink-no-cache "),
```
vladaionescu, line 622: Same here

earthfile2llb/converter.go
alexcb: ouch, that was a bad find-and-replace. Good catch.

vladaionescu (approved):

sdaschner:
Stumbled upon this only by accident (which luckily solved my issue).

Can this be added to the documentation? 🙂

PR #1078, opened 2021-06-28
Use Mirror

dchw:
This is a sandbox for using the pull-through cache until its ready to go.

### DIFF
--- a/.buildkite/tests.sh
+++ b/.buildkite/tests.sh
@@ -51,6 +51,11 @@ sleep 1 # WSL2 requires a wait after to avoid "Text file busy".
 "$earthly" --version
 "$earthly" config global.local_registry_host 'tcp://127.0.0.1:8371'
 
+# Yes, there is a bug in the upstream YAML parser. Sorry about the jank here.
+"$earthly" earthly config global.buildkit_additional_config "'[registry.\"docker.io\"]
+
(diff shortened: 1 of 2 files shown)


.buildkite/tests.sh
```diff
@@ -51,6 +51,11 @@ sleep 1 # WSL2 requires a wait after to avoid "Text file busy".
 "$earthly" --version
 "$earthly" config global.local_registry_host 'tcp://127.0.0.1:8371'
 
+# Yes, there is a bug in the upstream YAML parser. Sorry about the jank here.
+"$earthly" config global.buildkit_additional_config "'[registry.\"docker.io\"]
+
+ mirrors = [\"registry-1.docker.io.mirror.corp.earthly.dev\"]'"
```
dchw, line 57: This can be fixed here and in the YAML if https://github.com/go-yaml/yaml/issues/423 ever gets fixed.

vladaionescu (approved):

.buildkite/tests.sh
alexcb: can you add a link to that github issue in the comment, so that way if we read the comment later, we can reference the github issue more easily and updated it if it's fixed?

alexcb (approved):

PR #1080, opened 2021-06-29
Only sending files that are referenced by COPY

alexcb:
bugfix: the entire directory was being sent to buildkit whenever a COPY
was referenced because no IncludePattern was given.

This is a rather ugly hack, and could probably be improved.


alexcb:
this PR is old and will be replaced by https://github.com/earthly/earthly/pull/1085

PR #1083, opened 2021-07-06
Allow user to override container and volume names

alexcb:
This allows a user to switch between volume caches, which increases the
speed of testing operations where one needs to reset the volume cache.

### DIFF
--- a/buildkitd/settings.go
+++ b/buildkitd/settings.go
@@ -28,6 +28,7 @@ type Settings struct {
 	ServerTLSKey         string
 	UseTCP               bool
 	UseTLS               bool
+	VolumeName           string
 }
 
 // Hash returns a secure hash of the settings.
--- a/cmd/earthly/main.go
+++ b/cmd/earthly/main.go
@@ -72,6 +72,13 @@ import (
 	"github.com/earthly/earthly/variables"
 )
 
+const (
+	// DefaultBuildkitdContainerName is the name of the buildkitd container.
+	DefaultBuildkitdContainerName = "earthly-buildkitd"
+	// DefaultBuildkitdVolumeName is the name of the docker volume used for storing the cache.
+	DefaultBuildkitdVolumeName = "earthly-cache"
+)
+
 var dotEnvPath = ".env"
 
 type earthlyApp struct {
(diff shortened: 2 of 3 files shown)


buildkitd/settings.go
```diff
@@ -28,6 +28,7 @@ type Settings struct {
 	ServerTLSKey         string
 	UseTCP               bool
 	UseTLS               bool
+	VolumeName           string
```
vladaionescu, line 31: Should `ContainerName` be part of settings too? 🤔

cmd/earthly/main.go
```diff
@@ -72,6 +72,13 @@ import (
 	"github.com/earthly/earthly/variables"
 )
 
+const (
+	// DefaultContainerName is the name of the buildkitd container.
+	DefaultContainerName = "earthly-buildkitd"
+	// DefaultVolumeName is the name of the docker volume used for storing the cache.
+	DefaultVolumeName = "earthly-cache"
```
vladaionescu, line 79: Might want to make these variable names a little more specific given that they're no longer under the buildkitd package: `DefaultBuildkitContainerName` & `DefaultBuildkitVolumeName`?

vladaionescu (commented):
Looks good - @dchw might want to look at this too.

Another possible issue: clashing port numbers. You might want to try to run multiple local buildkits and see if everything works correctly. Possible issues: debugger, TCP connectivity, export via local registry. Might be worth testing out these features to make sure they work as expected.

alexcb:
> clashing port numbers

unfortunately that did happen with the interactive debugger. I'll see if I can override it and use a random port if it's already in use.

buildkitd/settings.go
alexcb: I considered that; however we need the container name when making a call such as `GetSettingsHash(ctx, containerName)`.

alexcb:
I had quick go to see if we could have the OS assign us a random free port, but it's non-trivial and will take some effort. Do we want to address this in a follow up PR? The benefit is we make use of switching volumes which means we can reset one cache for testing without having it affect build times while compiling earthly.

vladaionescu (approved):
SGTM

vladaionescu:
Another possibility could be to define a port range to use for the debugger, local registry etc. Something like `1230-1239`. Though I agree that random port seems best if it's possible.

PR #1098, opened 2021-07-13
dont pass empty variables to earthly-buildkit run

alexcb:


vladaionescu (approved):
This should be fine.. maybe @dchw should review too just in case.

dchw (approved):

PR #1100, opened 2021-07-14
Allow overriding features via env

alexcb:

### DIFF
--- a/cmd/earthly/main.go
+++ b/cmd/earthly/main.go
@@ -150,6 +150,7 @@ type cliFlags struct {
 	certPath                  string
 	keyPath                   string
 	disableAnalytics          bool
+	featureFlagOverrides      string
 }
 
 var (
@@ -562,6 +563,13 @@ func newEarthlyApp(ctx context.Context, console conslogging.ConsoleLogger) *eart
 			EnvVars:     []string{"EARTHLY_DISABLE_ANALYTICS", "DO_NOT_TRACK"},
 			Usage:       "Disable collection of analytics",
 			Destination: &app.disableAnalytics,
+		},
+		&cli.StringFlag{
+			Name:        "feature-flag-overrides",
+			EnvVars:     []string{"EARTHLY_FEATURE_FLAG_OVERRIDES"},
+			Usage:       "Apply additional flags after each VERSION command across all Earthfiles, multiple flags can be seperated by commas",
+			Destination: &app.featureFlagOverrides,
+			Hidden:      true, // used for feature-flipping from ./earthly dev script
 		},
 	}
 
@@ -2780,6 +2788,7 @@ func (app *earthlyApp) actionBuildImp(c *cli.Context, flagArgs, nonFlagArgs []st
 		ParallelConversion:     (app.conversionParllelism != 0),
 		Parallelism:            parallelism,
 		LocalRegistryAddr:      localRegistryAddr,
+		FeatureFlagOverrides:   app.featureFlagOverrides,
 	}
 	b, err := builder.NewBuilder(c.Context, builderOpts)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


features/features.go
```diff
@@ -18,7 +18,8 @@ import (
 type Features struct {
 	// DoSaves flags the feature to only save artifacts under directly referenced targets
 	// of those referenced by BUILDs
-	ReferencedSaveOnly bool `long:"referenced-save-only" description:"only save artifacts that are directly referenced"`
+	ReferencedSaveOnly     bool `long:"referenced-save-only" description:"only save artifacts that are directly referenced"`
+	UseCopyIncludePatterns bool `long:"use-copy-include-patterns" description:"specify an include pattern to buildkit when performing copies"`
```
alexcb, line 22: This isn't currently referenced anywhere else; however, I needed a second flag for unit testing the comma splitting.

Once this PR is in, i'll rebase https://github.com/earthly/earthly/pull/1085 against this code which will then make use of this field for feature flipping.

earthfile2llb/earthfile2llb.go
```diff
@@ -105,6 +106,11 @@ func Earthfile2LLB(ctx context.Context, target domain.Target, opt ConvertOpt) (m
 		return nil, errors.Wrapf(err, "resolve feature set for version %v for target %s", bc.Earthfile.Version.Args, target.String())
 	}
 
+	err = features.ApplyFlagOverrides(ftrs, os.Getenv("EARTHLY_VERSION_FLAGS"))
```
alexcb, line 109: unsure if we should call this `EARTHLY_VERSION_FLAGS` or `EARTHLY_VERSION_FLAG_OVERRIDES`

dchw: Would this make more sense next to the rest of our config in `main.go`?

dchw (approved):

dchw:
What is our intended lifetime of these feature flags? Just hoping we dont accumulate too many that stick around for too long

earthfile2llb/earthfile2llb.go
alexcb: That's a good point, I can leave it Hidden as it's only (mostly?) a dev feature.

alexcb:
> What is our intended lifetime of these feature flags? Just hoping we dont accumulate too many that stick around for too long

Good question, under https://github.com/earthly/earthly/issues/991 we outlined a rollout strategy for the version numbering. I suspect the feature flags will be treated in a similar way:

> We aim to support at least 4 minor versions behind and also at least a whole year since the last release of that minor version. This should be a nice balance between keeping Earthly development nimble enough, while not breaking people's builds abruptly.

earthfile2llb/earthfile2llb.go
dchw: 100% was assuming it would be hidden.

dchw:
So the flags will stay supported for as long as the version they were introduced in stay supported. Feels fair. :+1:

cmd/earthly/main.go
```diff
@@ -563,6 +564,13 @@ func newEarthlyApp(ctx context.Context, console conslogging.ConsoleLogger) *eart
 			Usage:       "Disable collection of analytics",
 			Destination: &app.disableAnalytics,
 		},
+		&cli.StringFlag{
+			Name:        "feature-flag-overrides",
+			EnvVars:     []string{"EARTHLY_FEATURE_FLAG_OVERRIDES"},
+			Usage:       "Apply additional flags after the VERSION command is parsed, multiple flags can be seperated by commas",
+			Destination: &app.featureFlagOverrides,
+			Hidden:      true, // used for feature-flipping from ./earthly dev script
+		},
```
vladaionescu, line 573: Would these apply to all Earthfiles encountered during the build?

vladaionescu (approved):
:+1:

cmd/earthly/main.go
alexcb: yes, does something like `Override all VERSION commands with these additional flags, multiple flags can be seperated by commas` make it more clear?

vladaionescu: Yes - that makes sense

PR #1105, opened 2021-07-15
WSL cannot use `iptables-nft`.

dchw:


buildkitd/entrypoint.sh
```diff
@@ -32,8 +32,16 @@ if [ "$EARTHLY_RESET_TMP_DIR" = "true" ]; then
 fi
 
 if ! lsmod | grep -wq "^ip_tables"; then
-  echo "Legacy ip_tables is not loaded. Switching to iptables-nft."
-  ln -sf /sbin/iptables-nft /sbin/iptables
+  echo "Legacy ip_tables is not loaded."
+
+  if ! uname -a | grep -wiq "microsoft"; then
+    echo "WSL Detected! Keeping legacy iptables in place."
```
dchw, line 38: I could combine the conditions, but I also like the logging letting us know it is WSL too.

buildkitd/entrypoint.sh
```diff
@@ -32,8 +32,16 @@ if [ "$EARTHLY_RESET_TMP_DIR" = "true" ]; then
 fi
 
 if ! lsmod | grep -wq "^ip_tables"; then
-  echo "Legacy ip_tables is not loaded. Switching to iptables-nft."
-  ln -sf /sbin/iptables-nft /sbin/iptables
+  echo "Legacy ip_tables is not loaded."
+
+  if ! uname -a | grep -wiq "microsoft"; then
```
alexcb, line 37: Is it possible to use `$OSTYPE` here instead? or if that doesn't work, can we create our own OS detection and move it higher up then just have a `if [ "$OS" = "windows" ]; then` here?

That way we'll keep it fairly consistent in this script if it ever has to grow.

alexcb: This might do the trick:

```suggestion
  if [ "$OSTYPE" = "msys" ]; then
```

dchw: Does `$OSTYPE` work inside a Docker container? This is more or less from Microsoft on how to detect WSL.

alexcb: I think you might have to do a:

```
GUEST_OS="unknown"

if ! uname -a | grep -wiq "microsoft"; then
    GUEST_OS="windows"
fi
```

then just check 

```
if [ "$GUEST_OS" = "windows" ]; then
```

dchw: So just lift the OS detection up out of the if? Sure why not.

buildkitd/entrypoint.sh
```diff
@@ -32,8 +39,16 @@ if [ "$EARTHLY_RESET_TMP_DIR" = "true" ]; then
 fi
 
 if ! lsmod | grep -wq "^ip_tables"; then
-  echo "Legacy ip_tables is not loaded. Switching to iptables-nft."
-  ln -sf /sbin/iptables-nft /sbin/iptables
+  echo "Legacy ip_tables is not loaded."
+
+  if [ $KERNEL = "WSL" ]; then
```
alexcb, line 44: just in case $KERNEL ever contains a space:

```suggestion
  if [ "$KERNEL" = "WSL" ]; then
```

dchw: I hope it won't since we control it; but more security is nice

alexcb (approved):

dchw:
Note that this passed Windows CI. Merging now.

Issue #961, opened 2021-04-29
change git protocol from string to new type


alexcb:
This might be a nit; but the only thing I would do in addition here is maybe enumerate the protocols to make it clearer down the line. Like:

```
type gitProtocol string

const (
	auto gitProtocol = "auto"
	ssh gitProtocol = "ssh"
        ...
)
```

This would ensure that git protocols stay on the rails? At the very least I would move them into a constants list and name them, the type just helps them only get used where they should be used.

_Originally posted by @dchw in https://github.com/earthly/earthly/pull/960#discussion_r623376019_

JosemyDuarte:
This one seems like a good first issue. Before considering contributing just a couple of clarifying questions: 

1. Do you want `gitProtocol` to be private to `buildcontext` package? Meaning, from [`main.go`](https://github.com/earthly/earthly/blob/aaffd2cc84bf7cb7a9cd7943602ecc44e46b2701/cmd/earthly/main.go#L2863) we will keep using Strings, or what would be the idea?

2. I guess `ssh` is not a feasible name for a variable without renaming [current ssh import](https://github.com/earthly/earthly/blob/aaffd2cc84bf7cb7a9cd7943602ecc44e46b2701/buildcontext/gitlookup.go#L16) or using another name for the variable. How would you proceed in this regard?

alexcb:
@JosemyDuarte That's great if you're willing to tackle this :raised_hands: 

1. I would be inclined to keep it as a private type and continue to pass in protocol as a string to `func (gl *GitLookup) AddMatcher(name, pattern, sub, user, password, suffix, protocol, keyScan string) error {`. Then inside that function we could parse the string value and convert it into a `gitProtocol` type. Keeping it as a string outside this package allows us to continue to define it as a string when we parse the yaml into the `GitConfig` structure. 
I don't think we have any public methods that export the configured protocol; if we do, then the extra complexity might be warranted.

2. How about calling it something like `sshProtocol`, `autoProtocol`, `httpProtocol`, and `httpsProtocol`?

Let us know if you run into any troubles.

Cheers,
Alex

JosemyDuarte:
Perfect. I agree with both ideas. I'll try to have a PR before the end of the week. Thanks!

PR #1111, opened 2021-07-20
Define gitProtocol variables

JosemyDuarte:
Contributing to solving https://github.com/earthly/earthly/issues/961

I've tried running tests locally with `./earthly -P +test` but after some minutes I got:

```
Error: build target: build main: bkClient.Build: failed to solve: failed to lookup secret "/earthly-technologies/dockerhub/user" from secrets server: no authorized public keys found
```
First I thought it was due to my changes, but I get the same error when I'm on `main` branch. So, not sure if these tests are meant to be run on CI only.

### DIFF
--- a/buildcontext/gitlookup.go
+++ b/buildcontext/gitlookup.go
@@ -31,17 +31,26 @@ type gitMatcher struct {
 	sub      string
 	user     string
 	suffix   string
-	protocol string
+	protocol gitProtocol
 	password string
 	keyScan  string
 }
+
+type gitProtocol string
+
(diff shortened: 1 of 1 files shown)


alexcb:
This looks great @JosemyDuarte; no need to worry about the few tests which have failed due to permission issues. Behind the scenes some of our tests want to perform a docker login, and the permission error occurred when earthly tried to authenticate.

I'll check out your branch and give it a try locally then merge it.

Thanks for your contribution :raised_hands:

PR #1113, opened 2021-07-22
fix homebrew release process and instructions

alexcb:
fixes issue where PR was from `griswoldthecat/homebrew-earthly` was
opened against `griswoldthecat/homebrew-earthly` rather than `earthly/homebrew-earthly`

Update instructions to document how to apply the `pr-pull` label.
This label must be applied manually once the test checks complete and is
equivalent to approving the PR. For this reason this label can not be
automatically applied.

### DIFF
--- a/release/README.md
+++ b/release/README.md
@@ -37,7 +37,15 @@
   ```bash
   ./earthly --build-arg RELEASE_TAG --push ./release+release-homebrew
   ```
-* Important: Subscribe to the PR that griswoldthecat created in homebrew-core, so that you can address any review comments that may come up.
+* Visit the newly created pull request (which is both listed in the previous command's stdout and the `#release` slack channel); there should be two checks that are automatically applied:
+  * `brew test-bot / test-bot (ubuntu-latest)`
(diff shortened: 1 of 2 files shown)


release/README.md
```diff
@@ -37,7 +37,15 @@
   ```bash
   ./earthly --build-arg RELEASE_TAG --push ./release+release-homebrew
   ```
-* Important: Subscribe to the PR that griswoldthecat created in homebrew-core, so that you can address any review comments that may come up.
+* Visit the newly created pull request at `https://github.com/earthly/homebrew-earthly/pull/XXX`; there should be two checks that are automatically applied:
```
vladaionescu, line 40: Maybe say to look in the `#release` internal Slack channel and click on the homebrew-earthly PR link?

vladaionescu (approved):

release/README.md
alexcb: sounds good, I changed it to that as well as point out that there's a link in the previous commands stdout.

vladaionescu (approved):

PR #1115, opened 2021-07-22
bugfix: insecure check was checking for https rather than http

alexcb:
https://github.com/earthly/earthly/pull/1111 accidentally swapped http for https, changing back.


vladaionescu (approved):

vladaionescu:
Does this require a release?

PR #1118, opened 2021-07-23
use --privileged flag instead of --no-cache for bool flag value test

alexcb:
The use of the --no-cache flag for this test was producing inconsistent results on cases where the cache was garbage collected between runs which led to a race condition. Instead use both --privileged and --no-cache to have a consistent result we can use to test.


alexcb:
This race condition was only ever observed on the circle-ci worker, which appears to be happy now.

vladaionescu (approved):

PR #1121, opened 2021-07-26
interactive flag should not be possible with ci flag

alexcb:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -119,7 +119,7 @@ jobs:
       - name: Enable local registry-based exporter
         run: ./build/linux/amd64/earthly config global.local_registry_host 'tcp://127.0.0.1:8371'
       - name: Execute interactive debugger test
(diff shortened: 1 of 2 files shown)


vladaionescu (approved):

dchw:
Shouldn't it technically be `--strict`? That would also get `--ci`?

alexcb:
> Shouldn't it technically be `--strict`? That would also get `--ci`?

Possibly. I just read over our description of `--strict`, and it reads `Disallow usage of features that may create unreproduceable builds`. the interactive debugger only occurs after a build has failed, and thus won't affect if a build is reproducible or not.

I could see a case where someone wants to use `--strict` while debugging a failure that occurred in ci; however, they might want to include the interactive debugger to drop them into a shell to debug the failure.

dchw:
Ah. I was thinking more along the lines of `RUN --interactive` which IIRC we already disallow. Tally ho, then!

dchw (approved):

PR #1124, opened 2021-07-27
Add docs for VERSION

alexcb:

### DIFF
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -37,6 +37,7 @@
 * [Earthfile reference](earthfile/earthfile.md)
     * [Builtin args](earthfile/builtin-args.md)
     * [Excluding patterns](earthfile/earthignore.md)
+    * [Experimental features](earthfile/features.md)
 * [The `earthly` command](earthly-command/earthly-command.md)
 * [Configuration reference](earthly-config/earthly-config.md)
 * [CI integration](./ci-integration.md)
--- a/docs/earthfile/earthfile.md
+++ b/docs/earthfile/earthfile.md
@@ -584,6 +584,47 @@ For more information see the [multi-platform guide](../guides/multi-platform.md)
 
 Same as [`FROM --allow-privileged`](#allow-privileged).
 
+## VERSION
+
+#### Synopsis
+
+* `VERSION [--use-copy-include-patterns] <version-number>`
+
+#### Description
+
+The command `VERSION` identifies which set of features to enable in Earthly while handling the corresponding Earthfile. The `VERSION` command is optional; however if specified, it must be the first command in the Earthfile.
+
+| Version number | enabled features |
+| --- | --- | --- |
+| `0.5` | _initial functionality will be preserved_ |
+
+#### Options
+
+##### `--use-copy-include-patterns`
+
+Instructs Earthly to set buildkit's `include pattern` when executing `COPY` commands; without this feature, Earthly sends all files in the current build context except ignored files.
+Setting this option may increase the speed of COPY commands by forcing Earthly to only transfered files needed for the specific COPY, rather than transfering all files initially.
+This is disabled by default under `0.5`.
+
+#### Overriding features globally
+
+It is possible to enable additional options to *all* Earthfiles, by including them in the `EARTHLY_VERSION_FLAG_OVERRIDES` environment variable. Multiple options can be seperated by commas.
+
+For example, to enable `--use-copy-include-patterns` across all Earthfiles, you can export:
+
+```bash
+export EARTHLY_VERSION_FLAG_OVERRIDES="use-copy-include-patterns"
+```
+
+{% hint style='danger' %}
+##### Important
+
+Overriding feature flags with the `EARTHLY_VERSION_FLAG_OVERRIDES` environment variable is discouraged, as it will cause the `Earthfile` to be executed in a non-reproducible way.
+
+Instead, the `VERSION` command should be used at the top of each `Earthfile`; this will ensure the Earthfile is executed using the same version of earthly features across multiple hosts
+even when different versions of earthly are installed.
+{% endhint %}
+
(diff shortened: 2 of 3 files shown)


agbell (approved):
Looks good. 

One thing I would love to see, and maybe its more suited to the blog, is the thinking behind this flag. It's an example of us taking 'repeatable' builds seriously and could be a good excuse to talk about that.

alexcb:
> One thing I would love to see, and maybe its more suited to the blog, is the thinking behind this flag. It's an example of us taking 'repeatable' builds seriously and could be a good excuse to talk about that.

During standup we talked a bit more about the environment override and #1125 and decided it's best to not even advertise this feature as it promotes non-repeatable builds. We'll leave the env variable override code in place; however will only use it as a dev feature, primarily to allow us to dogfood our features via the `./earthly` script in the root of the repo.

dchw:
I like it. If we are planning on phasing these out over time adding the expiration date might be nice too.

dchw (approved):

docs/earthfile/earthfile.md
```diff
@@ -584,6 +584,28 @@ For more information see the [multi-platform guide](../guides/multi-platform.md)
...
 
+## VERSION
+
+#### Synopsis
+
+* `VERSION [--use-copy-include-patterns] <version-number>`
+
+#### Description
+
+The command `VERSION` identifies which set of features to enable in Earthly while handling the corresponding Earthfile. The `VERSION` command is optional; however if specified, it must be the first command in the Earthfile.
```
vladaionescu, line 595: We should also say that the `VERSION` command will become mandatory in a future version of Earthly.

docs/earthfile/earthfile.md
```diff
@@ -584,6 +584,28 @@ For more information see the [multi-platform guide](../guides/multi-platform.md)
...
+| --- | --- | --- |
+| `0.5` | _initial functionality will be preserved_ |
+
+#### Options
+
+##### `--use-copy-include-patterns`
+
+Instructs Earthly to set buildkit's `include pattern` when executing `COPY` commands; without this feature, Earthly sends all files in the current build context except ignored files.
+Setting this option may increase the speed of COPY commands by forcing Earthly to only transfered files needed for the specific COPY, rather than transfering all files initially.
+This is disabled by default under `0.5`.
```
vladaionescu, line 607: I think we should have a separate page with a table of feature flags like this one. We should make it clear that all feature flags are experimental.

docs/earthfile/earthfile.md
```diff
@@ -584,6 +584,28 @@ For more information see the [multi-platform guide](../guides/multi-platform.md)
...
+
+| Version number | enabled features |
+| --- | --- | --- |
+| `0.5` | _initial functionality will be preserved_ |
+
+#### Options
+
+##### `--use-copy-include-patterns`
+
+Instructs Earthly to set buildkit's `include pattern` when executing `COPY` commands; without this feature, Earthly sends all files in the current build context except ignored files.
```
vladaionescu, line 605: We should explain this from the perspective of the user: what is it that the user gains. We should assume that people don't know about BuildKit's internals here.

A better explanation could be "experimental faster COPY algorithm" or something along those lines.

vladaionescu (changes requested):

docs/earthfile/earthfile.md
grindernlm: Ok

grindernlm (commented):
True

docs/earthfile/earthfile.md
alexcb: If we list them on a second page, would we then want to change 

```
VERSION [--use-copy-include-patterns] <version-number>
```

to something like

```
VERSION [<feature-flags>...] <version-number>
```

so we wouldn't have to manage this list across two different files?

alexcb:
I spun out the feature flags into a separate page:

image

Which then has the table listed here:

image

docs/SUMMARY.md
```diff
@@ -40,6 +40,7 @@
 * [The `earthly` command](earthly-command/earthly-command.md)
 * [Configuration reference](earthly-config/earthly-config.md)
 * [CI integration](./ci-integration.md)
+* [Experimental features](experimental-features/features.md)
```
vladaionescu, line 43: Wondering if this should go under the Earthfile section rather?

alexcb: I was torn about that too, as some of the features aren't exactly related to Earthfiles, but I'm fine moving them there if we want to limit the number of top-level side links on the summary.

alexcb: I just moved it.

vladaionescu: Forgot to update SUMMARY too?

alexcb: good catch :)

docs/earthfile/earthfile.md
```diff
@@ -584,6 +584,30 @@ For more information see the [multi-platform guide](../guides/multi-platform.md)
...
+| --- | --- | --- |
+| `0.5` | _initial functionality will be preserved_ |
+
+#### Options
+
+Individual features may be enabled by setting the corresponding feature flag.
+New features start off as experimental, which is why they are disabled by default.
+Once a feature reaches maturity, it will be enabled by default under a new version number.
+
+All features are described in [a corresponding table](features.md).
```
vladaionescu, line 609: This link doesn't work: https://app.gitbook.com/@earthly/s/earthly-staging/~/revisions/-Mft5yKY3dlzLbngYGJX/v/version-docs/docs/earthfile#version

I wonder if it should be `./features.md` instead? Or maybe just git book weirdness that will fix itself with another run.

alexcb: that definitely did the trick! thanks.

vladaionescu (approved):

PR #1125, opened 2021-07-27
ignore EARTHLY_VERSION_FLAG_OVERRIDES in strict mode

alexcb:
When running in --strict mode, EARTHLY_VERSION_FLAG_OVERRIDES
will be ignored, and a warning will be printed instead suggesting
the user set them via the VERSION command.

### DIFF
--- a/cmd/earthly/main.go
+++ b/cmd/earthly/main.go
@@ -1014,6 +1014,11 @@ func (app *earthlyApp) before(context *cli.Context) error {
 		return err
 	}
 
+	if app.strict == true && app.featureFlagOverrides != "" {
+		app.console.Warnf("EARTHLY_VERSION_FLAG_OVERRIDES (%s) will be ignored due to running in strict mode; they should be set via the VERSION Earthfile command instead\n", app.featureFlagOverrides)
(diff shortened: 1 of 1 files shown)


cmd/earthly/main.go
```diff
@@ -1014,6 +1014,11 @@ func (app *earthlyApp) before(context *cli.Context) error {
 		return err
 	}
 
+	if app.strict == true && app.featureFlagOverrides != "" {
+		app.console.Warnf("EARTHLY_VERSION_FLAG_OVERRIDES (%s) will be ignored due to running in strict mode; they should be set via the VERSION Earthfile command instead\n", app.featureFlagOverrides)
+		app.featureFlagOverrides = ""
+	}
+
```
vladaionescu, line 1021: A warning doesn't feel right here. I would either fail altogether if this is not allowed, or allow it under strict mode.

What's the argument against allowing it?

vladaionescu (changes requested):

alexcb:
in the end we'll just allow `EARTHLY_VERSION_FLAG_OVERRIDES` to be set, but not document it and keep it as a hidden dev only option.

Issue #1029, opened 2021-06-02
support .earthlyignore


alexcb:
`.earthignore` was never renamed to `.earthlyignore`, perhaps we should support either names (and raise an error if both exist).

PR #1130, opened 2021-07-29
Incoorporating .earthlyignore - Issue #1029

Threpio:
Issue:
https://github.com/earthly/earthly/issues/1029

What?:
- Allow for the ignore file `.earthignore` to be incorporated/used as `.earthlyignore`.
- If both exist then return an error

How:
- Check both the possible filenames to see if they both exist
- If both Exists then return error and implicit exclusion
- If only one exists then treat it as the ignore file
- If neither exist then return just implicit exlusion (as was behaviour before)

Acceptance:
- `.earthlyignore` files can be used in the place of `.earthignore`files
-  If both exist then an error is thrown or registered

Notes:
- Not sure that I threw the error correctly - Maybe if both exists it should be a hard stop
- I am sure the code could be more consise and cleaner

### DIFF
--- a/buildcontext/excludes.go
+++ b/buildcontext/excludes.go
@@ -23,7 +23,7 @@ var ImplicitExcludes = []string{
 }
 
 func readExcludes(dir string) ([]string, error) {
-	ignoreFile := EarthIgnoreFile
+	var ignoreFile = EarthIgnoreFile
 
 	// if non (doesNotExist) errors appear then we need to return them.
 	//EarthIgnoreFile
@@ -42,7 +42,7 @@ func readExcludes(dir string) ([]string, error) {
 		// return just ImplicitExcludes if neither of them exist
 		return ImplicitExcludes, nil
 	} else if earthlyExists {
-		ignoreFile := EarthlyIgnoreFile
+		ignoreFile = EarthlyIgnoreFile
 	}
 
 	filePath := filepath.Join(dir, ignoreFile)
--- a/buildcontext/excludes.go
+++ b/buildcontext/excludes.go
@@ -1,6 +1,7 @@
 package buildcontext
 
 import (
+	"github.com/earthly/earthly/util/fileutil"
 	"os"
 	"path/filepath"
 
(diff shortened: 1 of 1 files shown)


buildcontext/excludes.go
```diff
@@ -35,3 +56,14 @@ func readExcludes(dir string) ([]string, error) {
 	}
 	return append(excludes, ImplicitExcludes...), nil
 }
+
+// cleanest way I could imagine to iterate and check if both of them exist - If the file is a bash/executable file then it could return an error other than (os.IsNotExist(err))
```
alexcb, line 60: instead of this function, would it be possible to use `fileutil/FileExists` instead? it exists under `util/fileutil/fileutil.go`. I think it would be fine to ignore the errors that are swallowed by `FileExists()`.

Threpio: My apologies - I thought I had checked if there was a function already - I'll apply that

buildcontext/excludes.go
```diff
@@ -1,6 +1,7 @@
 package buildcontext
 
 import (
+	"github.com/earthly/earthly/util/fileutil"
```
alexcb, line 4: very minor nitpick, can you add a newline between the standard imports and the github.com/earthly imports, like this:

```
	"os"
	"path/filepath"

	"github.com/earthly/earthly/util/fileutil"
```

buildcontext/excludes.go
```diff
@@ -10,23 +11,43 @@ import (
...
 	"Earthfile",
 	EarthIgnoreFile,
+	EarthlyIgnoreFile,
 }
 
 func readExcludes(dir string) ([]string, error) {
-	filePath := filepath.Join(dir, EarthIgnoreFile)
+	var ignoreFile = EarthIgnoreFile
+
+	// if non (doesNotExist) errors appear then we need to return them.
```
alexcb, line 28: is this comment still relevant or can it be removed?

buildcontext/excludes.go
Threpio: Apologies - Standard import with goland - Have moved and ordered :)

buildcontext/excludes.go
Threpio: Good spot - Removed and pushed.

alexcb:
Some of our tests need additional permissions that only members of earthly have access to, so we can ignore those for this PR.

Thanks for the contribution :raised_hands:

PR #1137, opened 2021-07-31
Add EARTHLY_SOURCE_DATE_EPOCH to support reproducible builds

pmembrey:
`EARTHLY_SOURCE_DATE_EPOCH` will be set to the timestamp of the last git commit. If that is not available then it defaults to the string `0`.

This ensures that there is always a reproducible timestamp to use, either the UNIX epoch itself, or the UNIX timestamp of the last git commit.

This PR also restores `EARTHLY_GIT_COMMIT_TIMESTAMP` to the same behaviour as other `EARTHLY_GIT_` args that return an empty string when the value is unavailable.

Background: https://reproducible-builds.org/docs/source-date-epoch/

### DIFF
--- a/variables/builtin.go
+++ b/variables/builtin.go
@@ -43,6 +43,15 @@ func BuiltinArgs(target domain.Target, platform specs.Platform, gitMeta *gitutil
 		ret.AddInactive("EARTHLY_GIT_ORIGIN_URL_SCRUBBED", stringutil.ScrubCredentials(gitMeta.RemoteURL))
 		ret.AddInactive("EARTHLY_GIT_PROJECT_NAME", getProjectName(gitMeta.RemoteURL))
 		ret.AddInactive("EARTHLY_GIT_COMMIT_TIMESTAMP", gitMeta.Timestamp)
+
+		if gitMeta.Timestamp == "" {
+			ret.AddInactive("EARTHLY_SOURCE_DATE_EPOCH", "0")
+		} else {
+			ret.AddInactive("EARTHLY_SOURCE_DATE_EPOCH", gitMeta.Timestamp)
+		}
+	} else {
+		// Ensure SOURCE_DATE_EPOCH is always available
+		ret.AddInactive("EARTHLY_SOURCE_DATE_EPOCH", "0")
 	}
 	// Note: Please update targetinput.go BuiltinVariables if adding more builtin variables.
 	for _, key := range ret.SortedAny() {
--- a/states/dedup/targetinput.go
+++ b/states/dedup/targetinput.go
@@ -160,6 +160,7 @@ var BuiltinVariables = map[string]bool{
 	"TARGETOS":                        true,
 	"TARGETARCH":                      true,
 	"TARGETVARIANT":                   true,
+	"EARTHLY_SOURCE_DATE_EPOCH":       true,
 }
 
 // IsDefaultValue returns whether the value of the BuildArgInput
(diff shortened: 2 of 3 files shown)


vladaionescu:
Is the main difference between `EARTHLY_SOURCE_DATE_EPOCH ` and `EARTHLY_GIT_COMMIT_TIMESTAMP` the default value?

I would only have one of these for simplicity. Perhaps instead we make `EARTHLY_GIT_COMMIT_TIMESTAMP` have a default value of `0` and always be defined.

pmembrey:
At the moment, the only difference is that one has a default value. I am a big fan of simplicity, so I don't want two variables that are almost the same. We can definitely get away with keeping it as is.

There are a couple of reasons why I think we should have a separate variable for simplicity's sake though 🙂

First, it means that the git commit timestamp field is always just that, git's latest timestamp. You don't need to interpret the value, if it's there you can use it, else it's blank like the others.

Second, although right now it's just the git timestamp, the epoch could come from other sources. For example, if Eathly honours the standard environment variable, we would want that to be used for the epoch. If we overload the git timestamp, then it is no longer possible to get the actual timestamp. If we don't, then you need conditional logic to handle the case where both are set, which means more logic in the Earthfile.

Third, we don't want users to have to guess or do logic to use this. If they've never heard of source date epoch, we don't want to override anything from the git variables else it will be very confusing. Ideally we follow the path of least surprise.

So basically, whilst I fully agree we don't want two variables that are almost exactly the same, I think in this case what we're really doing is providing a consistent interface that behaves as expected, whilst laying the groundwork to support source date epoch and timestamps from other sources.

Apologies for the long reply 😅

variables/builtin.go
```diff
@@ -43,6 +43,15 @@ func BuiltinArgs(target domain.Target, platform specs.Platform, gitMeta *gitutil
...
 		ret.AddInactive("EARTHLY_GIT_PROJECT_NAME", getProjectName(gitMeta.RemoteURL))
 		ret.AddInactive("EARTHLY_GIT_COMMIT_TIMESTAMP", gitMeta.Timestamp)
+
+		if gitMeta.Timestamp == "" {
+			ret.AddInactive("EARTHLY_SOURCE_DATE_EPOCH", "0")
+		} else {
+			ret.AddInactive("EARTHLY_SOURCE_DATE_EPOCH", gitMeta.Timestamp)
+		}
+	} else {
+		// Ensure SOURCE_DATE_EPOCH is always available
```
vladaionescu, line 53:
```suggestion
		// Ensure EARTHLY_SOURCE_DATE_EPOCH is always available
```

vladaionescu (commented):
I think the reasoning makes sense.

I also contemplated whether this should be named just `SOURCE_DATE_EPOCH` to comply with the standard. But then I thought that there may be cases where `SOURCE_DATE_EPOCH` needs to be constructed according to some more custom logic and the user should have freedom to do that.

Could you also augment the remote test to do some basic verification of these two new variables?

The test is here: https://github.com/earthly/test-remote/blob/main/builtin-args/Earthfile
And it's called from here: https://github.com/earthly/earthly/blob/main/examples/tests/Earthfile#L226 (you can run this via `./earthly -P ./examples/tests+remote-test`)

pmembrey:
I wasn't able to get the tests to run as it said I didn't have credentials. I'll add the tests and add the new epoch tomorrow :)

vladaionescu:
> I wasn't able to get the tests to run as it said I didn't have credentials. I'll add the tests and add the new epoch tomorrow :)

Right I forgot - you need to do `./earthly --build-arg DOCKERHUB_AUTH=false -P ./examples/tests+remote-test`

Oh also this test would be useful to augment in order to lock down the default zero value of each build arg (this runs without a git context): https://github.com/earthly/earthly/blob/main/examples/tests/builtin-args.earth

vladaionescu (approved):

vladaionescu:
I ended up merging this because we want to create a release. I'll add those tests in a separate PR.

PR #1142, opened 2021-08-03
Implement `FOR ... IN ...`

vladaionescu:
Part of https://github.com/earthly/earthly/issues/779

### DIFF
--- a/examples/tests/for.earth
+++ b/examples/tests/for.earth
@@ -1,0 +1,69 @@
+FROM alpine:3.13
+
+WORKDIR /test
+
+all:
+    BUILD +test-for-ls
+    BUILD +test-for-comma
+    BUILD +test-for-comma-colon
+    BUILD +test-for-constant
+    # TODO @# how to test parallelizability?
+    # @#
+    #BUILD +test-for-ls-locally
+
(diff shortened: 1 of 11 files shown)


dchw (approved):
LGTM. Wonder if we can use the SetArg stuff to also help overriding/re-setting build args?

examples/tests/for.earth
```diff
@@ -0,0 +1,80 @@
...
+        RUN echo "variable=$variable"
+        RUN echo "$variable" >>./output.txt
+    END
+    RUN cat ./output.txt
+    RUN cat ./output.txt | tr '\n' ' '
+    RUN test "$(cat ./output.txt | tr '\n' ' ')" = "a b c d "
+
+test-for-comma:
+    RUN echo "foo,bar:buz:zub,oof" >./input.txt
+    FOR --sep="," what IN $(cat ./input.txt)
```
dchw, line 30: Nervous that this turns into IFS someday

vladaionescu:
> Wonder if we can use the SetArg stuff to also help overriding/re-setting build args?

Yes, it should. We just need to think about the semantics carefully.

PR #1143, opened 2021-08-03
Redirects updates

dchw:


vladaionescu (approved):

dchw:
There are many more redirects now, requesting a re-review.

Here are all the problematic URLs in case you want to sample or explore:

vladaionescu (approved):

PR #1145, opened 2021-08-04
Ensure mirror is user for CI by removing service account direct login

dchw:
I have also manually logged out on each of the macOS/Windows build agents; and kept the mirror login. The thought is that we can't get rate limited if we never log in :thinking: 

I plan to abuse and re-run these over and over to see if I can force rate limiting.


dchw:
This took _way_ more lifting than expected, and I found some interesting edge cases. Might be worth turning these into some issues to clean up later.

vladaionescu (approved):

PR #1147, opened 2021-08-06
Docs for FOR and for --use-copy-include-patterns

vladaionescu:
Part of #779

### DIFF
--- a/docs/earthfile/earthfile.md
+++ b/docs/earthfile/earthfile.md
@@ -950,7 +950,7 @@ ELSE
 END
 ```
 
-The reason this is invalid is because the `IF` condition is actually running the `/usr/bin/[` executable to test if the condition is true or false, and therefore requires a valid build environment has been initialized.
+The reason this is invalid is because the `IF` condition is actually running the `/usr/bin/[` executable to test if the condition is true or false, and therefore requires that a valid build environment has been initialized.
 
 Here is how this might be fixed.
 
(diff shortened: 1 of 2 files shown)


docs/earthfile/earthfile.md
```diff
@@ -994,6 +994,83 @@ Same as [`RUN --mount <mount-spec>`](#mount-less-than-mount-spec-greater-than).
...
+  FOR [<options>...] <variable-name> IN <expression>
+    <for-block>
+  END
+  ```
+
+#### Description
+
+The `FOR` clause can iterate over the items resulting from the expression `<expression>`. On each iteration, the value of `<variable-name>` is set to the current item in the iteration and the block of commands `<for-block>` is executed in the context of that variable set as a build arg.
+
+The expression may be either a constant list of items (e.g. `foo bar buz`), or the output of a command (e.g. `$(echo foo bar buz)`), or a paramterized list of items (e.g. `foo $BARBUZ`). The result of the expression is then tokenized using the list of separators provided via the `--sep` option. If unspecified, the separator list defaults to `[tab]`, `[new line]` and `[space]` (`\t\n `).
```
dchw, line 1023:
```suggestion
The expression may be either a constant list of items (e.g. `foo bar buz`), or the output of a command (e.g. `$(echo foo bar buz)`), or a parameterized list of items (e.g. `foo $BARBUZ`). The result of the expression is then tokenized using the list of separators provided via the `--sep` option. If unspecified, the separator list defaults to `[tab]`, `[new line]` and `[space]` (`\t\n `).
```

docs/earthfile/earthfile.md
```diff
@@ -994,6 +994,83 @@ Same as [`RUN --mount <mount-spec>`](#mount-less-than-mount-spec-greater-than).
...
+
+{% hint style='danger' %}
+##### Important
+Changes to the filesystem in expressions are not preserved. If a file is created as part of a `FOR` expression, then that file will not be present in the build environment for any subsequent commands.
+{% endhint %}
```
dchw, line 1028: Made me think (and I did not test)... did we (and should we) block `RUN --interactive` in a `FOR` loop?

dchw (approved):

docs/earthfile/earthfile.md
vladaionescu: Ah right... good idea.

vladaionescu: Or maybe there are valid use-cases for it. Like the user making a series of decisions about something. Like a series of yes/no questions.

vladaionescu: I just tested this and it errors out with a reasonable message:

```
Error: build target: build main: bkClient.Build: failed to solve: w/Earthfile line 8:4 unset i: If present, a single --interactive command must be the last command in a target
in              github.com/earthly/earthly/w:main+test
```

PR #1150, opened 2021-08-06
Add login so we can push examples on merge to main

dchw:
On pushing to main, we run a `--push`, which requires the service account to be logged in. Add back the login in just these two cases.


vladaionescu (approved):

vladaionescu:
Note that push is disabled in PR builds. So it needs to be merged to really test that.

Issue #938, opened 2021-04-20
Under LOCALLY commands from different targets should not alternate


vladaionescu:
Reasons:
* Easier to follow the output
* Less likely that the targets conflict with each other in ways that make the build inconsistent

Part of #580 

CC @alexcb

Issue #580, opened 2020-12-04
Proposal: LOCALLY


vladaionescu:
This proposal builds upon a previous (now abandoned) version described in https://github.com/earthly/earthly/issues/167

## Background

Not everything a build system does is suitable for a container. There are certain tasks which are either too cumbersome for a container or Earthly just doesn't support them yet and may never will, for build repeatability reasons.

Because of the above, many Earthly users end up wrapping Earthly in other scripts (bash or Makefile) to be able to complete their development workflow.

### Use-cases

* In general, wrap common development commands that can only run on the host
* docker-compose up commands
* Watch type of commands, where mounting the source code is necessary - so it’s best done via a docker run
* Commands interacting with local config / secrets - example: grabbing local AWS config or grabbing the local maven credentials
* Situations where using the host docker daemon is necessary / preferred

## Proposal

An "escape hatch" is provided to be able to execute commands on the host system rather than in a container. Targets that execute on the host are marked as `FROM LOCAL`.

The intended use of `FROM LOCAL` is for **development commands only** and not for use in CI. To make this distinction for the user, a new *strict* mode is introduced (default: off) in which `FROM LOCAL` commands are not allowed. A warning is displayed for cases where strict mode is off but the build includes `FROM LOCAL`.

### Example

```Dockerfile
some-target:
  FROM LOCAL
  RUN docker-compose up
```

### Notes

* Executes commands directly on the host machine
* The current directory used is the directory in which the Earthfile is defined (not the directory it is called from)
* If the `FROM LOCAL` target is a remote target (referenced from github), then the current directory used is a directory under `~/.earthly/tmp/<unique-id>/github.com/bla/bla/deep/dir`, where the remote git dir is cloned into.
* No caching, no layers, no `SAVE IMAGE`
* `SAVE ARTIFACT` can be used to expose files to regular targets. Regular (containerized) targets can reference the artifact in the usual way - this results in builds that partly execute on the host, and partly in containers.
* Ordering of commands are guaranteed - eg `BUILD +one` and `BUILD +two` on consecutive lines, will build and output sequentially, guaranteed
* `COPY`, `LABEL`, `CMD`, `ENTRYPOINT` don’t work in this context (we throw an error)
* Because `FROM LOCAL` targets interact with the local system, perhaps we turn off any parallelization when executing these. (More reason for the user to avoid these for anything heavy-duty)
[...]

PR #1151, opened 2021-08-06
Refactor implementation of the various RUN flavors

vladaionescu:
* Unify all implementations of run into a single big, general function
* Ensures `LOCALLY` commands run sequentially (Fixes https://github.com/earthly/earthly/issues/938, part of https://github.com/earthly/earthly/issues/580)
* Removes support for `RUN --push` under `LOCALLY`


dchw (approved):
Hard to follow the diffs! BUt I like pushing it all into `internalRun` more or less to keep behavior consistent.

I feel like we should get some unit tests around this function going forward.

alexcb (approved):

earthfile2llb/converter.go
```diff
@@ -1517,21 +1439,43 @@ func (c *Converter) internalRun(ctx context.Context, args, secretKeyValues []str
...
 }
 
+func (c *Converter) forceExecution(ctx context.Context, state pllb.State) error {
+	ref, err := llbutil.StateToRef(ctx, c.opt.GwClient, state, c.opt.Platform, c.opt.CacheImports.AsMap())
+	if err != nil {
+		return errors.Wrap(err, "run locally state to ref")
+	}
+	// We're not really interested in reading the dir - we just
+	// want to un-lazy the ref so that the local commands have executed.
+	_, err = ref.ReadDir(ctx, gwclient.ReadDirRequest{Path: "/"})
```
alexcb, line 1472: nice hack :)

PR #1157, opened 2021-08-11
enable use-copy-include-patterns in ci and ./earthly

alexcb:
feature flip --use-copy-include-patterns in ci and ./earthly script.


vladaionescu (approved):

vladaionescu:
This needs a rebase from `main`?

vladaionescu:
It might be useful to also enable this in buildkite for windows & mac testing

alexcb:
That's a good plan, I wonder if we could define these flags in a single location and have both GHA and buildkite make use of them? I'll take a look and see if I can do that.

PR #1158, opened 2021-08-11
Add retries for server dial that fails sometime in unit tests

dchw:


debugger/server/server_test.go
```diff
@@ -24,11 +24,22 @@ func TestServer(t *testing.T) {
 
 	time.Sleep(10 * time.Millisecond)
 
-	// first open terminal
-	termConn, err := net.Dial("tcp", addr)
+	numRetries := 3
```
vladaionescu, line 27: Nit: Slightly more readable

```suggestion
	const numRetries = 3
```

vladaionescu (approved):

alexcb (approved):

PR #1163, opened 2021-08-13
test for sequential runs

alexcb:
This test is to ensure all LOCALLY blocks are executed sequentially.

### DIFF
--- a/examples/tests/Earthfile
+++ b/examples/tests/Earthfile
@@ -716,7 +716,7 @@ true-false-flag-invalid:
     RUN cat output3.txt | grep 'invalid argument for flag .*--no-cache.*expected bool'
 
 sequential-locally-test:
-    DO +RUN_EARTHLY --earthfile=sequential-locally.earth --use_tmpfs=false --target="+run-lots" --extra_args="--no-cache" --post_command="2>output.txt"
+    DO +RUN_EARTHLY --earthfile=sequential-locally.earth --use_tmpfs=false --target="+run-lots" --post_command="2>output.txt"
     RUN cat output.txt | grep -vw RUN | grep -vw char | grep -o '\(start\|mid\|end\).*' > output-filtered.txt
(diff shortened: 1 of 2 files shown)


examples/tests/sequential-locally.earth
```diff
@@ -0,0 +1,33 @@
...
+    RUN echo "mid $char"
+    RUN echo "mid $char"
+    RUN echo "mid $char"
+    RUN echo "mid $char"
+    RUN echo "mid $char"
+    RUN echo "mid $char"
+    RUN echo "mid $char"
+    RUN echo "mid $char"
+    RUN echo "mid $char"
+    RUN echo "end $char"
```
vladaionescu, line 25: LoL. That should do it.

examples/tests/Earthfile
```diff
@@ -714,6 +715,42 @@ true-false-flag-invalid:
     DO +RUN_EARTHLY --earthfile=true-false-flag-invalid.earth --should_fail=true --target=+run-maybe --post_command="2>output3.txt"
     RUN cat output3.txt | grep 'invalid argument for flag .*--no-cache.*expected bool'
 
+sequential-locally-test:
+    DO +RUN_EARTHLY --earthfile=sequential-locally.earth --use_tmpfs=false --target="+run-lots" --extra_args="--no-cache" --post_command="2>output.txt"
```
vladaionescu, line 719: `--no-cache` isn't really needed given that these are LOCALLY targets, right?

```suggestion
    DO +RUN_EARTHLY --earthfile=sequential-locally.earth --use_tmpfs=false --target="+run-lots" --post_command="2>output.txt"
```

vladaionescu (approved):

examples/tests/Earthfile
alexcb: good point, it won't have any effect as LOCALLY commands are never cached.

PR #1164, opened 2021-08-13
Prevent saving artifacts to parent directories

alexcb:
Currently if a user specifies

    SAVE ARTIFACT some-file AS LOCAL .

This error occurs:

    $ ./earthly ~/broken/save-dot+sd
      /h/a/b/save-dot+sd | --> FROM alpine
      /h/a/b/save-dot+sd | [██████████] resolve docker.io/library/alpine@sha256:eb3e4e175ba6d212ba1d6e04fc0782916c08e1c9d7b45892e9796141b1d379ae ... 100%
      /h/a/b/save-dot+sd | --> RUN ls -la > output
      /h/a/b/save-dot+sd | --> SAVE ARTIFACT output /home/alex/broken/save-dot+sd/output AS LOCAL .
                  output | [██████████] copying files ... 100%
    ================================ SUCCESS [main] ================================
    Error: build target: rm /home/alex/broken/save-dot: remove /home/alex/broken/save-dot: directory not empty

Instead the user most likely wanted to specify

    SAVE ARTIFACT some-file AS LOCAL ./

Either we should raise a more appropriate error (rather than attempt to
delete the current directory, which fortunately fails), or as this PR
suggests, we can just assume the user meant save "some-file" to the
current directory.

This makes it more consistent with the current
non-local SAVE ARTIFACT commands:

    SAVE ARTIFACT some-file .
    SAVE ARTIFACT some-file ./
    SAVE ARTIFACT some-file ./some-file

which are currently interpreted the same way.

### DIFF
--- a/earthfile2llb/interpreter.go
+++ b/earthfile2llb/interpreter.go
@@ -798,7 +798,7 @@ func (i *Interpreter) handleSaveArtifact(ctx context.Context, cmd spec.Command) 
 		return nil
 	}
 
-	err = i.converter.SaveArtifact(ctx, saveFrom, saveTo, saveAsLocalTo, opts.KeepTs, opts.KeepOwn, opts.IfExists, opts.SymlinkNoFollow, i.pushOnlyAllowed)
+	err = i.converter.SaveArtifact(ctx, saveFrom, saveTo, saveAsLocalTo, opts.KeepTs, opts.KeepOwn, opts.IfExists, opts.SymlinkNoFollow, opts.Force, i.pushOnlyAllowed)
 	if err != nil {
 		return i.wrapError(err, cmd.SourceLocation, "apply SAVE ARTIFACT")
 	}
--- a/earthfile2llb/converter.go
+++ b/earthfile2llb/converter.go
@@ -22,6 +22,7 @@ import (
 	"github.com/earthly/earthly/states"
 	"github.com/earthly/earthly/states/dedup"
 	"github.com/earthly/earthly/states/image"
+	"github.com/earthly/earthly/util/fileutil"
 	"github.com/earthly/earthly/util/gitutil"
 	"github.com/earthly/earthly/util/llbutil"
 	"github.com/earthly/earthly/util/llbutil/llbfactory"
@@ -69,6 +70,7 @@ const (
 
 // Converter turns earthly commands to buildkit LLB representation.
 type Converter struct {
+	target              domain.Target
 	gitMeta             *gitutil.GitMetadata
 	opt                 ConvertOpt
 	mts                 *states.MultiTarget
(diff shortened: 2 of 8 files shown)


earthfile2llb/interpreter.go
```diff
@@ -787,6 +787,10 @@ func (i *Interpreter) handleSaveArtifact(ctx context.Context, cmd spec.Command)
 	saveTo = i.expandArgs(saveTo, false)
 	saveAsLocalTo = i.expandArgs(saveAsLocalTo, false)
 
+	if saveAsLocalTo == "." {
+		saveAsLocalTo = "./"
+	}
```
vladaionescu, line 792: Mind moving this to the converter? Best to keep the interpreter as agnostic as possible.

vladaionescu (approved):

alexcb:
@vladaionescu We might want to iterate on this PR a bit more.

Should we prevent a user from saving to `..`? because here's a very scary situation where one could accidentally delete their parent directory:

consider this script, let's save it to `~/bin/scary.sh`

```
#!/bin/sh
set -e
rm -rf /tmp/scary
mkdir -p /tmp/scary/dot-dot
echo "FROM alpine
whoops:
    RUN mkdir mydir
    RUN echo whoops > mydir/myfile
    SAVE ARTIFACT mydir AS LOCAL ..
" > /tmp/scary/dot-dot/Earthfile

echo important > /tmp/scary/important-file

ls -la /tmp/scary/important-file

earthly /tmp/scary/dot-dot+whoops

ls -la /tmp/scary/important-file
```

now when we run it:

```
$ scary.sh
-rw-rw-r-- 1 alex alex 10 Aug 13 17:38 /tmp/scary/important-file     <--- my important file exists
   /t/s/dot-dot+base | --> FROM alpine
   /t/s/dot-dot+base | [██████████] resolve docker.io/library/alpine@sha256:eb3e4e175ba6d212ba1d6e04fc0782916c08e1c9d7b45892e9796141b1d379ae ... 100%
 /t/s/dot-dot+whoops | --> RUN mkdir mydir
 /t/s/dot-dot+whoops | --> RUN echo whoops > mydir/myfile
 /t/s/dot-dot+whoops | --> SAVE ARTIFACT mydir /tmp/scary/dot-dot+whoops/mydir AS LOCAL ..
              output | [██████████] copying files ... 100%
================================ SUCCESS [main] ================================
 /t/s/dot-dot+whoops | Artifact /tmp/scary/dot-dot+whoops/mydir as local ..
ls: cannot access '/tmp/scary/important-file': No such file or directory   <--- after earthly exits, my parent directory has been deleted
```

and now `/tmp/scary` contains:

```
$ ls /tmp/scary/
myfile
```

should we disallow a case where we're outputting somewhere above the current directory of the Earthfile? I would hate to think what would happen if someone saved to `SAVE ARTIFACT dir AS LOCAL /.`

vladaionescu:
Those are some really good points. Maybe we disallow dir artifacts outside of the Earthfile's dir. Earthly deletes the destination only for those.

alexcb:
Assuming an Earthfile under `/root/some/path`; this PR now causes a command like `SAVE ARTIFACT /data AS LOCAL /root` to fail.

and when the `--force` flag is added, e.g. `SAVE ARTIFACT --force /data AS LOCAL /root`, the command will succeed and overwrite the destination.

vladaionescu (approved):
💯

The reasoning makes sense

PR #1171, opened 2021-08-18
move github specific password-based https tests to selfhosted

alexcb:
move our password-based github specific tests to a containerized
self-hosted test which are run within earthly instead of a GHA-specific
test.
    

examples/tests/private-https/Earthfile
```diff
@@ -0,0 +1,77 @@
+FROM ..+base
+
+all:
+    BUILD +git-clone-private-https-set-by-args
+    BUILD +git-clone-private-https-set-by-config-command
+    BUILD +git-clone-private-https-set-by-config-file
+    BUILD +git-clone-private-https-set-by-netrc
+
+
+git-clone-private-https-base:
```
dchw, line 10: Maybe this target would be a cool UDC to share with the community?

dchw (approved):

examples/tests/private-https/Earthfile
alexcb: it's pretty hacky right now, but maybe we can get there eventually?

Maybe we could incorporate it into a blog post regarding these self-signed cert tests.

PR #1172, opened 2021-08-19
Add config option escape hatch for iptables detection

dchw:


buildkitd/entrypoint.sh
```diff
@@ -38,19 +31,51 @@ if [ "$EARTHLY_RESET_TMP_DIR" = "true" ]; then
...
+
+    if lsmod | grep -wq "^ip_tables"; then
+        echo "Detected iptables-legacy module"
+        IP_TABLES="iptables-legacy"
+
+    elif lsmod | grep -wq "^nf_tables"; then
+        echo "Detected iptables-nft module"
+        IP_TABLES="iptables-nft"
+    else
+        echo "Could not find an ip_tables module. Using heuristics."
```
alexcb, line 45: How about re-wording this to a single sentence:

```suggestion
        echo "Could not find an ip_tables module; falling back to heuristics."
```

that way it's less likely a user will misread it as `Could not find an ip_tables module using heuristics.`.

dchw: Sure. I question people missing punctuation though :P

buildkitd/entrypoint.sh
```diff
@@ -38,19 +31,51 @@ if [ "$EARTHLY_RESET_TMP_DIR" = "true" ]; then
...
+        echo "Detected iptables-legacy module"
+        IP_TABLES="iptables-legacy"
+
+    elif lsmod | grep -wq "^nf_tables"; then
+        echo "Detected iptables-nft module"
+        IP_TABLES="iptables-nft"
+    else
+        echo "Could not find an ip_tables module; falling back to heuristics."
+
+        legacylines=$(iptables-legacy -t nat -S --wait | wc -l | cut -d' ' -f1)
```
alexcb, line 47: is the `cut` really needed here; is there a case where `wc -l` outputs more than one result?
```suggestion
        legacylines=$(iptables-legacy -t nat -S --wait | wc -l)
```

dchw: `wc -l` returns a filename too. I wanted just the number of lines.

dchw: See my edit

alexcb (approved):

PR #1174, opened 2021-08-19
Extend COPY --if-exists to classical COPY

alexcb:
extend --if-exists to support the clasic COPY command for copying files
from the local host; if the file doesn't exist, the copy doesn't execute
and earthly continues to the next command.

### DIFF
--- a/ast/tests/if-exists.ast.json
+++ b/ast/tests/if-exists.ast.json
@@ -290,6 +290,32 @@
           }
         }
       ]
+    },
+    {
+      "name": "classic-copy",
+      "recipe": [
+        {
+          "command": {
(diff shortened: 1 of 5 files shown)


dchw (approved):

aaronrenner:
I was just looking for the is exact flag and I'm excited to see you've already built it! Thanks for all of your great work on earthly! :heart:  I can hardly wait until this feature is released!

alexcb:
> I can hardly wait

@aaronrenner if you don't want to wait, we have [a script](https://github.com/earthly/earthly/blob/main/earthly) which will allow you to run on the latest pre-release version.

Issue #1108, opened 2021-07-19
interactive debugger should not run (or display errors) when running in non-interactive session


jclangst:
Version: v0.5.18

In our CI environments, the Docker daemon exists on a separate host and we communicate over TCP. In this type of setup, every earthly command generates the following error message; however, functionality is NOT impaired.

`internal-term | Failed to connect to terminal: dial tcp 127.0.0.1:8373: connect: connection refused`

alexcb:
Thanks for the report; it appears that any interactive RUN commands (or the interactive debugger) won't work with a non-local docker daemon. Fortunately in the context of CI there won't be an interactive commands which won't affect your use-case.

It's definitely something we'll want to fix, but it might be a lower priority.

PR #1176, opened 2021-08-23
only connect to interactive terminal when running under tty

alexcb:
This will prevent internal-term errors from occuring (and being printed
to stderr) when running under CI.


alexcb:
This attempts to fix #1108

vladaionescu (approved):

PR #1179, opened 2021-08-23
set feature flag overrides in a single location

alexcb:
This ensures that both GHA and buildkite are using the same defined
features.

### DIFF
--- a/.buildkite/tests.sh
+++ b/.buildkite/tests.sh
@@ -6,6 +6,9 @@ trap 'kill $(jobs -p); wait' SIGINT SIGTERM
 
 export EARTHLY_CONVERSION_PARALLELISM=5
 
+EARTHLY_VERSION_FLAG_OVERRIDES="$(tr -d '\n' < .earthly_version_flag_overrides)"
+export EARTHLY_VERSION_FLAG_OVERRIDES
+
 case "$EARTHLY_OS" in
     darwin)
         download_url="https://github.com/earthly/earthly/releases/latest/download/earthly-darwin-amd64"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,6 @@ jobs:
     env:
       FORCE_COLOR: 1
       EARTHLY_CONVERSION_PARALLELISM: "5"
-      EARTHLY_VERSION_FLAG_OVERRIDES: "referenced-save-only,use-copy-include-patterns"
       EARTHLY_TOKEN: "${{ secrets.EARTHLY_TOKEN }}"
       EARTHLY_INSTALL_ID: "earthly-githubactions"
     steps:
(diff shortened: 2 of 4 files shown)


vladaionescu:
Wondering if there may be cases where we'd want the list to be different for a specific test.

.buildkite/tests.sh
```diff
@@ -6,6 +6,9 @@ trap 'kill $(jobs -p); wait' SIGINT SIGTERM
 
 export EARTHLY_CONVERSION_PARALLELISM=5
 
+EARTHLY_VERSION_FLAG_OVERRIDES="$(tr -d '\n' < .earthly_version_flag_overrides)"
+export EARTHLY_VERSION_FLAG_OVERRIDES
```
vladaionescu, line 10:
```suggestion
export EARTHLY_VERSION_FLAG_OVERRIDES="$(tr -d '\n' < .earthly_version_flag_overrides)"
```

.github/workflows/ci.yml
```diff
@@ -110,6 +112,11 @@ jobs:
         if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name == github.repository
       - name: Build latest earthly using released earthly
         run: earthly --use-inline-cache +for-linux
+      - name: Set EARTHLY_VERSION_FLAG_OVERRIDES env
+        run: |-
+            set -euo pipefail
+            EARTHLY_VERSION_FLAG_OVERRIDES="$(tr -d '\n' < .earthly_version_flag_overrides)"
+            echo "EARTHLY_VERSION_FLAG_OVERRIDES=$EARTHLY_VERSION_FLAG_OVERRIDES" >> "$GITHUB_ENV"
```
vladaionescu, line 119: Would this work on a single line?

```suggestion
            echo "EARTHLY_VERSION_FLAG_OVERRIDES=$(tr -d '\n' < .earthly_version_flag_overrides)" >> "$GITHUB_ENV"
```

earthly
```diff
@@ -96,7 +97,9 @@ case "$1" in
         fi
 
         export EARTHLY_CONVERSION_PARALLELISM=5
-        export EARTHLY_VERSION_FLAG_OVERRIDES="referenced-save-only,use-copy-include-patterns"
+        EARTHLY_VERSION_FLAG_OVERRIDES="$(tr -d '\n' < "$script_dir/.earthly_version_flag_overrides")"
+        export EARTHLY_VERSION_FLAG_OVERRIDES
```
vladaionescu, line 101:
```suggestion
        export EARTHLY_VERSION_FLAG_OVERRIDES="$(tr -d '\n' < "$script_dir/.earthly_version_flag_overrides")"
```

vladaionescu (approved):
Minor suggestions

.buildkite/tests.sh
alexcb: I had to break it up in order to have error checking for a case where the `.earthly_version_flag_overrides` file doesn't exist:

```
#!/bin/bash
set -euo pipefail
export EARTHLY_VERSION_FLAG_OVERRIDES="$(tr -d '\n' < this-file-doesnt-exist)"
echo "EARTHLY_VERSION_FLAG_OVERRIDES was set to $EARTHLY_VERSION_FLAG_OVERRIDES"


EARTHLY_VERSION_FLAG_OVERRIDES="$(tr -d '\n' < this-file-doesnt-exist)"
export EARTHLY_VERSION_FLAG_OVERRIDES
echo "this shouldnt be called"
```

When run this will display:
```
$ ./break.sh 
./break.sh: line 3: this-file-doesnt-exist: No such file or directory
EARTHLY_VERSION_FLAG_OVERRIDES was set to 
./break.sh: line 7: this-file-doesnt-exist: No such file or directory
```

showing that the all-in-one command doesn't fail when the file doesn't exist.

I initial discovered this via: https://github.com/koalaman/shellcheck/wiki/SC2155

.github/workflows/ci.yml
alexcb: similarly I had to break this up to address error checking

.buildkite/tests.sh
vladaionescu: ah gotcha

PR #1180, opened 2021-08-24
Fix config, unify feature flag and URL scheme.

dchw:
Also add unit tests for unified config.

### DIFF
--- a/cmd/earthly/main.go
+++ b/cmd/earthly/main.go
@@ -1168,6 +1168,10 @@ func parseAndvalidateURL(addr string) (*url.URL, error) {
 	}
 
 	if parsed.Scheme != "tcp" && parsed.Scheme != "docker-container" {
+		if parsed.Scheme == "" {
+			// This text should satisfy the homebrew test, as it pases a schemeless URL
+			return nil, fmt.Errorf("buildkitd failed to start: %w", errURLValidationFailure)
+		}
 		return nil, fmt.Errorf("%s is not a valid scheme. Only tcp or docker-container is allowed at this time: %w", parsed.Scheme, errURLValidationFailure)
 	}
 
--- a/cmd/earthly/main.go
+++ b/cmd/earthly/main.go
@@ -1132,7 +1132,7 @@ func (app *earthlyApp) getAndValidateAddresses(context *cli.Context) (addresses,
 	}
 
 	if bkURL.Hostname() == dbURL.Hostname() && bkURL.Port() == dbURL.Port() {
-		return addresses{}, fmt.Errorf("Debugger and Buildkit ports are the same: %w", urlValidationFailure)
+		return addresses{}, fmt.Errorf("Debugger and Buildkit ports are the same: %w", errURLValidationFailure)
 	}
 
 	return addresses{
@@ -1164,21 +1164,21 @@ func (app *earthlyApp) handleTLSCertificateSettings(context *cli.Context) {
 	app.buildkitdSettings.ServerTLSKey = app.cfg.Global.ServerTLSKey
 }
 
-var urlParseFailure = errors.New("Invalid URL")
-var urlValidationFailure = errors.New("URL did not pass validation")
+var errURLParseFailure = errors.New("Invalid URL")
+var errURLValidationFailure = errors.New("URL did not pass validation")
 
 func parseAndvalidateURL(addr string) (*url.URL, error) {
 	parsed, err := url.Parse(addr)
 	if err != nil {
-		return nil, fmt.Errorf("%s: %w", addr, urlParseFailure)
+		return nil, fmt.Errorf("%s: %w", addr, errURLParseFailure)
 	}
 
 	if parsed.Scheme != "tcp" && parsed.Scheme != "docker-container" {
-		return nil, fmt.Errorf("%s is not a valid scheme. Only tcp or docker-container is allowed at this time: %w", parsed.Scheme, urlValidationFailure)
+		return nil, fmt.Errorf("%s is not a valid scheme. Only tcp or docker-container is allowed at this time: %w", parsed.Scheme, errURLValidationFailure)
 	}
 
 	if parsed.Port() == "" && parsed.Scheme == "tcp" {
-		return nil, fmt.Errorf("%s does not contain a port number: %w", addr, urlValidationFailure)
+		return nil, fmt.Errorf("%s does not contain a port number: %w", addr, errURLValidationFailure)
 	}
 
 	return parsed, nil
--- a/cmd/earthly/main.go
+++ b/cmd/earthly/main.go
@@ -1020,21 +1020,19 @@ func (app *earthlyApp) before(context *cli.Context) error {
 	}
 
 	var addrs addresses
-	switch app.cfg.Global.BuildkitScheme {
-	case "tcp":
-		addrs, err = app.getAddressesForTCP(context)
-		if err != nil {
-			return err
-		}
-		app.handleTLSCertificateSettings(context)
 
-	case "docker-container":
-		addrs, err = app.getAddressesOriginal()
-		if err != nil {
-			return err
-		}
-	default:
-		return fmt.Errorf("%s is not a valid buildkit scheme", app.cfg.Global.BuildkitScheme)
+	addrs, err = app.getAndValidateAddresses(context)
+	if err != nil {
+		return err
+	}
+
+	bkURL, err := parseAndvalidateURL(addrs.buildkit)
+	if err != nil {
+		return errors.Wrapf(err, "invalid buildkit_host: %s", app.buildkitHost)
+	}
+
+	if bkURL.Scheme == "tcp" {
+		app.handleTLSCertificateSettings(context)
 	}
 
 	app.debuggerHost = addrs.debugger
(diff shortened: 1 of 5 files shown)


cmd/earthly/main.go
```diff
@@ -1020,21 +1020,19 @@ func (app *earthlyApp) before(context *cli.Context) error {
...
-		return fmt.Errorf("%s is not a valid buildkit scheme", app.cfg.Global.BuildkitScheme)
+	addrs, err = app.getAndValidateAddresses(context)
+	if err != nil {
+		return err
+	}
+
+	bkURL, err := parseAndvalidateURL(addrs.buildkit)
+	if err != nil {
+		return errors.Wrapf(err, "invalid buildkit_host: %s", app.buildkitHost)
+	}
```
dchw, line 1032: This has to happen after so we are sure we got the intended scheme.

cmd/earthly/main.go
```diff
@@ -1087,21 +1085,12 @@ type addresses struct {
...
-}
-
-func (app *earthlyApp) getAddressesForTCP(context *cli.Context) (addresses, error) {
+func (app *earthlyApp) getAndValidateAddresses(context *cli.Context) (addresses, error) {
 	if !context.IsSet("buildkit-host") {
 		if app.cfg.Global.BuildkitHost != "" {
 			app.buildkitHost = app.cfg.Global.BuildkitHost
 		} else {
-			app.buildkitHost = buildkitd.TCPAddress
+			app.buildkitHost = buildkitd.DockerAddress
```
dchw, line 1093: In keeping with this as our default address scheme, because this is the only codepath now.

cmd/earthly/main.go
```diff
@@ -1114,7 +1103,11 @@ func (app *earthlyApp) getAddressesForTCP(context *cli.Context) (addresses, erro
 		if app.cfg.Global.DebuggerHost != "" {
 			app.debuggerHost = app.cfg.Global.DebuggerHost
 		} else {
-			app.debuggerHost = fmt.Sprintf("tcp://%s:%v", bkURL.Hostname(), config.DefaultDebuggerPort)
+			if app.cfg.Global.DebuggerPort == config.DefaultDebuggerPort && bkURL.Scheme == "tcp" {
+				app.debuggerHost = fmt.Sprintf("tcp://%s:%v", bkURL.Hostname(), config.DefaultDebuggerPort)
+			} else {
+				app.debuggerHost = fmt.Sprintf("tcp://127.0.0.1:%v", app.cfg.Global.DebuggerPort)
+			}
```
dchw, line 1110: Since the old codepath needed to honor the port, we now support both. The  new, non-deprecated setting takes precedence, and the CLI wins over all. Port appears to no longer be available on the CLI?

cmd/earthly/main.go
```diff
@@ -1123,20 +1116,23 @@ func (app *earthlyApp) getAddressesForTCP(context *cli.Context) (addresses, erro
...
-		return addresses{}, err
+	if buildkitd.IsLocal(app.buildkitHost) && app.cfg.Global.LocalRegistryHost != "" {
+		// Local registry only matters when local, and specified.
+		lrURL, err := parseAndvalidateURL(app.cfg.Global.LocalRegistryHost)
+		if err != nil {
+			return addresses{}, err
+		}
+		if bkURL.Hostname() != lrURL.Hostname() {
+			app.console.Warnf("Buildkit and Local Registry URLs are pointed at different hosts (%s vs. %s)", bkURL.Hostname(), lrURL.Hostname())
+		}
```
dchw, line 1127: No checks and no validation if local. We assume that registry code downstream is smart enough.

An alternative here is if its non-local we return an empty string for the address.

cmd/earthly/main.go
```diff
@@ -1123,20 +1116,23 @@ func (app *earthlyApp) getAddressesForTCP(context *cli.Context) (addresses, erro
...
 	if bkURL.Hostname() != dbURL.Hostname() {
 		app.console.Warnf("Buildkit and Debugger URLs are pointed at different hosts (%s vs. %s)", bkURL.Hostname(), dbURL.Hostname())
 	}
-	if bkURL.Hostname() != lrURL.Hostname() {
-		app.console.Warnf("Buildkit and Local Registry URLs are pointed at different hosts (%s vs. %s)", bkURL.Hostname(), lrURL.Hostname())
-	}
 
 	if bkURL.Hostname() == dbURL.Hostname() && bkURL.Port() == dbURL.Port() {
-		return addresses{}, errors.New("Debugger and Buildkit ports are the same")
+		return addresses{}, fmt.Errorf("Debugger and Buildkit ports are the same: %w", urlValidationFailure)
```
dchw, line 1135: Arguable use of this sentinel error. It fit well enough and I didn't want to add a single-use sentinel just for one test..

cmd/earthly/main.go
```diff
@@ -1251,8 +1250,8 @@ func (app *earthlyApp) processDeprecatedCommandOptions(context *cli.Context, cfg
 		app.buildkitdSettings.CacheSizeMb = cfg.Global.BuildkitCacheSizeMb
 	}
 
-	if cfg.Global.DebuggerPort != config.DefaultDebuggerPort && cfg.Global.BuildkitScheme == "tcp" {
-		app.console.Warnf("Warning: specifying the port using the debugger-port setting is deprecated. Set it in ~/.earthly/config.yml under the global debugger_port setting.\n")
+	if cfg.Global.DebuggerPort != config.DefaultDebuggerPort {
+		app.console.Warnf("Warning: specifying the port using the debugger-port setting is deprecated. Set it in ~/.earthly/config.yml as part of the debugger_host variable; see https://docs.earthly.dev/earthly-config for reference.\n")
```
dchw, line 1254: More generally deprecated now.

conslogging/conslogging.go
```diff
@@ -126,6 +126,13 @@ func (cl ConsoleLogger) WithFailed(isFailed bool) ConsoleLogger {
 	return ret
 }
 
+func (cl ConsoleLogger) WithWriter(w io.Writer) ConsoleLogger {
+	ret := cl.clone()
+	ret.outW = w
+	ret.errW = w
+	return ret
+}
+
```
dchw, line 135: Needed a way to be able to also read logs from tests. You could also use this with `https://pkg.go.dev/io#MultiWriter` to effectively tee logs. Could be useful when implementing truncation schemes later.

dchw (commented):
Thoughts I thought were relevant

cmd/earthly/main.go
vladaionescu: Wondering if we should write out an error (or at least a warning?) if it's a remote buildkit and the local registry address is set.

vladaionescu (approved):

cmd/earthly/main.go
dchw: The more I think about it, the more I would like to return an empty string to force the issue when non local.

> Wondering if we should write out an error (or at least a warning?) if it's a remote buildkit and the local registry address is set.

I could see people having it set for local dev; and then pointing at a big buildkitd in the sky. It would be annoying for a config file value to prevent this easy switching. A warning seems best to me? WYT @vladaionescu ?

vladaionescu: Yeah, I can see how that can be annoying.

I wonder if the config should be a bit different to make the switch easier, without the gotcha implications. Maybe in the future, we can have a single host + port and we multiplex the different functionalities through that (buildkit, debugger, registry).

In the meantime, I think either a warning or no message is fine, given that it's anyway called **local** registry.

cmd/earthly/settings_test.go
```diff
@@ -0,0 +1,354 @@
...
+
+		logger := conslogging.Current(conslogging.NoColor, conslogging.DefaultPadding, false)
+		logger = logger.WithWriter(&logs)
+
+		earthlyApp := newEarthlyApp(ctx, logger)
+		earthlyApp.cfg = &config.Config{Global: tt.config}
+		earthlyApp.cliApp.Writer = &trash    // Just chuck the help output
+		earthlyApp.cliApp.ErrWriter = &trash // All of it, we dont care
+
+		// Before is called at about the time that we would parse these, plus it a nice place to hook.
```
alexcb, line 191: interesting hook. This test almost feels like an integration test that could be done via earthly-in-earthly which might provide more test coverage; but that would definitely take more execution time. I think it's fine here though.

cmd/earthly/main.go
dchw: Warning added.

cmd/earthly/settings_test.go
dchw: Yeah. I also would like too be able to debug these.

cmd/earthly/settings_test.go
```diff
@@ -0,0 +1,354 @@
...
+				debugger:      "tcp://127.0.0.1:5678",
+				localRegistry: "",
+			},
+		},
+		{
+			"Debugger host and port specified",
+			config.GlobalConfig{
+				BuildkitHost:      "",
+				DebuggerHost:      "tcp://127.0.0.1:1234",
+				DebuggerPort:      5678,
```
alexcb, line 105: is this test supposed to pass when `1234` doesn't match `5678`?

vladaionescu (approved):

cmd/earthly/settings_test.go
dchw: Yes. Specifying both means the host setting wins. The port setting is deprecated.

dchw: More clarification:

The old code would only use port. If the feature flag was on, this setting would generate a deprecation notice. When unifying the codepaths, I made the preferably-deprecated setting fully deprecated, and the new setting always preferred. You can get old behavior by omitting the new setting and only specifying port.

cmd/earthly/settings_test.go
dchw: Little more clarification:

I only hooked here because it was easier than building the `cli.Context` and all the stuff along with it (like a `FlagSet`). Waaay easier to allow the CLI framework to give me one.

alexcb (approved):

Issue #1114, opened 2021-07-22
./earthly -P github.com/earthly/earthly/examples/tests/dockerfile2/subdir:main+test fails


alexcb:
```
$ ./earthly -P github.com/earthly/earthly/examples/tests/dockerfile2/subdir:main+test 
           buildkitd | Found buildkit daemon as docker container (earthly-buildkitd)
             context | --> git context github.com/earthly/earthly/examples/tests/dockerfile2/subdir:main+test
         alpine:3.13 | --> Load metadata linux/amd64
Error: build target: build main: bkClient.Build: failed to solve: open /tmp/buildkit-mount447021056/examples/tests/dockerfile2/subdir/Dockerfile: no such file or directory
```

However, when I run it locally:

```
alex@mah:~/gh/earthly/earthly$ ./earthly -P ./examples/tests/dockerfile2/subdir+test 
           buildkitd | Found buildkit daemon as docker container (earthly-buildkitd)
         alpine:3.13 | --> Load metadata linux/amd64
                 1/3 | --> FROM docker.io/library/alpine:3.13@sha256:f51ff2d96627690d62fee79e6eecd9fa87429a38142b5df8a3bfbb26061df7fc
             context | --> local context ./examples/tests/dockerfile2/subdir
                 1/3 | [██████████] resolve docker.io/library/alpine:3.13@sha256:f51ff2d96627690d62fee79e6eecd9fa87429a38142b5df8a3bfbb26061df7fc ... 100%
             context | transferred 0 file(s) for context ./examples/tests/dockerfile2/subdir (24 B, 1 file/dir stats)
                 2/3 | *cached* --> COPY a.txt .
                 3/3 | *cached* --> RUN cat a.txt
 ./e/t/d/subdir+test | *cached* --> RUN --entrypoint cat a.txt
              output | --> exporting outputs
              output | [██████████] exporting layers ... 100%
              output | [██████████] exporting manifest sha256:45bf6f102b9313c153758e1eddd5862f21f290e365da524f25221342d137e806 ... 100%
              output | [██████████] exporting config sha256:374f2a777e26578f9cdc0cbb12159c233bff4fbc5d85a368d1c6ff9ce6e845ca ... 100%
              output | [██████████] transferring docker.io/library/test-dockerfile2:latest ... 100%
================================ SUCCESS [main] ================================
[...]

PR #1199, opened 2021-09-03
fix FROM DOCKERFILE for remote targets

alexcb:
fixes bug where earthly ignored the "FROM DOCKERFILE" dockerfile path
override flag value.


alexcb:
this fixes #1114

util/stringutil/str_or_default.go
```diff
@@ -0,0 +1,9 @@
+package stringutil
+
+// StrOrDefault returns str or defaultStr when str is empty.
+func StrOrDefault(str, defaultStr string) string {
+	if str == "" {
+		return defaultStr
+	}
+	return str
+}
```
vladaionescu, line 9: Minor comment re filename: we haven't used `_` nor `-` in go file names before. Maybe rename to `strordefault.go`?

vladaionescu (approved):

Issue #1201, opened 2021-09-07
Surprising ordering of tasks when using LOCALLY


mortenlj:
This might be more of a problem with docs/syntax than an actual bug, depends a bit on how `LOCALLY` is meant to be used.

I was experimenting with ways of using earthly to replace more of my `scripts/` folder, which includes scripts to start dependencies when debugging. So I added a task to build a target, and then execute it. What surprised me is that the `RUN` under  `LOCALLY` was executed before the preceeding `BUILD`.

I've reproduced it in a minimal `Earthfile`:

```
artifact:
	FROM busybox:latest
	RUN date > /tmp/mydate && cat /tmp/mydate
	SAVE ARTIFACT /tmp/mydate AS LOCAL mydate

launch:
	BUILD +artifact
	LOCALLY
		RUN cat mydate
```

Log output when trying to "build" the `+launch` task:

```
           buildkitd | Found buildkit daemon as docker container (earthly-buildkitd)
           buildkitd | Comparing running container image ("sha256:e87ea162ed9860b45d111261bfd5839da4ae460cb645079ab109193955f53a77\n") with available image ("sha256:e87ea162ed9860b45d111261bfd5839da4ae460cb645079ab109193955f53a77\n")
           buildkitd | Settings hashes match ("38e570d02f7a2b3e\n"), no restart required
       internal-term | connecting to shellrepeater on 127.0.0.1:8373
      busybox:latest | --> Load metadata linux/amd64
      busybox:latest | Completed in 1.23175818s
             +launch *local* | --> RUN cat mydate
             +launch *local* | Completed in 24.785µs
             +launch *local* | cat: mydate: No such file or directory
             +launch *local* | WARN: (RUN cat mydate) error calling LocalhostExec: rpc error: code = Unknown desc = exit status 1
             +launch *local* | Completed in 18.594993ms
Summary of timing information
Note that the times do not include the expansion of commands like BUILD, FROM, COPY (artifact).
      busybox:latest | () 1.23175818s
             +launch | () 18.619778ms
===============================================================
Total       	1.250377958s
Total (real)	1.365108791s
Error stack trace:
github.com/moby/buildkit/util/stack.Enable
	/go/pkg/mod/github.com/earthly/buildkit@v0.7.1-0.20210728003253-199ad6a5c213/util/stack/stack.go:77
github.com/moby/buildkit/util/grpcerrors.FromGRPC
[...]

PR #1204, opened 2021-09-10
Clarifications for `LOCALLY` and for `SAVE ARTIFACT` of dirs

vladaionescu:
Fixes https://github.com/earthly/earthly/issues/1201


adamgordonbell (approved):
Nice Addition. I can see how this behaviour could be confusing at first, so it is great to call it out with examples.

Issue #1202, opened 2021-09-08
Panic on launch if Earthly doesn't have permission to read .env 


indygriffiths:
OS: Ubuntu 18.04

The `FileExists` and `DirExists` functions within fileutil.go don't handle errors other than when a file cannot be found, meaning that you get a nice-looking panic if you try running Earthly if it doesn't have access to read `.env` in the current working directory.

```
$ stat .env
stat: cannot stat '.env': Permission denied

$ earthly
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0xcf8358]

goroutine 1 [running]:
github.com/earthly/earthly/util/fileutil.FileExists(0x11f535a, 0x4, 0xc00066c960)
	/earthly/util/fileutil/fileutil.go:17 +0x98
main.main()
	/earthly/cmd/earthly/main.go:208 +0x1f0
```

Depending on semantics these functions could return false if `err != nil` or the functions could return an error themselves and let the caller decide what to do. For my specific instance around `.env` an error could be ignored since it doesn't bother to pull in config if the file doesn't exist (or can't be read), but there could be other invocations where an error should be better-handled.

indygriffiths:
This also comes up when attempting to find the Earthly config directory in `GetEarthlyDir`, which gets called when attempting to detect the presence of `config.yaml` vs `config.yml`. 

The function first looks for the `.earthly` folder from the `SUDO_USER` home directory if that envvar exists. In our case where Earthly is only be able to run from a specific user with no permissions to look at my home folder, using `sudo -u <earthly user> earthly` results in the same sort of panic due to a permission error.

vladaionescu:
Makes sense - seems like something we should fix

dbermond:
The tests fails to pass for me in earthly 0.5.23 when building the Arch Linux package on a chroot (which is the default way to build packages on Arch). In 0.5.22, the tests passes fine.

I think this is related to the current issue discussed here. Please correct me if I'm wrong.

Output:

```
--- FAIL: TestBuildArgMatrix (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x55d63a3df21c]

goroutine 10 [running]:
[...]

PR #1207, opened 2021-09-13
Perform error handling when checking if file or dir exists

alexcb:
Error handling needs to be performed to catch permission-related errors
from obscuiring if a file exists or not.

### DIFF
--- a/cmd/earthly/main.go
+++ b/cmd/earthly/main.go
@@ -1057,9 +1057,11 @@ func (app *earthlyApp) before(context *cli.Context) error {
 	var yamlData []byte
 	var err error
 	if app.configPath != "" {
-		yamlData, err = config.ReadConfigFile(app.configPath, context.IsSet("config"))
+		yamlData, err = config.ReadConfigFile(app.configPath)
 		if err != nil {
-			return errors.Wrapf(err, "read config")
+			if context.IsSet("config") || !errors.Is(err, os.ErrNotExist) {
+				return errors.Wrapf(err, "read config")
+			}
 		}
 	}
 
@@ -1268,7 +1270,8 @@ func (app *earthlyApp) warnIfEarth() {
 			return
 		}
 		earthlyPath := path.Join(path.Dir(absPath), "earthly")
-		if fileutil.FileExists(earthlyPath) {
+		earthlyPathExists, _ := fileutil.FileExists(earthlyPath)
+		if earthlyPathExists {
 			app.console.Warnf("Once you are ready to switch over to earthly, you can `rm %s`", absPath)
 		}
 	}
@@ -1426,12 +1429,20 @@ func (app *earthlyApp) insertBashCompleteEntry() error {
 	}
 	dirPath := filepath.Dir(path)
 
-	if !fileutil.DirExists(dirPath) {
+	dirPathExists, err := fileutil.DirExists(dirPath)
+	if err != nil {
+		return errors.Wrapf(err, "failed checking if %s exists", dirPath)
+	}
+	if !dirPathExists {
 		fmt.Fprintf(os.Stderr, "Warning: unable to enable bash-completion: %s does not exist\n", dirPath)
 		return nil // bash-completion isn't available, silently fail.
 	}
 
-	if fileutil.FileExists(path) {
+	pathExists, err := fileutil.FileExists(path)
+	if err != nil {
+		return errors.Wrapf(err, "failed checking if %s exists", path)
+	}
+	if pathExists {
 		return nil // file already exists, don't update it.
 	}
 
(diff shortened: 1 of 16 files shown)


alexcb:
fixes #1202

cmd/earthly/main.go
```diff
@@ -205,7 +205,12 @@ func main() {
 
 	// Load .env into current global env's. This is mainly for applying Earthly settings.
 	// Separate call is made for build args and secrets.
-	if fileutil.FileExists(dotEnvPath) {
+	dotPathExists, err := fileutil.FileExists(dotEnvPath)
+	if err != nil {
+		fmt.Printf("Error checking for file %s: %s\n", dotEnvPath, err.Error())
```
vladaionescu, line 210: If we don't have permission to read `.env` should we print out a warning and continue?

cmd/earthly/main.go
```diff
@@ -2652,7 +2682,11 @@ func (app *earthlyApp) actionBuildImp(c *cli.Context, flagArgs, nonFlagArgs []st
 	}
 
 	dotEnvMap := make(map[string]string)
-	if fileutil.FileExists(dotEnvPath) {
+	dotEnvPathExists, err := fileutil.FileExists(dotEnvPath)
+	if err != nil {
+		return errors.Wrapf(err, "failed to check if %s exists", dotEnvPath)
```
vladaionescu, line 2687: Same comment about `.env` being inaccessible.

vladaionescu (commented):
Not sure if we should keep going if we cannot read `.env`. Thoughts?

Another, perhaps even more important question: Do we keep going if we cannot access `~/.earthly`? Perhaps we should - what if the user just doesn't have a home dir at all - does that show up as a permission error?

cmd/earthly/main.go
alexcb: Treating it as a warning feels like we would mask a problem; I would rather not make an assumption on the user's behalf and error out. This feels more explicit and less error-prone.

vladaionescu: Yeah, I think I agree :+1:

alexcb:
> Do we keep going if we cannot access ~/.earthly? Perhaps we should - what if the user just doesn't have a home dir at all - does that show up as a permission error?

This seems a bit trickier; In the case of a user without a home directory; I think we'll stumble on this code first:

```
// ExpandPath expands the tilde in the path
func ExpandPath(s string) string {
        if !strings.HasPrefix(s, "~") {
                return s
        }
        homeDir, _, err := HomeDir()
        if err != nil || homeDir == "" {
                return s // best effort
        }
```

perhaps in that case we should check for `/etc/earthly/`?

vladaionescu:
> I think we'll stumble on this code first

That's from auto-completion, right?

I think there are other areas which we may encounter first in other situations. See e.g. this comment about Earthly trying to look for the home dir of the `SUDO_USER`: https://github.com/earthly/earthly/issues/1202#issuecomment-915559057

alexcb:
> That's from auto-completion, right?

Correct; I also thought it was used elsewhere, but I'm wrong.

Digging deeper into that comment, I can replicate that issue via this example:

```
alex@mah:~/gh/earthly/earthly$ chmod 700 /tmp/asdf/
alex@mah:~/gh/earthly/earthly$ cd /tmp/asdf/
alex@mah:/tmp/asdf$ sudo -u xena earthly
[sudo] password for alex: 
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0xcf8358]

goroutine 1 [running]:
github.com/earthly/earthly/util/fileutil.FileExists(0x11f535a, 0x4, 0xc000650a20)
	/earthly/util/fileutil/fileutil.go:17 +0x98
main.main()
	/earthly/cmd/earthly/main.go:208 +0x1f0
```

in this case earthly will panic when it is run under a directory it does not have access to.

where as with this fix, now it display an error rather than panic()ing:

```
$ sudo -u xena /home/alex/gh/earthly/earthly/build/linux/amd64/earthly 
Error checking for file .env: unable to stat .env: stat .env: permission denied
```

However, if it DOES have permission to list the dir, but not the .env.sh file, even before this change we got this error both before and after this PR:

```
alex@mah:/tmp/asdf$ chmod 755 /tmp/asdf
alex@mah:/tmp/asdf$ echo foo > /tmp/asdf/.env
alex@mah:/tmp/asdf$ chmod 600 /tmp/asdf/.env 
alex@mah:/tmp/asdf$ sudo -u xena earthly
Error loading dot-env file .env: open .env: permission denied
```

vladaionescu:
How about the Earthly config? What's the behavior with that?

alexcb:
on a side note, when running an earthly build in a non-writable directory, one might have to set `DOCKER_CONFIG=/tmp/.docker` will change the config directory which affects the `getSeeds()` function that calls a mkdir on the docker config directory. It might be possible to detect if this is set to a non-writable location and override it to a temp dir.

alexcb:
> How about the Earthly config? What's the behavior with that?
> 
> (Note that the comment I pointed to is about the Earthly config, not about `.env`)

When we read the config we'll ignore os.ErrNotExists errors:

```
		yamlData, err = config.ReadConfigFile(app.configPath)
		if err != nil {
			if context.IsSet("config") || !errors.Is(err, os.ErrNotExist) {
				return errors.Wrapf(err, "read config")
			}
		}
```

vladaionescu (approved):
A bit worried that some people might have relied on the "false on error" behavior (e.g. a dir exists, but is not accessible by the current user) and this could break something.

It's the right thing to do though, and I can't think of a better way.

alexcb:
> A bit worried that some people might have relied on the "false on error" behavior

I wouldn't be surprised if we see a similar error message from circle CI next, but that'll allow us to follow the path to see what's failing `os.stat`.

PR #1212, opened 2021-09-15
move DetectHomeDir to fileutil

alexcb:
Both cliutil and fileutil contain similar code to detect the home
directory; this combines the two functions into a single function.
This will ensure we consistently use the same home dir detection.

### DIFF
--- a/buildcontext/gitlookup.go
+++ b/buildcontext/gitlookup.go
@@ -18,7 +18,6 @@ import (
 	"golang.org/x/crypto/ssh/knownhosts"
 
 	"github.com/earthly/earthly/conslogging"
-	"github.com/earthly/earthly/util/cliutil"
 	"github.com/earthly/earthly/util/fileutil"
 
 	"github.com/jdxcode/netrc"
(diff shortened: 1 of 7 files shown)


alexcb:
Currently this is failing on
```
=== Test 6: works in read-only directory ===
           buildkitd | Found buildkit daemon as *** container (earthly-buildkitd)
      busybox:latest | --> Load metadata linux/amd64
               +test | --> FROM busybox:latest
               +test | [          ] resolve ***.io/library/busybox:latest@sha256:52f73a0a43a16cf37cd0720c90887ce972fe60ee06a687ee71fb93a7ca601df7 ... 0%
               +test | [██████████] resolve ***.io/library/busybox:latest@sha256:52f73a0a43a16cf37cd0720c90887ce972fe60ee06a687ee71fb93a7ca601df7 ... 100%
               +test | *cached* --> RUN echo "hello earth, this is major tom."
=================================== SUCCESS ====================================
Error: build target: unable to create dir .tmp-earthly-out: mkdir .tmp-earthly-out: permission denied
```
which requires the fix that is in #1214

alexcb:
This PR additionally will only return the SUDO_USER's home directory when the current user has access to it.

vladaionescu (approved):

PR #1214, opened 2021-09-15
Only create outdir when needed

alexcb:
The temp output directory should only be created when needed; without
this PR, running a target that doesn't output artifacts (or running
under --no-output), will still cause the temp directory to be created,
which will fail if earthly is being run on a read-only directory.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -392,21 +392,17 @@ func (b *Builder) convertAndBuild(ctx context.Context, target domain.Target, opt
 		})
 		return pipeW, nil
 	}
-	var outDir string
 	onArtifact := func(childCtx context.Context, index int, artifact domain.Artifact, artifactPath string, destPath string) (string, error) {
 		sp.printCurrentSuccess()
 		if !destPathWhitelist[destPath] {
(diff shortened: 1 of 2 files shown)


alexcb:
This is a part 1 of 2 PR; part 2 is in #1212 which implements a follow up integration test which covers this test case and ensures that the directory is writable.

builder/builder.go
```diff
@@ -443,13 +432,6 @@ func (b *Builder) convertAndBuild(ctx context.Context, target domain.Target, opt
 		return nil, errors.Wrapf(err, "build main")
 	}
 	sp.printCurrentSuccess()
-	if outDir == "" {
-		var err error
-		outDir, err = b.tempEarthlyOutDir()
-		if err != nil {
-			return nil, err
-		}
-	}
```
vladaionescu, line 452: The removal of this section is the key part of the PR, right?

vladaionescu (approved):
I wonder why we were doing all the `if outDir == ""` checks. They don't seem necessary...

builder/builder.go
alexcb: This is the section that was giving the most grief, as it was called even when we were running with `opt.NoOutput`.

PR #1216, opened 2021-09-16
Go Integration Test Example

### LINKED ISSUE #352: Go Integration Test Example
We have a simple go example being built by earthly here:
https://github.com/earthly/earthly/tree/master/examples/go

It would be nice to have a second, more detailed go example, that included unit tests and integration tests and build targets for each of these.  It would be more similar to this example https://github.com/earthly/earthly/blob/master/examples/integration-test/Earthfile but in go.

JosemyDuarte:
Should close #352


examples/go/main.go
```diff
@@ -1,6 +1,10 @@
 package main
 
-import "github.com/sirupsen/logrus"
+import (
+	"github.com/sirupsen/logrus"
+)
+
+var howCoolIsEarthly = "IceCool"
```
vladaionescu, line 7: 🧊

vladaionescu (approved):
Very nice!

PR #1218, opened 2021-09-16
pin ubuntu to 20.10

alexcb:
rather than using ubuntu:latest for our dind image, pin to a specific
version to speed up release times so we don't have to rebuild it as
frequently.

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -331,7 +331,7 @@ dind-alpine:
     SAVE IMAGE --push --cache-from=earthly/dind:main earthly/dind:$DIND_ALPINE_TAG
 
 dind-ubuntu:
-    FROM ubuntu:latest
+    FROM ubuntu:20.10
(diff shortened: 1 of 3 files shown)


Earthfile
```diff
@@ -331,7 +331,7 @@ dind-alpine:
     SAVE IMAGE --push --cache-from=earthly/dind:main earthly/dind:$DIND_ALPINE_TAG
 
 dind-ubuntu:
-    FROM ubuntu:latest
+    FROM ubuntu:20.10
```
vladaionescu, line 334: I think `latest` simply points to `20.10`... unless there's another release. It doesn't incrementally update between releases AFAIK

vladaionescu (approved):

Earthfile
alexcb: ah it points to `20.04`, which gets updated at the same cadence; I was hoping it was somehow less frequently. It's still probably good to pin to a specific version, but it might not give us as much of a speedup as I was hoping for.

PR #1219, opened 2021-09-21
Fix `IF` after `WITH DOCKER`

dchw:
When an `IF` occurred after a `WITH DOCKER`, the exported images were not avaliable to Buildkit because the target had not been finalized yet. This would result in faliures like:

```
Error: build target: build main: bkClient.Build: failed to solve: NotFound: no access allowed to dir "a:latest-5c5691246807332bf2a2320bc0065ede3b51617b9fbce723c804d3634dcee50e"

```

This fixes them by adding the directories immediately when exporting the tar, rather than piggybacking on the local dirs.


buildcontext/provider/provider.go
```diff
@@ -67,6 +67,12 @@ func NewBuildContextProvider(console conslogging.ConsoleLogger) *BuildContextPro
 
 // AddDirs adds local directories to the context.
 func (bcp *BuildContextProvider) AddDirs(dirs map[string]string) {
+	for dirName, dir := range dirs {
+		bcp.AddDir(dirName, dir)
+	}
```
dchw, line 72: The real main difference here is that we don't hold the lock the whole time, rather we acquire/release it for each dir. Doesn't appear to be a problem in my cursory testing.

examples/tests/with-docker/Earthfile
```diff
@@ -70,3 +71,13 @@ one-target-many-names:
...
+
+if-after:
+    FROM earthly/dind:alpine
+    WITH DOCKER --load a:latest=+multi-from-one
+        RUN docker run a:latest
+    END
+
+    IF [ "true" = "true" ]
+        RUN true
+    END
```
dchw, line 83: This reliably reproduces it locally on older Earthly.

alexcb (approved):

examples/tests/with-docker/Earthfile
alexcb: nice repro-case!

earthfile2llb/withdockerrun.go
```diff
@@ -335,7 +335,8 @@ func (wdr *withDockerRun) solveImage(ctx context.Context, mts *states.MultiTarge
 			llb.Platform(llbutil.DefaultPlatform()),
 			llb.WithCustomNamef("[internal] docker tar context %s %s", opName, sessionID),
 		)
-		wdr.c.mts.Final.LocalDirs[string(solveID)] = outDir
+		// Add directly to build context so that if a later statement forces execution, the images are avaliable.
+		wdr.c.opt.BuildContextProvider.AddDir(string(solveID), outDir)
```
vladaionescu, line 339: Note that this isn't the only place where we add things to `LocalDirs` in the converter. `FromDockerfile` also does that and would be prone to the same kind of issue. We should try to remove `SingleState.LocalDirs` altogether and instead use `BuildContextProvider.AddDir` directly everywhere.

earthfile2llb/withdockerrun.go
```diff
@@ -335,7 +335,8 @@ func (wdr *withDockerRun) solveImage(ctx context.Context, mts *states.MultiTarge
 			llb.Platform(llbutil.DefaultPlatform()),
 			llb.WithCustomNamef("[internal] docker tar context %s %s", opName, sessionID),
 		)
-		wdr.c.mts.Final.LocalDirs[string(solveID)] = outDir
```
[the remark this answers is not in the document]
dchw: I did. It looked like the source directory is also part of the localdirs ecosystem... so I didnt chase it much further.

earthfile2llb/withdockerrun.go
dchw: FromDockerfile never adds things, it only copies what exists. Because its not adding anything new, it shouldnt be affected?

vladaionescu (approved):

zedtux:
Awesome ! When even you need me to test something, I'm available!

vladaionescu (approved):

PR #1224, opened 2021-09-24
Add Podman Support, Automatic Configuration

dchw:
Adds podman as a compatible frontend for Earthly to use when exporting images, and running buildkit. Podman works in rootless and rootful modes, daemonless or not. As far as I have tested.

Also, autodetects some compatibility issues and ensures they are properly dealt with rather than relying on user configuration.

---

Its a draft for now, a few more tweaks need to be made and GoDoc comments to pass the linter. Its necessarily a large PR (I mean, creating the seam for pluggable container frontends was just touching a _lot_ of code), so any head start is probably worth it in reviewing.

### DIFF
--- a/builder/docker.go
+++ b/builder/docker.go
@@ -4,13 +4,12 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"os"
-	"os/exec"
 	"strings"
 
 	"github.com/containerd/containerd/platforms"
 	"github.com/docker/distribution/reference"
 	"github.com/earthly/earthly/conslogging"
+	"github.com/earthly/earthly/util/containerutil"
 	"github.com/earthly/earthly/util/llbutil"
 	"golang.org/x/sync/errgroup"
 
@@ -40,7 +39,7 @@ func platformSpecificImageName(imgName string, platform specs.Platform) (string,
 	return reference.FamiliarString(r2), nil
 }
 
-func loadDockerManifest(ctx context.Context, console conslogging.ConsoleLogger, parentImageName string, children []manifest) error {
+func loadDockerManifest(ctx context.Context, console conslogging.ConsoleLogger, fe containerutil.ContainerFrontend, parentImageName string, children []manifest) error {
 	console = console.WithPrefix(parentImageName)
 	if len(children) == 0 {
 		return errors.Errorf("no images in manifest list for %s", parentImageName)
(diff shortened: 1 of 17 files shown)


builder/docker.go
```diff
@@ -69,58 +68,56 @@ func loadDockerManifest(ctx context.Context, console conslogging.ConsoleLogger,
...
-func loadDockerTar(ctx context.Context, r io.ReadCloser, console conslogging.ConsoleLogger) error {
-	cmd := exec.CommandContext(ctx, "docker", "load")
-	cmd.Stdin = r
-	output, err := cmd.CombinedOutput()
+func loadDockerTar(ctx context.Context, fe containerutil.ContainerFrontend, r io.ReadCloser, console conslogging.ConsoleLogger) error {
+	err := fe.ImageLoad(ctx, r)
 	if err != nil {
-		console.Warnf("%+v output:\n%s\n", cmd.Args, string(output))
-		return errors.Wrapf(err, "docker load")
+		console.Warnf(err.Error())
```
vladaionescu, line 84: I think this is no longer necessary. We used to have this in order to also get the output itself (otherwise the error alone would say just something along the lines of "exited with non-zero").

The wrapped error is probably reported somewhere upstream anyway.

buildkitd/buildkitd.go
```diff
@@ -37,17 +35,29 @@ var TCPAddress = "tcp://127.0.0.1:8372"
 // DockerAddress is the address at which the daemon is avaliable whe using a Docker Container directly
 var DockerAddress = "docker-container://earthly-buildkitd"
 
-// TODO: Implement all this properly with the docker client.
+// PodmanAddress is the address at which the daemon is avaliable whe using a Docker Container directly
+var PodmanAddress = "podman-container://earthly-buildkitd"
```
vladaionescu, line 39: Perhaps add a comment to say that this variable is not yet used?

buildkitd/buildkitd.go
```diff
@@ -125,28 +135,28 @@ func ResetCache(ctx context.Context, console conslogging.ConsoleLogger, image, c
...
-	isStarted, err := IsStarted(ctx, containerName)
+func MaybeStart(ctx context.Context, console conslogging.ConsoleLogger, image, containerName string, fe containerutil.ContainerFrontend, settings Settings, opts ...client.ClientOpt) (string, error) {
+	isStarted, err := IsStarted(ctx, containerName, fe)
 	if err != nil {
 		return "", errors.Wrap(err, "check is started buildkitd")
 	}
 	if isStarted {
 		console.
 			WithPrefix("buildkitd").
 			Printf("Found buildkit daemon as docker container (%s)\n", containerName)
```
vladaionescu, line 146: Could we parameterize this `Printf` to say `Found buildkit daemon as podman container` etc depending on frontend? Alternatively, we could also report the frontend completely separately, at the very start of Earthly.

buildkitd/buildkitd.go
```diff
@@ -226,184 +236,228 @@ func MaybeRestart(ctx context.Context, console conslogging.ConsoleLogger, image,
...
 
-		dbURL, err := url.Parse(settings.DebuggerAddress)
+		_, err := url.Parse(settings.DebuggerAddress)
 		if err != nil {
 			panic("Debugger address was not a URL when attempting to start buildkit")
 		}
-		args = append(args, "-p", fmt.Sprintf("127.0.0.1:%s:8373", dbURL.Port()))
+		portOpts = append(portOpts, containerutil.Port{
+			IP:            "127.0.0.1",
+			HostPort:      8373, //strconv.FormatInt(dbURL.Port(), 10),
```
vladaionescu, line 324: We should use the `dbURL.Port()` here?

util/containerutil/types.go
```diff
@@ -0,0 +1,113 @@
...
+
+	// We would like to shift to the non-shell providers. However, we do provide an option for supplying
+	// additional arguments to the CLI when starting buildkit. While this allowed great flexibility, we
+	// also do not know what or how it is being used. This gives us the option to support those users until
+	// we decide to pull the plug. This argument is ignored by non-shell providers.
+	AdditionalArgs []string
+}
+
+const (
+	FrontendAutomatic   = "automatic"
```
vladaionescu, line 108: Should we use "auto" here instead? Seems like a more common name for this.

config/config.go
```diff
@@ -37,6 +37,9 @@ const (
 
 	// DefaultServerTLSKey is the default path to use when looking for the Buildkit TLS key
 	DefaultServerTLSKey = "./certs/buildkit_key.pem"
+
+	// DefaultContainerFrontend is the default frontend program or interfacing with the running containers and saved images
+	DefaultContainerFrontend = "docker-shell"
```
vladaionescu, line 42: Would it be safe to use `"auto"` here already?

config/config.go
```diff
@@ -65,6 +68,7 @@ type GlobalConfig struct {
 	ServerTLSCert            string   `yaml:"buildkitd_tlscert"          help:"The path to the server cert for verification. Relative paths are interpreted as relative to ~/.earthly. Only used when Earthly manages buildkit."`
 	ServerTLSKey             string   `yaml:"buildkitd_tlskey"           help:"The path to the server key for verification. Relative paths are interpreted as relative to ~/.earthly. Only used when Earthly manages buildkit."`
 	TLSEnabled               bool     `yaml:"tls_enabled"                help:"If TLS should be used to communicate with Buildkit. Only honored when BuildkitScheme is 'tcp'."`
+	ContainerFrontend        string   `yaml:"container_frontend"         help:"What program should be used to start and stop buildkitd, save images. Default is 'docker'. Valid options are 'docker' and 'podman' (experimental)."`
```
vladaionescu, line 71:
```suggestion
	ContainerFrontend        string   `yaml:"container_frontend"         help:"What program should be used to start and stop buildkitd, save images. Default is 'docker-shell'. Valid options are 'docker-shell' and 'podman-shell' (experimental)."`
```

util/containerutil/types.go
dchw: I just made the string match the name

config/config.go
dchw: Probably, was wanting to minimize the number of new behaviors shipped. If youre comfortable, I am

vladaionescu: Yeah.. it's probably safest to be turned off completely by default given that we'll start in experimental status.

vladaionescu (approved):
This looks good to me.

PR #1227, opened 2021-09-28
Merge github, dockerhub, and homebrew release into one command

alexcb:
This PR introduces a new release target `+all`; which will
build earthly and release it to github and our homebrew repo.

In a follow up PR we will also chain up the apt and yum releases into a
single task.

This PR additionally requires this new task be merged before we perform a new deploy: https://github.com/earthly/homebrew-earthly/pull/12 ; that PR implements the new homebrew automatic build and merge on `release-v...` branches being pushed, without having to create a PR.

Note: This PR has a very ugly hack: the `+all` target is a `LOCALLY` target which calls earthly on the individual targets. This was done as we had to run `+release-github` along with the push, before `+release-homebrew` could begin (otherwise `+release-homebrew` would try to download files which had yet to be uploaded to github).

### DIFF
--- a/release/Earthfile
+++ b/release/Earthfile
@@ -3,6 +3,57 @@ FROM alpine:3.13
 RUN apk add --update --no-cache \
     curl
 
+all:
+    LOCALLY
+    ARG RELEASE_TAG
+    ARG GITHUB_USER=earthly
+    ARG DOCKERHUB_USER=earthly
+    ARG EARTHLY_REPO=earthly
+    ARG BREW_REPO=homebrew-earthly
+    RUN test -n "$RELEASE_TAG" || (echo "ERROR: RELEASE_TAG is not set" && exit 1)
+
+    # validate args: if doing a prod-release make sure no args are changed
+    # otherwise if a test release is detected, make sure all users are changed
+    # (this is to prevent a case where only the github user is changed, but dockerhub is not, and the dockerhub binary is accidentally overwritten)
+    RUN set -e; \
+        test -n "$RELEASE_TAG" || (echo "ERROR: RELEASE_TAG is not set" && exit 1); \
+        (echo "$RELEASE_TAG" | grep '^v[0-9]\+.[0-9]\+.[0-9]\+$' > /dev/null) || (echo "ERROR: RELEASE_TAG must be formatted as v1.2.3; instead got \"$RELEASE_TAG\""; exit 1); \
+        if [ "$GITHUB_USER" = "earthly" ] && [ "$DOCKERHUB_USER" = "earthly" ]; then \
+          test "$BREW_REPO" = "homebrew-earthly" || (echo "ERROR: prod release detected; expecting BREW_REPO=earthly, but got $BREW_REPO" && exit 1); \
+          test "$EARTHLY_REPO" = "earthly" || (echo "ERROR: prod release detected; expecting EARTHLY_REPO=earthly, but got $EARTHLY_REPO" && exit 1); \
+          test "$EARTHLY_GIT_BRANCH" = "main" || (echo "ERROR: performing release on non-main branch" && exit 1); \
+        else \
+          test "$GITHUB_USER" != "earthly" || (echo "ERROR: non-prod release detected but got GITHUB_USER=$GITHUB_USER" && exit 1); \
+          test "$DOCKERHUB_USER" != "earthly" || (echo "ERROR: non-prod release detected but got DOCKERHUB_USER=$DOCKERHUB_USER" && exit 1); \
+        fi; \
+        echo "performing release to repo=$GITHUB_USER/$EARTHLY_REPO; homebrew=$GITHUB_USER/$BREW_REPO; dockerhub=$DOCKERHUB_USER"
+
+    # terrible unsafe hack to determine --push flag
+    IF ps auxw | grep -v grep | grep earthly | grep -e --push
+        ARG PUSH="--push"
+    ELSE
+        ARG PUSH=""
+    END
+
+    RUN set -e; \
+        earthly_pid=$(ps axw | grep -v grep | grep earthly | grep 'all' | awk '{print $1}'); \
+        expected_hash=$(cat ~/.earthly/earthly-prerelease | md5sum); \
+        actual_hash=$(cat /proc/$earthly_pid/exe | md5sum); \
+        if [ "$expected_hash" != "$actual_hash" ]; then \
+          echo "\nERROR: +all requires the ./earthly script be used (rather than directly using any earthly binary).\n"; \
+          exit 1; \
+        fi;
+
+    RUN ../earthly $PUSH --build-arg DOCKERHUB_USER="$DOCKERHUB_USER" --build-arg RELEASE_TAG="$RELEASE_TAG" +release-dockerhub
+    RUN ../earthly $PUSH --build-arg GITHUB_USER="$GITHUB_USER" --build-arg EARTHLY_REPO="$EARTHLY_REPO" --build-arg BREW_REPO="$BREW_REPO" --build-arg DOCKERHUB_USER="$DOCKERHUB_USER" --build-arg RELEASE_TAG="$RELEASE_TAG" +release-github
+    RUN ../earthly $PUSH --build-arg GITHUB_USER="$GITHUB_USER" --build-arg EARTHLY_REPO="$EARTHLY_REPO" --build-arg BREW_REPO="$BREW_REPO" --build-arg DOCKERHUB_USER="$DOCKERHUB_USER" --build-arg RELEASE_TAG="$RELEASE_TAG" +release-homebrew
+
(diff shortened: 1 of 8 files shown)


release/Earthfile
```diff
@@ -99,102 +186,55 @@ release-homebrew:
...
     RUN sed -i \
-        -e 's^url ".*"^url "'"$NEW_URL"'"^' \
-        -e 's^sha256 ".*"$^sha256 "'$(cat /params/downloadsha256)'"^' \
-        -e 's^tags = ".*"^tags = "'"$(cat /params/tags)"'"^' \
-        -e 's^ldflags = ".*"^ldflags = "'"$(cat /params/ldflags1)"'"^' \
-        -e '/ldflags = ".*"/{n;s^".*"^"'"$(cat /params/ldflags2)"'"^}' \
+        -e 's^\burl ".*"^url "'"$NEW_URL"'"^' \
+        -e 's^\bsha256 ".*"$^sha256 "'$(cat /params/downloadsha256)'"^' \
+        -e 's^\btags = ".*"^tags = "'"$(cat /params/tags)"'"^' \
+        -e 's^\bldflags = ".*"^ldflags = "'"$(cat /params/ldflags)"'"^' \
```
alexcb, line 230: we no longer have to worry about the homebrew checks telling us this line is too long :tada:

release/Earthfile
```diff
@@ -99,102 +186,55 @@ release-homebrew:
...
-            -m '* `NEW_URL='"$NEW_URL"'`' \
-            -m '* `NEW_SHA256='"$(cat /params/downloadsha256)"'`' \
-            -m '* `TAGS='"$(cat /params/tags)"'`' \
-            -m '* `LDFLAGS='"$(cat /params/ldflags)"'`' \
-            -m '* `LDFLAGS1='"$(cat /params/ldflags1)"'`' \
-            -m '* `LDFLAGS2='"$(cat /params/ldflags2)"'`') && \
-        echo $output && \
-        curl -X POST -H 'Content-type: application/json' --data '{"text":"Successfully released `'$RELEASE_TAG'`: https://github.com/earthly/earthly/releases/tag/'$RELEASE_TAG'"}' $SLACK_WEBHOOK_URL && \
-        curl -X POST -H 'Content-type: application/json' --data '{"text":"Successfully released homebrew PR for `'$RELEASE_TAG'`: '"$output"'"}' $SLACK_WEBHOOK_URL
+        git commit -a -m "earthly $version" || true
```
alexcb, line 235: `|| true` required to allow this code to be re-entrant in the case of the version already being released.

release/Earthfile
```diff
@@ -3,6 +3,57 @@ FROM alpine:3.13
...
+          test "$DOCKERHUB_USER" != "earthly" || (echo "ERROR: non-prod release detected but got DOCKERHUB_USER=$DOCKERHUB_USER" && exit 1); \
+        fi; \
+        echo "performing release to repo=$GITHUB_USER/$EARTHLY_REPO; homebrew=$GITHUB_USER/$BREW_REPO; dockerhub=$DOCKERHUB_USER"
+
+    # terrible unsafe hack to determine --push flag
+    IF ps auxw | grep -v grep | grep earthly | grep -e --push
+        ARG PUSH="--push"
+    ELSE
+        ARG PUSH=""
+    END
```
vladaionescu, line 36: This doesn't look great :D

Should we add a builtin arg `EARTHLY_PUSH` which would be set to either `true` or `false` accordingly?

release/Earthfile
```diff
@@ -63,18 +124,43 @@ release-github:
...
+        # next make sure any previous tag is deleted (otherwise github will create an untagged release name vX.Y.Z)
+        curl -H "Authorization: token $GITHUB_TOKEN" --request DELETE "https://api.github.com/repos/$GITHUB_USER/$EARTHLY_REPO/git/refs/tags/$RELEASE_TAG"; \
+        # next, upload binaries
         github-release upload \
-        --owner earthly \
-        --repo earthly \
-        --prerelease true \
+        --owner "$GITHUB_USER" \
+        --repo "$EARTHLY_REPO" \
+        --prerelease false \
```
vladaionescu, line 146: Is this right? We should first write the release notes and only afterward uncheck the prerelease flag.

release/Earthfile
```diff
@@ -99,102 +186,55 @@ release-homebrew:
...
 
+    RUN --secret GIT_PASSWORD="$GITHUB_TOKEN_SECRET_PATH" --no-cache \
+        git clone "https://$GITHUB_USER@github.com/$GITHUB_USER/$BREW_REPO.git" .
+
+    ARG RELEASE_BRANCH="release-$RELEASE_TAG"
+    RUN git switch -c "$RELEASE_BRANCH"
+
     # Make the change in a new branch.
-    RUN git checkout -b "release-$RELEASE_TAG"
+    #RUN git checkout -b "release-$RELEASE_TAG"
```
vladaionescu, line 216: Clean this up?

release/Earthfile
vladaionescu: This is probably better

```suggestion
        git commit -a --allow-empty -m "earthly $version"
```

vladaionescu (commented):
It's not clear to me what the release process would look like after this? Can you clarify? For example, when do release notes get written.

release/Earthfile
alexcb: good catch.

release/Earthfile
alexcb: If we leave this as true, then we must manually go in and uncheck the pre-release flag between the github release and the homebrew release.

I was thinking that we could write the release notes in parallel, and once the release shows up in github we can edit it there. The downside with this is it will introduce a race-condition of sorts where users could download a version of earthly that does not contain release notes.

Another idea is we could start committing release notes into a changelog in git, and this release script could pull out the appropriate details for the release we are doing. If the changelog notes don't exist, then the release would exit out. Maintaining a changelog would give users a single file to look over in case they wanted to know what feature introduced a change that may have created (or fixed) a bug.

release/Earthfile
alexcb: it's downright terrible :grimacing: 

That's a very interesting idea regarding that flag. I could see us abusing it like:

```
IF $EARTHLY_PUSH
    RUN --no-cache do-a-push
END
RUN echo a non-push command comes after.
```

That alone might allow us to get rid of the `../earthly` calling `../earthly` hack too.

Would allowing us (and our users) to adopt such a pattern cause any issues?

I don't think so, as a user could currently introduce their own `--build-arg MY_PUSH_FLAG=true` to implement something very similar.

alexcb: this LOCALLY target has been moved into a seperate `./release.sh` script which will call `earthly` several times. We will tackle supporting this via a single earthly command in the future.

release/Earthfile
alexcb: we now have a `CHANGELOG.md` and corresponding `changelogparser.py` which passes this as the body for the release text.

release/changelogparser.py
```diff
@@ -0,0 +1,122 @@
...
+
+class DuplicateVersionError(KeyError):
+    pass
+
+
+def parse_line(line):
+    '''
+    parses lines of the form "# <title>", "## <sub title>", etc.
+    if line is not a header, a regular string is returned.
+    headers must contain exactly once space between the '#' and title, and may not contain trailling space.
```
vladaionescu, line 27:
```suggestion
    headers must contain exactly one space between the '#' and title, and may not contain trailing spaces.
```

release/release.sh
```diff
@@ -0,0 +1,52 @@
...
+
+../earthly --push --build-arg DOCKERHUB_USER --build-arg RELEASE_TAG +release-dockerhub
+../earthly --push --build-arg GITHUB_USER --build-arg EARTHLY_REPO --build-arg BREW_REPO --build-arg DOCKERHUB_USER --build-arg RELEASE_TAG $GITHUB_SECRET_PATH_BUILD_ARG +release-github
+../earthly --push --build-arg GITHUB_USER --build-arg EARTHLY_REPO --build-arg BREW_REPO --build-arg DOCKERHUB_USER --build-arg RELEASE_TAG $GITHUB_SECRET_PATH_BUILD_ARG +release-homebrew
+
+# TODO pass along a RELEASE_REPO_TEST_SUFFIX which would cause us to host our yum/apt repos under https://test-pkg.earthly.dev/$RELEASE_REPO_TEST_SUFFIX/...
+# and when it is empty, we would use https://pkg.earthly.dev/...
+#../earthly --push --build-arg GITHUB_USER --build-arg EARTHLY_REPO --build-arg BREW_REPO --build-arg DOCKERHUB_USER --build-arg RELEASE_TAG +release-repo
+# until then, we will just print this out:
+echo "TODO: the apt/yum release must be triggered seperately; once we get https://test-pkg.earthly.dev/ setup"
```
vladaionescu, line 52:
```suggestion
echo "TODO: the apt/yum release must be triggered seperately; until we get https://test-pkg.earthly.dev/ setup"
```

vladaionescu (approved):
This looks good as far as I can tell. Perhaps we can do the release together when you're ready, as a way for me to review this further.

PR #1229, opened 2021-09-28
new integrations and ref diagrams, tutorial text revisions

mirho:
Docs tutorial text revisions, and new integrations and ref diagrams.

Doing a draft PR so I can review the changes, and refine and tweak where necessary.


docs/basics/final-words.md
```diff
@@ -1,32 +1,35 @@
 # Final words
 
-Congratulations, you made it this far!
+Congratulations, you completed the Earthly tutorial! 
 
-To learn more about Earthly, take a look at the [examples directory on GitHub](https://github.com/earthly/earthly/tree/main/examples), where you will find the complete code used in this guide:
+Learning Earthly does not stop here. Discover more of what Earthly can do by exploring the documentation.
 
+**Recommended reading:**
```
vladaionescu, line 7: Nit: might be better to have these as a subtitles?

```suggestion
##### Recommended reading:
```

docs/basics/part-1-a-simple-earthfile.md
```diff
@@ -191,66 +193,25 @@ print("hello world")
...
-earthly +docker
-```
-
-The output might look like this:
-
-![Earthly build output](../guides/img/go-example.png)
+From the example above, you may notice that an Earthfile is very similar to a Dockerfile. This is an intentional design decision. Existing Dockerfiles can easily be ported to Earthly by copying them to an Earthfile and tweaking them slightly.
 
-Notice how to the left of `|`, within the output, we can see some targets like `+base`, `+build` and `+docker` . Notice how the output is interleaved between `+docker` and `+build`. This is because the system executes independent build steps in parallel. The reason this is possible effortlessly is because only very few things are shared between the builds of the recipes and those things are declared and obvious. The rest is completely isolated.
+<!-- Compared to Dockerfile syntax, there are some additional commands specific to Earthly (like `SAVE ARTIFACT`), some commands have additional semantics (like `COPY +target/some-artifact`) and other semantics have been removed (like `FROM ... AS ...` and `COPY --from`). -->
```
vladaionescu, line 213: We could remove this entirely perhaps?

vladaionescu (approved):
This looks great!

For `integration-diagram-v2.png`, just a nit:

`CI Failures difficult to reproduce` -> `CI failures difficult to reproduce` (lowercase `f`)

BTW, you may need to configure your local `git` client to use an email that GitHub recognizes, for the CLA (you can use `mirho@users.noreply.github.com` and GH will recognize that properly).

docs/basics/part-1-a-simple-earthfile.md
mirho: I like that, simply it more

docs/basics/final-words.md
mirho: That makes sense will make the change.

vladaionescu (approved):

PR #1235, opened 2021-10-01
Add C example (#357)

sea-grass:
Adds a C example with an Earthfile that can build a Docker image and run unit tests.


alexcb:
@sea-grass Thanks for the submission! Looks great.

Issue #904, opened 2021-04-01
Required ARGs


vladaionescu:
Some UDCs (or even targets) could benefit from strictly required ARGs - ARGs that must be passed in, in order for the command (or target) to work.

This helps eliminate cases where the user is chasing their tail trying to figure out what's wrong when an ARG is unexpectedly set to `""` when declared as `ARG something` and no value is provided on the call itself.

Sytanx suggestion:

```
ARG --required something
```

Required ARGs can never have a default value.

CC @jazzdan

alexcb:
This might be a good first issue if anyone wants to take it on. We would need a new struct for ARG options under https://github.com/earthly/earthly/blob/main/earthfile2llb/flags.go

then, under `func (i *Interpreter) handleArg(ctx context.Context, cmd spec.Command) error {`, one would have to reference the struct to parse the opts. then if the required flag is set, check the value is not empty.

alexcb:
this has been released and is available under `v0.6.0`.

PR #1244, opened 2021-10-04
Implement Required ARGs #904

camerondurham:
Add optional `--required` flag for Earthfile `ARG`s. Based on implementation description from https://github.com/earthly/earthly/issues/904

Unable to run integration tests using `./build/linux/amd64/earthly -P +test-all` after building with `./build/linux/amd64/earthly +for-linux` due to [dockerhub auth errors](https://hastebin.com/idonahepiy.php). Is there Earthly specific setup I need to do to run these integ tests locally?

I would be more than happy to help update [ARG reference](https://docs.earthly.dev/docs/earthfile#arg) if this implementation is what the maintainers are looking for. Should I create or add to existing unit/integ tests in this repo?

For contacting me directly, my Slack member ID is `U02GHP4SYN7`. Thanks!

### DIFF
--- a/earthfile2llb/flags.go
+++ b/earthfile2llb/flags.go
@@ -110,3 +110,7 @@ type doOpts struct {
 type importOpts struct {
 	AllowPrivileged bool `long:"allow-privileged" description:"Allow targets to assume privileged mode"`
 }
+
+type argOpts struct {
+	Required bool `long:"required" description:"Require argument to be non-empty"`
+}
--- a/earthfile2llb/interpreter.go
+++ b/earthfile2llb/interpreter.go
@@ -1028,22 +1028,30 @@ func (i *Interpreter) handleArg(ctx context.Context, cmd spec.Command) error {
 	if i.pushOnlyAllowed {
 		return i.pushOnlyErr(cmd.SourceLocation)
 	}
+	opts := argOpts{}
+	args, err := flagutil.ParseArgs("ARG", &opts, getArgsCopy(cmd))
+	if err != nil {
+		return i.wrapError(err, cmd.SourceLocation, "invalid ARG arguments %v", cmd.Args)
+	}
 	var key, value string
-	switch len(cmd.Args) {
+	switch len(args) {
 	case 3:
-		if cmd.Args[1] != "=" {
+		if args[1] != "=" {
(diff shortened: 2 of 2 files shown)


vladaionescu:
Hi @camerondurham - you may need to edit the main Earthfile to edit some of this logic here https://github.com/earthly/earthly/blob/main/Earthfile#L322-L329

(Remove `registry-1.docker.io.mirror.corp.earthly.dev` and also point credentials to credentials to login to your DockerHub personally). Sorry this is a bit of a pain currently.

vladaionescu:
(BTW this is looking pretty good so far - I've assigned @alexcb to help you out further)

camerondurham:
Hello and thanks for the help @vladaionescu! Apologies for sending the partial CR and leaving it untouched for a week.

I have added the implementation and manually tested with the following Earthfile:

```
FROM alpine:3.13
test:
    ARG validEmpty
    ARG validDefault=val
    ARG --required req
    ARG --required shouldNotHaveDefaultValue=default
    RUN echo "$SOMEARG" >file.txt
    SAVE ARTIFACT file.txt AS LOCAL "$SOMEARG"
```
Here is the public recording of testing with this file: https://asciinema.org/a/441110

Remaining tasks, judging from existing docs/tests:
- [ ] update `docs/earthfile/earthfile.md` section with new `--required` flag in `ARG` section
- [ ] add feature test in `examples/tests/required-arg.earth` (?)
- [ ] address any code-quality or implementation comments

While the suggestions for adjusting docker login procedure helped, I'm sadly still getting `401` errors when trying to run the integ test stages. I'll keep trying and maybe commenting out tests I don't think need to be run to get this to work but have hit a wall.

Regarding feature tests, are there existing tests I could use to automate the manual test I did (i.e. verify errors are thrown when build-arg not provided)?

Apologies if this is already documented somewhere, I will keep looking in the repository to figure things out. Thanks for your help!

vladaionescu:
> Regarding feature tests, are there existing tests I could use to automate the manual test I did (i.e. verify errors are thrown when build-arg not provided)?

Check out this test: https://github.com/earthly/earthly/blob/main/examples/tests/Earthfile#L345

You could run it twice: once with the build arg provided and once with it not provided and make sure that it fails the second time.

PR #1245, opened 2021-10-04
Update and change emojis in  README.md

idivyanshbansal:


adamgordonbell:
Hi @idivyanshbansal ,
I don't think we would like to introduce this change. It doesn't add any value to the readme.

idivyanshbansal:
No problem @adamgordonbell . Have a great day.

PR #1250, opened 2021-10-04
Fix up new Frontend output

dchw:
Error cases looked bad. This gives better output, all the info on Verbose, adds a validation check, and adds some checks deeper in, to prevent panics if this happens again on accident.

### DIFF
--- a/cmd/earthly/main.go
+++ b/cmd/earthly/main.go
@@ -1022,7 +1022,8 @@ func (app *earthlyApp) before(context *cli.Context) error {
 
 	fe, err := containerutil.FrontendForSetting(context.Context, app.cfg.Global.ContainerFrontend)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


cmd/earthly/main.go
```diff
@@ -1022,7 +1022,8 @@ func (app *earthlyApp) before(context *cli.Context) error {
 
 	fe, err := containerutil.FrontendForSetting(context.Context, app.cfg.Global.ContainerFrontend)
 	if err != nil {
-		app.console.Warnf("%s frontend could not be initialized, using stub: %s", app.cfg.Global.ContainerFrontend, err.Error())
+		app.console.Warnf("%s frontend could not be initialized, but trying anyways", app.cfg.Global.ContainerFrontend)
```
alexcb, line 1025:
```suggestion
		app.console.Warnf("%s frontend could not be initialized, but trying anyway", app.cfg.Global.ContainerFrontend)
```

PR #1251, opened 2021-10-04
Add user platform built-in vars

### LINKED ISSUE #1228: Add built-in vars to expose the host os/platform
Presently it is not easily possible to know what the host os/platform is.  This is particularly importatnt when working on a machine like an m1 macbook that tends to run lots of portions under emulation.

For example you may want to do something like the following (mostly to avoid golang crashing instant under qemu)

```
go-cross-amd64:
    FROM --platform=linux/${HOST_PLATFORM} golang:1.17.1
    # Setup cross compiler toolchain if not running on host platform
    IF [ ${HOST_PLATFORM} != "amd64" ]
        RUN dpkg --add-architecture amd64
        RUN apt-get update
        RUN apt-get install -y gccgo-x86-64-linux-gnu
        ENV CC=x86_64-linux-gnu-gcc
        ENV PKG_CONFIG_PATH=/usr/lib/x86_64-linux-gnu/pkgconfig
    END
    
	COPY . /work
	WORKDIR /work
    RUN GOOS=linux GOARCH=amd64 CGO=1 go build \
        -o ./processor cmd/main.go
```

akrantz01:
resolves #1228 

This adds the built-ins for `USERPLATFORM`, `USEROS`, `USERARCH`, and `USERVARIANT` for the host platform.

Are there any other places that these need to be documented? I added it in the `built-ins.md`, but I'm not sure if there's anywhere else it needs to be added. Also, should these be considered experimental similar to how `TARGET*` are?

### DIFF
--- a/docs/earthfile/builtin-args.md
+++ b/docs/earthfile/builtin-args.md
@@ -41,6 +41,10 @@ The following builtin args are available
 | `TARGETARCH` | (**experimental**) The target processor architecture the target is being built for. | `arm`, `amd64`, `arm64` |
 | `TARGETVARIANT` | (**experimental**) The target processor architecture variant the target is being built for. | `v7` |
 | `EARTHLY_SOURCE_DATE_EPOCH` | The timestamp, as unix seconds, of the git commit detected within the build context directory. If no git directory is detected, then the value is `0` (the unix epoch) | `1626881847`, `0` |
+| `USERPLATFORM` | (**experimental**) The platform the target is being built on. | `linux/arm/v7`, `linux/amd64`, `linux/arm64` |
+| `USEROS` | (**experimental**) The OS the target is being built on. | `linux` |
+| `USERARCH` | (**experimental**) The processor architecture the target is being built on. | `arm`, `amd64`, `arm64` |
+| `USERVARIANT` | (**experimental**) The processor architecture variant the target is being built on. | `v7` |
 
 {% hint style='info' %}
 ##### Note
--- a/docs/earthfile/builtin-args.md
+++ b/docs/earthfile/builtin-args.md
@@ -41,10 +41,10 @@ The following builtin args are available
 | `TARGETARCH` | (**experimental**) The target processor architecture the target is being built for. | `arm`, `amd64`, `arm64` |
 | `TARGETVARIANT` | (**experimental**) The target processor architecture variant the target is being built for. | `v7` |
 | `EARTHLY_SOURCE_DATE_EPOCH` | The timestamp, as unix seconds, of the git commit detected within the build context directory. If no git directory is detected, then the value is `0` (the unix epoch) | `1626881847`, `0` |
-| `USERPLATFORM` | (**experimental**) The platform the target is being built on. | `linux/arm/v7`, `linux/amd64`, `linux/arm64` |
-| `USEROS` | (**experimental**) The OS the target is being built on. | `linux` |
-| `USERARCH` | (**experimental**) The processor architecture the target is being built on. | `arm`, `amd64`, `arm64` |
-| `USERVARIANT` | (**experimental**) The processor architecture variant the target is being built on. | `v7` |
+| `USERPLATFORM` | (**experimental**) The platform the target is being built from. | `linux/arm/v7`, `linux/amd64`, `linux/arm64` |
+| `USEROS` | (**experimental**) The OS the target is being built from. | `linux` |
+| `USERARCH` | (**experimental**) The processor architecture the target is being built from. | `arm`, `amd64`, `arm64` |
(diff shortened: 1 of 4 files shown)


docs/earthfile/builtin-args.md
```diff
@@ -41,6 +41,10 @@ The following builtin args are available
 | `TARGETARCH` | (**experimental**) The target processor architecture the target is being built for. | `arm`, `amd64`, `arm64` |
 | `TARGETVARIANT` | (**experimental**) The target processor architecture variant the target is being built for. | `v7` |
 | `EARTHLY_SOURCE_DATE_EPOCH` | The timestamp, as unix seconds, of the git commit detected within the build context directory. If no git directory is detected, then the value is `0` (the unix epoch) | `1626881847`, `0` |
+| `USERPLATFORM` | (**experimental**) The platform the target is being built on. | `linux/arm/v7`, `linux/amd64`, `linux/arm64` |
```
vladaionescu, line 44: Adjust wording slightly to account for the remote case too.

```suggestion
| `USERPLATFORM` | (**experimental**) The platform the target is being built from. | `linux/arm/v7`, `linux/amd64`, `linux/arm64` |
```

docs/earthfile/builtin-args.md
```diff
@@ -41,6 +41,10 @@ The following builtin args are available
...
+| `USEROS` | (**experimental**) The OS the target is being built on. | `linux` |
+| `USERARCH` | (**experimental**) The processor architecture the target is being built on. | `arm`, `amd64`, `arm64` |
+| `USERVARIANT` | (**experimental**) The processor architecture variant the target is being built on. | `v7` |
```
vladaionescu, line 47: Ditto for the others too.

variables/builtin.go
```diff
@@ -70,6 +71,15 @@ func SetPlatformArgs(s *Scope, platform specs.Platform) {
...
 }
 
+// SetUserPlatformArgs sets the user's platform-specific built-in args.
+func SetUserPlatformArgs(s *Scope) {
+	platform := platforms.DefaultSpec()
+	s.AddInactive("USERPLATFORM", platforms.Format(platform))
+	s.AddInactive("USEROS", platform.OS)
+	s.AddInactive("USERARCH", platform.Architecture)
+	s.AddInactive("USERVARIANT", platform.Variant)
+}
```
vladaionescu, line 81: Can you also add these to targetinput.go?

variables/builtin.go
```diff
@@ -70,6 +71,15 @@ func SetPlatformArgs(s *Scope, platform specs.Platform) {
 	s.AddInactive("TARGETVARIANT", platform.Variant)
 }
 
+// SetUserPlatformArgs sets the user's platform-specific built-in args.
+func SetUserPlatformArgs(s *Scope) {
+	platform := platforms.DefaultSpec()
```
vladaionescu, line 76: This won't work very well as we default to `linux/*` on a Mac for example. We probably need a new function in `platforms` for this.

vladaionescu (commented):
A good start! Thank you for the contribution!

Some comments below...

variables/builtin.go
akrantz01: I think I already did. It's [here](https://github.com/earthly/earthly/blob/9f5eddafd447182dc8c950f7a1bc972df311a9a6/states/dedup/targetinput.go#L164-L167), right?

vladaionescu: Ah yes - indeed

variables/builtin.go
akrantz01: There's the function `llbutil.DefaultPlatform` that looks to default to `linux/*` for Mac. However, it also does the same for Windows, which I don't think is desired in this case.

Would it be worth creating a different variant that does not change the OS for Windows?

vladaionescu: Yes, different flavor is this would make sense

vladaionescu: Actually, your original implementation was correct here. Sorry for misleading. I was thinking of our `llbutil` package, but actually you were correctly using the containerd `platforms`, which doesn't override mac nor windows.

akrantz01: Ah ok, no worries.

vladaionescu (approved):
This is looking great to me! Thank you for the contribution (and your patience :D )!

@alexcb - mind validating that this is working as intended?

alexcb (approved):
I gave it a test with:

```
FROM alpine:3.13
user-args-test:
    ARG USERPLATFORM
    ARG USEROS
    ARG USERARCH
    ARG USERVARIANT
    RUN test "$USERPLATFORM" = "linux/amd64"
    RUN test "$USEROS" = "linux"
    RUN test "$USERARCH" = "amd64"
    RUN test "$USERVARIANT" = ""
```

on my linux machine, looks good. I will open up a follow-up PR to include the above with our tests.

PR #1256, opened 2021-10-05
fix typos in docs

garanews:


adamgordonbell (approved):
Thanks @garanews! 
If you'd like some earthly stickers checkout this blog post: https://earthly.dev/blog/hacktoberfest-2021/

docs/guides/registries/self-signed.md
```diff
@@ -68,7 +68,7 @@ global:
       http = true|false
       insecure = true|false
       ca=["<ca-path-pem>"]
-      [[registry."<registry-hostname>".keypair]]
+      [[registry."<registry-hostname>".key pair]]
```
alexcb, line 71: This shouldn't be changed as it introduces whitespace.

alexcb (changes requested):
thanks for the fixes, just one thing should be left as-is.

docs/guides/registries/self-signed.md
garanews: Oh sorry, didn't notice it was a field.

PR #1259, opened 2021-10-05
Add Spellchecking of markdown files

alexcb:
Adds `vale` for spellchecking of some markdown files;
many other files have been temporarily omitted from spelling
until we fix them.


.vale/styles/HouseStyle/TLA.yml
```diff
@@ -0,0 +1,173 @@
+extends: conditional
```
adamgordonbell, line 1: FYI: I found this rule to be a bit of a pain and was thinking of removing it from the blog. Lots of false positives.

adamgordonbell (approved):
Nice!
At some point, we might want to consider putting these rules in a git submodule that the blog/website and docs can all share, but I think this is great step forward.

alexcb:
> git submodule

we could also use a remote target such as `FROM github.com/earthly/earthly+vale` which would pull in the rules.

However, this would tie us to using the same rules, but maybe we could do something similar for just the custom dictionary?

PR #1260, opened 2021-10-05
test user-args

alexcb:
This PR implements tests for https://github.com/earthly/earthly/pull/1251


alexcb:
I tried using 

```
# This is used to detect if we are on a linux amd64 machine
# which will then be used to determin if we can call the test
# to validate $USERPLATFORM is correct; therefore we can't use
# $USERPLATFORM here.
local-uname:
    LOCALLY
    RUN uname -a > /tmp/uname-data-for-earthly-test
    SAVE ARTIFACT /tmp/uname-data-for-earthly-test output

user-arg-test:
    COPY +local-uname/output uname
    IF grep "Linux" uname && grep "x86_64" uname
        DO +RUN_EARTHLY --earthfile=user-arg.earth
    ELSE
        ARG EARTHLY_TARGET
        RUN echo "skipping $EARTHLY_TARGET due to non-amd64-linux host"
    END
```

however this does not work under `--strict`, so I decided to just call `+user-arg-test` directly from GHA.

PR #1264, opened 2021-10-06
Only store mirror password in GitHub secrets

alexcb:
The mirror username is "docker" and is currently stored
as a secret, this causes the earthly output to replace
all instances of "docker" with "***", e.g.

    test-***-compose failed with exit code=1

Once this PR is merged, i'll remove `DOCKERHUB_MIRROR_USERNAME`
from our secrets.


alexcb:
The username has now been changed to something much less common, `docker` will now be visible in our logs.

PR #1271, opened 2021-10-07
Make secrets and mirror optional, to ease contribution.

dchw:
You can invoke it like this:

```
earthly -P --build-arg DOCKERHUB_AUTH=true --build-arg DOCKERHUB_USER_SECRET=+secrets/earthly-technologies/dockerhub-mirror/user --build-arg DOCKERHUB_TOKEN_SECRET=+secrets/earthly-technologies/dockerhub-mirror/pass --build-arg DOCKERHUB_MIRROR=registry-1.docker.io.mirror.corp.earthly.dev +test
```

Or just specify the default secrets at a path:

```
earthly -P --secret DOCKERHUB_USER=me --secret DOCKERHUB_TOKEN=hunter2 +test
```

In addition, remove lots of the `if`s in GHA that did different things depending on where the branch came from. We are doing this in favor of the maintainer approval button per earlier discussion in Slack.


vladaionescu:
Yeah, maybe. Basically a struct :-)

Could be worth a proposal to explore some options

dchw:
> Yeah, maybe. Basically a struct :-)
> 
> Could be worth a proposal to explore some options

Gonna be tricky to avoid creating

PR #1281, opened 2021-10-08
buildkite test job killer

alexcb:
If no jobs are running, we get the errornous error:

    kill: usage: kill [-s sigspec | -n signum | -sigspec] pid | jobspec ... or kill -l [sigspec]

which adds noise to the error logs.

Additionally, we should kill jobs on all exits (including clean), not just sigint and sigterm

### DIFF
--- a/.buildkite/tests.sh
+++ b/.buildkite/tests.sh
@@ -2,7 +2,16 @@
 
 set -xeu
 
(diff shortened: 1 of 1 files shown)


alexcb:
example buildkite failure:

image

from https://buildkite.com/earthly-technologies/earthly-mac-scheduled/builds/13#88c0bc85-77cc-4dcf-9d65-38a62ddb6abd

Issue #707, opened 2021-01-12
Include secrets when RUN --secrets is run


agbell:
To run a command with a secret ENV set, I need to specify it twice, once when I run earthly and once when I run the command.  It would be useful if there were a terser syntax.

Now:
```
earthly -P \
    --secret AWS_ACCESS_KEY_ID=foo \
    --secret AWS_SECRET_ACCESS_KEY=bar \
    --secret AWS_SESSION_TOKEN=baz \
```

```
 RUN \
            --secret AWS_ACCESS_KEY_ID=+secrets/AWS_ACCESS_KEY_ID \
            --secret AWS_SECRET_ACCESS_KEY=+secrets/AWS_SECRET_ACCESS_KEY \
            --secret AWS_SESSION_TOKEN=++secrets/AWS_SESSION_TOKEN \
            printenv
```

Potential Simplification:

```
earthly -P \
    --secret AWS_ACCESS_KEY_ID=foo \
    --secret AWS_SECRET_ACCESS_KEY=bar \
    --secret AWS_SESSION_TOKEN=baz \
```

```
 RUN --secrets printenv
```

vladaionescu:
This has the disadvantage that the build is less readable - it's less clear which secrets are actually used.

An alternative, which simplifies the usage is registering [cloud secrets](https://docs.earthly.dev/guides/cloud-secrets). With that, secrets are only referenced in the Earthfile only.

kevinkjt2000:
Cloud secrets may not be a possibility for some projects. Imagine laws/regulations that enforce storage of secrets to particular systems, relying on a cloud for privileged information that is outside of those systems could be illegal.

As for this issue, I don't know how I feel about exposing all secrets to a command when some may want to be more selective about which secrets go to which commands. I concur that it is also less clear which secrets are being used in this way.

For secrets that have exactly the same environment variable name as their counterpart, how about reducing the repetition by changing this
```
 RUN \
            --secret AWS_ACCESS_KEY_ID=+secrets/AWS_ACCESS_KEY_ID \
            --secret AWS_SECRET_ACCESS_KEY=+secrets/AWS_SECRET_ACCESS_KEY \
            --secret AWS_SESSION_TOKEN=+secrets/AWS_SESSION_TOKEN \
            printenv
```
to this
```
 RUN \
            --secret AWS_ACCESS_KEY_ID \
            --secret AWS_SECRET_ACCESS_KEY \
            --secret AWS_SESSION_TOKEN \
            printenv
```
[...]

PR #1284, opened 2021-10-09
enhancement: add short version for secrets

dhaven:
Implements https://github.com/earthly/earthly/issues/707#issuecomment-764693208

Give the user the option to reference secrets in a compact way when the env var name is the same as the secret-id.


wesselvdv:
Was a release ever cut for this? I states in the docs that this should work, but I'm getting a `invalid secret definition ` when using this secret form.

PR #1312, opened 2021-10-15
Separate earthly tests from examples

andrewsykim:
This PR separates `examples/tests` into its own directory `tests/`. I'm not actually sure if keeping our tests in `examples/tests` could still be useful because the tests do serve as good examples sometimes. But the changes required to make it work were pretty simple so I figured I would open the PR anyways.


andrewsykim:
Admittedly this is more of a cosmetic fix, so I think I'm okay to close it if it would cause more damage like broken links and references.

andrewsykim:
@alexcb PR has updates to `.github/workflows/ci.yml` already, anything I missed there?

andrewsykim:
> That's a lot of diffs

Fwiw most of the diffs are just moving a bunch of the Earthfiles from `example/tests` to `tests/` :P

PR #1341, opened 2021-10-22
Introduce an internal cache version number

alexcb:
Whenever the number increases, the previous cache will be reset.


vladaionescu:
I wonder if the cloud cache could cause any issues. Hopefully, that's compatible between versions.

PR #1342, opened 2021-10-22
0.6: Update docs and examples

vladaionescu:
Docs preview: https://app.gitbook.com/o/-MB5s2W0v8S4xFU5ELti/s/WAEFmRlTkh9KoLoExeDB/

Our own builds have not been updated to avoid conflicts with the main branch. Only examples and docs.

This removes **beta** and **experimental** labels in the docs for a number of features that are being GA'd. It also enables by default the necessary feature flags (@alexcb - mind reviewing this part to make sure that I got it right? See `features.go` changes).

Not yet done (will be done in a separate PR. EDIT: See https://github.com/earthly/earthly/pull/1344):
* Enabling local registry exporter by default
* Enabling client conversion parallelism by default
* Enabling --use-inline-cache for all builds and deprecating the flag

Question regarding `FROM DOCKERFILE`: IIRC this is left as beta only because it doesn't quite work on windows. But windows is already in the experimental bucket - so I think we can call `FROM DOCKERFILE` ga then? Thoughts @dchw?

### DIFF
--- a/docs/earthfile/earthfile.md
+++ b/docs/earthfile/earthfile.md
@@ -814,7 +814,7 @@ Sets an initialization time period in which failures are not counted towards the
 
 Sets the number of retries before a container is considered `unhealthy`. Defaults to `3`.
 
-## FROM DOCKERFILE (**beta**)
+## FROM DOCKERFILE
 
 #### Synopsis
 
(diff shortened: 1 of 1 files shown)


alexcb:
> Do we already support returning an error as described?
> 

@vladaionescu we will need to insert `0.6` here: https://github.com/earthly/earthly/blob/main/ast/validator.go#L15-L17

Issue #1221, opened 2021-09-22
Support the ability to change build context


bkendzior:
I'm curious if there's a way to change the build context when running earthly.

i.e. my repository is organized as such:
```
/repo_root
|- ./docker
 |- Earthfile
 |- configurations, scripting, etc
|- ./src
 |- code I want to COPY into my container
```

docker build allows for you to specify a build context of `..` for this particular use-case, curious if Earthly supports anything similar or if the build context is purely an artifact of the placement of the Earthfile

alexcb:
as a work around, the following might work until we have a better solution:

```
run:
    FROM alpine
    COPY --dir +hack/src .
    RUN cat src/code

hack:
    LOCALLY
    SAVE ARTIFACT ../src src
```

vladaionescu:
We might not want to support this as a matter of convention. We want the context of the build to always be the current directory to make it easy for the user to follow the logic.

bjaanes:
Or perhaps make it even more explicit instead with `--context`, `-c` set to the path where you want the context to be run from.
That way it is still possible to run the Earthfile from anywhere with the `earthly ./path/to/earthfile/dir+target` (and not having two ways of specificing where the Earthfile is) and it would have a slightly similar way of doing it to Docker (only the context is not required and is by default the location of the Earthfile).

andrewsykim:
After mulling over this for a few hours, I think we need to give this more thought. Being able to set the local context can be useful for many cases as we already mentioend but it can negatively impact the portability and simplicity of Earthfiles. I moved the issue back to "Under Consideration" as we think about the right way to approach this.

adamgordonbell:
Rather than copy from a parent, it would be useful to have the source control root as a var in Earthfiles. Then in a mono repo setup, you could refer to other source dependencies, but still, you wouldn't harm reproducibility because you are never outside of your source control.
```
/repo_root
|- ./docker
 |- Earthfile
 |- configurations, scripting, etc
|- ./src
 |- code
```

Then you could do something like :  
```
COPY $ROOT/src/code .
```

I hit this issue with building this[ go project](https://github.com/adamgordonbell/cloudservices/tree/main/activity-client) but I suspect having a way to reference paths from the root would be useful in many mono-repo setup where their are dependencies between sibling paths.

brandonSc:
Wonder if both options should be supported. 

`../my-other-dir` is pretty intuitive and I think a lot of users may try that and expect that it might work. 

[...]

PR #1391, opened 2021-11-03
Support COPY-ing parent directories from local sources

andrewsykim:
Fixes https://github.com/earthly/earthly/issues/1221

(this PR is still a WIP but looking for some early feedback)

This PR adds the ability to use the `COPY` command against local sources that are referenced from parent directories of the current build context. 

Some notable things about the changes:
* COPY has validation such that a single COPY command can only operate against a single build context at a time. In other words, Earthly would disallow a COPY command such as `COPY ./test ../test1 ../../test2 ./`. However, you can specify multiple COPY commands in the same Earthfile that operate against different build contexts each.
* COPY-ing parent directories will be an experimental feature initially and will requires the `--copy-parent-dirs` feature flag at the top of the Earthfile
* COPY-ing whole parent directories like `COPY .. .` is disallowed since that could recursively copy files from the main local build context.
* When copying parent directories, we also assume `--use-copy-with-include-patterns` so that only referenced sources are added to the build context dynamically. This is to ensure the build context does not explode in size when referencing parent directories.

EDIT (vladaionescu):
* An additional limitation might be that `COPY`ing `../*` or `../service-*` (where the pattern matches directories too), will also end up recursively copying, resulting in reaching over more than one build context in a single copy operation


tests/copy/with-parents.earth
```diff
@@ -0,0 +1,45 @@
+VERSION --copy-parent-dirs 0.6
```
andrewsykim, line 1: Not sure how we would write a test to validate the files we sent to the build context, but here's some manual validation I did as suggested by @alexcb :

```
./t/copy+copy-with-parents |          alpine:3.13 | --> Load metadata linux/amd64
./t/copy+copy-with-parents |          alpine:3.13 | Completed in 2.008680026s
./t/copy+copy-with-parents |              context | --> local context .
./t/copy+copy-with-parents |              context | Completed in 37.11µs
./t/copy+copy-with-parents |              context | --> local context .
./t/copy+copy-with-parents |              context | Completed in 21.29µs
./t/copy+copy-with-parents |                +base | --> FROM alpine:3.13
./t/copy+copy-with-parents |                +base | [██████████] 100% resolve docker.io/library/alpine:3.13@sha256:2582893dec6f12fd499d3a709477f2c0c0c1dfcd28024c93f1f0626b9e3540c8
./t/copy+copy-with-parents |                +base | Completed in 1.540602ms
./t/copy+copy-with-parents |                +base | Completed in 27.25µs
./t/copy+copy-with-parents |                +base | Completed in 376.805µs
./t/copy+copy-with-parents |              context | [          ]   0% transferring ../..:
./t/copy+copy-with-parents |              context | sent file stat for ../test1
./t/copy+copy-with-parents |              context | sent file stat for ../test1/foo1
./t/copy+copy-with-parents |              context | sent data for ../test1/foo1 (5 B)
./t/copy+copy-with-parents |              context | [██████████] 100% transferring ..:
./t/copy+copy-with-parents |              context | Completed in 1.555302ms
./t/copy+copy-with-parents |              context | sent file stat for ../../opt
./t/copy+copy-with-parents |              context | sent file stat for ../../opt/test2
./t/copy+copy-with-parents |              context | sent file stat for ../../opt/test2/foo2
./t/copy+copy-with-parents |              context | sent data for ../../opt/test2/foo2 (5 B)
./t/copy+copy-with-parents |                +base | [          ]   0% sha256:4e9f2cdf438714c2c4533e28c6c41a89cc6c1b46cf77e54c488db30ca4f5b6f3
./t/copy+copy-with-parents |              context | [██████████] 100% transferring ../..:
./t/copy+copy-with-parents |              context | Completed in 170.921911ms

```

alexcb: > Not sure how we would write a test to validate the files we sent to the build context, but here's some manual validation I did as suggested by @alexcb :

Maybe we can extend https://github.com/earthly/earthly/blob/main/tests/Earthfile#L168-L178 to include copying a file from a parent directory?

andrewsykim: Ah great catch, let me add that

andrewsykim: updated

andrewsykim:
@alexcb @vladaionescu still need to polish up a few things and add more tests but looking for some early feedback on the general direction of this patch,

earthfile2llb/converter.go
```diff
@@ -442,13 +442,59 @@ func (c *Converter) CopyClassical(ctx context.Context, srcs []string, dest strin
 		return err
 	}
 
+	var buildContext string
+	for _, src := range srcs {
+		// local sources that copy entire parent directores like ".." or "../" are not allowed
```
alexcb, line 447: What would happen if a user is working under `/home/user/some/path/Earthfile` and does a `COPY ../../some`, would that still hit this recursion issue?

andrewsykim: I don't think so because we force `--use-copy-include-patterns` when copying from parent dirs -- but let me add a test case for this to be sure

alexcb:
> @alexcb @vladaionescu still need to polish up a few things and add more tests but looking for some early feedback on the general direction of this patch,

It's looking great. Good call moving those copy integration tests into a separate directory -- they are looking nice and thorough.

vladaionescu:
Closing as obsolete.

PR #1394, opened 2021-11-03
show hidden help when running dev versions

alexcb:
autocomplete already suggests hidden flags; however when running --help
those flags are not displayed. This extends the logic to cover both
autocompletion and help text.

### DIFF
--- a/cmd/earthly/main.go
+++ b/cmd/earthly/main.go
@@ -73,6 +73,7 @@ import (
 	"github.com/earthly/earthly/util/containerutil"
 	"github.com/earthly/earthly/util/fileutil"
 	"github.com/earthly/earthly/util/llbutil"
+	"github.com/earthly/earthly/util/reflectutil"
 	"github.com/earthly/earthly/util/termutil"
 	"github.com/earthly/earthly/variables"
 )
@@ -246,6 +247,7 @@ func main() {
 	}
 
 	app := newEarthlyApp(ctx, conslogging.Current(colorMode, padding, false))
+	app.unhideFlags(ctx)
 	app.autoComplete(ctx)
 
 	exitCode := app.run(ctx, os.Args)
@@ -615,7 +617,7 @@ func newEarthlyApp(ctx context.Context, console conslogging.ConsoleLogger) *eart
 		},
 		{
 			Name:        "docker",
-			Usage:       "Build a Dockerfile without converting to an Earthfile",
+			Usage:       "Build a Dockerfile without converting to an Earthfile *experimental*",
 			Description: "Builds a dockerfile",
 			Hidden:      true, // Experimental.
 			Action:      app.actionDocker,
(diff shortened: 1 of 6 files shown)


cmd/earthly/main.go
```diff
@@ -1278,6 +1280,44 @@ func (app *earthlyApp) processDeprecatedCommandOptions(context *cli.Context, cfg
 	return nil
 }
 
+func (app *earthlyApp) unhideFlags(ctx context.Context) error {
+	var err error
+	showHidden := strings.HasPrefix(Version, "dev-")
```
andrewsykim, line 1285: Part of me thinks we should keep these flags hidden so we don't get confused with what is hidden or not in the actual released versions. The `EARTHLY_SHOW_HIDDEN` would be there to always show it otherwise

alexcb: If we run `EARTHLY_SHOW_HIDDEN=0 earthly` it'll force them to be hidden even when running a dev version. When `EARTHLY_SHOW_HIDDEN` isn't set, we default to this auto behaviour, and `EARTHLY_SHOW_HIDDEN=1` we will always show them.

I have `EARTHLY_SHOW_HIDDEN=1` set in my `.bashrc`, so I don't have a strong opinion on removing the auto heuristic.

alexcb: @andrewsykim it's been removed. Can you take another look?

cmd/earthly/main.go
```diff
@@ -1278,6 +1280,43 @@ func (app *earthlyApp) processDeprecatedCommandOptions(context *cli.Context, cfg
 	return nil
 }
 
+func (app *earthlyApp) unhideFlags(ctx context.Context) error {
+	var err error
+	if os.Getenv("EARTHLY_AUTOCOMPLETE_HIDDEN") != "" && os.Getenv("COMP_POINT") == "" { // TODO delete this check after 2021-03-01
```
alexcb, line 1285: randomly picked date; this isn't an advertised feature -- no need to support long term.

util/reflectutil/setbool.go
```diff
@@ -0,0 +1,20 @@
+package reflectutil
+
+import (
+	"reflect"
+)
+
+// SetBool looks for a boolean struct value named `structName` in `iface` and sets it to `value`.
+// Upon success a true value is returned, otherwise false.
+func SetBool(iface interface{}, structName string, value bool) bool {
```
andrewsykim, line 9: minor nit, `s/structName/fieldName`

andrewsykim (approved):
LGTM, just left some minor non-blocking comments

util/reflectutil/setbool.go
alexcb: hah, such an obvious variable name. good point :)

cmd/earthly/main.go
alexcb: That seems much longer term than needed. It's only really a developer-feature, I don't think it's something that we need to worry about. I suspect I could just have skipped this warning all together and nobody would have noticed.

andrewsykim: yeah fair -- 0.7 was just an example, saying "after v0.6" could work too

PR #1395, opened 2021-11-03
Allow setting env variables before starting buildkit

alexcb:
Introduces a new `buildkit_additional_entrypoint` global argument that
can be configured in `~/.earthly/config.yml`, which is evaluated before
running buildkit. This allows a user to pass arbitratry commands to the
entrypoint.sh script, e.g.

    buildkit_additional_entrypoint: 'export BUILDKIT_SCHEDULER_DEBUG=1'

or it can be extended to help with debugging, for example displaying all
environment variables containing the string debug:

    buildkit_additional_entrypoint: 'echo ---start---; env | grep -i debug; echo ---end---'

### DIFF
--- a/buildkitd/buildkitd.go
+++ b/buildkitd/buildkitd.go
@@ -306,6 +306,10 @@ func Start(ctx context.Context, console conslogging.ConsoleLogger, image, contai
 		envOpts["EARTHLY_ADDITIONAL_BUILDKIT_CONFIG"] = settings.AdditionalConfig
 	}
 
+	if settings.AdditionalEntrypoint != "" {
+		envOpts["EARTHLY_ADDITIONAL_ENTRYPOINT"] = settings.AdditionalEntrypoint
+	}
+
 	if settings.IPTables != "" {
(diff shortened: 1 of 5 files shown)


alexcb:
The underlying reason for this, is due to a user having buildkit producing `Error: build target: build main: bkClient.Build: failed to solve: buildkit scheduler error: return leaving incoming open. Please report this with BUILDKIT_SCHEDULER_DEBUG=1`. Now we can do such a thing. (Unfortunately this setting is only set via an env and can't be done via buildkit's config file).

vladaionescu:
I think you can already do this via `buildkit_additional_args `. Those args are actually args passed to the `docker run` command, which means you can also pass `"-e", "BUILDKIT_SCHEDULER_DEBUG=1"`, I believe.

alexcb:
Ah ha, you're correct! Not sure how I missed that.

```
global:
    buildkit_additional_args: ['-e', 'BUILDKIT_SCHEDULER_DEBUG=1']
```

did the trick!

PR #1400, opened 2021-11-08
fix cache export when using --max-remote-cache

andrewsykim:
In `cmd/earthly/main.go`, we set either the `cacheExport` or the `maxCacheExport` field into the Builder. But when generating cache export options for buildkit, we we're always using cacheExport even when `--max-remote-cache` is set.

Signed-off-by: Andrew Sy Kim <andrew@earthly.dev>

### DIFF
--- a/builder/solver.go
+++ b/builder/solver.go
@@ -205,7 +205,7 @@ func (s *solver) newSolveOptMulti(ctx context.Context, eg *errgroup.Group, onIma
 		cacheExports = append(cacheExports, newCacheExportOpt(s.cacheExport, false))
 	}
 	if s.maxCacheExport != "" {
-		cacheExports = append(cacheExports, newCacheExportOpt(s.cacheExport, true))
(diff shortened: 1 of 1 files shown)


builder/solver.go
```diff
@@ -205,7 +205,7 @@ func (s *solver) newSolveOptMulti(ctx context.Context, eg *errgroup.Group, onIma
 		cacheExports = append(cacheExports, newCacheExportOpt(s.cacheExport, false))
 	}
 	if s.maxCacheExport != "" {
-		cacheExports = append(cacheExports, newCacheExportOpt(s.cacheExport, true))
+		cacheExports = append(cacheExports, newCacheExportOpt(s.maxCacheExport, true))
```
andrewsykim, line 208: In main.go we only set one of them:

https://github.com/earthly/earthly/blob/17feb937be8f295f0e094a09b6999c82ad3a6a40/cmd/earthly/main.go#L2846-L2851

vladaionescu (approved):

PR #1409, opened 2021-11-12
Bump buildkit

dchw:
Include Kind fix


alexcb:
Something looks off with this, I would expect to see `go.sum` and `go.mod` updated too, similar to this PR: https://github.com/earthly/earthly/pull/1404/files

dchw:
Just ran that target without specifying the SHA since it looked like it autodetected that

dchw:
@alexcb I wasn't using the latest version, just been rockin' v0.5.24, turns out

alexcb (approved):

Issue #1439, opened 2021-11-25
TLS error when enabling buildkit mTLS


andrewsykim:
Recently updated to use v0.6.0 and I started getting this error when enabling mTLS:
```
$ earthly +for-linux
...
...
Error: build target: build main: bkClient.Build: listing workers for Build: failed to list workers: Unavailable: connection error: desc = "transport: authentication handshake failed: remote error: tls: no application protocol"
```

The same issue did not occur using v0.6.0-rc3 so maybe there was a buildkit update introducing this regression. Here are some steps I took to validate that my setup is not misconfigured:
```
# config
$ cat ~/.earthly/config.yml 
global:
  tls_enabled: true

$ echo $EARTHLY_BUILDKIT_HOST 
tcp://127.0.0.1:8372

# checking that certs exist
$ ls ~/.earthly/certs/
buildkit_cert.pem  buildkit_key.pem  ca_cert.pem  ca_key.pem  earthly_cert.pem  earthly_key.pem

# checking that buildkit config has TLS enabled:
$ docker exec -ti earthly-buildkitd sh -c "cat /etc/buildkitd.toml"
debug = false
root = "/tmp/earthly/buildkit"
insecure-entitlements = [ "security.insecure" ]

[grpc]
  address = [ "tcp://0.0.0.0:8372" ]
# Please note the required indentation to fit in buildkit.toml.template accordingly.

  [grpc.tls]
    cert = "/etc/cert.pem"
    key = "/etc/key.pem"
    ca = "/etc/ca.pem"

[worker.oci]
  enabled = true
  snapshotter = "auto"
  max-parallelism = 20
  gc = true
  networkMode = "cni"
  cniBinaryPath = "/usr/libexec/cni"
  cniConfigPath = "/etc/cni/cni-conf.json"
```

This is using the certificate generation logic built into earthly so I don't think there's anything wrong with the certificates.

mxey:
I enabled gRPC logging on both earthly and buildkitd with GRPC_GO_LOG_SEVERITY_LEVEL=info GRPC_GO_LOG_VERBOSITY_LEVEL=99.

buildkitd:

```
Nov 25 12:24:36 s100205 buildkitd[29785]: WARNING: 2021/11/25 11:24:36 [core] grpc: Server.Serve failed to create ServerTransport:  connection error: desc = "transport: http2Server.HandleStreams failed to receive the preface from client: tls: client requested unsupported application protocols ([h2])"
```

earthly:

```
[...]

PR #1442, opened 2021-11-25
update buildkit to fix remote buildkit mTLS support

alexcb:
fixes https://github.com/earthly/earthly/issues/1439


.github/workflows/ci.yml
```diff
@@ -242,6 +242,8 @@ jobs:
       - name: Execute registry-certs test (Fork Only)
         run: ./tests/registry-certs/test.sh --build-arg DOCKERHUB_AUTH=false
         if: github.event_name != 'push' && github.event.pull_request.head.repo.full_name != github.repository
+      - name: Execute remote buildkit test
```
andrewsykim, line 245: Minor comment here on wording -- since we're still using localhost I the "remote" is a bit misleading. Maybe this test should be named something like `Execute local buildkit with mTLS test` or something like that

andrewsykim (approved):
Nice! Thanks for adding the test!

Issue #1448, opened 2021-11-26
ARG is evaluated twice (once under +base, once under +target)


alexcb:
```
foo:
    FROM alpine
    ARG myvar
    WORKDIR /test
    RUN echo "myvar is $myvar\n; dir contents is $(ls -la)"

run:
    FROM alpine
    RUN echo bar > data
    BUILD +foo --myvar="$(cat data)"
```

under `0.5.24`, it worked:

```
$ earthly-v0.5.24 --buildkit-volume-name=five-two-four +run
           buildkitd | Starting buildkit daemon as a docker container (earthly-buildkitd)...
           buildkitd | ...Done
              alpine | --> Load metadata linux/amd64
                +run | --> FROM alpine
                +run | [██████████] resolve docker.io/library/alpine@sha256:21a3deaa0d32a8057914f36584b5288d2e5ecc984380bc0118285c70fa8c9300 ... 100%
                +run | [██████████] sha256:59bf1c3509f33515622619af21ed55bbe26d24913cedbca106468a5fb37a50c3 ... 100%
                +run | [██████████] extracting sha256:59bf1c3509f33515622619af21ed55bbe26d24913cedbca106468a5fb37a50c3 ... 100%
                +run | --> RUN echo bar > data
                +run | --> ARG myvar = RUN $(cat data)
                +foo | myvar=bar
                +foo | --> WORKDIR /test
                +foo | myvar=bar
                +foo | --> RUN echo "myvar is $myvar\n; dir contents is $(ls -la)"
                +foo | myvar is bar\n; dir contents is total 8
                +foo | drwxr-xr-x    2 root     root          4096 Nov 26 19:04 .
                +foo | drwxr-xr-x    1 root     root          4096 Nov 26 19:04 ..
```

However under `0.6.0`, it now fails:

```
 1. Init 🚀
————————————————————————————————————————————————————————————————————————————————

           buildkitd | Found buildkit daemon as docker container (earthly-buildkitd)
           buildkitd | Updated image available. Restarting buildkit daemon...
           buildkitd | ...Done


 2. Build 🔧
————————————————————————————————————————————————————————————————————————————————

Error: build target: build main: bkClient.Build: failed to solve: parse build args: parse build arg myvar=$(cat data): the first command has to be FROM, FROM DOCKERFILE, LOCALLY, ARG, BUILD or IMPORT
```

If however we introduce a `FROM ...` under the base, like:

```
FROM alpine

foo:
    FROM alpine
[...]

PR #1450, opened 2021-11-27
Prevent parallel builds when arg is dynamic

alexcb:
This fixes https://github.com/earthly/earthly/issues/1448

### DIFF
--- a/earthfile2llb/interpreter.go
+++ b/earthfile2llb/interpreter.go
@@ -122,6 +122,15 @@ func (i *Interpreter) handleBlock(ctx context.Context, b spec.Block) error {
 	return nil
 }
 
+func containsDynamicBuildArgs(args []string) bool {
+	for _, arg := range args {
+		if strings.Contains(arg, "$(") {
+			return true
+		}
+	}
+	return false
+}
+
 func (i *Interpreter) handleBlockParallel(ctx context.Context, b spec.Block, startIndex int) error {
 	if i.local {
 		// Don't do any preemptive execution for LOCALLY targets.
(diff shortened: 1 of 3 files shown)


alexcb:
I need to dig in more -- there's already `isSafeAsyncBuildArgs` which should prevent this.

alexcb:
this fixes #1448, but I need to write a test before this PR is ready.

tests/build-arg-dynamic-with-empty-base.earth
```diff
@@ -0,0 +1,10 @@
+# This test requires that the base be empty (i.e. no FROM)
```
andrewsykim, line 1: What's the significance of the base being empty here? Does it impact how the ARG is evaluated?

earthfile2llb/interpreter.go
```diff
@@ -1561,10 +1561,25 @@ func parseParans(str string) (string, []string, error) {
...
-		if strings.HasPrefix(v, "$(") {
+		if strings.HasPrefix(v, "$(") || strings.HasPrefix(v, "\"$(") {
+			return false
+		}
+	}
+	return true
+}
+
+// isSafeAsyncBuildArgsNewStyle is used for "BUILD +target --key=value" style buildargs
+func isSafeAsyncBuildArgsNewStyle(args []string) bool {
```
andrewsykim, line 1576: Really minor non-blocking comment:

It might be slightly better to rename this to `isSafeAsyncBuildArgs` and rename the other one `isSafeAsyncBuildArgsDeprecated`

andrewsykim (approved):

tests/build-arg-dynamic-with-empty-base.earth
alexcb: This was part of the repo-case for this bug. With this fix this becomes irrelevant.

earthfile2llb/interpreter.go
alexcb: Good call, code looks funny when there's a `NewStyle` without any `OldStyle`, then v3 introduces `NewerStyle` :laughing:

vladaionescu (approved):

PR #1461, opened 2021-11-30
favor ssh-rsa keyscans

alexcb:
earthly comes pre-configured with ssh-rsa keyscans for github, gitlab,
bitbucket.

GitHub must have recently added a ecdsa-sha2-nistp256 keyscan, which
the crypto package favoured when passing it to our custom HostKeyCallback
function. This PR instructs the callback to favour passing the ssh-rsa
key when it exists.

### DIFF
--- a/buildcontext/gitlookup.go
+++ b/buildcontext/gitlookup.go
@@ -221,6 +221,7 @@ func (gl *GitLookup) detectProtocol(host string) (protocol gitProtocol, err erro
 
 	sshAgent, err := net.Dial("unix", gl.sshAuthSock)
 	if err != nil {
+		gl.console.VerbosePrintf("failed to connect to ssh-agent (using %s) due to %s, falling back to https", gl.sshAuthSock, err.Error())
 		protocol = httpsProtocol
 		err = nil
 		return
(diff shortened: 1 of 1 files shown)


buildcontext/gitlookup.go
```diff
@@ -230,11 +231,24 @@ func (gl *GitLookup) detectProtocol(host string) (protocol gitProtocol, err erro
 		Auth: []ssh.AuthMethod{
 			ssh.PublicKeysCallback(agent.NewClient(sshAgent).Signers),
 		},
+		HostKeyAlgorithms: []string{
```
andrewsykim, line 234: So to clarify, this list is the same as supportedHostKeyAlgos but with `ssh.KeyAlgoRSA` first for compatibility?

alexcb: yeah, I took the entire list from `list from golang.org/x/crypto/ssh/common.go supportedHostKeyAlgos` and moved `ssh.KeyAlgoRSA` to the top.

andrewsykim: The list seems different from the one referenced here? https://cs.opensource.google/go/x/crypto/+/ae814b36:ssh/common.go;l=69-81;bpv=0;bpt=0. Could it have changed in a newer version or something?

alexcb: ah ha! yeah mine came from `mod/golang.org/x/crypto@v0.0.0-20200820211705-5c72a883971a/ssh/common.go`.

Ultimately I'm not sure it matters that much in the context of this PR, as we just need to have `ssh.KeyAlgoRSA` at the top.

When we loop back and allow support for multiple pre-configured key scans, I want to remove this list.

vladaionescu (approved):

buildcontext/gitlookup.go
andrewsykim: sgtm, thanks!

Issue #1452, opened 2021-11-28
Create builtin ARG EARTHLY_VERSION


cehoffman:
This would be useful for referencing the docker setup script to make an image suitable for use `WITH DOCKER`, e.g. `  DO github.com/earthly/lib:$EARTHLY_VERSION+INSTALL_DIND`

alexcb:
https://github.com/earthly/earthly/blob/main/variables/builtin.go#L18 is where we can add a new entry for `EARTHLY_VERSION`.

Perhaps we should also include `EARTHLY_BUILD_SHA`, which would point to the corresponding git sha that was used for the build?

Here are the relevant strings that are set at build time and would need to be propigated into `builtin.go`: https://github.com/earthly/earthly/blob/main/cmd/earthly/main.go#L168-L172

PR #1478, opened 2021-12-03
Added Built-in ARGs: EARTHLY_VERSION and EARTHLY_BUILD_SHA

brandonSc:
Fixes: https://github.com/earthly/earthly/issues/1452

Example usage: 

Earthfile:
```Earthfile
VERSION --earthly-version-arg 0.6

FROM busybox:latest

build:
    ARG EARTHLY_VERSION
    ARG EARTHLY_BUILD_SHA
    RUN echo "$EARTHLY_VERSION"
    RUN echo "$EARTHLY_BUILD_SHA"
```
Output:
```
earthly +build
> 
              ...
              +build | --> RUN echo "$EARTHLY_VERSION"
              +build | dev-brandon_version-builtin-args
              +build | --> RUN echo "$EARTHLY_BUILD_SHA"
              +build | 51205ecda2da2f6dbe98850033a7e6d9394d4fbd
              ...
```


docs/earthfile/builtin-args.md
```diff
@@ -36,6 +36,8 @@ The following builtin args are available
 | `EARTHLY_GIT_ORIGIN_URL` | The git URL detected within the build context directory. If no git directory is detected, then the value is an empty string. Please note that this may be inconsistent, depending on whether an HTTPS or SSH URL was used. | `git@github.com:bar/buz.git` or `https://github.com/bar/buz.git` |
 | `EARTHLY_GIT_PROJECT_NAME` | The git project name from within the git URL detected within the build context directory. If no git directory is detected, then the value is an empty string. | `bar/buz` |
 | `EARTHLY_GIT_COMMIT_TIMESTAMP` | The timestamp, as unix seconds, of the git commit detected within the build context directory. If no git directory is detected, then the value is an empty string. | `1626881847` |
+| `EARTHLY_VERSION` | The version of Earthly currently running. | `v0.6.2`, `dev-main` |
+| `EARTHLY_BUILD_SHA` | The git hash of the commit which built the currently running version of Earthly. | `1a9eda7a83af0e2ec122720e93ff6dbe9231fc0c` |
```
vladaionescu, line 40: Mind splitting the docs changes into another PR that branches off of `next`?

docs/earthfile/builtin-args.md
```diff
@@ -36,6 +36,8 @@ The following builtin args are available
 | `EARTHLY_GIT_ORIGIN_URL` | The git URL detected within the build context directory. If no git directory is detected, then the value is an empty string. Please note that this may be inconsistent, depending on whether an HTTPS or SSH URL was used. | `git@github.com:bar/buz.git` or `https://github.com/bar/buz.git` |
 | `EARTHLY_GIT_PROJECT_NAME` | The git project name from within the git URL detected within the build context directory. If no git directory is detected, then the value is an empty string. | `bar/buz` |
 | `EARTHLY_GIT_COMMIT_TIMESTAMP` | The timestamp, as unix seconds, of the git commit detected within the build context directory. If no git directory is detected, then the value is an empty string. | `1626881847` |
+| `EARTHLY_VERSION` | The version of Earthly currently running. | `v0.6.2`, `dev-main` |
```
alexcb, line 39: Since this new feature won't be available until it is released, we have to be careful about when we update docs.

Any new docs for unreleased code needs to be submitted in a seperate PR against the `next` branch: https://github.com/earthly/earthly/tree/next

Then when we perform the release we merge next into main, which causes the docs to get updated.

Sorry I forgot to mention this when we were pairing on the issue.

We might also want to change the example to only list `v0.6.2` and omit `dev-main`, since most users won't see that.

vladaionescu (commented):
Awesome - this looks good!

One thought... wonder if we should have version flags for all language additions from now on. This way the user would not be surprised if Earthly `v0.6.2` behaves differently from Earthly `v0.6.3` (e.g. different people running the same Earthfile but getting different results unexpectedly). Thoughts @alexcb?

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to [Earthly](https://github.com/earthly/earthly) will be doc
 
 ## Unreleased
 
+### Added
+
+- Earthly now provides the following [builtin ARGs](https://docs.earthly.dev/docs/earthfile/builtin-args): `EARTHLY_VERSION` and `EARTHLY_BUILD_SHA` [#1452](https://github.com/earthly/earthly/issues/1452).
```
alexcb, line 9: very nice :tada: (I always forget to update this)

builder/builder.go
```diff
@@ -88,6 +88,8 @@ type BuildOpt struct {
 	OnlyArtifact               *domain.Artifact
 	OnlyArtifactDestPath       string
 	EnableGatewayClientLogging bool
+	EarthlyVersion             string
```
alexcb, line 91: What if we combine these two fields into `BuiltinArgs variables.DefaultArgs`? That way we would be consistent with how we define them in the `BuildOpt` and `ConvertOpt` structs.
```suggestion
	BuiltinArgs                variables.DefaultArgs
```

tests/builtin-args.earth
```diff
@@ -16,3 +18,5 @@ builtin-args-test:
     RUN test -z "$EARTHLY_TARGET_TAG"
     RUN test "$EARTHLY_SOURCE_DATE_EPOCH" = "0"
     RUN test -z "$EARTHLY_GIT_COMMIT_TIMESTAMP"
+    RUN test "$EARTHLY_VERSION" != ""
```
alexcb, line 21: I think our code base tends to use `-n` more frequently:
```suggestion
    RUN test -n "$EARTHLY_VERSION"
```

tests/builtin-args.earth
```diff
@@ -16,3 +18,5 @@ builtin-args-test:
...
+    RUN test "$EARTHLY_BUILD_SHA" != ""
```
alexcb, line 22:
```suggestion
    RUN test -n "$EARTHLY_BUILD_SHA"
```

alexcb:
> One thought... wonder if we should have version flags for all language additions from now on. This way the user would not be surprised if Earthly `v0.6.2` behaves differently from Earthly `v0.6.3` (e.g. different people running the same Earthfile but getting different results unexpectedly). Thoughts @alexcb?

That's a real good point. So much for this being an easy PR :sweat_smile: 

We'll need to add a new entry to https://github.com/earthly/earthly/blob/main/features/features.go#L25

```
```

Then you'll need to pass the features struct along:

```
	vc := variables.NewCollection(opt.Console,
		target, llbutil.PlatformWithDefault(opt.Platform), bc.GitMetadata, opt.OverridingVars,
		opt.GlobalImports, bc.features)
```

Then check the feature flag before including them:

```
if features.EarthlyVersionArg {
	ret.AddInactive("EARTHLY_VERSION", defaultArgs.EarthlyVersion)
	ret.AddInactive("EARTHLY_BUILD_SHA", defaultArgs.EarthlyBuildSha)
}
```

and then for testing, you'll need to update `.earthly_version_flag_overrides` to contain:

```
referenced-save-only,use-copy-include-patterns,earthly-version-arg
```

docs/earthfile/builtin-args.md
brandonSc: Makes sense!

brandonSc: Moved to new branch: https://github.com/earthly/earthly/tree/brandon/docs-builtin-args
Will PR that docs branch once we're settled on the changes here

builder/builder.go
brandonSc: Thanks, added this

features/features.go
```diff
@@ -159,13 +160,23 @@ func GetFeatures(version *spec.Version) (*Features, error) {
...
-	if (ftrs.Major == 0 && ftrs.Minor >= 6) || ftrs.Major > 1 { // 0.6+
+	switch {
+	case versionAtLeast(ftrs, 6): // 0.6+
 		ftrs.ReferencedSaveOnly = true
 		ftrs.UseCopyIncludePatterns = true
 		ftrs.ForIn = true
 		ftrs.RequireForceForUnsafeSaves = true
 		ftrs.NoImplicitIgnore = true
+	case versionAtLeast(ftrs, 7): // 0.7+
+		ftrs.EarthlyVersionArg = true
```
brandonSc, line 171: Does it make sense to add this preemptively now?

variables/collection.go
```diff
@@ -42,9 +43,10 @@ type Collection struct {
 }
 
 // NewCollection creates a new Collection to be used in the context of a target.
-func NewCollection(console conslogging.ConsoleLogger, target domain.Target, platform specs.Platform, gitMeta *gitutil.GitMetadata, overridingVars *Scope, globalImports map[string]domain.ImportTrackerVal) *Collection {
+func NewCollection(console conslogging.ConsoleLogger, target domain.Target, platform specs.Platform, gitMeta *gitutil.GitMetadata,
+	defaultArgs DefaultArgs, overridingVars *Scope, globalImports map[string]domain.ImportTrackerVal, ftrs *features.Features) *Collection {
```
brandonSc, line 47: This function signature is getting quite long.

Thoughts on budling them into a struct?

```golang
type NewCollectionParams struct {
    console conslogging.ConsoleLogger
    target    domain.Target
    // etc
}

func NewCollection(params NewCollectionParams) *Collection { 
    // ...
}

File: docs/earthfile/builtin-args.md (hunk 1 above)
Comment (brandonSc) [reply docs/earthfile/builtin-args.md]: Done! https://github.com/earthly/earthly/tree/brandon/docs-builtin-args

File: features/features.go (hunk 7 above)
Comment (vladaionescu) [reply features/features.go:171]: Yeah.. it's probably a good idea.

File: variables/collection.go (hunk 8 above)
Comment (vladaionescu) [reply variables/collection.go:47]: Sounds good to me. We've used the term `opts` or `options` instead of `params` in other places in the code, but the idea is the same.

File: CHANGELOG.md (hunk 3 above)
Comment (vladaionescu) [reply CHANGELOG.md:9]: This might need to be updated to say that this needs to be enabled with a certain feature flag.

File: features/features.go (hunk 7 above)
Comment (vladaionescu) [reply features/features.go:171]: Using `VERSION 0.7` now will fail in `earthly` v0.6, right @alexcb? (It's a good idea if it fails, so the script doesn't behave differently if ran with an Earthly v0.6.* vs v0.7.*)

File: features/features.go [hunk 9]
```diff
@@ -159,13 +160,23 @@ func GetFeatures(version *spec.Version) (*Features, error) {
...
 	}
 
 	return &ftrs, nil
 }
+
+// versionAtLeast returns true if the version configured in `ftrs` are greater than or equal to
+// the provided `minorVersion`, where `minorVersion` is the middle digit in a version string (e.g. '6' in 'v0.6.0').
+// This function probably needs to be expanded if our version checks become more complicated in future releases of Earthly.
+func versionAtLeast(ftrs Features, minorVersion int) bool {
+	return (ftrs.Major == 0 && ftrs.Minor >= minorVersion) || ftrs.Major > 1
```
vladaionescu, line 181:
```suggestion
	return (ftrs.Major == 0 && ftrs.Minor >= minorVersion) || ftrs.Major >= 1
```

vladaionescu (approved):
Looks good :+1:

A few minor things...

features/features.go
alexcb:
```
VERSION 0.7
FROM alpine
RUN echo the future!
```

yup!

```
Error: build target: build main: bkClient.Build: failed to solve: resolve build context for target +base: 1 validation issues.
- Earthfile version is invalid, supported versions are [0.5 0.6]
```

features/features.go
```diff
@@ -159,13 +160,23 @@ func GetFeatures(version *spec.Version) (*Features, error) {
...
+		ftrs.EarthlyVersionArg = true
 	}
 
 	return &ftrs, nil
 }
+
+// versionAtLeast returns true if the version configured in `ftrs` are greater than or equal to
+// the provided `minorVersion`, where `minorVersion` is the middle digit in a version string (e.g. '6' in 'v0.6.0').
+// This function probably needs to be expanded if our version checks become more complicated in future releases of Earthly.
+func versionAtLeast(ftrs Features, minorVersion int) bool {
```
alexcb, line 180: `1.0` is a long ways away, but it still might be good to pass in a major version here:
```suggestion
func versionAtLeast(ftrs Features, majorVersion, minorVersion int) bool {
```

Then we can do:

```
case versionAtLeast(ftrs, 0, 6):
```

rather than 

```
case versionAtLeast(ftrs, 6): // 0.6+
```

which resolves the need to comment it's actually `0.6` and not just `6`.

brandonSc: We could also just use a library like https://github.com/hashicorp/go-version if we want more complete version comparison, although I think it's not worth the added dependency for our simple use case right now

alexcb (approved):

PR #1482, opened 2021-12-06
support mix of sshd algorithms

alexcb:
This fixes a bug where the ssh-rsa known_host entry was prioritized and results in errors when an sshd host returns both an `ssh-rsa` and `ecdsa-sha2-nistp256` key (or smaller keys such as `ssh-ed25519`), but the user only has an entry in known_hosts for ssh-ed25519.

This fix corrects the work-around that was added in 1d644dad1430041117b8e030e79fc1ccffd5a9b7

Additionally adds a new config option to disable `known_host` checking for specific git hosts by setting `insecure_allow_unknown_hosts` to `true` under the `git` section of `earthly/config.yml`.


tests/git-ssh-server/setup/Earthfile
```diff
@@ -0,0 +1,90 @@
+VERSION 0.6
```
alexcb, line 1: this file was moved from `tests/git-ssh-server/Earthfile` to `tests/git-ssh-server/setup/Earthfile`

tests/Earthfile
```diff
@@ -195,30 +199,6 @@ cache-test:
 git-clone-test:
     DO +RUN_EARTHLY --earthfile=git-clone.earth --target=+test
 
-git-clone-private-ssh:
```
alexcb, line 198: moved to `./git-ssh-server/Earthfile`

Earthfile
```diff
@@ -1,3 +1,5 @@
+VERSION 0.6
```
vladaionescu, line 1: Yay! We should probably do this for all our Earthfiles. But that could be a separate PR

alexcb: and converting `--build-args` too :tada:

buildcontext/git.go
```diff
@@ -153,9 +153,10 @@ func (gr *gitResolver) resolveGitProject(ctx context.Context, gwClient gwclient.
 			llb.WithCustomNamef("[internal] GIT CLONE %s", stringutil.ScrubCredentials(gitURL)),
 			llb.KeepGitDir(),
 		}
-		if keyScan != "" {
-			gitOpts = append(gitOpts, llb.KnownSSHHosts(keyScan))
+		if len(keyScans) > 0 {
+			gitOpts = append(gitOpts, llb.KnownSSHHosts(strings.Join(keyScans, "\n")))
```
andrewsykim, line 157: Maybe a naive question, but I didn't get the impression that llb.KnownSSHHosts receives multiple keys separated by new lines, is that documented somewhere else? Or were you assuming it accepts any content compatible with `~/.ssh/authorized_keys?

config/config.go
```diff
@@ -91,13 +91,14 @@ type GitConfig struct {
...
+	Suffix            string `yaml:"suffix"                       help:"The git repository suffix, like .git."`                                       // .git
+	Auth              string `yaml:"auth"                         help:"What authentication method do you use? Valid options are: http, https, ssh."` // http, https, ssh
+	User              string `yaml:"user"                         help:"The username to use when auth is set to git or https."`
+	Password          string `yaml:"password"                     help:"The https password to use when auth is set to https. This setting is ignored when auth is ssh."`
+	KeyScan           string `yaml:"serverkey"                    help:"SSH fingerprints, like you would add in your known hosts file, or get from ssh-keyscan."`
+	AllowEmptyKeyScan bool   `yaml:"insecure_allow_unknown_hosts" help:"Allow ssh access to hosts with unknown server keys (e.g. no entries in known_hosts)."`
```
andrewsykim, line 101: Should the Go field should match the yaml field name?

buildcontext/gitlookup.go
```diff
@@ -226,24 +369,25 @@ func (gl *GitLookup) detectProtocol(host string) (protocol gitProtocol, err erro
 		err = nil
 		return
 	}
+
+	algs, keys, err := gl.getHostKeyAlgorithms(host)
+	if err != nil {
+		gl.console.VerbosePrintf("failed to get accepted host key algorithms for %s: %s", host, err.Error())
```
andrewsykim, line 375: Should there be a return here?

buildcontext/git.go
alexcb: it passes it along to buildkit which writes it out to a temp file which is then passed to `git` when cloning: https://github.com/moby/buildkit/blob/a0afb6902ad2014f702f196ae2e597c1768a763b/source/git/gitsource.go#L283

perhaps the original author of the known_hosts implementation in buildkit could have done a better job documenting that: https://github.com/moby/buildkit/commit/f2c8eb177c83f1381e69966bdd0312dd9d86e1ff :laughing:

config/config.go
alexcb: possibly? do you think we should include the `Insecure` prefix? i.e. `InsecureAllowUnknownHosts` or just `AllowUnknownHosts`?

I don't have any strong preference of renaming.

andrewsykim: Not sure what the best name is to be honest, but the name of the Go field (AllowEmptyKeyScan) and the yaml field (insecure_allow_unknown_hosts) should be consistent at least IMO. I guess it would be useful to prefix with "Insecure" if we really thought it was and people shouldn't use it :thinking:

buildcontext/git.go
andrewsykim: LOLL -- thanks good to know :)

config/config.go
alexcb: I renamed it to `InsecureAllowUnknownHosts`, and additionally changed `KeyScan` to `ServerKey`.

buildcontext/gitlookup.go
alexcb: great catch! :raised_hands:

buildcontext/gitlookup.go
```diff
@@ -25,14 +28,14 @@ import (
...
-	password string
-	keyScan  string
+	name              string
+	re                *regexp.Regexp
+	sub               string
+	user              string
+	suffix            string
+	protocol          gitProtocol
+	password          string
+	allowEmptyKeyScan bool
```
andrewsykim, line 38: nit: insecureAllowUnknownHosts

buildcontext/gitlookup.go
```diff
@@ -116,8 +107,20 @@ func (gl *GitLookup) DisableSSH() {
...
+		if s != "" && !strings.HasPrefix(s, "#") {
+			keyScans = append(keyScans, s)
+		}
+	}
+	return keyScans
+}
+
 // AddMatcher adds a new matcher for looking up git repos
-func (gl *GitLookup) AddMatcher(name, pattern, sub, user, password, suffix, protocol, keyScan string) error {
+func (gl *GitLookup) AddMatcher(name, pattern, sub, user, password, suffix, protocol, knownHosts string, allowEmptyKeyScan bool) error {
```
andrewsykim, line 123: nit: insecureAllowUnknownHosts

buildcontext/gitlookup.go
```diff
@@ -137,49 +140,189 @@ func (gl *GitLookup) AddMatcher(name, pattern, sub, user, password, suffix, prot
...
-		protocol: p,
-		keyScan:  keyScan,
+		name:              name,
+		re:                re,
+		sub:               sub,
+		user:              user,
+		password:          password,
+		suffix:            suffix,
+		protocol:          p,
+		allowEmptyKeyScan: allowEmptyKeyScan,
```
andrewsykim, line 150: nit: insecureAllowUnknownHosts

tests/git-ssh-server/Earthfile
```diff
@@ -1,67 +1,354 @@
...
+ssh-add -l
+
+# first check that known_hosts are hashed (hashed entries start with |1|)
+cat ~/.ssh/known_hosts | awk '{print \$1}' | grep '|1|'
+
+# convert hashed hostnames into unhashed hostnames
+cp ~/.ssh/known_hosts /tmp/known_hosts.old
+cat /tmp/known_hosts.old | awk '{printf \"git.example.com %s %s\\n\", \$2, \$3}' > ~/.ssh/known_hosts
+
+echo here we go 1
```
andrewsykim, line 269: guessing this is debugging logs we can remove now?

tests/git-ssh-server/Earthfile
```diff
@@ -1,67 +1,354 @@
...
+# first ensure these two /etc/hosts entries are working
+ping -c 1 git.example.com
+ping -c 1 buildkitsandbox
+
+# load in preauthorized key
+eval \$(ssh-agent)
+ssh-add /root/self-hosted-rsa-key
+ssh-add -l
+
+earthly --config \$earthly_config --verbose -D +test
```
andrewsykim, line 30: What's the reason for having a local +test target? Is it testing ssh behavior still?

andrewsykim: Ah I see the GIT CLONE in git-clone-private-ssh.earth now

andrewsykim (approved):

buildcontext/gitlookup.go
alexcb: I'm hesitant to extend these changes here, as it makes the rest of the code rather verbose to read, for example:

```
		if len(keyScans) == 0 && !m.insecureAllowEmptyKeyScan {
			return "", nil, errors.Errorf("no known_hosts entries exist for %s", host)
		}
```

which I then read as "if not insecure" which reads like a double-negative.

setting `allowEmptyKeyScan` is essentially ssh's `StrictHostKeyChecking` setting. Maybe it's better to just use the later as the variable name?

Then this becomes:

```
		if len(keyScans) == 0 && m.StrictHostKeyChecking {
			return "", nil, errors.Errorf("no known_hosts entries exist for %s", host)
		}
```

alexcb: I made the above change, I think it's easier to explain to our users what is actually happening (better to re-use existing terminology).

andrewsykim: Nice, I like StrictHostKeyChecking better too :+1:

Issue #1449, opened 2021-11-26
earthly config key "" panics


alexcb:
running:
```
earthly config global.conversion_parallelism ''
panic: runtime error: index out of range [0] with length 0
```
fails with panic.

It's also not obvious how config options can be deleted.

alexcb:
To expand on this:

If my config is initially:
```
$ cat ~/.earthly/config.yml
global:
    buildkit_additional_args:
        - '-e'
        - 'BUILDKIT_SCHEDULER_DEBUG=1'
```

Then i call `earthly config global.conversion_parallelism 5`, it turns into:

```
$ cat ~/.earthly/config.yml
global:
    buildkit_additional_args:
        - '-e'
        - 'BUILDKIT_SCHEDULER_DEBUG=1'
    conversion_parallelism: 5
```

How can I delete it? Let's try running: `earthly config global.conversion_parallelism ""`

that produces:

```
panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
github.com/earthly/earthly/config.valueToYaml({0x7ffe6611dbc5, 0x0})
	/earthly/config/config.go:266 +0x125
github.com/earthly/earthly/config.UpsertConfig({0xc00039e000, 0x1e, 0x200}, {0x7ffe6611dba7, 0x1d}, {0x7ffe6611dbc5, 0xf23440})
	/earthly/config/config.go:180 +0x3bb
main.(*earthlyApp).actionConfig(0xc0000d7200, 0xc000110300)
	/earthly/cmd/earthly/main.go:2576 +0x1e5
github.com/urfave/cli/v2.(*Command).Run(0xc00014f320, 0xc00003cf40)
	/go/pkg/mod/github.com/urfave/cli/v2@v2.3.0/command.go:163 +0x64a
github.com/urfave/cli/v2.(*App).RunContext(0xc000583040, {0x11f77a8, 0xc0002e0a80}, {0xc000110040, 0x4, 0x4})
	/go/pkg/mod/github.com/urfave/cli/v2@v2.3.0/app.go:313 +0x81e
main.(*earthlyApp).run(0xc0000d7200, {0x11f77a8, 0xc0002e0a80}, {0xc000110040, 0x4, 0x4})
	/earthly/cmd/earthly/main.go:1513 +0xad
main.main()
	/earthly/cmd/earthly/main.go:253 +0x538
```

I would have expected the above to cleanly detect the error such as what happens when you pass it a non-int:

```
[...]

PR #1490, opened 2021-12-09
Support --delete flag with earthly config

brandonSc:
Fixes https://github.com/earthly/earthly/issues/1449

Examples:
```bash
# Starting with empty config
❯ earthly config global.disable_analytics true
❯ cat ~/.earthly/config.yml
global:
    disable_analytics: true

❯ earthly config global.conversion_parallelism 5
❯ cat ~/.earthly/config.yml
global:
    disable_analytics: true
    conversion_parallelism: 5

# Fixed panic when providing empty-string in this case
❯ earthly config global.conversion_parallelism ""
Error: upsert config: cannot set global.conversion_parallelism to , as the types are incompatible

# In case empty-string is actual desired value for string-type configs
❯ earthly config global.tlscert ""
❯ cat ~/.earthly/config.yml
global:
    disable_analytics: true
    conversion_parallelism: 5
    tlscert: ""

# Removing config option
❯ earthly config global.tlscert --delete
❯ cat ~/.earthly/config.yml
global:
    disable_analytics: true
    conversion_parallelism: 5

# Removing a config which doesn't exist is a no-op
❯ earthly config global.cache_size_mb --delete
❯ cat ~/.earthly/config.yml
global:
    disable_analytics: true
    conversion_parallelism: 5
```


cmd/earthly/main.go
```diff
@@ -2573,9 +2573,24 @@ func (app *earthlyApp) actionConfig(c *cli.Context) error {
 		return errors.Wrap(err, "read config")
 	}
 
-	outConfig, err := config.UpsertConfig(inConfig, args[0], args[1])
-	if err != nil {
-		return errors.Wrap(err, "upsert config")
+	var outConfig []byte
+
+	switch {
+	case args[1] == "--delete":
```
brandonSc, line 2579: Tbh I think providing the flags as the first argument like `earthly config --delete global.conversion_parallelism` is a bit more natural, but the `--help` tag already worked this way.

Would be easy to support both ways (and maybe deprecate the old way) if we want?

cmd/earthly/main.go
```diff
@@ -2573,9 +2573,24 @@ func (app *earthlyApp) actionConfig(c *cli.Context) error {
 		return errors.Wrap(err, "read config")
 	}
 
-	outConfig, err := config.UpsertConfig(inConfig, args[0], args[1])
-	if err != nil {
-		return errors.Wrap(err, "upsert config")
+	var outConfig []byte
+
+	switch {
```
vladaionescu, line 2578: Could use `switch args[1]` to simplify this.

cmd/earthly/main.go
alexcb: I think moving it to https://github.com/earthly/earthly/blob/main/cmd/earthly/main.go#L970-L974 is the right call.

The only reason we support `--help` as a command afterwards is to prevent people from accidentally writting the string `"--help"` as a config value.

As an extra ask, can you extend `case args[1] == "--help":` to include `-h`.

vladaionescu (commented):
I wonder if we should have some simple unit tests for config manipulation.

cmd/earthly/main.go
brandonSc: > As an extra ask, can you extend case args[1] == "--help": to include -h.

Good call, that's also in the help printout but currently doesn't work on main either.

What do we think about supporting both positions for --help or --delete?

alexcb (commented):
looks great!

can we add a new test to cover `earthly config ...`. We currently use it in some tests such as https://github.com/earthly/earthly/blob/main/tests/scrub-https-credentials/Earthfile#L58-L59

Maybe we can test it in such a way where we set, update, delete items then check the contents of the actual `~/.earthly/config.yml` (or `\$earthly_config` in the case of the linked test).

edit/update: here's another two cases where we do something similar: https://github.com/earthly/earthly/blob/main/tests/private-https/Earthfile#L52-L76

brandonSc:
> I wonder if we should have some simple unit tests for config manipulation.

I was thinking the same. Was picturing a target in the Earthfile that used a freshly compiled earthly to run a bash script that ran and tested config commands directly. Could also of course also do unit tests in Go that call the config functions.

What do you think?

cmd/earthly/main.go
alexcb: I have no strong opinion, supporting both does lead to more code, but perhaps a better user experience. Your call.

brandonSc: Supporting both wasn't as easy as it seemed so I figured not worth the effort

config/config.go
```diff
@@ -264,7 +297,21 @@ func valueToYaml(value string) (*yaml.Node, error) {
 	}
 	fixStyling(valueNode)
 
-	return valueNode.Content[0], nil
+	contentNode := &yaml.Node{}
+	switch {
```
alexcb, line 301: nit: given that we aren't using the switch with a single value, I think using a if/elseif/else is more readable.

config/config.go
```diff
@@ -346,6 +393,47 @@ func setYamlValue(node *yaml.Node, path []string, value *yaml.Node) []string {
...
+	case yaml.DocumentNode:
+		for _, c := range node.Content {
+			path = deleteYamlValue(c, path)
+		}
+
+	case yaml.MappingNode:
+		for i := 0; i < len(node.Content); i += 2 {
+			// Keys/Values are inline. Count by twos to get it right.
+			key := node.Content[i]
+			val := node.Content[i+1]
```
alexcb, line 408: what if the input is malformed and contains an odd number of elements? Would we index out of bounds in such a case?

tests/config/Earthfile
```diff
@@ -0,0 +1,35 @@
+VERSION 0.6
+
+FROM ..+base
+
+test:
+    COPY expected-*.yml .
+    RUN touch config.yml
```
alexcb, line 7: I wonder if we should include a

```
RUN earthly --config config.yml config global.disable_analytics true 2>&1 | grep 'failed to read from config.yml: open config.yml: no such file or directory'
```

to ensure earthly doesn't create overridden config files when they don't exist (which makes it more obvious when someone overrides a config file but typos it and points the the wrong location).

tests/config/Earthfile
```diff
@@ -0,0 +1,35 @@
...
+    RUN earthly --config config.yml config global.conversion_parallelism 5
+    RUN test "$(cat config.yml)" = "$(cat expected-4.yml)"
+
+    # --delete should remove a config
+    RUN earthly --config config.yml config global.conversion_parallelism --delete
+    RUN test "$(cat config.yml)" = "$(cat expected-5.yml)"
+
+    # --help and -h should succeed
+    RUN earthly config global.conversion_parallelism --help
+    RUN earthly config global.conversion_parallelism -h
```
alexcb, line 28: might be preduent to check `-h` doesn't change the config file, perhaps add the same check below:

```
RUN test "$(cat config.yml)" = "$(cat expected-5.yml)"
```

tests/config/Earthfile
```diff
@@ -0,0 +1,35 @@
...
+
+    # Edge cases
+    RUN earthly --config config.yml config global.conversion_parallelism oops; test $? = 1
+    RUN earthly --config config.yml config global.conversion_parallelism ""; test $? = 1
+    RUN earthly --config config.yml config global.buildkit_image ""
+
+
```
alexcb, line 35: nit: left-over newlines at end of file (we have a linter that should catch this)

alexcb (commented):
great work, just a couple minor suggestions.

config/config.go
brandonSc: Since they don't require a single value like other languages, I've personally found switch statements in Go to be more readable than a series of if/else-if/else, and have tended to use them liberally.

Using switch:
```golang
	switch {
	case args[1] == "--delete":
		outConfig, err = config.Delete(inConfig, args[0])
		if err != nil {
			return errors.Wrap(err, "delete config")
		}
	case args[1] == "--help" || args[1] == "-h":
		if err = config.PrintHelp(args[0]); err != nil {
			return errors.Wrap(err, "help")
		}
		return nil // exit now without writing any changes to config
	default: // args are key/value pairs, e.g. ["global.conversion_parallelism","5"]
		outConfig, err = config.Upsert(inConfig, args[0], args[1])
		if err != nil {
			return errors.Wrap(err, "upsert config")
		}
	}
```
Using if:
```golang
	if args[1] == "--delete" {
		outConfig, err = config.Delete(inConfig, args[0])
		if err != nil {
			return errors.Wrap(err, "delete config")
		}
	} else if args[1] == "--help" || args[1] == "-h" {
		if err = config.PrintHelp(args[0]); err != nil {
			return errors.Wrap(err, "help")
		}
		return nil // exit now without writing any changes to config
	} else { // args are key/value pairs, e.g. ["global.conversion_parallelism","5"]
		outConfig, err = config.Upsert(inConfig, args[0], args[1])
		if err != nil {
			return errors.Wrap(err, "upsert config")
		}
	}
```

Maybe it's just personal taste but I find the switch statement looks a bit cleaner. Happy to change it though if we have that preference for using if's in these cases throughout the project already though

config/config.go
brandonSc: Good question, I copied the logic from `setYamlValue` above but will investigate further

config/config.go
alexcb: I like how it was used in 
```
	// Enable version-specific features.
	switch {
	case versionAtLeast(ftrs, 0, 6):
		ftrs.ReferencedSaveOnly = true
		ftrs.UseCopyIncludePatterns = true
		ftrs.ForIn = true
		ftrs.RequireForceForUnsafeSaves = true
		ftrs.NoImplicitIgnore = true
	case versionAtLeast(ftrs, 0, 7):
		ftrs.EarthlyVersionArg = true
	}
```

and in the example you shared, it can further be reduced into `switch args[1] {` (which then switches work great for).

In both cases the switch is centred around different cases for a single condition (major/minor feature, or arg).

I think the part that jumped out at me is that `valueNode.Content` and `value` aren't as tightly connected to a single variable, or even variable type.

For this discussion, it might be more useful to include the code that I was commenting on:

```
	switch {
	case len(valueNode.Content) > 0:
		// ContentNode contains the user-provided value with it's type etc
		contentNode = valueNode.Content[0]
	case value == "":
		// Edge case where the yaml.Unmarshal above results in no nodes in valueNode.Content.
		// The code below ensures we can write an actual empty string to our yaml as requested.
		contentNode.SetString("")
	default:
		// Very unlikely
		return nil, errors.New("failed setting value in yaml")
	}
```

vs

```
	if len(valueNode.Content) > 0 {
		// ContentNode contains the user-provided value with it's type etc
		contentNode = valueNode.Content[0]
	} else if value == "" {
		// Edge case where the yaml.Unmarshal above results in no nodes in valueNode.Content.
		// The code below ensures we can write an actual empty string to our yaml as requested.
		contentNode.SetString("")
	} else {
		// Very unlikely
		return nil, errors.New("failed setting value in yaml")
	}
```

Maybe @andrewsykim or @vladaionescu have thoughts on adopting this coding style?

config/config.go
brandonSc: Even if I have no value (nil) in my yaml it seems to work

config:
```
global:
    conversion_parallelism:
```
then
```
earthly config global.conversion_parallelism --delete
```
runs cleanly. printed the values of `[i]` and `[i+1]` when we get to that spot in the yaml
```
key: &{Kind:8 Style:0 Tag:!!str Value:conversion_parallelism Anchor: Alias:<nil> Content:[] HeadComment: LineComment: FootComment: Line:2 Column:5}
val: &{Kind:8 Style:0 Tag:!!null Value: Anchor: Alias:<nil> Content:[] HeadComment: LineComment: FootComment: Line:2 Column:28}
```
and even if there's nothing there
```
key: &{Kind:8 Style:0 Tag:!!str Value:global Anchor: Alias:<nil> Content:[] HeadComment: LineComment: FootComment: Line:1 Column:1}
val: &{Kind:4 Style:32 Tag:!!map Value: Anchor: Alias:<nil> Content:[] HeadComment: LineComment: FootComment: Line:1 Column:9}
```
Seems to always have a key and val with at least nil within the yaml node's Content array.

What do you think @alexcb can you think of a way it might not be the case?

config/config.go
vladaionescu: It's a nit, so any option is fine TBH. In any case, I think I like if/elseif/else when the conditions are on different variables and switch when the conditions are about the same thing.

brandonSc: Yeah I totally pasted the wrong example 😅 

The example I did paste though might be worse as `switch args[1]` as we may have to duplicate the code for the --help case. e.g.
```golang
switch args[1]:
case "-h"
 // help
case "--help"
 // help again
case "--delete"
```
So I think I would prefer keeping as switch without arg or just use if/else.

Anyway, will patch both to use ifs instead

vladaionescu: You can still use switch for that https://www.geeksforgeeks.org/golang-program-that-uses-switch-multiple-value-cases/

config/config.go
alexcb: that's good to know, it was just something that jumped out, but if our tests check for that case and they pass, they seem like the best source of truth :+1:

cmd/earthly/main.go
```diff
@@ -2575,18 +2575,21 @@ func (app *earthlyApp) actionConfig(c *cli.Context) error {
...
-	case args[1] == "--delete":
-		outConfig, err = config.Delete(inConfig, args[0])
-		if err != nil {
-			return errors.Wrap(err, "delete config")
-		}
-	case args[1] == "--help" || args[1] == "-h":
+	switch args[1] {
+	case "-h":
+		fallthrough
+	case "--help":
```
alexcb, line 2581: you can combine these into a single line:
```suggestion
	case "--help", "-h":
```

brandonSc: 💯

alexcb (approved):
:tada:

Issue #1491, opened 2021-12-09
Some config commands shown in --help don't work


brandonSc:
Running `earthly config --help` prints the following summary:
```
NAME:
   earthly config - Edits your Earthly configuration file

USAGE:
   This command takes a path, and a value and sets it in your configuration file.

   As the configuration file is YAML, the key must be a valid key within the file. You can specify sub-keys by using "." to separate levels.
   If the sub-key you wish to use has a "." in it, you can quote that subsection, like this: git."github.com".

   Values must be valid YAML, and also be deserializable into the key you wish to assign them to.
   This means you can set higher level objects using a compact style, or single values with simple values.

   Only one key/value can be set per invocation.

   To get help with a specific key, do "config [key] --help". Or, visit https://docs.earthly.dev/earthly-config for more details.

DESCRIPTION:
   Set your cache size:

     config global.cache_size_mb 1234

   Set additional buildkit args, using a YAML array:

     config global.buildkit_additional_args ['userns', '--host']

   Set a key containing a period:

     config git."example.com".password hunter2

   Set up a whole custom git repository for a server called example.com, using a single-line YAML literal:
     * which stores git repos under /var/git/repos/name-of-repo.git
     * allows access over ssh
     * using port 2222
     * sets the username to git
     * is recognized to earthly as example.com/name-of-repo

     config git "{example: {pattern: 'example.com/([^/]+)', substitute: 'ssh://git@example.com:2222/var/git/repos/\$1.git', auth: ssh}}"

OPTIONS:
   --dry-run   Print the changed config file to the console instead of writing it out (default: false)
   --help, -h  show help (default: false)
```

However, some of the examples shown in this summary do not work for me.

For example:
```bash
❯ earthly config global.buildkit_additional_args ['userns', '--host']
zsh: bad pattern: [userns,

❯ earthly config git."example.com".password hunter2
Error: upsert config: path is not valid: no path for com.password
```

Perhaps the commands are not correct, or they should work and there may be a bug.

PR #1495, opened 2021-12-09
Update help doc for earthly config

brandonSc:
Fixes: https://github.com/earthly/earthly/issues/1491

### DIFF
--- a/cmd/earthly/main.go
+++ b/cmd/earthly/main.go
@@ -951,11 +951,11 @@ func newEarthlyApp(ctx context.Context, console conslogging.ConsoleLogger) *eart
 
 Set additional buildkit args, using a YAML array:
 
-	config global.buildkit_additional_args ['userns', '--host']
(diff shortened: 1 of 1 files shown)


cmd/earthly/main.go
```diff
@@ -951,11 +951,11 @@ func newEarthlyApp(ctx context.Context, console conslogging.ConsoleLogger) *eart
 
 Set additional buildkit args, using a YAML array:
 
-	config global.buildkit_additional_args ['userns', '--host']
+	config global.buildkit_additional_args "['userns', '--host']"
```
alexcb, line 954: nit: how about we use single quotes for the outer yaml string, then use double quotes for the inner parts, that way it's also valid json.
```suggestion
	config global.buildkit_additional_args '["userns", "--host"]'
```

alexcb (approved):

PR #1496, opened 2021-12-09
Error out if interactive and buildkit-host flags

alexcb:
Currently --interactive does not work with a remote buildkit instance,
until this is implemented, we should display a more clear error message.

### DIFF
--- a/cmd/earthly/main.go
+++ b/cmd/earthly/main.go
@@ -2606,9 +2606,15 @@ func (app *earthlyApp) actionBuild(c *cli.Context) error {
 			return errors.New("unable to use --ci flag in combination with --interactive flag")
 		}
 	}
-	if !termutil.IsTTY() && app.interactiveDebugging {
-		return errors.New("A tty-terminal must be present in order to the --interactive flag")
+	if app.interactiveDebugging {
+		if !termutil.IsTTY() {
+			return errors.New("A tty-terminal must be present in order to the --interactive flag")
+		}
+		if !buildkitd.IsLocal(app.buildkitHost) {
+			return errors.New("the --interactive flag is not currently supported with non-local buildkit servers")
+		}
 	}
+
 	if app.imageMode && app.artifactMode {
 		return errors.New("both image and artifact modes cannot be active at the same time")
 	}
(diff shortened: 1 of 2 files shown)


cmd/earthly/main.go
```diff
@@ -2606,9 +2606,15 @@ func (app *earthlyApp) actionBuild(c *cli.Context) error {
 			return errors.New("unable to use --ci flag in combination with --interactive flag")
 		}
 	}
-	if !termutil.IsTTY() && app.interactiveDebugging {
-		return errors.New("A tty-terminal must be present in order to the --interactive flag")
+	if app.interactiveDebugging {
+		if !termutil.IsTTY() {
+			return errors.New("A tty-terminal must be present in order to the --interactive flag")
+		}
+		if c.IsSet("buildkit-host") {
```
andrewsykim, line 2613: Should this also apply if buildkit-host points to localhost?

alexcb: technically `--interactive` would work in that case, as would `127.0.0.1` or any other local host entry under `/etc/hosts`, but if that's the case I don't understand why someone would want to pass that value in rather than just leaving it empty.

andrewsykim: I generally agree that the use-case for setting buildkit host to localhost seems limited, but just calling out there are some places where we explicitly check if buildkit host is local (via IsLocal function https://github.com/earthly/earthly/blob/2888c9c530d994324c224654a40e895ed378aac5/buildkitd/buildkitd.go#L753). 

Maybe the notable thing is mTLS support locally requires you to set BUILDKIT_HOST to point to localhost, not sure how many people really want mTLS locally though.

andrewsykim: Maybe something like this in the meantime might be fine?
```
if c.IsSet("buildkit-host") && !buildkitd.IsLocal(app.buildkitHost) {
    ...
}
```

alexcb: ah great suggestion, I didn't realise we had a function to check that.

vladaionescu (approved):

PR #1497, opened 2021-12-09
Add some examples using earthly secrets

andrewsykim:
I wasn't able to find a lot of examples of secrets except in our tests, but those are a bit verbose and use earthly-in-earthly. This PR just adds a simple Earthfile that uses Secrets in a very basic way. 

Signed-off-by: Andrew Sy Kim <andrew@earthly.dev>

### DIFF
--- a/examples/secrets/Earthfile
+++ b/examples/secrets/Earthfile
@@ -1,0 +1,23 @@
+FROM alpine:latest
+
+# This target can be called in the following ways:
+# 1. $ earthly --secret MY_SECRET=foobar +secrets
+# 2. $ export MY_SECRET=foobar
+#    $ earthly --secret MY_SECRET +secrets
+# 3. $ export EARTHLY_SECRETS="MY_SECRET=foobar"
+#    $ earthly +secrets
+# 4. $ echo "MY_SECRET=foobar" > .env
+#    $ earthly +secrets
+secrets:
+  ARG EARTHLY_SECRET_PATH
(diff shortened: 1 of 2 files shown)


brandonSc:
When I first read the Earthly docs I found secrets difficult to get my head around, so I think it's great to have this example!

examples/secrets/Earthfile
```diff
@@ -0,0 +1,14 @@
+FROM alpine:latest
+
+# This target can be called in the following ways:
+# 1. $ earthly --secret MY_SECRET=foobar +secrets
+# 2. $ export MY_SECRET=foobar
+#    $ earthly --secret MY_SECRET +secrets
+# 3. $ export EARTHLY_SECRETS="MY_SECRET=foobar"
+#    $ earthly +secrets
```
alexcb, line 8:
```
4. $ echo "MY_SECRET=foobar" > .env
   $ earthly +secrets
````

alexcb (approved):
nice example! typically I would suggest making a reference to this via our root [examples target](https://github.com/earthly/earthly/blob/main/Earthfile#L513) however since this target requires secrets that won't work.

What if we instead do `BUILD ./examples/secrets+base` so we can at least ensure this Earthfile is parsable.

vladaionescu (approved):

examples/secrets/Earthfile
andrewsykim: Good catch -- added this one

PR #1504, opened 2021-12-13
fix plus escaping when copying files

alexcb:
Fixes escaping of plus characters which broke when `--use-copy-include-patterns` was
introduced (and enabled by default in 0.6).


tests/escape-v0.5.earth
```diff
@@ -0,0 +1,69 @@
+VERSION 0.5
```
alexcb, line 1: this file is a direct copy of `escape.earth`, except set to `0.5`.

earthfile2llb/local_state_cache.go
```diff
@@ -72,7 +72,9 @@ func createIncludePatterns(incl []string) []string {
 			inc = "./*"
 		} else if strings.HasSuffix(inc, "/.") {
 			inc = inc[:len(inc)-1] + "*"
+		} else {
```
andrewsykim, line 75: Why was this empty else added?

alexcb: leftover debug that wasn't fully removed, good catch.

tests/escape.earth
```diff
@@ -1,3 +1,4 @@
+VERSION 0.6
```
andrewsykim, line 1: So if I'm understanding correctly, we already tested escaping + but because we default to VERSION 0.5 it wasn't caught? So just bumping this version gives us test coverage for this?

earthfile2llb/local_state_cache_escape.go
```diff
@@ -0,0 +1,50 @@
+package earthfile2llb
+
+import (
+	"unicode/utf8"
+)
+
+// This file is based on regexp.QuoteMeta and helper functions, but has been adapted
```
andrewsykim, line 7: So buildkit also ported some functions from regexp.QuoteMeta?

andrewsykim (approved):

tests/escape.earth
alexcb: that's correct, `0.6` enables `--use-copy-include-patterns` by default. prior to this https://github.com/earthly/earthly/blob/main/tests/copy-verbose.earth was the only test we had providing us coverage (which wasn't ample)

earthfile2llb/local_state_cache_escape.go
alexcb: nope, buildkit passes this copy string to https://pkg.go.dev/path/filepath#Match and doesn't provide any escaping.

Issue #1410, opened 2021-11-15
`FROM DOCKERFILE` Usage Difference Between `some/path/*` and `some/path`


user404d:
I have a very simple use case where I want to slightly alter a Dockerfile before building the container image. I tried using the `FROM DOCKERFILE` command, but ran into some issues with the build context with the following Earthfile:

```
FROM bash:latest
WORKDIR /minecraft

modify:
    GIT CLONE https://github.com/itzg/docker-minecraft-server.git docker-minecraft-server
    # remove volume mount since it can't be removed later
    RUN sed -i 's/VOLUME \["\/data"\]//g' docker-minecraft-server/Dockerfile
    RUN grep -L 'VOLUME \["/data"\]' docker-minecraft-server/Dockerfile
    SAVE ARTIFACT docker-minecraft-server

docker:
    FROM DOCKERFILE --platform linux/amd64 -f +modify/docker-minecraft-server/Dockerfile +modify/docker-minecraft-server
    SAVE IMAGE minecraft-server:latest
```

I expected that the path to the directory which I wanted to use would work, but it seems like the build context isn't transferred correctly. I was able to find issue #615 which has an example that uses path globbing. Simply adding `/*` to the end of the context path finally did the trick.

Side-note: I also tried giving a path to a tarball, but that didn't work either.

I'm using `earthly version v0.5.24 linux/amd64; Debian 9.13`. Let me know what you think.

PR #1506, opened 2021-12-13
Document Issues with FROM Dockerfile

BjoernAkAManf:
Hi!

Issue #1410 is not intuitive for a user. Besides a reference in the documentation that "WITH DOCKER" and "docker build" should not be used together, the documentation currently lacks any direction in allowing the generation of a Dockerfile (e.g. in combination with GIT CLONE and RUN git apply <patch>).

Without the Workaround discussed in the above issues, users would be forced to work around earthly rather than utilize it correctly. Usability wise i'd argue FROM DOCKERFILE +target/repo should automatically imply or at the very least should be functionally identical to FROM DOCKERFILE -f +target/repo/Dockerfile +target/repo/*.

In case you disagree with that interpretation, the documentation should be adapted accordingly.

Thanks

PS: Otherwise been using earthly in some production-ish workloads and it's now been a breeze.

### DIFF
--- a/docs/earthfile/earthfile.md
+++ b/docs/earthfile/earthfile.md
@@ -731,6 +731,8 @@ The `FROM DOCKERFILE` command initializes a new build environment, inheriting fr
 
 The `<context-path>` is the path where the Dockerfile build context exists. By default, it is assumed that a file named `Dockerfile` exists in that directory. The context path can be either a path on the host system, or an [artifact reference](../guides/target-ref.md#artifact-reference), pointing to a directory containing a `Dockerfile`.
 
+Transfer of the build context when using [artifact references](../guides/target-ref.md#artifact-reference) can be achieved by specifying the -f flag manually and appending a /* to the context-path argument. For example `FROM DOCKERFILE -f +target/dir/Dockerfile +target/dir`. Other specifications may currently face issues as documented in [#1410](https://github.com/earthly/earthly/issues/1410).
+
 #### Options
 
 ##### `-f <dockerfile-path>`


alexcb:
Thanks for the suggestion.

What if we expanded this into a larger example we could include under the `##### -f <dockerfile-path>` heading?

Something along the lines of:

```
##### `-f <dockerfile-path>`

Specify an alternative Dockerfile to use. The `<dockerfile-path>` can be either a path on the host system, relative to the current Earthfile, or an [artifact reference](../guides/target-ref.md#artifact-reference) pointing to a Dockerfile.

{% hint style='info' %}
It is possible to split the `Dockerfile` and the build context across two seperate [artifact references](../guides/target-ref.md#artifact-reference):

<backtick><backtick><backtick>Dockerfile #replace with ```Dockerfile
FROM alpine

mybuildcontext:
    WORKDIR /mydata
    RUN echo mydata > myfile
    SAVE ARTIFACT /mydata

mydockerfile:
    RUN echo "
FROM busybox
COPY myfile .
RUN cat myfile" > Dockerfile
    SAVE ARTIFACT Dockerfile

docker:
    FROM DOCKERFILE -f +mydockerfile/Dockerfile +mybuildcontext/mydata/*
    SAVE IMAGE testimg:latest
<backtick><backtick><backtick> #replace with ```

Alternatively, if both the `Dockerfile` and build context are inside the same target, one must reference the same target twice, e.g. `FROM DOCKERFILE -f +target/dir/Dockerfile +target/dir`. Other specifications may currently face issues as documented in [#1410](https://github.com/earthly/earthly/issues/1410).

{% endhint %}
```

(note that I had to include `<backtick>` rather than a single backtick in order to make the formatting in this github comment work)

BjoernAkAManf:
I agree in general. Though i feel it would be beneficial to add more such examples in general. Might also prove beneficiary to document practical styles for each.

While im not at a computer right now, ill update the PR to include your suggestions.

alexcb:
The docs were updated in 224fab95486112cbd6b582e8494cb4dd0803cd6e to include this example.

PR #1509, opened 2021-12-14
fix ssh keyscan warning when port doesnt exist

alexcb:
keyscan warnings such as:

    failed to parse key scan "...": SplitHostPort on "github.com" failed: address github.com: missing port in address

were being incorrectly warned when a missing port (which defaults to 22) is allowed.

### DIFF
--- a/buildcontext/gitlookup.go
+++ b/buildcontext/gitlookup.go
@@ -177,6 +177,16 @@ var errUnsupportedHash = errors.New("unsupported keyscan hash")
 var errInvalidScan = errors.New("invalid keyscan")
 var errKeyScanNoMatch = errors.New("keyscan does not match")
 
+func hasPort(s string) bool {
+	if len(s) == 0 {
+		return false
+	}
+	if s[0] == '[' {
+		return strings.Contains(s, "]:")
+	}
+	return strings.Contains(s, ":")
+}
+
 func isHashedHost(hashAndSalt, hostname string) (bool, error) {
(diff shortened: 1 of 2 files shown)


buildcontext/gitlookup.go
```diff
@@ -177,6 +177,16 @@ var errUnsupportedHash = errors.New("unsupported keyscan hash")
 var errInvalidScan = errors.New("invalid keyscan")
 var errKeyScanNoMatch = errors.New("keyscan does not match")
 
+func hasPort(s string) bool {
```
andrewsykim, line 180: There's nothing in the net package to check this?

andrewsykim, line 180: There's nothing in the net package to check this?

andrewsykim: Maybe SplitHostPort returns a specific error if the port is missing?

alexcb: `SplitHostPort` isn't great; I tried that originally. 

https://cs.opensource.google/go/go/+/refs/tags/go1.17.5:src/net/ipsock.go;l=167 sets `tooManyColons = "too many colons in address"` then creates an error on the fly `return addrErr(hostport, tooManyColons)`. My only other idea was to try `strings.Contains(err.Error, "too many colons in address")`, but that feels more error-prone as that error string might change.

brandonSc (approved):
🎉

PR #1510, opened 2021-12-15
fix: --no-cache cli flag didn't apply to IF

alexcb:
This fixes a bug where IF statements were still being cached when a user
provided the --no-cache flag.

### DIFF
--- a/buildcontext/git.go
+++ b/buildcontext/git.go
@@ -86,7 +86,7 @@ func (gr *gitResolver) resolveEarthProject(ctx context.Context, gwClient gwclien
 		gr.cleanCollection.Add(func() error {
 			return os.RemoveAll(earthfileTmpDir)
 		})
-		gitState, err := llbutil.StateToRef(ctx, gwClient, rgp.state, nil, nil)
+		gitState, err := llbutil.StateToRef(ctx, gwClient, rgp.state, false, nil, nil)
 		if err != nil {
 			return nil, errors.Wrap(err, "state to ref git meta")
 		}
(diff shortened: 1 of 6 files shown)


earthfile2llb/converter.go
```diff
@@ -550,6 +550,9 @@ func (c *Converter) RunExitCode(ctx context.Context, opts ConvertRunOpts) (int,
 			return 0, errors.Wrap(err, "read exit code file")
 		}
 	} else {
+		if c.opt.NoCache {
+			state = state.SetMarshalDefaults(llb.IgnoreCache)
+		}
```
vladaionescu, line 555: I would maybe add a `noCache` parameter to `StateToRef` and pass in `c.opt.NoCache` for that in all these places. That should catch all other similar places that have this issue: `FOR`, dynamic args, `WITH DOCKER` etc.

buildcontext/git.go
```diff
@@ -86,7 +86,7 @@ func (gr *gitResolver) resolveEarthProject(ctx context.Context, gwClient gwclien
 		gr.cleanCollection.Add(func() error {
 			return os.RemoveAll(earthfileTmpDir)
 		})
-		gitState, err := llbutil.StateToRef(ctx, gwClient, rgp.state, nil, nil)
+		gitState, err := llbutil.StateToRef(ctx, gwClient, rgp.state, false, nil, nil)
```
vladaionescu, line 89: Maybe add TODO comment for all the places where we hard-code `noCache`?

vladaionescu (approved):
One more minor remark...

PR #1512, opened 2021-12-16
fix autocompletion of ARGs containing flags

alexcb:
autocompletion for ARGS with flags (e.g. ARG --required FOO)
was incorrectly suggesting

    +target ----required

rather than

    +target --FOO`.

### DIFF
--- a/earthfile2llb/interpreter.go
+++ b/earthfile2llb/interpreter.go
@@ -1027,31 +1027,47 @@ func (i *Interpreter) handleEnv(ctx context.Context, cmd spec.Command) error {
 	return nil
 }
 
+var errInvalidSyntax = errors.New("invalid syntax")
+var errRequiredArgHasDefault = errors.New("required ARG cannot have a default value")
+
+// parseArgArgs parses the ARG command's arguments
+// and returns the argOpts, key, value (or nil if missing), or error
+func parseArgArgs(ctx context.Context, cmd spec.Command) (argOpts, string, *string, error) {
+	opts := argOpts{}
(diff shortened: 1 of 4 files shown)


earthfile2llb/interpreter.go
```diff
@@ -1027,31 +1027,47 @@ func (i *Interpreter) handleEnv(ctx context.Context, cmd spec.Command) error {
...
+	if i.pushOnlyAllowed {
+		return i.pushOnlyErr(cmd.SourceLocation)
+	}
+	opts, key, valueOrNil, err := parseArgArgs(ctx, cmd)
+	if err != nil {
+		return i.wrapError(err, cmd.SourceLocation, "invalid ARG arguments %v", cmd.Args)
+	}
+
+	var value string
+	if valueOrNil != nil {
```
brandonSc, line 1067: If you returned an empty-string from `parseArgArgs` instead would it be sufficient to just check `value != ""` here instead?

I think the pointer adds a _slight_ additional cost plus potential greater chance for panic in code if someone changes the code and mistakenly forgets to guard against nil somewhere

brandonSc (approved):
👍

earthfile2llb/interpreter.go
alexcb: my idea was to differentiate between `ARG key` and `ARG key = ""`. I figured a nil pointer kept the number of returned arguments smaller rather than introducing a `valueExists` bool. I could expand it to 5 returned values if you think that's more clear.

brandonSc: I think pointer makes more sense knowing now that you need to differentiate between empty string and lack of string (and 5 return values would be a bit much)

Issue #1399, opened 2021-11-08
Proposal: `CACHE` (extended Mount-based caching)


adamgordonbell:
## Problem Statement:

Some build tools are computationally expensive when run initially but quick on subsequent runs because they work incrementally. Earthly's layer-based caching ensures these tools are called no more than necessary, but, at present, when they are called the incremental features of these tools are lost. 

**Related Issues:**
* Earthly: #1139
* Earthly: #1129
* Earthly: #788
* Buildkit: https://github.com/moby/buildkit/issues/1512
* Buildkit: https://github.com/moby/buildkit/issues/1723
* Buildkit: https://github.com/moby/buildkit/issues/1545

**Maven Example:**
Maven’s `mvn install` is slow to run the first time because all packages are downloaded to a cache. If you add a new package, it will only need to download that one new package. However, in earthly, whenever a new package is added, `mvn install` will need to redownload every package.

**Dialyzer Example:**
Elixir’s Dialyzer PLT files are expensive to generate the first time, often taking 20 minutes to generate from scratch. But, subsequent runs take seconds because they use the output of the previous runs to prevent rework.

**Incremental Compilers:**
Some compilers compile slowly initially, but subsequent compiles use the intermediate files from the previous run to gain speed. With Earthly, every compile is the first run. 

An ideal solution would have the following attributes:

- Not harm repeatability. 
- Be easily understandable
- Work with existing cache options (layer cache, remote inline cache, remote explicit cache)

## Proposed Solution:

Introduce a new keyword `CACHE` which can be used inside a target like so:

```
target:
      FROM alpine
      CACHE /path/to/cache
      RUN ...
      RUN ...
```

`CACHE` mounts a shared volume to that path that will be available in all subsequent RUNs within that target. This volume will be reused across runs. With no additional flags provided this works identically adding a [mount-based caching](https://docs.earthly.dev/docs/guides/advanced-local-caching#option-2-mount-based-caching-advanced) statement to every run in a target.

Flags will extend this behavior further

## Cache Persistence

By default, volumes used for caching are not part of the resulting image. This can be a problem if you need the cached file in subsequent build targets. Cache persistence overcomes this by also copying the contents of the cache in the image itself at the end of the target as an additional layer:

```
one:
      FROM alpine
      CACHE /path/to/cache
      RUN touch /path/to/cache/one
      # at this point earthly copies the contents of the cache into /path/to/cache in the image,
[...]

PR #1514, opened 2021-12-17
Support CACHE command in Earthfiles

brandonSc:
"Phase I" of feature proposal here: https://github.com/earthly/earthly/issues/1399

Example usage:

```earthly
VERSION 0.6

FROM node:17-alpine3.12

npm-dep:
    CACHE ./node_modules
    COPY package*.json .
    RUN npm install

npm-build:
    FROM +npm-dep
    COPY *.js .
    ENTRYPOINT ["npm",  "start"]

another-test:
    WORKDIR /test
    CACHE ./something
    RUN echo "hello world" > ./something/hello

another-test-2:
    FROM +another-test
    RUN cat ./something/hello
```


earthfile2llb/converter.go
```diff
@@ -1765,6 +1792,45 @@ func (c *Converter) targetInputActiveOnly() dedup.TargetInput {
 	return c.mts.Final.TargetInput().WithFilterBuildArgs(activeBuildArgs)
 }
 
+// persistCacheVolume makes a temporary cache volume permanent by writing it's contents
+// from the volume to the host filesystem at the same directory.
+// Used when the Target contains a `CACHE --persist /my/directory` directive (with the --persist).
+func (c *Converter) persistCacheVolume() {
```
brandonSc, line 1798: Small amount of code, large amount of hassle. The operations in this function are a bit delicate and took a lot of trial and error to get right. 

For example, 
* it seems the cache volume isn't totally unmounted in a successive operation on the image without the run option that includes the volume. An empty directory seems to linger.
* Commands to `mv` or `cp` files to or from the directory where the volume is/was mounted fail for reasons I don't understand
  * e.g. before switching to llb.Copy, I was invoking `cp -r` to move the files to the directory that used to have the cache volume mounted to it - that would fail unless I put them in a sub directory, putting the files in the directory at top-level would fail for some reason

brandonSc: Regarding cache volumes not being totally unmounted, this example shows the same behaviour using `RUN --mount=type=cache` :
```Earthfile
mount-test:
    RUN --mount=type=cache,target=/mnt/test echo "hello world" > /mnt/test/hello
    RUN --mount=type=cache,target=/mnt/test cat /mnt/test/hello
    RUN ls /mnt
    RUN ls /mnt/test
```
=>
```
         +mount-test | --> RUN echo "hello world" > /mnt/test/hello
         +mount-test | --> RUN cat /mnt/test/hello
         +mount-test | hello world
         +mount-test | --> RUN ls /mnt
         +mount-test | test
         +mount-test | --> RUN ls /mnt/test
```

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ All notable changes to [Earthly](https://github.com/earthly/earthly) will be doc
 
 ### Added
 
+- Expirmental CACHE command with optional --persist flag can optimize projects by allowing incremental changes to files on host machine without rebuilding the entire cache. For example, a Maven project where `SNAPSHOT` dependencies are added frequently, an NPM project where `node_modules` changes frequently, or programming languages using incremental compilers. [#1399](https://github.com/earthly/earthly/issues/1399)
```
alexcb, line 13:
```suggestion
- Expirmental CACHE command with optional `--persist` flag can optimize projects by allowing incremental changes to files on host machine without rebuilding the entire cache. For example, a Maven project where `SNAPSHOT` dependencies are added frequently, an NPM project where `node_modules` changes frequently, or programming languages using incremental compilers. [#1399](https://github.com/earthly/earthly/issues/1399)
```

earthfile2llb/converter.go
```diff
@@ -1143,6 +1148,22 @@ func (c *Converter) Import(ctx context.Context, importStr, as string, isGlobal,
...
 }
 
+// Cache handles a `CACHE` command in a Target.
+// It appends run options to the Converter which will mount a cache volume in each successive `RUN` command.
+func (c *Converter) Cache(ctx context.Context, path string, isPersisted bool) error {
+	if err := c.checkAllowed(cacheCmd); err != nil {
+		return err
+	}
+	if isPersisted {
+		c.hasPersistentCache = true
```
alexcb, line 1158: will this work with multiple caches? perhaps with a mix of one that persists and another that doesn't?

tests/cache-cmd.earth
```diff
@@ -0,0 +1,20 @@
+VERSION 0.6
```
andrewsykim, line 1: I wonder if we should have tests or examples based on user feedback, like caching Elixir’s Dialyzer or Maven.

alexcb, line 1: Some users might want to have two (or more caches) within a target. It would be good to validate that works.

Another use-case could be defining the cache at the top of the Earthfile (under the implicit `+base` target). For example, we might need something along the lines of:

```
VERSION 0.6
CACHE /non-persistant-cache
CACHE --persists /persistant-cache

foo:
    RUN echo data >> /non-persistant-cache/data.txt
    RUN echo data >> /persistant-cache/data.txt
    RUN touch /fake-dep
    SAVE ARTIFACT /fake-dep

bar:
    RUN test $(cat /non-persistant-cache/data.txt | wc -l) -gt 0
    RUN test $(cat /persistant-cache/data.txt | wc -l) -gt 0
    SAVE IMAGE myimage:test

test:
    COPY +foo/fake-dep .
    IF [ -z "$(cat fake-dep)" ]
        BUILD +bar
    ELSE
        RUN false # should never happen
    END
```

Then on top of that, we should test the persisted cache made it to `myimage:test`:

```
docker run --rm myimage:test cat /persistant-cache/data.txt | grep 'data'
# and a similar one to make sure the non-persitant cache doesn't exist.
```

alexcb (changes requested):
looking good, but I think we need to expand it to support (and test for) multiple caches.

tests/cache-cmd.earth
alexcb: I would rather keep the testing as simple as possible (without over simplifying the requirements).

and then additionally we could create a real-world example (under `./examples`) to ensure it solves our use-cases.

andrewsykim: Yeah I think some Earthfiles in ./examples is fine too

earthfile2llb/converter.go
brandonSc: > looking good, but I think we need to expand it to support (and test for) multiple caches.

Thanks, I did not think about the multiple cache case. Update on the way.

brandonSc:
> looking good, but I think we need to expand it to support (and test for) multiple caches.

Thanks, I did not think about the multiple cache case. Update on the way.

tests/cache-cmd.earth
brandonSc: Good idea. I can easily add examples for Maven and NPM within this PR. I'll have to look into the Elixir's Dialyzer example a bit more but if its straight forward I can include that as well.

earthfile2llb/converter.go
```diff
@@ -1143,6 +1147,21 @@ func (c *Converter) Import(ctx context.Context, importStr, as string, isGlobal,
...
 }
 
+// Cache handles a `CACHE` command in a Target.
+// It appends run options to the Converter which will mount a cache volume in each successive `RUN` command.
+func (c *Converter) Cache(ctx context.Context, path string, isPersisted bool) error {
+	if err := c.checkAllowed(cacheCmd); err != nil {
+		return err
+	}
+	if isPersisted {
+		c.persistentCacheDirs = append(c.persistentCacheDirs, path)
```
alexcb, line 1157: What happens if a user defines the same cache path twice?

earthfile2llb/converter.go
```diff
@@ -1197,6 +1216,12 @@ func (c *Converter) FinalizeStates(ctx context.Context) (*states.MultiTarget, er
 		// Should never happen.
 		return nil, errors.New("internal error: stack not at base in FinalizeStates")
 	}
+
+	if len(c.persistentCacheDirs) > 0 {
```
alexcb, line 1220: nit: if guard seems unnessisary as the function does a `for _, cacheDir := range c.persistentCacheDirs {` which doesn't seem to have any side-effects when it's empty. I think simply calling `c.persistCacheVolumes()` would lead to less code-bloat.

tests/cache-cmd.earth
```diff
@@ -0,0 +1,32 @@
...
+    CACHE --persist ./my/data
+    CACHE --persist ./my/other
+    RUN echo "hello world" >> ./my/data/hello
+    RUN echo "hello again" >> ./my/other/hello
+    RUN echo "$(cat /base/persistent/hello), hello" > /base/persistent/hello
+
+test:
+    FROM +test-prev
+    # These files should contain only the contents written to it in +test-prev.
+    RUN test "$(cat ./my/data/hello)" = "hello world"
```
alexcb, line 25: will this still work when `./my/data/hello` contains multiple lines? should we be using `tail -n 1` instead?

earthfile2llb/converter.go
brandonSc: 💯 agree

earthfile2llb/converter.go
brandonSc: Good catch I think there would be risk of losing their data when we persist it. Will handle that more gracefully

tests/cache-cmd.earth
alexcb: In a similar vein, can we write a test that validates that the cache contains items written from a previous run of earthly?
e.g.

Consider earth1.test
```
foo:
    CACHE ./my/data
    RUN echo alpha >> ./my/data/hello
```

Consider earth2.test
```
foo:
    CACHE ./my/data
    RUN echo bravo >> ./my/data/hello
```

then run `cp earth1.test Earthfile; earthly +foo`, followed by `cp earth2.test Earthfile; earthly +foo`.

Can we validate `./my/data/hello` ends with `alpha\nbravo\n`?

brandonSc: Yes will look into that. Seems like a good test since that's an important part of the feature

earthfile2llb/converter.go
```diff
@@ -78,6 +79,8 @@ type Converter struct {
 	directDeps          []*states.SingleTarget
 	buildContextFactory llbfactory.Factory
 	cacheContext        pllb.State
+	runOpts             []llb.RunOption
+	persistentCacheDirs map[string]bool // Treated as a "set" of cache directories to persist in the image
```
brandonSc, line 83: @alexcb let me know what you think of this approach. I've found it to be an idiomatic way in Go to achieve set-like behavior and/or get O(1) lookup without looping through an array

alexcb: That sounds good, we do this in other places of the code too. I've also seen it as `map[string]struct{}`, which makes it even more clear that it's a set as there's no associated value with it. And according to https://itnext.io/set-in-go-map-bool-and-map-struct-performance-comparison-5315b4b107b using `struct{}` is actually faster and reduces memory. (not like memory is a concern when I suspect N<10 in most cases).

vladaionescu:
I believe the meaning of the `--persist` flag is different than originally intended in the proposal.

The contents of the cache are meant to be "persisted" in the image layers, not on the host system. Similar to what https://github.com/moby/buildkit/issues/1545 describes. In fact, I believe we need to implement that buildkit feature in order to support this.

(I agree that the name "persist" is maybe a bit misleading - we should probably think of a more suggestive alternative)

brandonSc:
> I believe the meaning of the `--persist` flag is different than originally intended in the proposal.
> 
> The contents of the cache are meant to be "persisted" in the image layers, not on the host system. Similar to what [moby/buildkit#1545](https://github.com/moby/buildkit/issues/1545) describes. In fact, I believe we need to implement that buildkit feature in order to support this.
> 
> (I agree that the name "persist" is maybe a bit misleading - we should probably think of a more suggestive alternative)

Maybe we can chat tomorrow to clarify how the "persist" works @vladaionescu @alexcb @adamgordonbell

adamgordonbell:
>The contents of the cache are meant to be "persisted" in the image layers, not on the host system. Similar to what moby/buildkit#1545 describes. In fact, I believe we need to implement that buildkit feature in order to support this.

+1 on that. `persist` needs to get the data into the image so that `load` can pull it back out. I *think* persist is also needed for some incremental compilers as well, where subsequent steps might need the intermediate files, but caching the intermediate files across runs makes things a lot faster.

earthfile2llb/converter.go
```diff
@@ -1765,6 +1789,47 @@ func (c *Converter) targetInputActiveOnly() dedup.TargetInput {
 	return c.mts.Final.TargetInput().WithFilterBuildArgs(activeBuildArgs)
 }
 
+// persistCacheVolumes makes temporary cache volumes permanent by writing their contents
+// from the volume to the host filesystem at the same directory.
```
brandonSc, line 1793: rephrase

brandonSc: note SAVE ARTIFICAT will not work with current solution

adamgordonbell:
Should we consider `--persist` being on by default? Users may not expect files to disappear at the end of a target when the mount is gone.

earthfile2llb/converter.go
```diff
@@ -1143,6 +1168,22 @@ func (c *Converter) Import(ctx context.Context, importStr, as string, isGlobal,
 	return c.varCollection.Imports().Add(importStr, as, isGlobal, currentlyPrivileged, allowPrivilegedFlag)
 }
 
+// Cache handles a `CACHE` command in a Target.
+// It appends run options to the Converter which will mount a cache volume in each successive `RUN` command,
+// and configures the `Converter` to persist the cache in the image at the end of the target.
+func (c *Converter) Cache(ctx context.Context, path string) error {
+	if err := c.checkAllowed(cacheCmd); err != nil {
```
vladaionescu, line 1175: Nit: Make the style consistent?

earthfile2llb/converter.go
```diff
@@ -1765,6 +1815,50 @@ func (c *Converter) targetInputActiveOnly() dedup.TargetInput {
 	return c.mts.Final.TargetInput().WithFilterBuildArgs(activeBuildArgs)
 }
 
+// persistCache makes temporary cache directories permanent by writing their contents
+// from the cached directory to the persistent image layers at the same directory.
+// Used when the Target contains at least one `CACHE /my/directory` command.
```
vladaionescu, line 1820:
```suggestion
// This only has an effect when the Target contains at least one `CACHE /my/directory` command.
```

util/llbutil/copyop.go
```diff
@@ -57,6 +58,37 @@ func CopyOp(srcState pllb.State, srcs []string, destState pllb.State, dest strin
 	return destState.File(fa, opts...)
 }
 
+// DockerfileCopy copies from `src` to `dest` using Docker's own custom image
+// for running the COPY command in Dockerfiles. This function is similar to
+// llb.Copy() but can accept RunOptions (for example, which may contain a mount).
+func DockerfileCopy(srcState pllb.State, src, dest string, opts []llb.RunOption, platform *specs.Platform) pllb.State {
```
vladaionescu, line 64: I would call this... `CopyViaRun` or something along those lines. The fact that the inspiration comes from an old Dockerfile-specific implementation is not particularly relevant.

earthfile2llb/converter.go
```diff
@@ -1765,6 +1815,50 @@ func (c *Converter) targetInputActiveOnly() dedup.TargetInput {
...
+			&llb.CopyInfo{
+				FollowSymlinks:     true,
+				CreateDestPath:     true,
+				AllowWildcard:      true,
+				AllowEmptyWildcard: true,
+			},
+		)
+
+		// Remove the temporary backup after we're done copying from it.
+		dest = dest.File(fa.Rm(tmpDir))
```
vladaionescu, line 1856: So this has 3*N layers:

1. Use copy+mounts to copy cache from mount into a tmp dir
2. Use an `llb.Copy` to copy from a tmp dir and into the previously mounted path
3. Rm tmp dir

I wonder if there's a way to reduce this as much as possible. For example, what if we could copy from the mount directly into the proper destination. Then we would need only a single copy and no rm.

I think we can achieve this by mounting the cache to a tmp location (as opposed to mounting in its usual location), and then copying from it to the real intended path.

examples/cache-command/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+A worse situation may occur if a build tool doesn't ignore a cached dependency that was removed from a manifest file (e.g. `pom.xml`, `package.json`, `build.gradle`). In such a case, a build  could behave differently (or break) on another machine if it unintentionally uses a cached dependency that should be removed.
+
+## More Detailed Examples
+
+Some fully functional example projects are contained within subdirectories to demonstrate the CACHE command in more detail. 
+
+Take a look at the Earthfiles in these examples to see how they work. 
+
+Feel free to reach out via Github Issue or on our [Community Slack channel](https://join.slack.com/t/earthlycommunity/shared_invite/zt-n5vw7fw6-7inWcey3_UNkx0FOm3bQBA) if you have any questions or ideas. Thanks for using Earthly!
```
vladaionescu, line 75:
```suggestion
Feel free to reach out via Github Issue or on our [Community Slack channel](https://earthly.dev/slack) if you have any questions or ideas. Thanks for using Earthly!
```

vladaionescu (commented):
Great stuff - I have some ideas to improve even further.

earthfile2llb/converter.go
brandonSc: Interesting idea, I think in theory that should work. I'll give it a go

earthfile2llb/converter.go
```diff
@@ -674,8 +693,14 @@ func (c *Converter) SaveArtifact(ctx context.Context, saveFrom string, saveTo st
...
 		if isPush {
+			pushState := persistCache(
+				c.mts.Final.RunPush.State,
+				c.persistentCacheDirs,
+				c.runOpts,
+				c.mts.Final.Platform,
+			)
 			separateArtifactsState = llbutil.CopyOp(
-				c.mts.Final.RunPush.State, []string{saveFrom}, separateArtifactsState,
+				pushState, []string{saveFrom}, separateArtifactsState,
```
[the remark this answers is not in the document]
brandonSc: I thought it would be necessary to copy the cached files into the state used here as well. I've been testing more thoroughly though, and I don't think I understand when this case gets executed.

Can you or @alexcb explain a bit more? I understand we may set `isPush` to `true` by using `--push` flag in RUN` or `SAVE IMAGE`, but what is it's relation to `SAVE ARTIFACT`?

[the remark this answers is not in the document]
vladaionescu: Ah yes, I missed the part above containing

```go
			pushState := persistCache(
				c.mts.Final.RunPush.State,
				c.persistentCacheDirs,
				c.runOpts,
				c.mts.Final.Platform,
			)
```

I think this makes sense now. This is the case of

```earthly
RUN --push ...
SAVE ARTIFACT ... AS LOCAL ...
```

The `RUN --push` only executes during the push phase of the build, and therefore `SAVE ARTIFACT` would need to execute off of that in this case. So this makes sense now that I look at it more carefully.

earthfile2llb/interpreter.go
```diff
@@ -1350,6 +1353,21 @@ func (i *Interpreter) handleImport(ctx context.Context, cmd spec.Command) error
...
 }
 
+func (i *Interpreter) handleCache(ctx context.Context, cmd spec.Command) error {
+	if len(cmd.Args) != 1 {
+		return errors.Errorf("invalid number of arguments for CACHE: %s", cmd.Args)
+	}
+	dir := cmd.Args[0]
+	if !path.IsAbs(dir) {
+		dir = strings.TrimLeft(dir, "./")
+		dir = path.Join("/", i.converter.mts.Final.MainImage.Config.WorkingDir, dir)
```
brandonSc, line 1363: Is this the best way to convert a relative path in `WORKDIR` to an absolute path?

tests/cache-cmd.earth
brandonSc: I attempted this test but it seems when we run the nested Earthfiles in the `tests/` director, they don't utilize cache (as if they're run with a `--no-cache` flag, but I don't see that being done explicitly anyway)

tests/cache-cmd.earth
brandonSc: Apparently it's not easy to save an image in our tests, as the operation fails with `io: read/write on closed pipe`. The other tests that utilize SAVE_IMAGE seem to be using `--no-output` (so they don't actually write the image) or testing something with pushing to a registry

util/llbutil/copyop.go
brandonSc: I called it `CopyWithRunOptions`

earthfile2llb/converter.go
brandonSc: Got this working, thanks for the suggestion!

earthfile2llb/interpreter.go
alexcb: maybe you can just do `path.Clean(path.Join("/", i.converter.mts.Final.MainImage.Config.WorkingDir, dir))` to avoid the call to `TrimLeft`, as well as account for cases where `dir` starts with `../`.

brandonSc: worked great, thanks for the suggestion

earthfile2llb/converter.go
```diff
@@ -1143,6 +1181,23 @@ func (c *Converter) Import(ctx context.Context, importStr, as string, isGlobal,
...
 }
 
+// Cache handles a `CACHE` command in a Target.
+// It appends run options to the Converter which will mount a cache volume in each successive `RUN` command,
+// and configures the `Converter` to persist the cache in the image at the end of the target.
+func (c *Converter) Cache(ctx context.Context, path string) error {
+	err := c.checkAllowed(cacheCmd)
+	if err != nil {
+		return err
+	}
```
vladaionescu, line 1191: I think that there needs to be a call to `c.nonSaveCommand()` as well here.

```suggestion
	}
	c.nonSaveCommand()
```

earthfile2llb/converter.go
```diff
@@ -1143,6 +1181,23 @@ func (c *Converter) Import(ctx context.Context, importStr, as string, isGlobal,
 	return c.varCollection.Imports().Add(importStr, as, isGlobal, currentlyPrivileged, allowPrivilegedFlag)
 }
 
+// Cache handles a `CACHE` command in a Target.
+// It appends run options to the Converter which will mount a cache volume in each successive `RUN` command,
+// and configures the `Converter` to persist the cache in the image at the end of the target.
+func (c *Converter) Cache(ctx context.Context, path string) error {
```
vladaionescu, line 1187: Should any new `FROM` encountered in the same target reset any cache dirs collected so far?

vladaionescu, line 1187: Also, should `LOCALLY` do the same?

vladaionescu, line 1187: Also, should there be an error if `CACHE` is seen after a `LOCALLY`?

vladaionescu (approved):
This is looking great! Almost there...

I have some additional fine points, just to make sure...

earthfile2llb/converter.go
```diff
@@ -78,12 +79,16 @@ type Converter struct {
 	directDeps          []*states.SingleTarget
 	buildContextFactory llbfactory.Factory
 	cacheContext        pllb.State
+	runOpts             []llb.RunOption
+	persistentCacheDirs map[string]struct{} // Treated as a "set" of cache directories to persist in the image
```
alexcb, line 83: nit: comment is no longer needed as `map[string]string` was changed to `map[string]struct{}`.
```suggestion
	persistentCacheDirs map[string]struct{}
```

alexcb (approved):
nice work.

earthfile2llb/converter.go
brandonSc: as-is currently, it seems to lose the cached files after another `FROM`. 

I think it's probably safest in general to not try and keep the files (e.g. if they went from linux to windows, could those files have unexpected issues?)

Open to your opinion here on if we should do anything more. Possibilities include:
* Try to support allowing multiple FROMs (don't like it, probably buggy and high effort for unlikely use case)
* Restrict it (throw an error if there's another FROM after cache)
* Better handle it: drop cache dirs etc since they're not used, being a bit more efficient. maybe whats going on so user understands
* Just leave it as-is for now, not worth effort for unlikely edge case on a new feature we don't the adoption rate of yet

brandonSc: That sounds like a good idea. Does that also cover [your question above](https://github.com/earthly/earthly/pull/1514#discussion_r780568982)?

brandonSc: Implemented this error

vladaionescu: I'd go with this option - it's probably what users would expect (`FROM` anyway resets everything else):

> Better handle it: drop cache dirs explicitly since they're not used (being a bit more efficient). maybe log what's going on so the user understands.

If someone wants to transfer the cache from one `FROM` to another, they should probably better use a different target.

vladaionescu: I'd do the same with `LOCALLY` - reset the cache dirs.

vladaionescu: Yeah, it does.

earthfile2llb/converter.go
```diff
@@ -1750,7 +1818,7 @@ func (c *Converter) checkAllowed(command cmdType) error {
 	}
 
 	switch command {
-	case fromCmd, fromDockerfileCmd, locallyCmd, buildCmd, argCmd, importCmd:
+	case fromCmd, fromDockerfileCmd, locallyCmd, buildCmd, argCmd, importCmd, cacheCmd:
```
vladaionescu, line 1821: I wouldn't allow cache before a `FROM`

earthfile2llb/converter.go
brandonSc: Added this

earthfile2llb/converter.go
```diff
@@ -1339,6 +1404,9 @@ func (c *Converter) internalRun(ctx context.Context, opts ConvertRunOpts) (pllb.
 		if opts.Transient {
 			return pllb.State{}, errors.New("Transient run not supported with LOCALLY")
 		}
+		if len(c.persistentCacheDirs) > 0 {
+			return pllb.State{}, errors.New("CACHE command not supported with LOCALLY")
+		}
```
vladaionescu, line 1409: I would make this check within the `Cache` function. I think there are edge cases that this doesn't catch (e.g. LOCALLY, then SAVE ARTIFACT, without RUN).

brandonSc: Understood - and I agree this does make more sense if there are edge cases that would be missed the other way

earthfile2llb/converter.go
brandonSc: resolving since other error check covers this

earthfile2llb/converter.go
brandonSc: Moved this check so it works in call cases (i.e. not just `RUN`)

vladaionescu (approved):
Looks good!

One more thing I realized we're missing (sorry!): it would be good to add a feature flag for the `CACHE` command and to return an error in `Cache` if the feature flag is not active.

Everything else looks good!

earthfile2llb/converter.go
brandonSc: Wondering though if Warning log is the right way to communicate this, as it looks more like a serious error in red text than perhaps it is. 

Is there a more appropriate way to print something at more of an info-level within the converter logic? Just `fmt.Println`?

vladaionescu: I would maybe not log it at all?.. we don't really have a good way of logging things from the converter.

brandonSc: removed the log

PR #1518, opened 2021-12-21
Support passing an env file

### LINKED ISSUE #1515: allow importing build args/env vars from file
Hello,
I noticed that when changing the value of a global arg, it invalidates cache for targets that do not reference it/depend on it.
For example:
1. Earthfile:
```
ARG FOO
hello:
    RUN echo "hello world"
```

2. `earthly +hello --FOO=bar`
```
...
              +hello | FOO=bar
              +hello | --> FROM alpine
              +hello | [██████████] 100% resolve docker.io/library/alpine@sha256:21a3deaa0d32a8057914f36584b5288d2e5ecc984380bc0118285c70fa8c9300
              +hello | FOO=bar
              +hello | --> RUN echo "hello world"
              +hello | hello world
```

3. `earthly +hello --FOO=baz`
```
...
              +hello | FOO=baz
              +hello | --> FROM alpine
              +hello | [██████████] 100% resolve docker.io/library/alpine@sha256:21a3deaa0d32a8057914f36584b5288d2e5ecc984380bc0118285c70fa8c9300
              +hello | FOO=baz
              +hello | --> RUN echo "hello world"
              +hello | hello world
```

Now, I’m guessing this is because the global args are set in a hidden `base` target that all other targets depend on, 
and I’m also guessing that the args cannot be differentiated according to whether they are referenced or not because they are auto exported as env vars as well, which means any command might be using it.
Is that right?

The reason I ask is that I’m interested in being able to set default values in one place for multiple args (preferably at the top of the Earthfile) even though some args won’t be used by all targets.

I wonder if it can become possible to set the args in one place while still expecting for caches to work for targets that do not expect to use a changed arg?

If not, the 2nd best option I can think of is to be able to pass an env var file to the earthly command, so that default values can still be managed in one place (this would mean of course that I would need to define the ARG in each target that wishes to use it, unlike the global arg approach). 

I've noticed there is an option to either set `$EARTHLY_BUILD_ARGS` as one string of key/value pairs or to set the env vars in a `.env`. However, I'd like to be to do something like 
`earthly  --env-file my_vars.env +my_target` (similar to the existing `docker run` option), since the other options do not allow me to set the values in an explicit file while still invoking earthly in one command.

Would love to hear your thoughts on this.
Thanks!

ido-namely:
Fixes #1515


cmd/earthly/main.go
```diff
@@ -588,6 +589,13 @@ func newEarthlyApp(ctx context.Context, console conslogging.ConsoleLogger) *eart
 			Destination: &app.featureFlagOverrides,
 			Hidden:      true, // used for feature-flipping from ./earthly dev script
 		},
+		&cli.StringFlag{
+			Name:        "env-file",
+			Aliases:     []string{"E"},
+			EnvVars:     []string{"EARTHLY_ENV_FILE"},
+			Usage:       "Specify additional build args and/or secrets via a given env file",
```
vladaionescu, line 596: I wonder if this setting should simply replace the default `./.env` location instead of adding yet another location.

docs/guides/build-args.md
```diff
@@ -197,7 +212,22 @@ This is possible in a few ways:
...
+   ```
+   
+   Then simply run earthly:
+   
+   ```bash
+   earthly --env-file my_env.env +hello
+   ```
+
+
+5. Via the `.env` file.
```
vladaionescu, line 230: Mind moving the docs to a new PR that is branched off the `next` branch?

vladaionescu (commented):
This is looking pretty good, though I haven't checked all the details. Leaving the main review the @alexcb.

cmd/earthly/main.go
ido-namely: The only issue I see with that is that .env has another role of setting earthly command env vars which are loaded before any of the flags are set - https://github.com/earthly/earthly/blob/37846e0c8a6b3f87823ffb7e24687abc8ffed125/cmd/earthly/main.go#L217

That is something that `--env-file` cannot do if it's to be treated as all the other flags.

vladaionescu: Oh yeah... there was a trick to make this work. We have a similar issue with `--config`. I think @alexcb may know the answer to this riddle.

alexcb: I'm not entirely sure what this comment means:
```
	// Load .env into current global env's. This is mainly for applying Earthly settings.
	// Separate call is made for build args and secrets.
```

in particular: `This is mainly for applying Earthly settings`.

It's not obvious why we have to do this under `before`, rather than in the main build method.

It was added in 708cb1ae5fe24020aba3778b73e04fec09326ed5

I'm happy to see if we can move this code into `actionBuildImp`, or figure out a way to bypass this if not.

vladaionescu: The entries in `.env` influence three things:

1. env vars that Earthly reacts to (e.g. `EARTHLY_CI=true`)
2. Build args
3. Secrets

I think 1. can be problematic from a chicken-and-egg perspective and that's why it may need to be applied in `before`.

alexcb: > 1. env vars that Earthly reacts to (e.g. EARTHLY_CI=true)

Wow, I had no idea, that definitely complicates things.

What would we want to do about the case where `.env` contains `EARTHLY_ENV_FILE=/other/path/to/.env`?

Do we follow the link?

Do we include both `.env` and `/other/path/to/.env`?

Do we allow `/other/path/to/.env` to reference yet a different env path (e.g. `.env` -> `/other/path/to/.env` -> `/somewhere/else`)?

vladaionescu: We could perhaps ignore that entry in that case... seems like an odd use-case.

alexcb: I took a closer look at the code and it's not quite the same pattern as the `--config` flag, which happens in `app.before`, which still occurs after `urfav/cli` is initialized (since it makes use of `context.IsSet("config")`, and context is part of the cli library).

It almost seems like we'll need to first scan all the cli args for `--env-file` before using `urfav/cli`; either via reaching into argv, or perhaps a second fake urfav/cli app?

I think argv seems like the simplest approach.

alexcb: update: argv won't work as we need to be able to differentiate between boolean flags and flags that take a value; therefore we need to know which flags expect a value.

ido-namely: I can work on that change, however, I gave it some more thought and while I can see why too many options to set a var can be confusing, I also think that the explicit and implicit env files serve a different purpose:
 - The `.env` is more for local dev in order to override some args/secrets in the `Earthfile`
  - The explicitly passed env var is to consolidate the project/app configuration in one file (at least that's how I envisioned it).

So if the explicit file is used, there will be no way to override vars as easily as before (changing the path of the env-file for local dev would break the repeatability and would create a different in how the command is executed locally and in the CI).

alexcb: I came back to the idea of scanning ahead for argv. It seems like somewhere we can tolerate some technical debt.

here's an example: https://github.com/earthly/earthly/pull/1520

The only failure case would be something like: `earthly --build-arg --env-file=some.other.env ...` where it will initially load in  `some.other.env`, where technically `--env-file=some.other.env` is being passed as a value to `--build-arg` (which is what urfav/cli will parse it as, and then fail upon). Seems like a very unlikely case, which we could fix up if anyone encounters it.

vladaionescu: > I also think that the explicit and implicit env files serve a different purpose

Are you suggesting that they stay separate then? (can use both .env and explicit --env-file?)

ido-namely: yes, I think so, though I'm not sure. What do you think?
One thing to notice is that if both are used at the same time, the impact of `.env` will be limited to things that are not in the explicit env file and I wonder if the `.env` is still effective in that case.

alexcb: > both are used at the same time

What if `--env-file` took multiple values, seperated by commas? Then you could include both: `--env-file=.env,~/.env`.

That would allow you to specify that if both `.env` and `~/.env` contained the same environment variable, the value from the later (`~/.env`) would be kept.

ido-namely: I think the idea behind using an `.env` file is to place it in the working directory and not have to worry about it.
I'm almost thinking to keep supporting both files and let `.env` win. 
The issue there is that it would probably be confusing to let an implicit file override the explicit file.

What if instead of calling it `env-file`, we call it `build-arg-file` (or something more inclusive if secrets are included), maybe that way it makes sense than environment variables can override a "config" file essentially.
It also limits the scope of the file to ARG (and SECRET) and without earthly cli options.

vladaionescu: > What if instead of calling it env-file, we call it build-arg-file (or something more inclusive if secrets are included), maybe that way it makes sense than environment variables can override a "config" file essentially.

Yeah.. perhaps. I think it's definitely easier to reason about the file if it doesn't influence Earthly settings too. Not sure what the best flag name should be if we went this way. An option could be to have these things separated into two sets of variables: `--secret-file` and `--build-arg-file`.

ido-namely: yeah, I think that should be ok.
I wonder if we should start with one (`--args-file`) and then wait if there is need for `--secrets-file`? 
Or alternatively for these options to be merged after all?
I'm not sure what people needs are.

ido-namely: I still like the idea of using such file by default if it exists in the Earthfile directory (`EarthBuildArgsFile`) to make the execution easier and less verbose.

vladaionescu: Re default file: Wouldn't it be possible for the user to specify those same entries as global ARGs with default values within the Earthfile itself? Or perhaps there's some use-case that I'm missing?

ido-namely: That's where I started from.
The problem was that whenever a value changed it invalidated the cache for all targets.

vladaionescu: Ah yes...

ido-namely: I'd still prefer to use the Earthfile otherwise.
Would it be possible to use a new keyword in the top of the file so a global ARG won't be considered global but rather a value that is loaded as if it was from the cli build-arg option and/or env?

vladaionescu: Perhaps an option could be to allow the user to explicitly declare which args are global.

```earthly
ARG not_global
ARG --global this_is_global
```

ido-namely: Yes, I was thinking of something similar. Is the breaking change acceptable though or should it be inverted (e.g 

```earthly
ARG --local not_global
```
?

Also would this still work?

```earthly
ARG --local not_global=foo # (or ARG --local not_global=foo)

my_target:
    FROM alpine
    ARG not_global
    RUN echo hello $not_global
```

`earthly +my_target`

expecting`hello foo`
?

vladaionescu: The breaking change should be acceptable, I believe. We can roll it out under `VERSION 0.7`.

In the meantime, perhaps it'll be possible to use the feature flag `VERSION --explicit-global 0.6`, in order to enable the new behavior.

Though we should throw an error if someone tries to use `ARG --global` within a non-base target.

vladaionescu: BTW, to add a feature flag, you can just add it to this struct: https://github.com/earthly/earthly/blob/main/features/features.go#L20

And you should have the flag available in `converter.go` via `c.opt.ftrs` or something like that.

alexcb:
We had a very similar request by another user in Slack, which resulted in a similar implementation: c056afc75f6daeb25cd1f7144acaa958dfcbbea4

PR #1523, opened 2021-12-22
set EARTHLY_VERSION_FLAG_OVERRIDES after initial build

alexcb:
Don't set `EARTHLY_VERSION_FLAG_OVERRIDES` until after the first build is
done with the released binary, since the values in `EARTHLY_VERSION_FLAG_OVERRIDES`
might not work with the released version of earthly and only apply to
the newly built binary.

### DIFF
--- a/.buildkite/tests.sh
+++ b/.buildkite/tests.sh
@@ -13,9 +13,6 @@ function cleanup() {
 }
 trap cleanup EXIT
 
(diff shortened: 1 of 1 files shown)


dchw (approved):
Now it matches GHA!

vladaionescu (approved):

PR #1528, opened 2021-12-28
Best practices page

vladaionescu:
Preview: https://app.gitbook.com/o/-MB5s2W0v8S4xFU5ELti/s/iZfkHzyR9lPa2mtTFR0D/best-practices

### DIFF
--- a/docs/best-practices/best-practices.md
+++ b/docs/best-practices/best-practices.md
@@ -46,6 +46,8 @@ COPY ./*.go ./
 RUN go build ...
 ```
 
+An additional way in which you can improve the precision of the `COPY` comamnd is to use the [`.earthlyignore`](../earthfile/earthignore.md) file. Note, however, that this is best left as a last resort, as new files added to the project (that may be irrelevant to builds) would need to be manually added to `.earthlyignore`, which may be error-prone. It is much better to have to include every new file manually into the build (by adding it to a `COPY` command), than to exclude every new file manually (by adding it to the `.earthlyignore`), as whenever any such new file *must* be included, then the build would typically fail, making it harder to make a mistake compared to the reverse.
+
 ## `ENV` for image env vars, `ARG` for build configurability
 
 `ENV` variables and `ARG` variables seem similar, however they are meant for different use-cases. Here is a breakdown of the differences, as well as how they differ from the Dockerfile-specific `ARG` command:
@@ -210,7 +212,7 @@ TODO ...
 
 TODO ...
 
-## Use `WITH DOCKER --load=+my-target` to pass files to `LOCALLY` targets
+## Use `WITH DOCKER --load=+my-target` to pass images to `LOCALLY` targets
 
 TODO ...
 
@@ -219,5 +221,9 @@ TODO ...
 TODO ....
 
 ## Use cross-repo references, avoid `GIT CLONE` if possible
+
+TODO ...
+
+## Technique: Use `earthly -i` to debug failures
 
 TODO ...
--- a/docs/best-practices/best-practices.md
+++ b/docs/best-practices/best-practices.md
@@ -107,7 +107,13 @@ As a best practice, it is recommended to use `RUN for` whenever possible (e.g. o
 
 ## Use `--ci` when running in CI
 
-...
+Build scripts serve slightly different purposes when they are run in CI compared to when they are executed for local development. Most of the logic is similar (hence Earthly attempts to unify the two concepts in a unified syntax), but there are some small differences.
+
+For example, for development purposes, you may use commands such as `LOCALLY`, which cause Earthly to be less repeatable, and yet might satisfy very much needed use-cases that are typically out of scope of a CI build.
+
+In addition, in CI it is much more likely that shared caching will be needed, while outputting artifacts and images locally would not be needed.
+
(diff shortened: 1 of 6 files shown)


docs/best-practices/best-practices.md
```diff
@@ -0,0 +1,370 @@
+# Best practices
+
+Although Earthly has been designed to be unambiguous about what command to use for the job, writing Earthfiles can sometimes still be tricky, when it comes to nuances. As you try to accomplish certain tasks, you may find that sometimes the same result can be achieved using more than one technique. Or so it might seem.
+
+Below we list some of the best practices that we have found to be useful in writing Earthfiles, with a focus on certain commands or techniques that seem similar, but aren't really, but also on some key points that we have seen newcomers stumble into.
+
+## `COPY` only the minimal amount of files. Avoid copying `.git`
+
+A typical mistake is to `COPY` entire large directories into the build environment and only using a subset of the files within them. Or worse, copying the entire repository (which might also include `.git`) for no good reason.
```
brandonSc, line 9: With Docker you might have a `.dockerignore` to exclude something like `.git`. Do we support something similar?

brandonSc:
Well written and super useful. I learned quite a few things from this guide!

docs/best-practices/best-practices.md
```diff
@@ -0,0 +1,1015 @@
...
+
+Note, however, that none of these cases are needed in a CI environment, and ultimately these commands are not regularly tested by a CI, which means they may break more frequently.
+
+## Pattern: Optionally `LOCALLY`
+
+In certain cases, it may be desirable to execute certain targets on the host machine, rather than in the sandboxed build environment, for debugging purposes. However, we need most of the target to execute in strict mode in CI. The solution to this is to use a target that can be optionally executed via `LOCALLY`. Here is an example:
+
+Suppose we wanted the following target to be executed on against the host's Docker daemon:
+
+```Dockerfile
```
ingwarsw, line 144: Use `Earthfile` syntax would work here?

docs/best-practices/best-practices.md
```diff
@@ -0,0 +1,1015 @@
...
+        --service db \
+        --load=+integration-test
+    RUN docker-compose up integration
+END
+```
+
+We could have an equivalent `LOCALLY` target:
+
+```Dockerfile
+LOCALLY
```
ingwarsw, line 158: Use `Earthfile` syntax would work here?

docs/best-practices/best-practices.md
```diff
@@ -0,0 +1,1015 @@
...
+WITH DOCKER --compose docker-compose.yml \
+        --service db \
+        --load=+integration-test
+    RUN docker-compose up integration
+END
+```
+
+We could have an equivalent `LOCALLY` target:
+
+```Dockerfile
```
ingwarsw, line 157: Use `Earthfile` syntax would work here?

docs/guides/udc.md
```diff
@@ -72,3 +72,28 @@ PRINT_VAR:
     COMMAND
     RUN echo "$a_global_var"
 ```
+
+## Targets vs UDCs
+
+Earthly targets and UDCs are Earthly's core primitives for organizing build recipes. They encapsulate build logic, and from afar they look pretty similar. However, the use-cases for each are vastly different.
+
+In general, targets are used to produce specific build results, while UDCs are used as a way to reuse build logic, when certain commands are repeated in multiple places. As a real-world analogy, targets are more like factories, while UDCs are more like components that are used to put together factories.
```
adamgordonbell, line 80: What about adding "UDCs work like function or methods call in an imperative programming language (c, javascript, python). Much like function calls it's helpful to imagine UDCs being executed by being inlined into the call site but in a separate variable scope"

docs/best-practices/best-practices.md
```diff
@@ -0,0 +1,1015 @@
...
+
+Examples of valid cases where `LOCALLY` may be used are:
+
+* installing dependencies on the host machine (e.g. to help IDEs provide better suggestions)
+* executing tests on the host docker daemon, to help with inspection and debugging
+* executing development commands which would otherwise require copying very large amounts of files to the sandboxed build environment
+
+Note, however, that none of these cases are needed in a CI environment, and ultimately these commands are not regularly tested by a CI, which means they may break more frequently.
+
+## Pattern: Optionally `LOCALLY`
```
adamgordonbell, line 138: Good tip!

docs/best-practices/best-practices.md
```diff
@@ -0,0 +1,1015 @@
...
+integration-test:
+    FROM some-other-image:latest
+    DO github.com/earthly/lib+INSTALL_DIND
+    COPY docker-compose.yml ./
+    WITH DOCKER --compose docker-compose.yml
+        RUN ...
+    END
+```
+
+A simple way to fix this is to use an earthly-provided [UDC](../guides/udc.md) to install Docker engine before the `COPY` command.
```
adamgordonbell, line 982: Worth mentioning that the UDC is fastest on alpine?

adamgordonbell (approved):
Great job. This will be helpful for sure.

brandonSc (approved):

docs/guides/udc.md
vladaionescu: Good idea

PR #1536, opened 2022-01-04
test against current branch rather than main

alexcb:
Currently if main is broken, and fixed in a PR,
this test will fail until main is fixed. This makes it impossible
to submit a PR and validate the PR fixes main until after it gets
merged to main.


brandonSc (approved):
💯

PR #1537, opened 2022-01-04
use tinyurl to reference Earthfile due to gitbook bug

alexcb:
GitBook currently has a bug where any references to an "Earthfile" gets confused with "docs/Earthfile" and somehow appends a /README.md

e.g. https://github.com/earthly/earthly/blob/main/Earthfile is changed to https://github.com/earthly/earthly/blob/main/Earthfile/README.md

This is a workaround they suggested we could use until they fix the
issue.

### DIFF
--- a/docs/best-practices/best-practices.md
+++ b/docs/best-practices/best-practices.md
@@ -843,12 +843,15 @@ As one example, you might find the [monorepo example](https://github.com/earthly
 
 For a real-world example, you can also take a look at Earthly's own build, where several Earthfiles are scattered across the repository to help organize build logic across modules, very much like regular code. Here are some examples:
 
+<!-- GitBook currently has a bug where any references to an "Earthfile" gets confused with "docs/Earthfile" and somehow appends a /README.md
+https://github.com/earthly/earthly/blob/main/Earthfile was changed to https://tinyurl.com/yt3d3cx6 -->
+
 * [`ast/parser`](https://github.com/earthly/earthly/tree/main/ast/parser) - Earthfile contains the logic for generating Go source code based on an ANTLR grammar.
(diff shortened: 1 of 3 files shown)


vladaionescu:
Mind fixing this one too?
image


It's on the best-practices.md page.

alexcb:
> It's on the best-practices.md page.

thanks for pointing that out!

I also found another instance under `guides/integration.md`.

adamgordonbell:
Would it work to rename this path:
`https://github.com/earthly/earthly/tree/main/docs/earthfile`

so it was `https://github.com/earthly/earthly/tree/main/docs/earth-file`? Is that what they are pattern matching on?

alexcb:
> Would it work to rename this path: https://github.com/earthly/earthly/tree/main/docs/earthfile to https://github.com/earthly/earthly/tree/main/docs/earth-file?

That's an interesting idea!

That would change our docs URL from `https://docs.earthly.dev/docs/earthfile` to `https://docs.earthly.dev/docs/earth-file`. Could that negatively affect our google search results? From a user's perspective it looks a bit weird too.

vladaionescu (approved):

adamgordonbell:
> That would change our docs URL from `https://docs.earthly.dev/docs/earthfile` to `https://docs.earthly.dev/docs/earth-file`. Could that negatively affect our google search results? From a user's perspective it looks a bit weird too.

I'm not sure about this idea, but they do seem to have redirects, so maybe we could redirect to earth-file from earthfile?

https://docs.gitbook.com/integrations/git-sync/content-configuration

I guess you already have the tinyurl fix done though.

adamgordonbell (approved):

PR #1539, opened 2022-01-05
Add a retry for starting `dockerd` when using `WITH DOCKER`

dchw:
Under heavy load, `dockerd` can't start `containerd` in time. The timeout is not configurable, so retries to catch it under lesser load seem simple enough.

The alternative would be to start `containerd` ourselves. We can go there and make sure its compatible... but it seems easier to let the dependency manage itself if we can get away with it.

### DIFF
--- a/buildkitd/dockerd-wrapper.sh
+++ b/buildkitd/dockerd-wrapper.sh
@@ -32,7 +32,7 @@ execute() {
             break
         else
             sleep 5
-      fi
+        fi
     done
 
     load_images
--- a/buildkitd/dockerd-wrapper.sh
+++ b/buildkitd/dockerd-wrapper.sh
@@ -67,7 +67,7 @@ EOF
     # Start with a rm -rf to make sure a previous interrupted build did not leave its state around.
     rm -rf "$EARTHLY_DOCKERD_DATA_ROOT"
     mkdir -p "$EARTHLY_DOCKERD_DATA_ROOT"
-    dockerd --data-root="$EARTHLY_DOCKERD_DATA_ROOT" --bip=172.20.0.1/16 >/var/log/docker.log 2>&1 &
+    dockerd --data-root="$EARTHLY_DOCKERD_DATA_ROOT" --bip=172.20.0.1/16 --log-level debug >/var/log/docker.log 2>&1 &
     dockerd_pid="$!"
     i=1
     timeout=300
--- a/buildkitd/dockerd-wrapper.sh
+++ b/buildkitd/dockerd-wrapper.sh
@@ -24,6 +24,7 @@ execute() {
         exit 1
     fi
 
+    # shellcheck disable=SC2015
     for i in 1 2 3 4 5; do start_dockerd && break || sleep 15; done
     load_images
     if [ "$EARTHLY_START_COMPOSE" = "true" ]; then
--- a/buildkitd/dockerd-wrapper.sh
+++ b/buildkitd/dockerd-wrapper.sh
@@ -31,6 +31,10 @@ execute() {
         if start_dockerd; then
             break
         else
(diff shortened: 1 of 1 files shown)


buildkitd/dockerd-wrapper.sh
```diff
@@ -24,7 +24,17 @@ execute() {
         exit 1
     fi
 
-    start_dockerd
+    # Sometimes, when dockerd starts containerd, it doesn't come up in time. This timeout is not configurable from
+    # dockerd, therefore we retry... since most instances of this timeout seem to be related to networking or scheculing
```
alexcb, line 28:
```suggestion
    # dockerd, therefore we retry... since most instances of this timeout seem to be related to networking or scheduling
```

buildkitd/dockerd-wrapper.sh
```diff
@@ -88,7 +98,7 @@ EOF
             cat /var/log/docker.log
             echo "==== End dockerd logs ===="
             echo "If you are having trouble running docker, try using the official earthly/dind image instead"
-            exit 1
+            return 1
```
alexcb, line 101: is this intentionally? Does this mean we need to ensure `execute()` checks for the return value now?

vladaionescu (commented):
This looks good, but I'm slightly worried that this will make legitimate failures really slow.

Is there any way to test for the specific failure that we want to retry on and let everything else fail immediately?

buildkitd/dockerd-wrapper.sh
```diff
@@ -24,7 +24,17 @@ execute() {
         exit 1
     fi
 
-    start_dockerd
+    # Sometimes, when dockerd starts containerd, it doesn't come up in time. This timeout is not configurable from
+    # dockerd, therefore we retry... since most instances of this timeout seem to be related to networking or scheduling
+    # when many WITH DOCKER commands are also running. Logs are printed for each failure.
+    for i in 1 2 3 4 5; do
+      if start_dockerd; then
```
alexcb, line 31: nit: indentation is inconsistent with the rest of the file (which uses 4 spaces).

buildkitd/dockerd-wrapper.sh
dchw: Yes, since exit... exits the script, and we cant retry it then. Unless we wrap the wrapper. Lets not go there.

dchw:
> This looks good, but I'm slightly worried that this will make legitimate failures really slow.
> 
> Is there any way to test for the specific failure that we want to retry on and let everything else fail immediately?

It would be really, really fragile, since it would be checking for specific error strings. We can do that if you want... its going to make the retry logic and surrounding stuff ickier though.

buildkitd/dockerd-wrapper.sh
dchw: I'll add an exit if retries are exhausted here instead.

brandonSc (approved):

buildkitd/dockerd-wrapper.sh
```diff
@@ -35,7 +35,15 @@ execute() {
                 # Exiting here on the last retry maintains prior behavior of exiting when this cant start.
                 exit 1
             fi
-            sleep 5
+
+            if grep -q "^failed to start containerd: timeout waiting for containerd to start$" /var/log/docker.log; then
```
alexcb, line 39: do we need to clear `/var/log/docker.log` between runs, so that when we grep it, we're only looking at output from the most recent run? maybe we need a `> /var/log/docker.log` here.

alexcb (approved):

buildkitd/dockerd-wrapper.sh
dchw: I checked with a test script and the log is overwritten with our current syntax.

PR #1542, opened 2022-01-05
Add more best practices

vladaionescu:


alexcb (approved):

docs/best-practices/best-practices.md
```diff
@@ -833,6 +835,67 @@ RUN go build ... && \
...
+    └── Earthfile
+```
+
+```Dockerfile
+# ./dir2/Earthfile
+# Bad: does not work
+COPY ../dir1/some-file.txt ./
+```
+
+In the above example, the file `some-file.txt` is copied from the sibling directory `dir1`. This will not work in Earthly as the file is not in the build context of `./dir2/Earthfile` (the build context in this case is `./dir2`). To address this issue, we can create an Earthfile in `dir1` that exports the file `some-file.txt` as an artifact.
```
brandonSc, line 856: @adamgordonbell this might approach might work for the activity tracker

adamgordonbell: yeah, good point! I might wait for the expanded build context stuff, but this would be a good backup plan. I didn't think of this.

vladaionescu: The expanded build context thingy might take longer... there are some key limitations with the prototype that we have so far, and we're not sure how to properly implement it.

Issue #1543, opened 2022-01-06
exporting outputs fails with "number of platforms does not match references" error for platform-specific builds


misterdjules:
**See some detailed information about this issue at https://github.com/misterdjules/earthly-platform-specific-repro.**

Earthly version 0.5.24 doesn't seem to be impacted, but version 0.6.2 is. I haven't tested versions between those two.

Quoting the README from https://github.com/misterdjules/earthly-platform-specific-repro:

When building a platform-specific image using a `COPY` command that copies the
artifact from a target that uses the `DO` command, and saving that image when
using `--push`, the earthly command sometimes fails with the following error
message:

```
Error: build target: build main: bkClient.Build: failed to solve: number of platforms does not match references 2 1
```

### Reproducing locally on MacOS

1. Make sure to use Earthly version 0.6.2.

2. Run `earthly --platform=linux/amd64 --push +images` from the root of this
   repository. Since the failure is not consistent, you might have to run this
   command at least 10 times before reproducing the problem. Running this
   command in a `while` loop is a good way to reproduce this quickly.

Note that I wasn't able to reproduce the problem when not passing a specific
platform using the `--platform` CLI option.

I was also not able to reproduce the problem when not using `--push`.

### Contributing factors

It seems that the problem comes from using a user-defined command in the Earthfile. Replacing:

```
COPY +usercommand-target/output.js .
```

with:

```
COPY ./output.js .
```

in this repository's `Earthfile` seems to make all runs succeed consistently.

vladaionescu:
Thanks for the detailed report - this is super useful. I'll play around with it today.

vladaionescu:
Ok, PR with fix is up.

This was actually not entirely deterministic. But the reproduction rate was pretty high. Adding `--no-cache` improves the reproduction rate, as a randomly successful run would not be cached that way.

The cause of this was the conversion parallelism feature and the way it detected whether two targets are the same. Conversion parallelism was eagerly issuing a duplicate ahead-of-time conversion of the target `+foo` (working as intended). But it seems that the ARGs within the UDCs were being considered as the ARGs of the target. Since these were being overridden in the UDC, it was looking like the target is being called with default values, but it had been previously called with overridden values, causing the same target to look like it were a different one. But these weren't the ARGs of the target - these were the ARGs of the UDC - so this assumption was incorrect.

[...]

PR #1544, opened 2022-01-06
Fix race condition caused by UDC args incorrectly causing the same target to look like a different one

vladaionescu:
Fixes https://github.com/earthly/earthly/issues/1543

Also, improve the error message when the same image is attempted to be pushed from more than one target

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -247,7 +247,7 @@ func (b *Builder) convertAndBuild(ctx context.Context, target domain.Target, opt
 				}
 
 				if !isMultiPlatform[saveImage.DockerTag] {
-					if singPlatImgNames[saveImage.DockerTag] {
+					if _, found := singPlatImgNames[saveImage.DockerTag]; found {
 						return nil, errors.Errorf(
 							"image %s is defined multiple times for the same default platform",
 							saveImage.DockerTag)
@@ -284,7 +284,7 @@ func (b *Builder) convertAndBuild(ctx context.Context, target domain.Target, opt
 					if err != nil {
 						return nil, err
 					}
-					if platformImgNames[platformImgName] {
+					if _, found := platformImgNames[platformImgName]; found {
 						return nil, errors.Errorf(
 							"image %s is defined multiple times for the same platform (%s)",
 							saveImage.DockerTag, platformImgName)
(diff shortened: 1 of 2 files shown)


builder/builder.go
```diff
@@ -245,6 +247,13 @@ func (b *Builder) convertAndBuild(ctx context.Context, target domain.Target, opt
 				}
 
 				if !isMultiPlatform[saveImage.DockerTag] {
+					if singPlatImgNames[saveImage.DockerTag] {
```
alexcb, line 250: nit: since we're using `singPlatImgNames` as a set, perhaps this would make it more obvious:

```suggestion
					if _, found := singPlatImgNames[saveImage.DockerTag]; found {
```

it would also allow us to change to `singPlatImgNames := map[string]struct{}` down the road if we wanted to: https://itnext.io/set-in-go-map-bool-and-map-struct-performance-comparison-5315b4b107b

(but given the size of the set in this code, it's pre-mature optimization)

builder/builder.go
```diff
@@ -271,6 +280,16 @@ func (b *Builder) convertAndBuild(ctx context.Context, target domain.Target, opt
 				} else {
 					platform := llbutil.PlatformWithDefault(sts.Platform)
 					platformStr := llbutil.PlatformWithDefaultToString(sts.Platform)
+					platformImgName, err := platformSpecificImageName(saveImage.DockerTag, platform)
+					if err != nil {
+						return nil, err
+					}
+					if platformImgNames[platformImgName] {
```
alexcb, line 287: same as above comment.
```suggestion
					if _, found := platformImgNames[platformImgName]; found {
```

alexcb (approved):

alexcb (approved):

adamgordonbell:
We have another report of this, but without using UDCs. I am wondering if it's the same issue or not.

https://earthlycommunity.slack.com/archives/C01DL2928RM/p1641964014031300

In this case, there are 1 image be produced for two platforms, and a third image be pushed as well and the error is:
`Error: build target: build main: bkClient.Build: failed to solve: number of platforms does not match references 3 2`

@vladaionescu  do you think that is a separate issue? I have the Earthfile, but can't share it publicly.

PR #1548, opened 2022-01-07
fixes IF statements not working as non-root

### LINKED ISSUE #1547: `IF` statement does not execute properly in a non-root container
The inverse is actually easier to state: that the `IF` statement only works in root-run containers.  It just seemed better to phrase it as a bug (that is doesn't work as expected/desired).

This appears fairly straightforward, so I won't go into too much detail.

Minimal reproduction (with 0.6.2):
```
FROM alpine
RUN adduser --system peon
USER peon

as-non-root:
    RUN whoami
    IF [ -f /etc/passwd ]
        RUN echo "got file as peon"
    END

as-root:
    USER root
    RUN whoami
    IF [ -f /etc/passwd ]
        RUN echo "got file as root"
    END
```
Executing the `as-root` target succeeds (and echos) as expected.  I would expect the behavior for the `as-non-root` target to be identical.  However:
```
$ earthly +as-non-root 

 1. Init 🚀

 2. Build 🔧

               +base | --> FROM alpine
               +base | [██████████] 100% resolve docker.io/library/alpine@sha256:21a3deaa0d32a8057914f36584b5288d2e5ecc984380bc0118285c70fa8c9300
               +base | *cached* --> RUN adduser --system peon
        +as-non-root | --> RUN whoami
        +as-non-root | peon
        +as-non-root | --> IF [ -f /etc/passwd ]
        +as-non-root | /bin/sh: can't create /run/exit_code: Permission denied
        +as-non-root | Command /bin/sh -c '[ -f /etc/passwd ] ; echo $? >'"'"'/run/exit_code'"'"'' failed with exit code 1
        +as-non-root | 
        +as-non-root | ERROR: Command exited with non-zero code: IF [ -f /etc/passwd ]

        +as-non-root *failed* | --> IF [ -f /etc/passwd ]
        +as-non-root *failed* | /bin/sh: can't create /run/exit_code: Permission denied
        +as-non-root *failed* | Command /bin/sh -c '[ -f /etc/passwd ] ; echo $? >'"'"'/run/exit_code'"'"'' failed with exit code 1
        +as-non-root *failed* |         +as-non-root *failed* | ERROR: Command exited with non-zero code: IF [ -f /etc/passwd ]
Error: build target: build main: bkClient.Build: failed to solve: process "/bin/sh -c PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin /usr/bin/earth_debugger /bin/sh -c '[ -f /etc/passwd ] ; echo $? >'\"'\"/run/exit_code\"'\"''" did not complete successfully: exit code: 1
```

It seems fairly obvious what's happening and why (i.e., https://github.com/earthly/earthly/blob/main/earthfile2llb/converter.go#L530), but will leave that for the actual devs to confirm.

That said, though this behavior is undesirable and unexpected... is it *really* a bug?  Or is my usage of a non-root container (for building) actually an anti-pattern in the Earthy world, thus more of a "we just never documented this behavior since nobody would actually do that"?

alexcb:
The IF command stores the exit code to a file.
This file must be writable by the current user.
/run is only writable by root; however /tmp
should be writable by all users (unless an earthly user goes out of
their way to change the permissions of /tmp).

Fixes #1547

### DIFF
--- a/earthfile2llb/converter.go
+++ b/earthfile2llb/converter.go
@@ -592,7 +592,7 @@ func (c *Converter) RunExpression(ctx context.Context, expressionName string, op
 		outputFile = path.Join(srcBuildArgDir, expressionName)
 		opts.statePrep = func(ctx context.Context, state pllb.State) (pllb.State, error) {
 			return state.File(
-				pllb.Mkdir(srcBuildArgDir, 0755, llb.WithParents(true)),
+				pllb.Mkdir(srcBuildArgDir, 0777, llb.WithParents(true)), // Mkdir is performed as root even when USER is set; we must use 0777
 				llb.WithCustomNamef("[internal] mkdir %s", srcBuildArgDir)), nil
 		}
 	}
--- a/earthfile2llb/converter.go
+++ b/earthfile2llb/converter.go
@@ -527,7 +527,7 @@ func (c *Converter) RunExitCode(ctx context.Context, opts ConvertRunOpts) (int, 
 			return os.RemoveAll(exitCodeDir)
 		})
 	} else {
-		exitCodeFile = "/run/exit_code"
+		exitCodeFile = "/tmp/earthly_if_statement_exit_code"
 		opts.statePrep = func(ctx context.Context, state pllb.State) (pllb.State, error) {
 			return state.File(
 				pllb.Mkdir("/run", 0755, llb.WithParents(true)),
--- a/ast/tests/build-arg.ast.json
+++ b/ast/tests/build-arg.ast.json
@@ -104,6 +104,14 @@
             ],
             "name": "BUILD"
           }
+        },
+        {
+          "command": {
+            "args": [
+              "+test-non-root"
+            ],
+            "name": "BUILD"
+          }
         }
       ]
     },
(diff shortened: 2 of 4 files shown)


earthfile2llb/converter.go
```diff
@@ -527,7 +527,7 @@ func (c *Converter) RunExitCode(ctx context.Context, opts ConvertRunOpts) (int,
 			return os.RemoveAll(exitCodeDir)
 		})
 	} else {
-		exitCodeFile = "/run/exit_code"
+		exitCodeFile = "/tmp/earthly_if_statement_exit_code"
```
vladaionescu, line 530: IIRC, we use `/run/` for other things too. Mind double-checking that this is the only spot?

vladaionescu (approved):

earthfile2llb/converter.go
alexcb: I was able to fix `ARG` too.

However, I see that `WITH DOCKER` doesn't work -- which makes a bit of sense since docker typically requires root access (or for the user to be part of the docker group, which doesn't exist under our DIND image).

I additionally created #1549 for creating ssh auth sockets which isn't a quick fix.

vladaionescu (approved):

PR #1551, opened 2022-01-09
Args in base target are optionally global and should be explicitly set if needed

ido-namely:
With this PR, to make an ARG global (in the base target only), it should be explicitly set with a `--global` flag.
(hidden behind a feature flag)


ido-namely:
Hi @vladaionescu ,
Following our discussion in #1518 I've attempted to implement the explicit global arg feature behind a feature flag.
Forcing the args to be global explicitly was pretty straight forward, but what I wasn't sure of was 
whether I set the non global args in the base target as overrides properly.
This is in order to have the effect of the args at the top of the file behaving as build-args passed to the earthly cli either from the command or from the environment variables.

Please let me know your thoughts.
Thanks!

ido-namely:
@vladaionescu , @alexcb just a friendly reminder in case you missed this PR.
I'd love to hear your thoughts about whether the idea and implementation make sense or not.
Thanks again!

vladaionescu:
Hi @ido-namely - thanks for reminding me. I lost track of PRs for a moment.

One thing I'm not sure about is this part:

> If a flag in the base target is not set as global, it is considered as an in-file override, meaning it's would be as if the arg was passed to the earthly cli.

I feel that this could lead to confusion as it's basically an `ARG` instruction that has an additional effect if it's placed in the base target. This can be too surprising for the user (especially if other targets have their own `FROM`s but not only).

earthfile2llb/interpreter.go
```diff
@@ -1026,15 +1026,29 @@ func (i *Interpreter) handleEnv(ctx context.Context, cmd spec.Command) error {
...
+func parseArgArgs(ctx context.Context, cmd spec.Command, isBaseTarget bool, explicitGlobal bool) (argOpts, string, *string, error) {
 	opts := argOpts{}
 	args, err := flagutil.ParseArgs("ARG", &opts, getArgsCopy(cmd))
 	if err != nil {
 		return argOpts{}, "", nil, err
 	}
+	if opts.Global {
+		// since the global flag is part of the struct, we need to manually return parsing error if it's used while the feature flag is off
+		if !explicitGlobal {
+			return argOpts{}, "", nil, errors.New("unknown flag `global'")
```
vladaionescu, line 1042:
```suggestion
			return argOpts{}, "", nil, errors.New("unknown flag --global")
```

earthfile2llb/interpreter.go
```diff
@@ -1026,15 +1026,29 @@ func (i *Interpreter) handleEnv(ctx context.Context, cmd spec.Command) error {
 
 var errInvalidSyntax = errors.New("invalid syntax")
 var errRequiredArgHasDefault = errors.New("required ARG cannot have a default value")
+var errGlobalArgNotInBase = errors.New("global ARG can only be set in the base target")
 
 // parseArgArgs parses the ARG command's arguments
 // and returns the argOpts, key, value (or nil if missing), or error
-func parseArgArgs(ctx context.Context, cmd spec.Command) (argOpts, string, *string, error) {
+func parseArgArgs(ctx context.Context, cmd spec.Command, isBaseTarget bool, explicitGlobal bool) (argOpts, string, *string, error) {
```
vladaionescu, line 1033: I would name `explicitGlobal` something like `explicitGlobalFeature` - just to make it clear that it represents the feature flag.

features/features.go
```diff
@@ -25,6 +25,7 @@ type Features struct {
 	NoImplicitIgnore           bool `long:"no-implicit-ignore" description:"disable implicit ignore rules to exclude .tmp-earthly-out/, build.earth, Earthfile, .earthignore and .earthlyignore when resolving local context"`
 	CheckDuplicateImages       bool `long:"check-duplicate-images" description:"check for duplicate images during output"`
 	EarthlyVersionArg          bool `long:"earthly-version-arg" description:"includes EARTHLY_VERSION and EARTHLY_BUILD_SHA ARGs"`
+	ExplicitGlobal             bool `long:"explicit-global" description:"require base target args to have explicit settings to be considered global args"`
```
vladaionescu, line 28: Mind enabling this in `0.7`? Look for `case versionAtLeast(ftrs, 0, 7):` in this file.

vladaionescu (approved):
This is looking great! Just a couple minor comments.

@alexcb should also take a look.

tests/build-arg-explicit-global.earth
```diff
@@ -0,0 +1,35 @@
...
+    RUN test "$local"   == ""
+
+test-global3:
+    FROM +dummy
+    RUN test "$global1" == "abc"
+    RUN test "$global2" == "def"
+    RUN test "$local"   == ""
+
+test-failure:
+    ARG --global global1=123
```
alexcb, line 35: nit: our linter will fail on this not ending with a newline.

tests/Earthfile
```diff
@@ -798,6 +806,10 @@ command:
     RUN echo "hello command" >./message.txt
     DO +RUN_EARTHLY --earthfile=command.earth
 
+command-explicit-global:
+    RUN echo "hello command-explicit-global" >./message.txt
```
alexcb, line 810: is this line required?

tests/build-arg-explicit-global.earth
ido-namely: thanks, I had a problem running all the tests locally so I waited for the CI to let me know how I was doing with the tests.

features/features.go
ido-namely: sure, so just to confirm, keep it as a feature flag for 0.6 but enable by default for 0.7?
Would it require a separate test file which will use 0.7?

tests/Earthfile
```diff
@@ -264,6 +266,12 @@ secrets-test:
 build-arg-test:
     DO +RUN_EARTHLY --earthfile=build-arg.earth
 
+build-arg-explicit-global-test:
+    DO +RUN_EARTHLY --earthfile=build-arg-explicit-global.earth
+    DO +RUN_EARTHLY --should_fail=true --earthfile=build-arg-explicit-global.earth \
```
alexcb, line 271: The DO and RUN grep can be combined into a single statement:
```
    DO +RUN_EARTHLY --should_fail=true --earthfile=build-arg-explicit-global.earth \
        --target=+test-failure \
        --output_contains="invalid ARG arguments*: global ARG can only be set in the base target"
```

tests/Earthfile
ido-namely: probably not, I think I thought this is some log, but I realize it was probably used by some test from the original target (`command`).

features/features.go
vladaionescu: Yes, exactly. No additional test needed for 0.7 for now.

tests/Earthfile
alexcb: yeah, it's used in the `+command` test: https://github.com/earthly/earthly/blob/main/tests/command.earth#L26 
this line can be removed.

earthfile2llb/interpreter.go
ido-namely: the error that is produced by the parser does not include `--`, so I was just trying to be consistent (including the backtick and the single quote)

vladaionescu: Oh I see - I think `--global` is better. The parser is sometimes a little weird anyway.

ido-namely:
@alexcb I think this is ready for another review. Thanks!

PR #1555, opened 2022-01-10
`COMMAND`s containing `LOCALLY ` running in incorrect directory

alexcb:
a `COMMAND` should always run in the same current directory as where they are called from (not where they are defined).

This PR fixes a bug where the COMMAND is executed within the directory where it was defined.


earthfile2llb/interpreter.go
```diff
@@ -631,6 +633,22 @@ func (i *Interpreter) handleFromDockerfile(ctx context.Context, cmd spec.Command
...
+	if n > 0 {
+		return i.workingDirStack[n-1], nil
+	}
+
+	workingDir, err := filepath.Abs(filepath.Dir(cmd.SourceLocation.File))
+	if err != nil {
+		return "", i.wrapError(err, cmd.SourceLocation, "failed to get absolute path for %s", cmd.SourceLocation.File)
+	}
+	return workingDir, nil
+}
```
vladaionescu, line 650: I would pass the source file location via the build context, from the resolver. It would save us from the complexity of tracking this stack here.

It is already returned here: https://github.com/earthly/earthly/blob/main/buildcontext/resolver.go#L31. We just need to plug it into the converter.

The converter already gets the build context data in its constructor: https://github.com/earthly/earthly/blob/main/earthfile2llb/converter.go#L92.

brandonSc: Would this new method also be a better way to get the working dir than what I had previously in `handleCache` below?
i.e.
```golang
dir = path.Clean(path.Join("/", i.converter.mts.Final.MainImage.Config.WorkingDir, dir))
```

vladaionescu: Not really... here we're interested in the dir where the Earthfile exists on the host, whereas `dir = path.Clean(path.Join("/", i.converter.mts.Final.MainImage.Config.WorkingDir, dir))` seems to be the work dir within the build environment.

vladaionescu (approved):

earthfile2llb/interpreter.go
brandonSc: I misunderstood that, thanks

brandonSc (approved):

earthfile2llb/interpreter.go
alexcb: > I would pass the source file location via the build context, from the resolver. It would save us from the complexity of tracking this stack here.

That's much simpler -- good suggestion. It turns out that the `Locally()` method took a workingDir as an argument, which is no longer needed.

> Would this new method also be a better way to get the working dir than what I had previously in handleCache below?

keeping the existing code for `handleCache` is good; `i.converter.mts.Final.MainImage.Config.WorkingDir` is relative to the containers, where as this PR only deals with the special case of local file system paths (which are required for `LOCALLY`).

PR #1557, opened 2022-01-11
`COPY --if-exists` Bug

adamgordonbell:
`COPY --if-exists` doesn't work if the path starts with './'

Adding a test case for this and adjusting the wildcard logic for if-exist to take this into account 

I'm not sure how to update the if-exists.ast.json?

Original PR: https://github.com/earthly/earthly/pull/683


util/llbutil/copyop.go
```diff
@@ -33,6 +33,11 @@ func CopyOp(srcState pllb.State, srcs []string, destState pllb.State, dest strin
 			// the filepath.Match syntax, so by simply creating a wildcard where the
 			// first letter needs to match the current first letter gets us the single
 			// match; and no error if it is missing.
+
+			//Normalize path by dropping './'
+			if src[0:2] == "./" {
+				src = src[2:]
+			}
```
vladaionescu, line 40:
```suggestion
			// Normalize path by dropping "./"
			strings.TrimPrefix(src, "./")
```

(You'll need to add the matching `"strings"` import)

vladaionescu (commented):
👍 

To update AST tests expectations, you can run `earthly ./ast/tests+update-all`

vladaionescu (approved):

PR #1559, opened 2022-01-11
Build Speed up

adamgordonbell:
This PR cuts `+all-buildkitd` into 3 steps in GHA.
This introduces some duplication but seems to take build time from 60 minutes to 30 minutes.

See here:
https://github.com/earthly/earthly/actions/runs/1684389244


alexcb (approved):
nice speedup change!

PR #1561, opened 2022-01-12
docs: FROM dockerfile with different build context

alexcb:
Adds an example of how to use FROM dockerfile with a different build
context.

### DIFF
--- a/docs/earthfile/earthfile.md
+++ b/docs/earthfile/earthfile.md
@@ -268,7 +268,7 @@ The `<mount-spec>` is defined as a series of comma-separated list of key-values.
 
 ###### Examples:
 
-Persisting cache for a single `RUN` command, even when it's dependencies change:
+Persisting cache for a single `RUN` command, even when its dependencies change:
 
 ```Dockerfile
 ENV GOCACHE=/go-cache
@@ -736,6 +736,35 @@ The `<context-path>` is the path where the Dockerfile build context exists. By d
 ##### `-f <dockerfile-path>`
 
 Specify an alternative Dockerfile to use. The `<dockerfile-path>` can be either a path on the host system, relative to the current Earthfile, or an [artifact reference](../guides/target-ref.md#artifact-reference) pointing to a Dockerfile.
+
+{% hint style='info' %}
+It is possible to split the `Dockerfile` and the build context across two seperate [artifact references](../guides/target-ref.md#artifact-reference):
+
+```Dockerfile
+FROM alpine
+
+mybuildcontext:
+    WORKDIR /mydata
+    RUN echo mydata > myfile
+    SAVE ARTIFACT /mydata
+
+mydockerfile:
+    RUN echo "
+FROM busybox
+COPY myfile .
+RUN cat myfile" > Dockerfile
+    SAVE ARTIFACT Dockerfile
+
+docker:
+    FROM DOCKERFILE -f +mydockerfile/Dockerfile +mybuildcontext/mydata/*
+    SAVE IMAGE testimg:latest
+```
+
+Note that `+mybuildcontext/mydata` on its own would copy the directory _and_ its contents; where as `+mybuildcontext/mydata/*` is required to copy all of the contents from within the `mydata` directory (
+without copying the wrapping `mydata` directory).
+
+If both the `Dockerfile` and build context are inside the same target, one must reference the same target twice, e.g. `FROM DOCKERFILE -f +target/dir/Dockerfile +target/dir`.
+{% endhint %}
 
 ##### `--build-arg <key>=<value>`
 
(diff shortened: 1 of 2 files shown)


docs/earthfile/earthfile.md
```diff
@@ -737,6 +737,35 @@ The `<context-path>` is the path where the Dockerfile build context exists. By d
 
 Specify an alternative Dockerfile to use. The `<dockerfile-path>` can be either a path on the host system, relative to the current Earthfile, or an [artifact reference](../guides/target-ref.md#artifact-reference) pointing to a Dockerfile.
 
+{% hint style='info' %}
+It is possible to split the `Dockerfile` and the build context across two seperate [artifact references](../guides/target-ref.md#artifact-reference):
+
+```Earthly
```
vladaionescu, line 743: This won't work in gitbook unfortunately... maybe use `Dockerfile` until it works?

docs/earthfile/earthfile.md
```diff
@@ -737,6 +737,35 @@ The `<context-path>` is the path where the Dockerfile build context exists. By d
...
+    SAVE ARTIFACT Dockerfile
+
+docker:
+    FROM DOCKERFILE -f +mydockerfile/Dockerfile +mybuildcontext/mydata/*
+    SAVE IMAGE testimg:latest
+```
+
+Alternatively, if both the `Dockerfile` and build context are inside the same target, one must reference the same target twice, e.g. `FROM DOCKERFILE -f +target/dir/Dockerfile +target/dir`.
+
+Other specifications may currently face issues as documented in [#1410](https://github.com/earthly/earthly/issues/1410).
```
vladaionescu, line 765: I would perhaps just say that `+mybuildcontext/mydata` (copies the whole dir, including the wrapping dir `mydata`) is not enough and that `/*` is necessary (copies the contents of the dir only), rather than reference this issue.

docs/earthfile/earthfile.md
```diff
@@ -737,6 +737,35 @@ The `<context-path>` is the path where the Dockerfile build context exists. By d
...
+RUN cat myfile" > Dockerfile
+    SAVE ARTIFACT Dockerfile
+
+docker:
+    FROM DOCKERFILE -f +mydockerfile/Dockerfile +mybuildcontext/mydata/*
+    SAVE IMAGE testimg:latest
+```
+
+Note that `+mybuildcontext/mydata` on its own would copy the directory _and_ its contents; where as `+mybuildcontext/mydata/*` is required to copy all of the contents from within the `mydata` directory (
+without copying the wrapping mydata directory).
```
vladaionescu, line 764:
```suggestion
without copying the wrapping `mydata` directory).
```

docs/earthfile/earthfile.md
```diff
@@ -737,6 +737,35 @@ The `<context-path>` is the path where the Dockerfile build context exists. By d
...
+
+If both the `Dockerfile` and build context are inside the same target, one must reference the same target twice, e.g. `FROM DOCKERFILE -f +target/dir/Dockerfile +target/dir`.
```
vladaionescu, line 766: Doesn't `FROM DOCKER +target/dir/*` work in this case? Also doesn't this still require the `/*` anyway?

Perhaps a better example is for the case when the Dockerfile is named differently, so that the `-f` is justified:

```suggestion
If both the `Dockerfile` and build context are inside the same target, one must reference the same target twice, e.g. `FROM DOCKERFILE -f +target/dir/my-service.Dockerfile +target/dir/*`.
```

alexcb: > Doesn't FROM DOCKER +target/dir/* work in this case?

```
FROM alpine

dir:
    RUN echo "
FROM busybox
COPY myfile .
RUN cat myfile" > Dockerfile
    RUN echo mydata > myfile
    SAVE ARTIFACT myfile
    SAVE ARTIFACT Dockerfile

docker:
    FROM DOCKERFILE +dir/*
    SAVE IMAGE testimg:latest
```

this results in 

```
Error: build target: build main: bkClient.Build: failed to solve: open /tmp/buildkit-mount1231517943/*/Dockerfile: no such file or directory
```

and similary, changing it to 

```
FROM alpine

dir:
    RUN echo "
FROM busybox
COPY myfile .
RUN cat myfile" > Dockerfile
    RUN echo mydata > myfile
    SAVE ARTIFACT myfile
    SAVE ARTIFACT Dockerfile

docker:
    FROM DOCKERFILE +dir
    SAVE IMAGE testimg:latest
```

produces

```
Error: build target: build main: bkClient.Build: failed to solve: Earthfile line 13:4 from dockerfile: resolve build context for dockerfile: unable to open "+dir/Dockerfile": open +dir/Dockerfile: no such file or directory
```

I suspect the two above errors were what led the initial github issues to request a better example for theses.

vladaionescu: Ah interesting.

Now that I think about it a bit more it feels like the second one not working should be a bug...

vladaionescu (approved):

docs/earthfile/earthfile.md
alexcb: I opened https://github.com/earthly/earthly/issues/1592 to track it.

PR #1569, opened 2022-01-12
Raise error if builtin ARG has value set by user

alexcb:
builtin ARGs can not be overridden. This introduces
an explicit check and raises an error. Previously
the overridden value was simply ignored.
    
### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -109,7 +109,11 @@ lint:
             echo "$output" ; \
             exit 1 ; \
         fi
+    # names.go defines some very obvious consts that to not need comments; however golint doesn't support disabling rules: https://github.com/golang/lint/issues/263
+    # therefore, we will hide this file from golint, and restore it after.
+    RUN mv variables/reserved/names.go variables/reserved/names.skip-go-lint && echo "package reserved" > variables/reserved/names.go
     RUN golint -set_exit_status ./...
+    RUN mv variables/reserved/names.skip-go-lint variables/reserved/names.go
     RUN output="$(go vet ./... 2>&1)" ; \
         if [ -n "$output" ]; then \
             echo "$output" ; \
--- a/variables/builtin.go
+++ b/variables/builtin.go
@@ -1,17 +1,17 @@
 package variables
 
 import (
-	"fmt"
 	"strings"
 
 	"github.com/containerd/containerd/platforms"
+	specs "github.com/opencontainers/image-spec/specs-go/v1"
+
 	"github.com/earthly/earthly/domain"
 	"github.com/earthly/earthly/features"
-	"github.com/earthly/earthly/states/dedup"
 	"github.com/earthly/earthly/util/gitutil"
 	"github.com/earthly/earthly/util/llbutil"
 	"github.com/earthly/earthly/util/stringutil"
-	specs "github.com/opencontainers/image-spec/specs-go/v1"
+	arg "github.com/earthly/earthly/variables/reserved"
 )
 
 // DefaultArgs contains additional builtin ARG values which need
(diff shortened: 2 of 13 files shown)


alexcb:
After typing out the various suggestions, I'm thinking the 3rd option might be best as it keeps the code bloat to a minimum.

variables/builtin.go
```diff
@@ -21,23 +21,54 @@ type DefaultArgs struct {
...
 
+const (
+	VarNameEarthlyTarget             = "EARTHLY_TARGET"
+	VarNameEarthlyTargetProject      = "EARTHLY_TARGET_PROJECT"
+	VarNameEarthlyTargetProjectNoTag = "EARTHLY_TARGET_PROJECT_NO_TAG"
+	VarNameEarthlyTargetName         = "EARTHLY_TARGET_NAME"
+	VarNameEarthlyTargetTag          = "EARTHLY_TARGET_TAG"
+	VarNameEarthlyTargetTagDocker    = "EARTHLY_TARGET_TAG_DOCKER"
+	VarNameEarthlyVersion            = "EARTHLY_VERSION"
+	VarNameEarthlyBuildSha           = "EARTHLY_BUILD_SHA"
```
vladaionescu, line 32: Note that there are also the target platform args and the user platform args.

vladaionescu: There's a complete set here: https://github.com/earthly/earthly/blob/main/states/dedup/targetinput.go#L144

I wonder if we should unify the definition here with the set in `targetinput.go`. Ideally, they should be defined in a single common location. It might have to be a new package though - IIRC there was some circular dependency issue.

earthfile2llb/interpreter.go
```diff
@@ -1040,6 +1041,9 @@ func parseArgArgs(ctx context.Context, cmd spec.Command) (argOpts, string, *stri
 		if args[1] != "=" {
 			return argOpts{}, "", nil, errInvalidSyntax
 		}
+		if variables.IsBuiltIn(args[0]) {
+			return argOpts{}, "", nil, errBuiltInArgHasDefault
+		}
```
vladaionescu, line 1046: I would put this check in `variables.ParseArgs` rather than in the interpreter, because ParseArgs is already called from this common point: https://github.com/earthly/earthly/blob/main/earthfile2llb/converter.go#L1296, that is used whenever another target is referenced (FROM, COPY, BUILD, WITH DOCKER --load, etc).

vladaionescu: The other relevant place to add the check to is `variables.ParseCommandLineArgs` (for the args passed via the command-line). This is called from here: https://github.com/earthly/earthly/blob/main/cmd/earthly/main.go#L2693.

variables/builtin.go
alexcb: Thanks for pointing that out. I see `variables/builtin.go` imports `dedup`.

I can try creating a new package such as `variables/reserved`, and move list all the strings there as consts.

vladaionescu: Makes sense!

earthfile2llb/interpreter.go
vladaionescu: Oh, I just realized, I think we'd also need to do a check in `variables.DeclareArg` (called from here: https://github.com/earthly/earthly/blob/main/earthfile2llb/converter.go#L1057) so that the user cannot assign a default value (which would be ignored) to a builtin variable.

Earthfile
```diff
@@ -109,7 +109,11 @@ lint:
             echo "$output" ; \
             exit 1 ; \
         fi
+    # names.go defines some very obvious consts that to not need comments; however golint doesn't support disabling rules: https://github.com/golang/lint/issues/263
+    # therefore, we will hide this file from golint, and restore it after.
+    RUN mv variables/reserved/names.go variables/reserved/names.skip-go-lint && echo "package reserved" > variables/reserved/names.go
```
alexcb, line 114: The alternative is:

```
const (
        // EarthlyBuildSha is a string
	EarthlyBuildSha             = "EARTHLY_BUILD_SHA"
	// EarthlyGitBranch is a string
	EarthlyGitBranch            = "EARTHLY_GIT_BRANCH"
	// EarthlyGitCommitTimestamp is a string ...
	EarthlyGitCommitTimestamp   = "EARTHLY_GIT_COMMIT_TIMESTAMP"
	EarthlyGitHash              = "EARTHLY_GIT_HASH"
	EarthlyGitOriginURL         = "EARTHLY_GIT_ORIGIN_URL"
	EarthlyGitOriginURLScrubbed = "EARTHLY_GIT_ORIGIN_URL_SCRUBBED"
	EarthlyGitProjectName       = "EARTHLY_GIT_PROJECT_NAME"
	EarthlyGitShortHash         = "EARTHLY_GIT_SHORT_HASH"
	EarthlyGitTag               = "EARTHLY_GIT_TAG"
	EarthlySourceDateEpoch      = "EARTHLY_SOURCE_DATE_EPOCH"
	EarthlyTarget               = "EARTHLY_TARGET"
	EarthlyTargetName           = "EARTHLY_TARGET_NAME"
	EarthlyTargetProject        = "EARTHLY_TARGET_PROJECT"
	EarthlyTargetProjectNoTag   = "EARTHLY_TARGET_PROJECT_NO_TAG"
	EarthlyTargetTag            = "EARTHLY_TARGET_TAG"
	EarthlyTargetTagDocker      = "EARTHLY_TARGET_TAG_DOCKER"
	EarthlyVersion              = "EARTHLY_VERSION"
	TargetArch                  = "TARGETARCH"
	TargetOS                    = "TARGETOS"
	TargetPlatform              = "TARGETPLATFORM"
	TargetVariant               = "TARGETVARIANT"
	UserArch                    = "USERARCH"
	UserOS                      = "USEROS"
	UserPlatform                = "USERPLATFORM"
	UserVariant                 = "USERVARIANT"
)
```

which I find is really difficult to look at.

vladaionescu (approved):

PR #1570, opened 2022-01-14
Add output_contains ARG to RUN_EARTHLY

alexcb:
Introduces a new ARG to RUN_EARTHLY which can be used to verify a string
(grep pattern) exists in the earthly output.

Previously we used post_command to call a perl regex (or grep); however
that approach did not allow us to test the exit code for cases where we
wanted to test that earthly failed AND displayed a correct message in
the log (testing that it failed the correct way).


vladaionescu (approved):
Very nice

PR #1579, opened 2022-01-15
Tests for parallel conversion

vladaionescu:
These tests are meant to verify that the recently fixed parallel conversion bugs don't return.

However, they don't work for their intended purpose currently, as they always pass even if the bugs are reintroduced. The reason is that the bugs reproduce only if the images are exported, and our integration testing capabilities don't support exports yet (they are called with `--no-output`).

This PR is WIP.

Possible way to fix this: run the tests one by one directly from GHA.


vladaionescu:
Abandoned

PR #1593, opened 2022-01-20
--env-file arg to support different .env file

alexcb:
Allow a user to specify an alternative `.env` file to source
environment variables from.

A user may load multiple `.env` files using process
substitution. For example, if a user has `.env` committed to git
containing:

    PUBLIC_VALUE=this-is-public-information
    MY_TOKEN=

And a second `.env.local` which is not committed to git:

    MY_TOKEN=secretvalue

A user can then have earthly combine these two files using:

    earthly --env-file <(cat .env && cat .env.local) +some-target

When multiple values are defined, the last one takes precedence.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to [Earthly](https://github.com/earthly/earthly) will be doc
 
 ## Unreleased
 
+### Added
+
+- Ability to load a different `.env` file via the `--env-file` flag.
+
 ### Changed
 
 - updated buildkit to include changes up to 17c237d69a46d61653746c03bcbe6953014b41a5
--- a/cmd/earthly/main.go
+++ b/cmd/earthly/main.go
@@ -7,6 +7,7 @@ import (
 	"crypto/rand"
 	"encoding/base64"
 	"encoding/json"
+	"flag"
 	"fmt"
 	"io"
 	"net"
@@ -84,9 +85,10 @@ const (
 	DefaultBuildkitdContainerName = "earthly-buildkitd"
 	// DefaultBuildkitdVolumeName is the name of the docker volume used for storing the cache.
 	DefaultBuildkitdVolumeName = "earthly-cache"
-)
 
-var dotEnvPath = ".env"
+	defaultEnvFile = ".env"
+	envFileFlag    = "env-file"
+)
 
 type earthlyApp struct {
 	cliApp      *cli.App
@@ -159,6 +161,7 @@ type cliFlags struct {
 	disableAnalytics          bool
 	featureFlagOverrides      string
 	localRegistryHost         string
+	envFile                   string
 	lsShowLong                bool
 	lsShowArgs                bool
 	containerFrontend         containerutil.ContainerFrontend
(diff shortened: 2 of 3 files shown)


util/flagutil/parse.go
```diff
@@ -42,3 +42,17 @@ func ParseArgsWithValueModifier(command string, data interface{}, args []string,
 	}
 	return res, nil
 }
+
+// GetArgBestEffort returns the next element that follows argName
+func GetArgBestEffort(args []string, argName string) (string, bool) {
```
vladaionescu, line 47: This feels a bit hacky. I'm thinking of an edge case where the value (not the flag itself) of another flag is `--env-file` and this catches that incorrectly?

alexcb: yeah, `--build-arg --env-file +some-target` is such an edge case.

I can take a deeper look at `urfave/cli` to see if there's a different way we can parse once, then re-parse a second time once we pull out the `--env-file` value.

alexcb: It looks like the parsing is done via a private method: https://github.com/urfave/cli/blob/master/app.go#L241 which then calls https://github.com/urfave/cli/blob/e855c4cd1d5f87ae2a12c0e2620c0b747101c2bc/flag.go#L130-L140

I'll see if I can try a similar for loop in earthly where we call `Apply`.

vladaionescu: I'm wondering if it's actually not possible, without knowing the possible options first...

alexcb: It was easier than I thought, the only ugly part is we need to run newEarthlyApp twice:

```
for _, f := range newEarthlyApp(ctx, conslogging.ConsoleLogger{}).cliApp.Flags {
```

but fortunately `newEarthlyApp` is re-entrant.

vladaionescu (approved):

scripts/tests/auth/setup.sh
```diff
@@ -14,9 +14,9 @@ echo "running tests with $earthly"
 "$earthly" account login
 
 # fetch shared secret key (this step assumes your personal user has access to the /earthly-technologies/ secrets org
-ID_RSA=$($earthly secrets get -n /earthly-technologies/github/other-service+github-cinnamon@earthly.dev/id_rsa)
-GITHUB_PASSWORD=$($earthly secrets get -n /earthly-technologies/github/other-service+github-cinnamon@earthly.dev/password)
-REPO_ACCESS_TOKEN=$($earthly secrets get -n /earthly-technologies/github/other-service+github-cinnamon@earthly.dev/repo-access-token)
+#ID_RSA=$($earthly secrets get -n /earthly-technologies/github/other-service+github-cinnamon@earthly.dev/id_rsa)
+#GITHUB_PASSWORD=$($earthly secrets get -n /earthly-technologies/github/other-service+github-cinnamon@earthly.dev/password)
+#REPO_ACCESS_TOKEN=$($earthly secrets get -n /earthly-technologies/github/other-service+github-cinnamon@earthly.dev/repo-access-token)
```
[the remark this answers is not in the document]
alexcb: :facepalm: thanks for catching that.

PR #1598, opened 2022-01-20
Rework server client to use JWT 

brandonSc:
Saves a new JWT to `~/.earthly/auth.v2.token` and migrates old-style tokens from `~/.earthly/auth.token` to `~/.earthly/auth.credentials`. 

The contents of `auth.credentials` are now used as login credentials to the server in order to get the new-style JWT. After that, the JWT is used as authentication to the rest of the server endpoints.

The reason for the `v2` in `auth.v2.token` is to ensure backwards compatibility - i.e. if a user downgrades back to an old version of Earthly, that old version won't try to parse a v2/jwt as an old-style token.

JWT's expire and can be revoked (for example, if user permissions change) - so the logic will refresh the tokens as necessary.

I also did some light refactoring, like renaming the package from `secretsclient` to `cloud` as we are starting to use the server for more features than just secrets.


brandonSc:
Still in draft mode as I have a couple quirks left in the code to work out, and a test failure. It works for the most part though, and the code is pretty settled. Some reviews at this point would be helpful.

cloud/client.go
```diff
@@ -142,35 +148,71 @@ func (c *client) doCall(method, url string, opts ...requestOpt) (int, string, er
...
+		if err := c.Authenticate(); err != nil {
+			return 0, "", errors.Wrap(err, "failed refreshing expired auth token")
+		}
+	}
+
 	var status int
 	var body string
 	var err error
 	duration := time.Millisecond * 100
+	alreadyReuthed := false
```
alexcb, line 161:
```suggestion
	alreadyReAuthed := false
```

cmd/earthly/main.go
```diff
@@ -2379,31 +2379,22 @@ func (app *earthlyApp) actionAccountLogin(c *cli.Context) error {
...
-					err := sc.SetLoginSSH(email, foundSSHKey)
-					if err != nil {
-						return err
-					}
-					fmt.Printf("Logged in as %q using ssh auth\n", email)
-					return nil
-				}
-			}
+		if err := cc.FindSSHCredentials(email); err != nil {
+			return errors.Wrap(err, "failed finding ssh keys")
```
alexcb, line 2389: Why is this returning an error now?

cloud/client.go
```diff
@@ -408,7 +565,17 @@ func (c *client) signChallenge(challenge string, key *agent.Key) (string, error)
 	return s, nil
 }
 
-func (c *client) tryAuth(challenge string, key *agent.Key) (string, string, error) {
+func (c *client) getSSHCredentials(challenge string, key *agent.Key) (credentials string, err error) {
```
alexcb, line 568: Can this be refactored so we avoid duplicating the credential string formatting?

For example, we now have `fmt.Sprintf("ssh-rsa %s %s", blob, sig)` showing up twice in our code-base, where as previously we only had it in a single location.

cloud/client.go
```diff
@@ -861,15 +1027,30 @@ func (c *client) WhoAmI() (string, string, bool, error) {
...
-func (c *client) getAuthTokenPath(create bool) (string, error) {
-	confDirPath := c.authTokenDir
+func (c *client) migrateV1Token() error {
+	confDirPath := c.authDir
+	if confDirPath == "" {
+		confDirPath = cliutil.GetEarthlyDir()
+	}
+	tokenPath := filepath.Join(confDirPath, "auth.token")
+	newPath := filepath.Join(confDirPath, "auth.credentials")
+	if fileutil.FileExists(tokenPath) {
```
alexcb, line 1044: what if `newPath` exists too?

cloud/client.go
```diff
@@ -861,15 +1027,30 @@ func (c *client) WhoAmI() (string, string, bool, error) {
...
+		if err := os.Rename(tokenPath, newPath); err != nil {
+			return errors.Wrapf(err, "failed to move v1 token from '%s' to '%s'", tokenPath, newPath)
```
alexcb, line 1046: `v1` might confuse earthly users. Currently we only have `v0.6.4`.

cloud/client.go
```diff
@@ -900,29 +1095,67 @@ func (c *client) loadAuthToken() error {
...
+	}
+	return nil
+}
+
+func (c *client) loadCredentials() error {
+	credPath, err := c.getCredentialsPath(false)
+	if err != nil {
+		return err
+	}
+	if !fileutil.FileExists(credPath) {
```
alexcb, line 1112: Is it possible to do this after the `ReadFile` via `errors.Is(err, os.ErrNotExist)`?

cmd/earthly/main.go
brandonSc: You're right it didn't return the error here before, and there's a switch case below that's set up to handle this error. Must've been a mistake, I'll fix that

cloud/client.go
```diff
@@ -936,8 +1169,23 @@ func IsValidEmail(email string) bool {
...
+func (c *client) saveToken() error {
+	path, err := c.getTokenPath(true)
+	if err != nil {
+		return err
+	}
+	data := []byte(fmt.Sprintf(
+		"%s %s",
+		c.authToken,
+		c.authTokenExpiry.Format(tokenExpiryLayout)))
+	if err = os.WriteFile(path, data, 0644); err != nil {
```
alexcb, line 1181: keep it secret:
```suggestion
	if err = os.WriteFile(path, data, 0600); err != nil {
```

cloud/client.go
brandonSc: Good idea 👍

cloud/client.go
brandonSc: Yeah I can see that, I was thinking `v1` and `v2` versions of the underlying authentication method. I wonder if even the file name `auth.v2.token` could be confusing as well, we could just call it `auth.jwt`

util/llbutil/secrets.go
```diff
@@ -5,20 +5,19 @@ import (
...
 	"github.com/moby/buildkit/session"
 	"github.com/moby/buildkit/session/secrets"
 	"github.com/pkg/errors"
 	"google.golang.org/grpc"
 )
 
-// ErrNoSecretsClient occurs when the secrets client is referenced but was never provided
-var ErrNoSecretsClient = errors.Errorf("no secrets client provided")
+// ErrNoCloudClient occurs when the secrets client is referenced but was never provided
+var ErrNoCloudClient = errors.Errorf("no secrets client provided")
```
alexcb, line 16:
```suggestion
var ErrNoCloudClient = errors.Errorf("no cloud client provided")
```

cloud/client.go
alexcb: `auth.jwt` sounds like a good idea.

brandonSc: Only thing I don't like is that it's the only place in the code that knows it's jwt. Otherwise, you could swap the jwt form some shiny new auth type one day and everything else here should be totally valid (except for maybe a comment or two)

brandonSc: but maybe that's overly optimistic anyway, using a new auth type might not be so simple

cloud/client.go
brandonSc: under normal circumstances it shouldn't be an issue because `auth.token` is removed. 

A more complicated scenario: you upgrade to jwt, downgrade and changed your password, upgrade again back to jwt. The file renaming still succeeds, and ends up overwriting theold  `auth.credentials` with the new set of credentials containing the new password. Seems like that might actually be a good thing.

Any scenarios you can think of that could be an issue?

alexcb: Possibly down the road (e.g. in v1) we decide we need new data in `auth.credential`, then a user goes from earthly `v1` back to `v0.6.3`, then moves on to `v0.6.5` (the release which will contain this change), it'll then destroy data that was set in `v1`.

cloud/client.go
brandonSc: actually, the other place was unused code so it's been removed.

cloud/client.go
brandonSc: I'm on the fence with this one. Seems like there's pros/cons either way. I'm perhaps more tempted to leave it as-is, since users downgrading back from JWT and changing their credentials or auth type seems more likely immediately than an incompatible change to how we store the `auth.credentials` file.

cloud/client.go
```diff
@@ -295,6 +338,120 @@ func (c *client) RegisterEmail(email string) error {
...
+		err = c.loginWithPassowrd()
+	case c.authCredToken != "":
+		err = c.loginWithToken()
+	default:
+		err = c.loginWithSSH()
+	}
+	if err != nil {
+		return err
+	}
+	if err = c.saveToken(); err != nil {
```
alexcb, line 359: nit: `return c.saveToken()`

alexcb (approved):
nice work.

Issue #1589, opened 2022-01-19
Unexpected lack of parallelization using WITH DOCKER


jdewar:
I am seeing some unexpected serialization in an earthly build, and have made a gist of a simple example: https://gist.github.com/jdewar/991f2c91c6f1550c446d454fed0a57ad

The gist's Earthfile runs both a test with a docker container  (+fast) and a test that takes a while to build an artifact, then uses the artifact to analyze the sourcecode (+slow).  Both of these paths depend on dependencies (+deps) being available, but after that they should be able to run in parallel.

If I remove the 'WITH DOCKER' block, then I see the parallel paths (+lib, +artifact) outputting as expected after +deps.  I found some docs that earthly's docker-in-docker may be needed, but adding `DO github.com/earthly/lib+INSTALL_DIND` did not seem to affect the result.

vladaionescu:
Thanks for the reproduction code @jdewar! Very useful - I was able to run it and verify reproduction.

I was also able to simplify it to the following:

Earthfile

```Earthly
VERSION 0.6

FROM alpine:3.13
WORKDIR /work
DO github.com/earthly/lib+INSTALL_DIND

all:
    BUILD +slow
    BUILD +fast

artifact: 
    FROM +deps
    RUN echo "ARTIFACT"; sleep 10; echo "ARTIFACT done"
    RUN echo "test" >plts
    SAVE ARTIFACT plts

slow:
    FROM +deps
    COPY +artifact/plts plts
    RUN echo "SLOW"; sleep 10; echo "SLOW done"

lib:
    FROM +deps
    RUN echo "LIB"; sleep 10; echo "LIB done"

fast:
    FROM +lib
    COPY docker-compose.yml ./
    WITH DOCKER --compose docker-compose.yml
        RUN echo "DOCKER"; sleep 10; echo "DOCKER done"
    END

deps:
    RUN echo "DEPS"; sleep 10; echo "DEPS done"
```

docker-compose.yml

```YML
version: '3'
services:
  postgres:
    image: postgres
    ports:
      - "5432:5432"
    environment:
      POSTGRES_USER: postgres
      POSTGRES_PASSWORD: postgres
```

It seems that replacing `--compose` with a simple `--pull postgres` makes the problem go away. I suspect this is related to the way we inspect the `docker-compose.yml` file (e.g. to extract information about which images to pull).

PR #1602, opened 2022-01-21
Use force execution after parallel build conversion

vladaionescu:
Fixes https://github.com/earthly/earthly/issues/1589

In the original repro script, it seems that `+slow` and `+fast` do actually convert in parallel, but there is nothing triggering the execution for `+slow` after the conversion, which leads to limited parallelism.

This change forces execution after each parallel conversion, to encourage more parallelism. This seems to address the problem in #1589. The parallel workers are now also started via a global error group and the top-most earthfile2llb will wait for all children to finish before returning control.

### DIFF
--- a/.earthly_version_flag_overrides
+++ b/.earthly_version_flag_overrides
@@ -1,1 +1,1 @@
-referenced-save-only,use-copy-include-patterns,earthly-version-arg,use-cache-command,check-duplicate-images
+referenced-save-only,use-copy-include-patterns,earthly-version-arg,use-cache-command,check-duplicate-images,exec-after-parallel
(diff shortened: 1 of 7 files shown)


vladaionescu:
It seems that there are some execution issues as tests seem to take forever now.

vladaionescu:
This is working well now.... and it seems that there's a significant boost in performance.

* `+test` is 28% faster
* `+examples1` is 23% faster
* `+examples2` is 15% faster (this is a smaller build)

brandonSc (approved):

alexcb (approved):

Issue #1574, opened 2022-01-14
Buildkitd's `GIT_URL_INSTEAD_OF` variable set from an unrelated setting


rrjjvv:
This seemed obvious enough that I didn't think it warranted a reproduction:

https://github.com/earthly/earthly/blob/961d3c55e08e57591ee1e3761bd51534a4a857d7/buildkitd/buildkitd.go#L427-L428

I assume the value of `settings.GitURLInsteadOf` is what is *supposed* to be applied.

I didn't actually experience an issue because of this.  I was going through the source trying to answer a question of mine (totally unrelated to this setting) and spotted this.  Viewing the diff that introduced it (#1224) suggests a simple copy/paste accident.

vladaionescu:
Good catch - `GIT_URL_INSTEAD_OF` is a deprecated feature (it's done a bit differently now). This is done differently now.

I guess now we know nobody is using this anymore as we haven't heard of anyone complaining.

vladaionescu:
Thanks for reporting it!

alexcb:
@vladaionescu do you think this warrants fixing, or should we just remove it entirely?

@rrjjvv does https://docs.earthly.dev/docs/guides/auth#git-authentication achieve what you are after?

vladaionescu:
I think we can remove it altogether. It's been like that for quite a while.

rrjjvv:
@alexcb I should have been clearer in my report.  I was not actually impacted by this; I was just reading the source and it was "obviously" wrong, so I wanted to report it before somebody actually ran into it.  (I actually do have a question around git auth, though it shouldn't be at all related to this report; I'll likely bring up in Slack at some point.  And my reason for looking at that code had nothing to do with git auth either.)

alexcb:
perfect, we'll remove it then: https://github.com/earthly/earthly/pull/1609

PR #1609, opened 2022-01-24
remove GIT_URL_INSTEAD_OF (which was never set correctly)

alexcb:
as discussed in https://github.com/earthly/earthly/issues/1574 we will remove this rather than fix it.


buildkitd/buildkitd.go
```diff
@@ -424,10 +424,6 @@ func Start(ctx context.Context, console conslogging.ConsoleLogger, image, contai
 		envOpts["CACHE_SIZE_MB"] = strconv.FormatInt(int64(settings.CacheSizeMb), 10)
 	}
 
-	if settings.GitURLInsteadOf != "" {
```
vladaionescu, line 427: We should probably also remove `settings.GitURLInsteadOf` and also whatever flags or env vars are setting it on the other side.

alexcb: :facepalm: good call.

scripts/tests/auth/test-with-git-url-instead-of.sh
```diff
@@ -1,24 +0,0 @@
...
-
-GITHUB_SERVER_URL="${VARIABLE:-https://github.com}"
-GITHUB_REPOSITORY="${GITHUB_REPOSITORY:-earthly/earthly}"
-github_server_without_protocol=$(echo "$GITHUB_SERVER_URL" | sed -e 's#^https\?://##; s#//$##')
-earthly_ref="$github_server_without_protocol/$GITHUB_REPOSITORY/examples/cpp:$gitsha+docker"
-
-docker image rm -f earthly/examples:cpp
-echo "running $earthly -VD $earthly_ref"
-GIT_URL_INSTEAD_OF="https://github.com/=git@github.com:" "$earthly" -VD "$earthly_ref"
-docker run --rm earthly/examples:cpp | grep fib
```
vladaionescu, line 24: Wonder how this test was passing given that the feature is broken?

vladaionescu (approved):

PR #1611, opened 2022-01-25
remove SUDO_USER from circleci test

alexcb:
circle CI is being run as user `circleci` with `SUDO_USER="root"`; however the circleci user does not have access to `/root/.earthly/config.yml` therefore, we must unset the `SUDO_USER`; otherwise we will get this error: `open /root/.earthly/config.yml: permission denied`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,7 +13,14 @@ jobs:
           command: "sudo /bin/sh -c 'wget https://github.com/earthly/earthly/releases/latest/download/earthly-linux-amd64 -O /usr/local/bin/earthly && chmod +x /usr/local/bin/earthly'"
       - run:
           name: Build latest earthly
-          command: earthly --use-inline-cache +for-linux
+          # earthly v0.6.5 has a bug that causes a panic when SUDO_USER is set to a user that the current user does not have read permission to. To Work-around this,
+          # we need to set SUDO_USER to "".
+          #
+          # This is due to circle CI running under the user "circleci" with SUDO_USER="root"; however the circleci user does not have access to /root/.earthly/config.yml
+          # therefore, we must unset the SUDO_USER; otherwise we will get this error: "open /root/.earthly/config.yml: permission denied"
+          #
+          # Once this bugfix is released, this can be undone.
+          command: SUDO_USER="" earthly --use-inline-cache +for-linux
       - run:
           name: Bootstrap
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -38,3 +40,4 @@ workflows:
             branches:
               only:
                 - main
+                - acb/fix-circleci-test
```
alexcb, line 43: I will remove this once PR is approved

.circleci/config.yml
```diff
@@ -13,14 +13,16 @@ jobs:
...
+          # therefore, we must unset the SUDO_USER; otherwise we will get this error: "open /root/.earthly/config.yml: permission denied"
+          command: SUDO_USER="" earthly --use-inline-cache +for-linux
       - run:
           name: Bootstrap
           command: sudo ./build/linux/amd64/earthly bootstrap
       - run:
           name: Run tests
           command: |-
-              ./build/linux/amd64/earthly --ci -P \
+              SUDO_USER="" ./build/linux/amd64/earthly --ci -P \
```
vladaionescu, line 25: Do we need to update our docs to encourage our users to do this too?

vladaionescu (commented):
BTW should something like this `/root/.earthly/config.yml: permission denied` not fail and rather print a warning? (Just wondering if our users are relying on this not crashing the build)

alexcb:
> BTW should something like this `/root/.earthly/config.yml: permission denied` not fail and rather print a warning? (Just wondering if our users are relying on this not crashing the build)

What's odd is https://github.com/earthly/earthly/blob/cb2e4f7f93e569d08de2b7e3fcef15c829157afd/util/fileutil/homedir.go#L20 has a check in place to skip `SUDO_USER` if the home dir isn't writable.

I can spend some more time debugging it to see why that's not being hit if you think that's useful.

.circleci/config.yml
alexcb: If that's the case, then maybe we should just remove the logic that makes use of `SUDO_USER`, or we can fine-tune it to only work when the current user is root and the `SUDO_USER` is not root, then the need for circle-ci-specific entry is removed.

vladaionescu:
> I can spend some more time debugging it to see why that's not being hit if you think that's useful.

Yeah - I think it is. We don't want users to have a bad experience when trying Earthly in Circle CI

vladaionescu (approved):

PR #1612, opened 2022-01-26
Add ability to write logs to disk.

dchw:
Initial code to generate static log bundles. These pass server validation. Still needs some cleanup, but initial feedback and thoughts are welcome.

I hooked the console directly to ensure I got the output as seen by the user; progress bars, colors, and all. This probably isn't extendable without a structured format the console and disk versions could render.


conslogging/conslogging.go
```diff
@@ -380,9 +391,71 @@ func (cl ConsoleLogger) prettyPrefix() string {
...
+
+			truncated += letter + "/"
+		}
+
+		prettyPrefix = truncated + target
+	}
+
+	formatString := fmt.Sprintf("%%%vv", prefixPadding)
+	return fmt.Sprintf(formatString, fmt.Sprintf("%s%s", prettyPrefix, brackets))
+}
```
dchw, line 424: This implementation needs to be merged with the other one still

conslogging/bundle.go
```diff
@@ -0,0 +1,246 @@
...
+	Size     int    `json:"size"`
+	Command  string `json:"command,omitempty"`
+	Summary  string `json:"summary,omitempty"`
+}
+
+type Permissions struct {
+	Version int      `json:"version"`
+	Users   []uint64 `json:"users"`
+	Orgs    []uint64 `json:"orgs"`
+}
```
dchw, line 246: How should we share these structs? Perhaps a target to copy them from our server code?

conslogging/conslogging.go
```diff
@@ -74,12 +75,14 @@ func (cl ConsoleLogger) clone() ConsoleLogger {
...
 		prefixPadding:  cl.prefixPadding,
 		mu:             cl.mu,
+		bb:             cl.bb,
 	}
 }
 
 // WithPrefix returns a ConsoleLogger with a prefix added.
 func (cl ConsoleLogger) WithPrefix(prefix string) ConsoleLogger {
 	ret := cl.clone()
+	ret.errW = io.MultiWriter(cl.errW, cl.bb.PrefixWriter(prefix))
```
dchw, line 85: Ran some experiments. We don't appear to ever call this a second time, but past performance does not predict future results (or at least thats what Vanguard tells me). To protect against this, we would need to make a new multiwriter using the original intended `os.Stderr` and the new prefix.

conslogging/bundle.go
```diff
@@ -0,0 +1,246 @@
...
+	if err != nil {
+		return errors.Wrap(err, "failed to serialize bundle permissions placeholder")
+	}
+
+	return nil
+}
+
+// Nobody expects ANSI in the command/summary.
+// So, even if we don't inject color we should strip it since a tool inside could have done an ANSI too. *SIGH*
+const ansi = "[\u001B\u009B][[\\]()#;?]*(?:(?:(?:[a-zA-Z\\d]*(?:;[a-zA-Z\\d]*)*)?\u0007)|(?:(?:\\d{1,4}(?:;\\d{0,4})*)?[\\dA-PRZcf-ntqry=><~]))"
```
dchw, line 153: I am sorry for this and all the other regex. Willing to sign documents saying they are my problem now.

conslogging/conslogging.go
vladaionescu: We should probably allow this to be called multiple times cleanly, for that "future" possibility :-)

cmd/earthly/main.go
```diff
@@ -2787,6 +2786,19 @@ func (app *earthlyApp) actionBuildImp(c *cli.Context, flagArgs, nonFlagArgs []st
...
+
+	logTemp, err := os.MkdirTemp("", "earthly-log")
+	if err != nil {
+		return errors.Wrap(err, "make temp log dir")
+	}
+	app.console = app.console.WithLogBundleWriter(logTemp, target.String())
+	defer func(console conslogging.ConsoleLogger) {
+		err := console.FlushBundleBuilder()
+		if err != nil {
+			fmt.Println(err.Error())
```
vladaionescu, line 2798: You were probably meaning to fix this part?

conslogging/bundle.go
dchw: Was toying with the idea of a shared proto here.

dchw: This may still need to be solved before we merge.

dchw:
There are for sure some things we may want to fix; and also some tests to write (mostly a test for uploading a log). But the code proper is ready at this point.

cmd/earthly/main.go
```diff
@@ -2801,6 +2802,45 @@ func (app *earthlyApp) actionBuildImp(c *cli.Context, flagArgs, nonFlagArgs []st
...
+
+	cc, err := cloud.NewClient(app.apiServer, app.sshAuthSock, app.authToken, app.console.Warnf)
+	if err != nil {
+		return errors.Wrap(err, "failed to create cloud client")
+	}
+
+	// Default upload logs, unless explicitly configured
+	if !app.cfg.Global.DisableLogSharing {
+		_, _, _, err := cc.WhoAmI()
+		isLoggedIn := err == nil
```
vladaionescu, line 2817: We should perhaps detect that this error is actually about not being logged in, and not 500 / 404 or other such cases.

config/config.go
```diff
@@ -77,6 +77,7 @@ type GlobalConfig struct {
 	TLSEnabled               bool     `yaml:"tls_enabled"                help:"If TLS should be used to communicate with Buildkit. Only honored when BuildkitScheme is 'tcp'."`
 	ContainerFrontend        string   `yaml:"container_frontend"         help:"What program should be used to start and stop buildkitd, save images. Default is 'docker'. Valid options are 'docker' and 'podman' (experimental)."`
 	IPTables                 string   `yaml:"ip_tables"                  help:"Which iptables binary to use. Valid values are iptables-legacy or iptables-nft. Bypasses any autodetection."`
+	DisableLogSharing        bool     `yaml:"disable_log_sharing"        help:"Disable cloud log sharing whe logged in with an Earthly account, see https://ci.earthly.dev for details."`
```
vladaionescu, line 80: when *

cmd/earthly/main.go
```diff
@@ -2801,6 +2802,45 @@ func (app *earthlyApp) actionBuildImp(c *cli.Context, flagArgs, nonFlagArgs []st
...
+
+		if isLoggedIn {
+			// If you are logged in, then add the bundle builder code, and configure cleanup and post-build messages.
+			app.console = app.console.WithLogBundleWriter(target.String(), cleanCollection)
+
+			defer func() { // Defer this to keep log upload code together
+				logPath, err := app.console.WriteBundleToDisk()
+				if err != nil {
+					fmt.Println(err.Error())
```
vladaionescu, line 2826: I believe we need to use console.Warn

... or at least use stderr if not

vladaionescu, line 2826: Should there be a return here?

cmd/earthly/main.go
dchw: I suppose we could append a warning if its anything other than a 401 next to the advertising message?

dchw: I tried my hand at this. If its unauthorized, then we advertise. Otherwise we say something else.

vladaionescu: Did you mean to make a code change for this as well?

BTW, a `// TODO` is also fine if we want to just get this out

conslogging/bundle.go
dchw: I broke these out into their own file, with a comment. Once its merged, I'll import this from our server-side code.

vladaionescu (approved):
Looks good - but we should at the very least add a TODO for the `isLoggedIn` comment.

cmd/earthly/main.go
dchw: The code change is lower in the file. It seemed like we would want to treat any error here as if we weren't logged in, and avoid attaching a logger because later upload would likely fail as well.

So, I added a conditional further down that branches on what kind of error we got. If it is `ErrUnauthorized`, we print the invitation message, otherwise, we print an error saying we could not upload the log.

cmd/earthly/main.go
```diff
@@ -2801,6 +2802,51 @@ func (app *earthlyApp) actionBuildImp(c *cli.Context, flagArgs, nonFlagArgs []st
...
+			}()
+		} else {
+			// If you are not logged in, then advertise the service, since they probably turned it on to try it.
+			defer func() { // Defer this to keep log upload code together
+				switch err {
+				case cloud.ErrUnauthorized:
+					app.console.Printf("Share your logs with an Earthly account! Register for one at https://ci.earthly.dev.")
+				default:
+					app.console.Warnf("Logs for this run were not shared, since earthly couldn't login. Error: %s", err.Error())
+				}
```
dchw, line 2845: @vladaionescu Here is where we print the different messages, and differentiate between real errors and unauthorized users.

vladaionescu: Ah yes - I see it now. We should probably rename that `err` so that it doesn't accidentally get overridden by subsequent commands (and by the time we get to execute the defer it's no longer set).

cmd/earthly/main.go
```diff
@@ -2801,6 +2802,51 @@ func (app *earthlyApp) actionBuildImp(c *cli.Context, flagArgs, nonFlagArgs []st
...
+				if err != nil {
+					app.console.Warnf(err.Error())
+					return
+				}
+				app.console.Printf("Share your build log with this link: %s\n", id)
+			}()
+		} else {
+			// If you are not logged in, then advertise the service, since they probably turned it on to try it.
+			defer func() { // Defer this to keep log upload code together
+				switch err {
```
alexcb, line 2840: This seems fragile, as I could see `err` potentially being changed in a future PR. I think an explcit error variable such as:
```suggestion
				switch whoamiErr {
```
would make this more robust.

cmd/earthly/main.go
```diff
@@ -2801,6 +2802,51 @@ func (app *earthlyApp) actionBuildImp(c *cli.Context, flagArgs, nonFlagArgs []st
...
+				case cloud.ErrUnauthorized:
+					app.console.Printf("Share your logs with an Earthly account! Register for one at https://ci.earthly.dev.")
+				default:
+					app.console.Warnf("Logs for this run were not shared, since earthly couldn't login. Error: %s", err.Error())
```
alexcb, line 2844: Should we make this more concise? 
```suggestion
					app.console.Warnf("Logs were not shared, due to earthly login error: %s", err.Error())
```

dchw: I'll leave wording up to those with stronger opinions. If nobody else has suggestions, I will use this.

conslogging/bundle.go
```diff
@@ -0,0 +1,266 @@
...
+
+		if tm.Status != StatusComplete {
+			manifest.Status = tm.Status
+		}
+	}
+
+	return manifest
+}
+
+func (bb *BundleBuilder) buildPermissions() *Permissions {
```
alexcb, line 175: something feels odd with this function since it returns an near-empty struct. Is it possible to omit this entirely?

dchw: Yeah, I agree, somewhat. Its there for completeness, since we do have to build a permissions struct. This makes it fit the pattern of the other parts; waiting until we are ready to fully implement permissions.

dchw: Inlining it, in context, would look like this:

cmd/earthly/main.go
dchw: Renamed, Alex caught it too.

vladaionescu (approved):

conslogging/bundle.go
alexcb: How does the server interpret a `permissions` file containing `{"Version": 1, "Users": [], "Orgs": []}`?

It seems like a potential security issue down the road if we default to an empty list meaning "publicly-for-anyone-with-log-link". Maybe we should have an explicit field stating this is the intended behaviour.

Either that or just omit including a permission file all together if we're not able to handle permissions yet.

dchw: Right now, the server only requires that the file exist, and it doesn't attempt to interpret the file too much. Logs age out, and we are able to migrate them behind the scenes if needed... so I am not too worried.

cmd/earthly/main.go
alexcb: @vladaionescu thoughts on the reword?

alexcb (approved):

conslogging/bundle.go
alexcb: Our user IDs have been switched to UUIDs this morning (still waiting on Orgs), adding a `*` seems like a good thing to do now, especially if it currently ignores it.

dchw: Yeah, I was thinking the same thing as I typed that.

cmd/earthly/main.go
vladaionescu: Looks good to me

dchw: Wording changed.

Issue #1621, opened 2022-01-31
Critical vulnerabilities in earthly/earthly docker image


heavelock:
A critical vulnerability came up in a scan of a docker image that is based on earthly/earthly:

```shell
$ trivy image --exit-code 1 --severity "$TRIVY_FAIL_SEVERITY" --ignore-unfixed "$TRIVY_TARGET"
2022-01-30T19:49:46.455Z	INFO	Using your github token
2022-01-30T19:49:47.325Z	INFO	Detected OS: alpine
2022-01-30T19:49:47.325Z	INFO	Detecting Alpine vulnerabilities...
2022-01-30T19:49:47.331Z	INFO	Number of language-specific files: 23
2022-01-30T19:49:47.331Z	INFO	Detecting gobinary vulnerabilities...
earthly_base:latest (alpine 3.14.0)
=====================================================================================
Total: 3 (CRITICAL: 3)
+--------------+------------------+----------+-------------------+---------------+---------------------------------------+
|   LIBRARY    | VULNERABILITY ID | SEVERITY | INSTALLED VERSION | FIXED VERSION |                 TITLE                 |
+--------------+------------------+----------+-------------------+---------------+---------------------------------------+
| apk-tools    | CVE-2021-36159   | CRITICAL | 2.12.5-r1         | 2.12.6-r0     | libfetch before 2021-07-26, as        |
|              |                  |          |                   |               | used in apk-tools, xbps, and          |
|              |                  |          |                   |               | other products, mishandles...         |
|              |                  |          |                   |               | -->avd.aquasec.com/nvd/cve-2021-36159 |
+--------------+------------------+          +-------------------+---------------+---------------------------------------+
| libcrypto1.1 | CVE-2021-3711    |          | 1.1.1k-r0         | 1.1.1l-r0     | openssl: SM2 Decryption               |
|              |                  |          |                   |               | Buffer Overflow                       |
|              |                  |          |                   |               | -->avd.aquasec.com/nvd/cve-2021-3711  |
+--------------+                  +          +                   +               +                                       +
| libssl1.1    |                  |          |                   |               |                                       |
|              |                  |          |                   |               |                                       |
|              |                  |          |                   |               |                                       |
+--------------+------------------+----------+-------------------+---------------+---------------------------------------+
```

vladaionescu:
Hi @heavelock - thanks for reporting!

Not clear how easy it is to exploit these vulns - I don't think these are really critical in our context. In any case, we should definitely patch them.

[...]

PR #1623, opened 2022-01-31
update alpine to 3.15

alexcb:


alexcb:
This is to address #1621

vladaionescu (approved):

PR #1624, opened 2022-01-31
Use ARG --required

alexcb:
Updates Earthfiles to make use of the ARG --required flag.

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -398,6 +398,7 @@ dind-ubuntu:
     FROM ubuntu:20.04
     COPY ./buildkitd/docker-auto-install.sh /usr/local/bin/docker-auto-install.sh
     RUN docker-auto-install.sh
+    ARG EARTHLY_TARGET_TAG_DOCKER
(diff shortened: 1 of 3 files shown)


Earthfile
```diff
@@ -389,7 +389,7 @@ dind:
 dind-alpine:
     FROM docker:dind
     RUN apk add --update --no-cache docker-compose
-    ARG EARTHLY_TARGET_TAG_DOCKER
+    ARG --required EARTHLY_TARGET_TAG_DOCKER
```
vladaionescu, line 392: Isn't this a builtin arg?

alexcb: ah right! I wonder if this should cause an error in this case?

vladaionescu:
BTW I think the release Earthfiles could benefit from some `--required`

Earthfile
vladaionescu: Yeah, it might be a good idea

alexcb: I created https://github.com/earthly/earthly/issues/1625 to track this.

vladaionescu (approved):

PR #1629, opened 2022-02-02
Display error when logout is called with EARTHLY_TOKEN

alexcb:

### DIFF
--- a/cmd/earthly/main.go
+++ b/cmd/earthly/main.go
@@ -178,6 +178,13 @@ var (
 	GitSha string
 )
 
+var (
+	errLoginFlagsHaveNoEffect            = errors.New("account login flags have no effect when --auth-token (or the EARTHLY_TOKEN environment variable) is set")
+	errLogoutHasNoEffectWhenAuthTokenSet = errors.New("account logout has no effect when --auth-token (or the EARTHLY_TOKEN environment variable) is set")
+	errURLParseFailure                   = errors.New("Invalid URL")
+	errURLValidationFailure              = errors.New("URL did not pass validation")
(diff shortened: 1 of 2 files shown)


cmd/earthly/main.go
```diff
@@ -178,6 +178,13 @@ var (
 	GitSha string
 )
 
+var (
+	errLoginFLagsHaveNoEffect            = errors.New("account login flags have no effect when --auth-token (or the EARTHLY_TOKEN environment variable) is set")
```
vladaionescu, line 182:
```suggestion
	errLoginFlagsHaveNoEffect            = errors.New("account login flags have no effect when --auth-token (or the EARTHLY_TOKEN environment variable) is set")
```

cmd/earthly/main.go
```diff
@@ -2405,7 +2409,7 @@ func (app *earthlyApp) actionAccountLogin(c *cli.Context) error {
 	// special case where global auth token overrides login logic
 	if app.authToken != "" {
 		if email != "" || token != "" || pass != "" {
-			return errors.New("account login flags have no effect when --auth-token (or the EARTHLY_TOKEN environment variable) is set")
+			return errLoginFLagsHaveNoEffect
```
vladaionescu, line 2412:
```suggestion
			return errLoginFlagsHaveNoEffect
```

cmd/earthly/main.go
alexcb: :facepalm: good catch. I accidentally propagated this error with vim's auto completion.

vladaionescu (approved):

PR #1630, opened 2022-02-03
Don't re-authenticate requests which don't have a JWT

brandonSc:
Edge case where this can cause an infinite recursive loop. 

Basically, an authenticated request would 401 (e.g. list secrets but not logged in) which would trigger a login request  (re-auth) -> which would 401 and trigger SSH discovery -> which would request an "auth challenge" -> which would 401, and trigger another login request (re-auth) -> which would trigger SSH discovery again -> ... etc

The solution here prevents the 401 on a login request from retrying.

### DIFF
--- a/cloud/client.go
+++ b/cloud/client.go
@@ -149,18 +149,21 @@ func (c *client) doCall(method, url string, opts ...requestOpt) (int, string, er
 		}
 	}
 
+	alreadyReAuthed := false
 	if r.hasAuth && time.Now().UTC().After(c.authTokenExpiry) {
 		if err := c.Authenticate(); err != nil {
+			if errors.Is(err, ErrUnauthorized) {
+				return 0, "", ErrUnauthorized
+			}
 			return 0, "", errors.Wrap(err, "failed refreshing expired auth token")
 		}
+		alreadyReAuthed = true
 	}
 
 	var status int
 	var body string
 	var err error
 	duration := time.Millisecond * 100
-	alreadyReAuthed := false
-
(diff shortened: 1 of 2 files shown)


cloud/client.go
```diff
@@ -169,6 +169,9 @@ func (c *client) doCall(method, url string, opts ...requestOpt) (int, string, er
 		}
 
 		if status == http.StatusUnauthorized {
+			if !r.hasAuth {
```
alexcb, line 172: Should we combine this with the below if? e.g.
```suggestion
			if !r.hasAuth || alreadyReAuthed {
```

it's not clear why `alreadyReAuthed` returns `err` and not `errUnauthorized`; I would have thought they would both return the same error.

alexcb:
It's not completely related to this PR, but I noticed `alreadyReAuthed := false` is set after the initial

```
if r.hasAuth && time.Now().UTC().After(c.authTokenExpiry) {
		if err := c.Authenticate(); err != nil {
			return 0, "", errors.Wrap(err, "failed refreshing expired auth token")
		}
	}
```

we should change this to

```
alreadyReAuthed := false
if r.hasAuth && time.Now().UTC().After(c.authTokenExpiry) {
		if err := c.Authenticate(); err != nil {
			return 0, "", errors.Wrap(err, "failed refreshing expired auth token")
		}
alreadyReAuthed = true
	}
```

cloud/client.go
brandonSc: Will give it a try

brandonSc: Added this

alexcb (approved):

PR #1633, opened 2022-02-05
explicitly allow sshd to accept ssh-rsa

alexcb:
Due to https://github.com/golang/go/issues/39885
if we only allow rsa-sha2-256,rsa-sha2-512 golang/x/crypto
doesn't work as it presents the key as a regular ssh-rsa (which
fallsback to using sha1 hashes which are insecure)


alexcb:
if I change it to
```
pubkeyacceptedkeytypes ssh-ed25519,rsa-sha2-512,rsa-sha2-256
```

it fails with:

```
failed to connect to git.example.com over ssh due to ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey], no supported methods remain; falling back to https
```

vladaionescu (approved):

PR #1641, opened 2022-02-07
Add better logs error messages

dchw:


cmd/earthly/main.go
```diff
@@ -2830,12 +2830,14 @@ func (app *earthlyApp) actionBuildImp(c *cli.Context, flagArgs, nonFlagArgs []st
...
 				if err != nil {
+					err := errors.Wrapf(err, "failed to write log to disk")
 					app.console.Warnf(err.Error())
 					return
 				}
 
 				id, err := cc.UploadLog(logPath)
 				if err != nil {
+					err := errors.Wrapf(err, "failed to upload log")
 					app.console.Warnf(err.Error())
```
brandonSc, line 2841: Wonder if this should check for unauthorized explicitly and log an info level instead. (i.e. getting a red error message even though the build succeeded looks misleading)

brandonSc:
ET-74

^ will this trigger Linear.app integration to clue in?

dchw:
No. Linear is not hooked up to this repo currently.

cmd/earthly/main.go
```diff
@@ -2821,7 +2821,7 @@ func (app *earthlyApp) actionBuildImp(c *cli.Context, flagArgs, nonFlagArgs []st
 	// Default upload logs, unless explicitly configured
 	if !app.cfg.Global.DisableLogSharing {
 		_, _, _, whoAmIErr := cc.WhoAmI()
-		isLoggedIn := err == nil
+		isLoggedIn := whoAmIErr == nil
```
dchw, line 2824: @brandonSc This is why you managed to get an unauthorized. Fixed now.

brandonSc: So we shouldn't even get the error when not logged in. I think that invalidates my other comment below.

dchw: Yeah. I was super confused and then realized my mistake :facepalm:

PR #1649, opened 2022-02-09
Fix some remote log manifest data

dchw:
Report timing in milliseconds, and exempt _full from affecting whole-build status.

cc @mirho - this should fix some builds showing as cancelled when they were really completed successfully.


vladaionescu (approved):

alexcb (approved):
my gut says seconds as a float64 and let JS convert it, as it would allow for sub-millisecond accuracy, but sub-millisecond would be really really good for a CI.

dchw:
> my gut says seconds as a float64 and let JS convert it, as it would allow for sub-millisecond accuracy, but sub-millisecond would be really really good for a CI.

You've gotten me curious. When would we care about precision that much in a CI :thinking:

PR #1650, opened 2022-02-09
Use JWT in Analytics Request

brandonSc:
When the JWT is set, send it in the analytics request, so the server can record the user's ID with the event.

### DIFF
--- a/cmd/earthly/main.go
+++ b/cmd/earthly/main.go
@@ -298,7 +298,7 @@ func main() {
 		cc, err := cloud.NewClient(app.apiServer, app.sshAuthSock, app.authToken, app.console.Warnf)
 		if err != nil && displayErrors {
 			app.console.Warnf("unable to start cloud client: %s", err)
-		} else {
+		} else if err == nil {
 			analytics.CollectAnalytics(
 				ctxTimeout, cc, displayErrors, Version, getPlatform(),
 				GitSha, app.commandName, exitCode, time.Since(startTime),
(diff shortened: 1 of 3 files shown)


cmd/earthly/main.go
```diff
@@ -3127,6 +3135,8 @@ func ifNilBoolDefault(ptr *bool, defaultValue bool) bool {
 }
 
 func (app *earthlyApp) actionListTargets(c *cli.Context) error {
+	app.commandName = "listTargets"
```
brandonSc, line 3138: noticed this was producing analytics event with key name `cli-` -> now it's `cli-listTargets`.

cmd/earthly/main.go
```diff
@@ -295,7 +295,15 @@ func main() {
 		ctxTimeout, cancel := context.WithTimeout(ctx, time.Millisecond*500)
 		defer cancel()
 		displayErrors := app.verbose
-		analytics.CollectAnalytics(ctxTimeout, app.apiServer, displayErrors, Version, getPlatform(), GitSha, app.commandName, exitCode, time.Since(startTime))
+		cc, err := cloud.NewClient(app.apiServer, app.sshAuthSock, app.authToken, app.console.Warnf)
+		if err != nil && displayErrors {
```
alexcb, line 299: If there's an error but the user is NOT in verbose mode, this will fall through to the `else` below and will use `cc` which will be nil.

You'll need to break them up like:
```
if err != nil {
    if displayErrors {
        app.console.Warnf("unable to start cloud client: %s", err)
    }
} else { ...
```

alexcb (approved):
looks good once the `if err != nil && displayErrors {` bug gets fixed.

cmd/earthly/main.go
brandonSc: oh yeah definitely, thanks for catching my mistake :)

PR #1656, opened 2022-02-10
include git sha in semver

alexcb:
This changes staging releases to be of the form

    0.<unix-timestamp>.<git-sha-in-decimal>

### DIFF
--- a/.github/workflows/staging-deploy.yml
+++ b/.github/workflows/staging-deploy.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - 'main'
+      - '**-staging-test'
     paths-ignore:
       - 'docs/**'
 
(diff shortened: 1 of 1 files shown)


alexcb:
This also changes the release text:

image

.github/workflows/staging-deploy.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - 'main'
+      - '**-staging-test'
```
alexcb, line 7: intentially leaving this here to allow us to trigger a stagging release for non-merged branches as long as we name with the `-staging-test` suffix, e.g. `acb/my-test-branch-staging-test`

vladaionescu (approved):

PR #1657, opened 2022-02-11
Small log bundle fixes

dchw:
Encode bundle target names with percent encoding. Also blacklist the `buildkitd` init prefix.

### DIFF
--- a/conslogging/bundle.go
+++ b/conslogging/bundle.go
@@ -118,7 +118,7 @@ func (bb *BundleBuilder) WriteToDisk() (string, error) {
 		targetData = append(targetData, mt)
 
 		trimmed := strings.TrimSpace(lines.prefix)
-		escaped := url.PathEscape(trimmed)
+		escaped := url.QueryEscape(trimmed)
 
 		tarWriter.WriteHeader(&tar.Header{
 			Name:       fmt.Sprintf("target/%s", escaped),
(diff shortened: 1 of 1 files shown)


conslogging/bundle.go
```diff
@@ -198,9 +198,9 @@ func (tl *targetLogger) toManifestTarget() (TargetManifest, error) {
 		return TargetManifest{}, errors.New("0 length target")
 	}
 
-	if tl.prefix == "ongoing" {
-		// The ongoing messages end up in here too. Since they are not updates from a vertex, we will never mark them as complete.
-		// Additionally, its not useful to have in the output. Ignore it here.
+	if tl.prefix == "ongoing" || tl.prefix == "buildkitd" {
+		// The ongoing & buildkitd init messages end up in here too. Since they are not updates from a vertex, we will
+		// never mark them as complete. Additionally, its not useful to have in the output. Ignore it here.
```
vladaionescu, line 203: I believe this stuff will get better when we switch to using the sts ID's as target IDs. Anything that doesn't have an sts ID isn't a real target, so we can ignore it automatically.

vladaionescu (approved):

brandonSc (approved):

conslogging/bundle.go
dchw: I would love if we had a way to include out of band things as "targets", though

PR #1673, opened 2022-02-15
fix RUN --ssh WITH DOCKER

alexcb:
This fixes an issue where `--ssh` was not correctly passed to the `WITH
DOCKER` handler.


vladaionescu (approved):
Does the `LOCALLY` version of `WITH DOCKER` need something similar?

alexcb:
> Does the `LOCALLY` version of `WITH DOCKER` need something similar?

the LOCALLY RUN doesn't support `--ssh`, and results in: `RUN: --ssh not supported with LOCALLY`

However since `LOCALLY` `RUN`s have access to the user's env variables, if they have ssh-agent running, then they can already access it.

vladaionescu (approved):

PR #1677, opened 2022-02-16
Add language sections to tutorial

jalletto:
[Gitbook link](https://docs.earthly.dev/~/revisions/rrmLhAaanneYk5ykOWnh/)

Biggest change was to make it easier to differentiate between examples based on programming language. It was particularly difficult to pick out the beginning and end of say, the Javascript section, in Part1b: Detailed Explanation, but thought this would be beneficial on all pages. 

It would have been cool to do this with [tabs](https://docs.gitbook.com/editing-content/rich-content/with-command-palette#tabs) but I could not figure out how to do that outside of the Gitbook UI. Do changes made in the UI re sync with this repo? Wasn't sure how much we wanted to stick to standard markdown for now.

Other than that I made a couple small changes to wording to add a little context.

### DIFF
--- a/docs/basics/part-1-a-simple-earthfile.md
+++ b/docs/basics/part-1-a-simple-earthfile.md
@@ -1,10 +1,17 @@
 # A simple Earthfile
 
-Earthfiles are always named `Earthfile`, regardless of their location in the codebase.
+All the magic of Earthly happens in the Earthfile. Earthfiles are always named `Earthfile`, regardless of their location in the codebase. Below you'll find several example Earthfiles. 
+
+- [Go](#go) 
+- [JavaScript](#javascript) 
+- [Java](#java)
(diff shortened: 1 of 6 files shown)


docs/basics/part-1-a-simple-earthfile.md
```diff
@@ -1,10 +1,17 @@
...
+- [Go](#go) 
+- [JavaScript](#javascript) 
+- [Java](#java)
+- [Python](#python)
+
+
+### Go
 
 {% method %}
 {% sample lang="Go" %}
```
vladaionescu, line 14: Would you mind removing these since they no longer work?

jalletto: Removed 👍

vladaionescu (approved):

PR #1681, opened 2022-02-17
Improve Windows compatibility

GMaxera:
I've ran into some issues when trying out Earthly native Windows binary.

This PR fixes some of them.

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -238,7 +238,7 @@ earthly:
     ARG VERSION="dev-$EARTHLY_TARGET_TAG_DOCKER"
     ARG EARTHLY_GIT_HASH
     ARG DEFAULT_BUILDKITD_IMAGE=docker.io/earthly/buildkitd:$VERSION # The image needs to be fully qualified for alternative frontend support.
-    ARG BUILD_TAGS=dfrunmount dfrunsecurity dfsecrets dfssh dfrunnetwork dfheredoc
+    ARG BUILD_TAGS=dfrunmount dfrunsecurity dfsecrets dfssh dfrunnetwork dfheredoc forceposix
     ARG GOCACHE=/go-cache
     RUN mkdir -p build
     RUN printf "$BUILD_TAGS" > ./build/tags && echo "$(cat ./build/tags)"
--- a/util/gitutil/detectgit.go
+++ b/util/gitutil/detectgit.go
@@ -284,9 +284,13 @@ func gitRelDir(basePath string, path string) (string, bool, error) {
 	if err != nil {
 		return "", false, errors.Wrapf(err, "stat for %s", basePath)
 	}
-	// `pathParts` here has lose the root filepath.Separator since built by strings.Split beforehand.
-	// so putting heading separator is required to make it absolute again.
-	b, err := os.Stat(string(filepath.Separator) + filepath.Join(pathParts[:len(basePathParts)]...))
+	rootPath := ""
+	if filepath.VolumeName(basePath) == "" {
+		// `pathParts` here has lose the root filepath.Separator since built by strings.Split beforehand.
+		// so putting heading separator is required to make it absolute again.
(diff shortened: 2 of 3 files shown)


Earthfile
```diff
@@ -238,7 +238,7 @@ earthly:
     ARG VERSION="dev-$EARTHLY_TARGET_TAG_DOCKER"
     ARG EARTHLY_GIT_HASH
     ARG DEFAULT_BUILDKITD_IMAGE=docker.io/earthly/buildkitd:$VERSION # The image needs to be fully qualified for alternative frontend support.
-    ARG BUILD_TAGS=dfrunmount dfrunsecurity dfsecrets dfssh dfrunnetwork dfheredoc
+    ARG BUILD_TAGS=dfrunmount dfrunsecurity dfsecrets dfssh dfrunnetwork dfheredoc forceposix
```
GMaxera, line 241: The https://github.com/jessevdk/go-flags package automatically uses the `/` for options when detect is building on Windows. This is, of course, undesired in the case of Earthfile. The `forceposix` makes go-flags always use the dashes for the options.

util/gitutil/detectgit.go
```diff
@@ -284,9 +284,13 @@ func gitRelDir(basePath string, path string) (string, bool, error) {
 	if err != nil {
 		return "", false, errors.Wrapf(err, "stat for %s", basePath)
 	}
-	// `pathParts` here has lose the root filepath.Separator since built by strings.Split beforehand.
-	// so putting heading separator is required to make it absolute again.
-	b, err := os.Stat(string(filepath.Separator) + filepath.Join(pathParts[:len(basePathParts)]...))
+	rootPath := ""
+	if filepath.VolumeName(basePath) == "" {
+		// `pathParts` here has lose the root filepath.Separator since built by strings.Split beforehand.
+		// so putting heading separator is required to make it absolute again.
```
GMaxera, line 290: This is only valid reasoning in case the basePath does not have a volume name (i.e.: C:), luckily filepath provides an easy way to detect when the path does not have a volume name.

GMaxera:
@brandonSc Thanks to your help, I was able to build Earthly for Windows by myself and then start to tweak and test on my setup the fixes.
I'm not an expert GoLang and I have no clue about code style and policy for PRs. So, please, indicate to me any changes required to meet the standards of Earthly.

util/gitutil/detectgit.go
brandonSc: Seems reasonable to me.

brandonSc:
Looking good so far, thanks for working on this!

I have a fix for another part to the issue, in particular, parsing the path to an Earthfile that is not in the current directory (like `earthly .\some\dir+target`) - in my PR here: https://github.com/earthly/earthly/pull/1683/files

I think these two PRs can work together to fix some of the Windows issues we're running into.

GMaxera:
> Looking good so far, thanks for working on this!

No worries. Earthly suits perfectly in our project and it will replace Makefile and Dockerfile, so I'm happy to give my help towards it.

> I think these two PRs can work together to fix some of the Windows issues we're running into.

Feel free to take the code from this PR and include in the other if you prefer :-)

brandonSc:
I tested our two branches merged together, and it seems to be working well for me locally in Powershell. For example, this command now runs successfully:
``` 
.\build\windows\amd64\earthly.exe .\examples\cache-command\npm+docker
```

Whereas on the current main branch, there are both issues with parsing the path `.\examples\cache-command\npm+docker`, and also issues finding the local `git` directory.

The next thing would be to figure out what's causing all the failures in Github Actions.

If you don't mind @GMaxera, I could just pull the commits from your branch into mine, and continue to work out the test failures there?

brandonSc:
Close PR since these commits are now included in https://github.com/earthly/earthly/pull/1683/files

Thanks again for your help, and glad to hear that Earthly is well suited to your project 😄

PR #1683, opened 2022-02-17
Various Windows Path Fixes

brandonSc:
There are three issue when running on Windows natively (i.e. not WSL) which have been fixed here:

1. Earthly didn't detect the local `git` path properly, since it tries to prepend file-separator to make a path absolute (on Windows that doesn't work since the absolute path is something like `C:/path/etc`)
2. Incorrect parsing of paths to Earthfiles that aren't in the current directory, for example, running `earthly .\path\to+target` was incorrectly treating the `\` as an escape key instead of a path separator.
3. When saving a local artifact, Earthly did not convert a path with `/` to `\` properly for Windows 

Some of the code here has been contributed by @GMaxera in [a separate PR](https://github.com/earthly/earthly/pull/1681), many thanks to them for all their help with this! 

Note that I've pulled the commits from #1681 into this PR to wrap up end-to-end testing and get the whole thing through the finish line.


brandonSc:
Relates to: https://github.com/earthly/earthly/pull/1681

Earthfile
```diff
@@ -238,7 +238,7 @@ earthly:
     ARG VERSION="dev-$EARTHLY_TARGET_TAG_DOCKER"
     ARG EARTHLY_GIT_HASH
     ARG DEFAULT_BUILDKITD_IMAGE=docker.io/earthly/buildkitd:$VERSION # The image needs to be fully qualified for alternative frontend support.
-    ARG BUILD_TAGS=dfrunmount dfrunsecurity dfsecrets dfssh dfrunnetwork dfheredoc
+    ARG BUILD_TAGS=dfrunmount dfrunsecurity dfsecrets dfssh dfrunnetwork dfheredoc forceposix
```
brandonSc, line 241: From @GMaxera:
> The https://github.com/jessevdk/go-flags package automatically uses the / for options when detect is building on Windows. This is, of course, undesired in the case of Earthfile. The forceposix makes go-flags always use the dashes for the options.

This seems to check out with the docs here: https://pkg.go.dev/github.com/jessevdk/go-flags#hdr-Supported_features

@alexcb what do you think?

domain/domain_windows_test.go
```diff
@@ -0,0 +1,160 @@
+//go:build windows
```
brandonSc, line 1: These tests only run on Windows since `filepath` behaviour is platform dependent.

Earthfile
GMaxera: To be clear, this fixes a third issue happening on Windows machines. When in an Earthfile you write a command like for example:
```
SAVE ARTIFACT --force ./some/path AS ...
```
without the "forceposix", the parser will expect `/force` to specify flags and it will mistake `/some/path` as a flag.
With "forceposix" the file is correctly interpreted in the same (exact) way both on Windows and non-Windows environments.

earthfile2llb/converter.go
```diff
@@ -779,7 +779,7 @@ func (c *Converter) canSave(ctx context.Context, saveAsLocalTo string) (bool, er
 	if !basePathExists {
 		return false, fmt.Errorf("no such directory: %s", basepath)
 	}
-	basepath += "/"
+	basepath += string(filepath.Separator)
```
brandonSc, line 782: This fixes a bug with `SAVE ARTIFACT` failing to write a specified path on Windows. I tried a simple example, on the current `main` version of earthly, I get an error like this on Powershell:
```
Error: build target: build main: bkClient.Build: failed to solve: Earthfile line 22:4 apply SAVE ARTIFACT: unable to save to ./hello/hi; path must be located under .
in              +save
```
But on this branch with the file separator fix, I see the output looks appropriate for Windows now:
```
Artifact +save/hello output as .\hello\hi
```

brandonSc: Note that this still uses forward-slash in the Earthfile, here's the one from my test:
```earthly
VERSION 0.6

save:
    FROM alpine
    RUN echo "hi" > hello
    SAVE ARTIFACT hello AS LOCAL ./hello/hi
```

domain/domain_windows_test.go
brandonSc: @dchw do we have a way to schedule tests like this to run on our Windows VM, or something like that?

earthfile2llb/converter.go
GMaxera: We have SAVE ARTIFACT ... AS LOCAL /with/path/like/that using forward slashes and this PR handles them both under Windows.
Or in other words, keeping forward slashes inside Earthfile are interpreted correctly as file path when running under Windows with the fixes of this PR 👍

alexcb (approved):
very nice tests!

Were you able to verify that something like

```Earthly
FROM alpine

a:
    WORKDIR /test
    RUN echo data > file\ with\ space
    SAVE ARTIFACT /test

b:
    COPY +a/test/file\ with\ space .
    RUN cat file\ with\ space | grep data
    RUN ls | grep 'file with space'
```

would continue to work?

GMaxera:
> 
> Were you able to verify that something like
> 
> 
> would continue to work?

Just tested on my Windows 10 machine with Earthly built from this PR:

```
               +base | --> FROM alpine
               +base | [██████████] 100% resolve docker.io/library/alpine@sha256:21a3deaa0d32a8057914f36584b5288d2e5ecc984380bc0118285c70fa8c9300
                  +a | --> WORKDIR /test
                  +a | --> RUN echo data > file\ with\ space
                  +a | --> SAVE ARTIFACT /test +a/test
                  +b | --> COPY +a/test/file with space ./
                  +b | --> RUN cat file\ with\ space | grep data
                  +b | data
                  +b | --> RUN ls | grep 'file with space'
                  +b | file with space
```

domain/domain_windows_test.go
dchw: No; we would need to configure an additional Windows runner.

brandonSc: Alright, perhaps a goal for us another day

dchw:
Knowing that the "FROM DOCKERFILE" tests will still not work, and that we do not yet have CI for native windows... @brandonSc can you run the examples and other integration tests?

brandonSc:
> Knowing that the "FROM DOCKERFILE" tests will still not work, and that we do not yet have CI for native windows... @brandonSc can you run the examples and other integration tests?

Yep, I'll kick that off on my Windows machine now and let you know the results

brandonSc:
> Knowing that the "FROM DOCKERFILE" tests will still not work, and that we do not yet have CI for native windows... @brandonSc can you run the examples and other integration tests?

Most of the examples work:
* `+examples1` works if I comment out the gRPC test (that one also fails for me on WSL for the same reason, it tries to download a version of `linux-libc-dev` that doesn't exist).
* `+examples2` works just fine.
* `+test` fails on the AST things.

PR #1685, opened 2022-02-18
Use new attachable when Podman is in use, Docker otherwise.

### LINKED ISSUE #1644: (podman) buildkitd is not able to pull private images
Hi there,

I'm trying to use earthly with podman (rootless), but I'm struggling to get our private registry ([ghcr.io](http://ghcr.io/)) to authenticate. Using `podman pull ghcr.io/username/repo` works as expected after podman login, but earthly - when trying to pull the same image - returns `failed to fetch anonymous token: unexpected status: 401 Unauthorized` .

- I'm running earthly 0.6.6 (installed via asdf) with podman 3.4.4 on Arch Linux.
- I had to modify my earthly config to define iptables-nft to get buildkitd to work with podman on my machine (because I use nf_tables  on my host)
- Running podman logs against buildkitd produces the following logs: https://termsend.pl/o/qq354

dchw:
Depends on this PR: https://github.com/earthly/buildkit/pull/72 to load Podman config from the default directory. Should fix #1644 once this and the referenced Buildkit PRs are merged.


cmd/earthly/main.go
```diff
@@ -2944,11 +2944,22 @@ func (app *earthlyApp) actionBuildImp(c *cli.Context, flagArgs, nonFlagArgs []st
...
 	}
 
+	switch app.containerFrontend.Config().Setting {
+	case containerutil.FrontendDocker, containerutil.FrontendDockerShell:
+		attachables = append(attachables, authprovider.NewDockerAuthProvider(os.Stderr))
+
+	case containerutil.FrontendPodman, containerutil.FrontendPodmanShell:
+		attachables = append(attachables, authprovider.NewPodmanAuthProvider(os.Stderr))
+
+	default:
```
brandonSc, line 2958: Nit: Do we need the `default` case? Might be more concise as if/else:
```golang
if containerutil.FrontendPodman || containerutil.FronendPodmanShell {
   // .. 
} else {
   // docker
}
```

brandonSc (approved):

cmd/earthly/main.go
dchw: We may get more frontends in the future (see things like nerdctl), and it really explicitly enumerates why each part is there. I prefer this over the `if` for explicitness in the _why_ each thing is there.

Issue #1168, opened 2021-08-16
add a HOSTS command to add entries to /etc/hosts


alexcb:
It's not trivial to modify `/etc/hosts`; it's currently set as read-only:

```
FROM alpine

addhost:
    RUN cat /etc/hosts
    RUN echo "myhost 127.0.0.1" >> /etc/hosts
    RUN cat /etc/hosts
```

results in:

```
$ earthly +addhost 
           buildkitd | Found buildkit daemon as docker container (earthly-buildkitd)
              alpine | --> Load metadata linux/amd64
               +base | --> FROM alpine
               +base | [██████████] resolve docker.io/library/alpine@sha256:eb3e4e175ba6d212ba1d6e04fc0782916c08e1c9d7b45892e9796141b1d379ae ... 100%
            +addhost | --> RUN cat /etc/hosts
            +addhost | 127.0.0.1	localhost buildkitsandbox
            +addhost | ::1	localhost ip6-localhost ip6-loopback
            +addhost | --> RUN echo "myhost 127.0.0.1" >> /etc/hosts
            +addhost | /bin/sh: can't create /etc/hosts: Read-only file system
            +addhost | Command /bin/sh -c 'echo "myhost 127.0.0.1" >> /etc/hosts' failed with exit code 1
            +addhost | 
            +addhost | WARN: (RUN echo "myhost 127.0.0.1" >> /etc/hosts) process "/bin/sh -c PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin /usr/bin/earth_debugger /bin/sh -c 'echo \"myhost 127.0.0.1\" >> /etc/hosts'" did not complete successfully: exit code: 1
Error: build target: build main: bkClient.Build: failed to solve: process "/bin/sh -c PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin /usr/bin/earth_debugger /bin/sh -c 'echo \"myhost 127.0.0.1\" >> /etc/hosts'" did not complete successfully: exit code: 1
```

The current workaround is to copy `/etc/hosts` elsewhere, then unmount the read-only version, then copy it back and then modify it:

```
FROM alpine

addhost:
    RUN --privileged cp /etc/hosts /tmp/hosts && \
        umount /etc/hosts && \
        cp /tmp/hosts /etc/hosts && \
        echo "127.0.0.1 myhost" >> /etc/hosts && \
        cat /etc/hosts && \
[...]

PR #1688, opened 2022-02-19
Add HOST command

dchw:
Implements the feature discussed in #1168. Pretty simple, direct mapping with no frills.

Since `/etc/hosts` tends to be controlled when launching built containers; these may not persist or be accessible once the image is built. However, they will persist for the duration of your build.

Example:

```
add-single-host:
    FROM alpine:3.15
    HOST example.com 1.2.3.4
    RUN wget example.com
```


features/features.go
```diff
@@ -27,6 +27,7 @@ type Features struct {
 	EarthlyVersionArg          bool `long:"earthly-version-arg" description:"includes EARTHLY_VERSION and EARTHLY_BUILD_SHA ARGs"`
 	ExplicitGlobal             bool `long:"explicit-global" description:"require base target args to have explicit settings to be considered global args"`
 	UseCacheCommand            bool `long:"use-cache-command" description:"allow use of CACHE command in Earthfiles"`
+	UseHostCommand             bool `long:"use-host-command" description:"allow use of HOST command in Earthfiles"`
```
vladaionescu, line 30: Mind adding this to 0.7? It's lower in this file somewhere.

vladaionescu (approved):
Very cool 🎉

vladaionescu:
Oh - mind adding an item in the changelog? We should mention how users can enable this experimental feature.

dchw:
Can do!

Issue #1647, opened 2022-02-08
WITH DOCKER use causing rm: can't remove /var/earthly/dind....: Resource busy


willysamz:
### Setups Tried:
- Jenkins running in EKS with k8s plugin & separate earthlly/buildkitd deployment (taken from the docs)
- Jenkins running in EKS with k8s plugin & single pod with both earthly/buildkitd & custom container with Earthly installed (will post this info if requested)

### Versions
Earthly v0.6.6, buildkitd/earthly v0.6.6
Docker Version from inside custom image with earthly installed
```
Client: Docker Engine - Community
 Version:           19.03.15
 API version:       1.40
 Go version:        go1.13.15
 Git commit:        99e3ed8919
 Built:             Sat Jan 30 03:17:11 2021
 OS/Arch:           linux/amd64
 Experimental:      false

Server:
 Engine:
  Version:          20.10.7
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.15.14
  Git commit:       b0f5bc3
  Built:            Wed Nov 17 03:06:14 2021
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.6
  GitCommit:        d71fcd7d8303cbf684402823e425e9dd2e99285d
 runc:
  Version:          1.0.0
  GitCommit:        84113eef6fc27af1b01b3181f31bbaf708715301
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```
### Custom image with Earthly installed which is used by the Jenkins plugin to run Earthly commands inside of
```yaml
apiVersion: v1
kind: Pod
spec:
  containers:
  - name: earthly
    image: ...customimage:tag
    command:
    - cat
    tty: true
    volumeMounts:
    - name: docker
      mountPath: '/var/run/docker.sock'
    - mountPath: /tmp/earthly
      name: buildkitd-temp
    securityContext:
      privileged: true
    resources:
      limits:
        memory: "768Mi"
        cpu: 2
      requests:
        memory: "512Mi"
        cpu: 2
    env:
      - name: BUILDKIT_HOST
        value: "tcp://earthly-buildkitd.buildkit.svc.cluster.local:8372"
      - name: EARTHLY_BUILDKIT_HOST
        value: "tcp://earthly-buildkitd.buildkit.svc.cluster.local:8372"
   volumes:
  - name: docker
    hostPath:
      path: '/var/run/docker.sock'
  - name: buildkitd-temp
[...]

PR #1692, opened 2022-02-23
DIND Cleanup Improvements

dchw:
While we can't say for certain that this will fix issues like #1647, It does fix all the ways I could make it replicate; and improve logging for when it may fail again.

### DIFF
--- a/buildkitd/dockerd-wrapper.sh
+++ b/buildkitd/dockerd-wrapper.sh
@@ -126,7 +126,7 @@ stop_dockerd() {
     # Kill dockerd in a subshell and wait for the PID to exit. This ensures that dockerd cannot keep any files in
     # EARTHLY_DOCKERD_DATA_ROOT open.
     (kill_dockerd)
-    wait $dockerd_pid || true
+    wait "$dockerd_pid" || true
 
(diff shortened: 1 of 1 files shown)


buildkitd/dockerd-wrapper.sh
```diff
@@ -106,32 +106,54 @@ EOF
...
+  echo "==== End dockerd logs ===="
+}
+
 stop_dockerd() {
     dockerd_pid="$(cat /var/run/docker.pid)"
+
+    # Kill dockerd in a subshell and wait for the PID to exit. This ensures that dockerd cannot keep any files in
+    # EARTHLY_DOCKERD_DATA_ROOT open.
+    (kill_dockerd)
+    wait "$dockerd_pid" || true
```
alexcb, line 129: can we move this lint into `kill_dockerd` (as well as the `dockerd_pid="$(cat /var/run/docker.pid)"`).

and then just change it to call and block on `kill_dockerd` (rather than `(kill_dockerd)`).

Then there's no need to have `dockerd_pid` set in this function.

alexcb (approved):

vladaionescu (approved):

Issue #1696, opened 2022-02-25
`WITH DOCKER --load` with build args is broken


vladaionescu:
For example, the following doesn't work:

```Earthfile
WITH DOCKER --load=(+test-img --workbox="$workbox")
    RUN docker run ...
END
```

A temporary workaround is to use the old `--build-arg` flag instead:

```Earthfile
WITH DOCKER --build-arg workbox="$workbox" --load=+test-img
    RUN docker run ...
END
```

PR #1697, opened 2022-02-26
Fix `WITH DOCKER --load` parsing

vladaionescu:
* Fixes https://github.com/earthly/earthly/issues/1696
* Update parser to a newer JDK, which results in less errors when emulated on the M1
* Fix a race condition that resulted in a `Canceled` error showing up instead of the real error


Earthfile
```diff
@@ -152,7 +152,7 @@ lint-newline-ending:
     # check for files with trailing newlines
     RUN set -e; \
         code=0; \
-        for f in $(find . -type f \( -iname '*.go' -o -iname 'Earthfile' -o -iname '*.earth' -o -iname '*.md' \) | grep -v "ast/tests/empty-targets.earth" ); do \
```
alexcb, line 155: haha.. guess antlr got an update :)

earthfile2llb/interpreter_test.go
```diff
@@ -70,3 +74,30 @@ func TestNegativeParseParans(t *testing.T) {
 		assert.Error(t, err)
 	}
 }
+
+func TestProcessParansAndQuotes(t *testing.T) {
```
alexcb, line 78: nice to see this corresponding test :sweat_smile:

alexcb (approved):

Issue #1575, opened 2022-01-14
Proposal: New `ARG` behavior and `SET`


vladaionescu:
##### Problem 1: Inefficient cache

* `ARG`s in Earthly differ from `ARG` in Dockerfiles by the fact that Earthly `ARG`s are also env vars
* Because they are also env vars, every `RUN` command automatically depend on the `ARG`s, even if the program(s) being run don't actually use those env vars
* This leads to inefficient use of cache
* This is most painful when there are global `ARG`s defined and their value is changed ever so slightly. Even if only a subset of the targets may be using that global `ARG`, most targets will need to be re-executed.

##### Problem 2: Global `ARG`s have non-obvious behaviors

* When an `ARG` is declared in a base target, it automatically becomes a global arg. This subtle difference in behavior is not obvious to a newcomer. Ideally, an `ARG` declared in a non-base target is the same as an arg declared in the base target.
* Combined with problem 1, this can lead to situations where the cache is used very inefficiently, if the user intends to declare a local `ARG` in base, but inadvertently declares a global (there is currently no way to declare a local `ARG` in base)

##### Problem 3: Redeclaring an ARG with a different value behaves in non-obvious ways

The following snippet does not work and there is no error thrown:

```Earthly
ARG FOO=1
IF ...
    ARG FOO=2
END
```

When an ARG is redeclared it actually does not get its value overridden. The reason is that the value on the right of `=` is actually a new **default value**, not a new true **value**. This can be very surprising for the user.

##### Problem 4: No way to change a value of an `ARG` within the same target

* There is currently no way to change the value of an `ARG` after it has been declared

### Proposal

To address the issues above, we propose the following changes:

Problem 1:

* `ARG`s don’t also set env vars by default
* `ARG --env FOO=bar` can be used to also set the env var (basically brings back the old functionality)
* We replace `ARG`s in `RUN` commands the same way we do for all other commands, and don’t rely on the shell to replace via env vars
* A benefit is that builtin args would not need to be pre-declared, because we could automatically detect where they are used exactly, and therefore we would not cause random cache busts
* The same is true for global `ARG`s - changing a value of a global `ARG` would only influence cache for the commands that actually use that `ARG`
* This all means that `ARG`s would work the same as Dockerfile `ARG`s (helps with better Dockerfile parity)

Problem 2:

* Global ARGs need to be declared as global explicitly:
   ```Earthly
   # Not a global
   ARG FOO
   # Global
   ARG --global FOO
   target:
      # Local
[...]

PR #1699, opened 2022-02-28
Change ARG shell-execution

alexcb:
ARG values such as `foo$(cat data)` were not being expanded.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ All notable changes to [Earthly](https://github.com/earthly/earthly) will be doc
 
 - Earthly is now 15-30% faster when executing large builds [#1589](https://github.com/earthly/earthly/issues/1589)
 - Experimental `HOST` command, which can be used like this: `HOST <domain> <ip>` to add additional hosts during the execution of your build. To enable this feature, use `VERSION --use-host-command 0.6`. [#1168](https://github.com/earthly/earthly/issues/1168)
+- Ability to shell-out in any Earthly command, (e.g. `SAVE IMAGE myimage:$(cat version)`), as well as in the middle of ARG strings. To enable this feature, use `VERSION --shell-out-anywhere 0.6`.
 
 ### Fixed
 
--- a/earthfile2llb/converter.go
+++ b/earthfile2llb/converter.go
@@ -311,8 +311,11 @@ func (c *Converter) FromDockerfile(ctx context.Context, contextPath string, dfPa
 		}
 		BuildContextFactory = data.BuildContextFactory
 	}
-	overriding, err := variables.ParseArgs(
-		buildArgs, c.processNonConstantBuildArgFunc(ctx), c.varCollection)
+	var pncvf variables.ProcessNonConstantVariableFunc
+	if !c.opt.Features.ShellOutAnywhere {
+		pncvf = c.processNonConstantBuildArgFunc(ctx)
+	}
+	overriding, err := variables.ParseArgs(buildArgs, pncvf, c.varCollection)
 	if err != nil {
 		return err
 	}
@@ -597,11 +600,31 @@ func (c *Converter) RunExitCode(ctx context.Context, opts ConvertRunOpts) (int, 
 // RunExpression runs an expression and returns its output. The run is transient - any state created
 // is not used in subsequent commands.
 func (c *Converter) RunExpression(ctx context.Context, expressionName string, opts ConvertRunOpts) (string, error) {
+	return c.runCommand(ctx, expressionName, true, opts)
+}
+
+// RunCommand runs a command and returns its output. The run is transient - any state created
+// is not used in subsequent commands.
+func (c *Converter) RunCommand(ctx context.Context, commandName string, opts ConvertRunOpts) (string, error) {
+	return c.runCommand(ctx, commandName, false, opts)
+}
+
+func (c *Converter) runCommand(ctx context.Context, outputFileName string, isExpression bool, opts ConvertRunOpts) (string, error) {
 	err := c.checkAllowed(runCmd)
 	if err != nil {
 		return "", err
 	}
 	c.nonSaveCommand()
+
+	if !opts.WithShell {
+		panic("runCommand must be called WithShell")
+	}
+	if opts.Locally == opts.Transient {
+		panic("runCommand Transient xor Locally must be true")
+	}
+	if opts.shellWrap != nil {
+		panic("runCommand expects shellWrap to be nil (as it is overridden)")
+	}
 
 	var outputFile string
 	if opts.Locally {
@@ -615,7 +638,7 @@ func (c *Converter) RunExpression(ctx context.Context, expressionName string, op
 		})
 	} else {
 		srcBuildArgDir := "/run/buildargs"
-		outputFile = path.Join(srcBuildArgDir, expressionName)
+		outputFile = path.Join(srcBuildArgDir, outputFileName)
 		opts.statePrep = func(ctx context.Context, state pllb.State) (pllb.State, error) {
 			return state.File(
 				pllb.Mkdir(srcBuildArgDir, 0777, llb.WithParents(true)), // Mkdir is performed as root even when USER is set; we must use 0777
@@ -626,10 +649,12 @@ func (c *Converter) RunExpression(ctx context.Context, expressionName string, op
 		}
 	}
 
-	// Perform execution, but append the command with the right shell incantation that
-	// causes it to output to a file. This is done via the shellWrap.
-	opts.shellWrap = withShellAndEnvVarsOutput(outputFile)
-	opts.WithShell = true // force shell wrapping
+	if isExpression {
+		opts.shellWrap = expressionWithShellAndEnvVarsOutput(outputFile)
+	} else {
+		opts.shellWrap = withShellAndEnvVarsOutput(outputFile)
+	}
+
 	state, err := c.internalRun(ctx, opts)
 	if err != nil {
 		return "", err
@@ -639,7 +664,7 @@ func (c *Converter) RunExpression(ctx context.Context, expressionName string, op
 	if opts.Locally {
 		outputDt, err = os.ReadFile(outputFile)
 		if err != nil {
-			return "", errors.Wrap(err, "read exit code file")
+			return "", errors.Wrap(err, "read output file")
 		}
 	} else {
 		ref, err := llbutil.StateToRef(ctx, c.opt.GwClient, state, c.opt.NoCache, c.opt.Platform, c.opt.CacheImports.AsMap())
@@ -1094,8 +1119,13 @@ func (c *Converter) Arg(ctx context.Context, argKey string, defaultArgValue stri
 		return err
 	}
 	c.nonSaveCommand()
-	effective, effectiveDefault, err := c.varCollection.DeclareArg(
-		argKey, defaultArgValue, opts.Global, c.processNonConstantBuildArgFunc(ctx))
+
+	var pncvf variables.ProcessNonConstantVariableFunc
+	if !c.opt.Features.ShellOutAnywhere {
+		pncvf = c.processNonConstantBuildArgFunc(ctx)
+	}
+
+	effective, effectiveDefault, err := c.varCollection.DeclareArg(argKey, defaultArgValue, opts.Global, pncvf)
 	if err != nil {
 		return err
 	}
@@ -1284,8 +1314,11 @@ func (c *Converter) EnterScopeDo(ctx context.Context, command domain.Command, ba
 		return err
 	}
 
-	overriding, err := variables.ParseArgs(
-		buildArgs, c.processNonConstantBuildArgFunc(ctx), c.varCollection)
+	var pncvf variables.ProcessNonConstantVariableFunc
+	if !c.opt.Features.ShellOutAnywhere {
+		pncvf = c.processNonConstantBuildArgFunc(ctx)
+	}
+	overriding, err := variables.ParseArgs(buildArgs, pncvf, c.varCollection)
 	if err != nil {
 		return err
 	}
@@ -1329,8 +1362,14 @@ func (c *Converter) FinalizeStates(ctx context.Context) (*states.MultiTarget, er
 }
 
 // ExpandArgs expands args in the provided word.
-func (c *Converter) ExpandArgs(word string) string {
-	return c.varCollection.Expand(word)
+func (c *Converter) ExpandArgs(ctx context.Context, runOpts ConvertRunOpts, word string) (string, error) {
+	if !c.opt.Features.ShellOutAnywhere {
+		return c.varCollection.ExpandOld(word), nil
+	}
+	return c.varCollection.Expand(word, func(cmd string) (string, error) {
+		runOpts.Args = []string{cmd}
+		return c.RunCommand(ctx, "internal-expand-args", runOpts)
+	})
 }
 
 func (c *Converter) prepBuildTarget(ctx context.Context, fullTargetName string, platform *specs.Platform, allowPrivileged bool, buildArgs []string, isDangling bool, cmdT cmdType) (domain.Target, ConvertOpt, bool, error) {
@@ -1351,7 +1390,12 @@ func (c *Converter) prepBuildTarget(ctx context.Context, fullTargetName string, 
 	}
 	target := targetRef.(domain.Target)
 
-	overriding, err := variables.ParseArgs(buildArgs, c.processNonConstantBuildArgFunc(ctx), c.varCollection)
+	var pncvf variables.ProcessNonConstantVariableFunc
+	if !c.opt.Features.ShellOutAnywhere {
+		pncvf = c.processNonConstantBuildArgFunc(ctx)
+	}
+
+	overriding, err := variables.ParseArgs(buildArgs, pncvf, c.varCollection)
 	if err != nil {
 		return domain.Target{}, ConvertOpt{}, false, errors.Wrap(err, "parse build args")
 	}
@@ -1779,6 +1823,7 @@ func (c *Converter) processNonConstantBuildArgFunc(ctx context.Context) variable
 			CommandName: fmt.Sprintf("ARG %s = RUN", name),
 			Args:        strings.Split(expression, " "),
 			Transient:   true,
+			WithShell:   true,
 		}
 		output, err := c.RunExpression(ctx, name, opts)
 		if err != nil {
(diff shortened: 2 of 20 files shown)


alexcb:
I added in

```
// containsShell returns true for strings containing $(
// except for cases where escaped: e.g. \$(
// or cases with singlquotes: '$(...'
func containsShell(s string) bool {
	var escaped bool
	var singlequoted bool
	var last rune
	for _, c := range s {
		fmt.Printf("got %s\n", string(c))
		if escaped {
			escaped = false
			last = 0
			continue
		}
		if c == '\\' {
			escaped = true
			last = 0
			continue
		}
		if c == '\'' {
			singlequoted = !singlequoted
			last = 0
			continue
		}
		if last == '$' && c == '(' && !singlequoted {
			return true
		}
		last = c
	}
	return false
}
```

however given a file like

```
test7:
    FROM scratch
    ARG VAR1="literal\$(string)"
    FROM alpine:3.15
    RUN env | grep lite
```

the parser is removing the escaping too early and by the time it gets to `parseArgValue(...)` it is `literal$(string)`, which then tries to execute (and fails since it's in the `SCRATCH` context), so we're going to have to change the parser to pass along the raw string, e.g. `"literal\$(string)"` rather than `literal$(string)``.

alexcb:
This PR will also allow the following to work:

```Earthfile
VERSION 0.6

test7:
    FROM alpine
    RUN echo magic > /my-image-name
    SAVE IMAGE $(cat /my-image-name):$(date +%s)
```

```
...
  /h/a/b/a/foo+test7 | --> RUN echo magic > /my-image-name
  /h/a/b/a/foo+test7 | --> ARG earthly-expand-arg = RUN $(cat /my-image-name):$(date +%s)      <---- this needs some fixing, since it's not an "ARG"

...

 4. Local Output 🎁

Image /home/alex/broken/args/foo+test7 output as magic:1646440873            <----- The image name
```

variables/parse.go
```diff
@@ -88,20 +83,36 @@ func parseArg(arg string, pncvf ProcessNonConstantVariableFunc, current *Collect
...
-	if strings.HasPrefix(value, "$(") {
-		// Variable build arg - resolve value.
-		var err error
-		value, _, err = pncvf(name, value)
-		if err != nil {
-			return "", err
+// ContainsShell returns true for strings containing $(
+// except for cases where escaped: e.g. \$(
+// or cases with singlquotes: '$(...'
+func ContainsShell(s string) bool {
```
alexcb, line 89: this should be made private moved into earthfile2llb.

earthfile2llb/converter.go
```diff
@@ -1300,8 +1300,16 @@ func (c *Converter) FinalizeStates(ctx context.Context) (*states.MultiTarget, er
 }
 
 // ExpandArgs expands args in the provided word.
-func (c *Converter) ExpandArgs(word string) string {
-	return c.varCollection.Expand(word)
+func (c *Converter) ExpandArgs(ctx context.Context, word string) (string, error) {
+	if !variables.ContainsShell(word) {
+		return c.varCollection.Expand(word), nil
+	}
+	pncvf := c.processNonConstantBuildArgFunc(ctx)
```
vladaionescu, line 1307: This doesn't feel right - this will replace the **entire** value, not just the `$(...)` part.

The code that detects whether a word contains shell, needs to be expanded to also detect which part of that to replace.

vladaionescu, line 1307: Another situation: what if there are multiple parts in the string to replace.

```
ARG greeting="hello $(id -F), the date is $(date)"
```

variables/parse.go
```diff
@@ -88,20 +83,36 @@ func parseArg(arg string, pncvf ProcessNonConstantVariableFunc, current *Collect
...
+			continue
+		}
+		if last == '$' && c == '(' && !singlequoted {
+			return true
+		}
+		last = c
 	}
-
-	return value, nil
+	return false
```
vladaionescu, line 115: I would try to reuse this code: https://github.com/moby/buildkit/blob/master/frontend/dockerfile/shell/lex.go#L13. It would require modifications, to look for `$(...)` and not just for `$ABC` and `${...}`, but I think it already covers a lot of edge cases and it's battle-tested.

earthfile2llb/converter.go
alexcb: it should keep the complete string in tact, here's a test I added https://github.com/earthly/earthly/pull/1699/files#diff-b3414e5141e314866d254a961baf56400082c346f72c0633d94b3ec745dfd93bR65-R68

it does "hello$(cat /data)" which ends up "helloworld".

alexcb: I'll create another test to verify this two-parter works.

vladaionescu: Oh I see - it works because they are part of different words, I think. There may be edge cases where they are in the same word, and that will break it.

alexcb: this handles multiple parts plus an env:
https://github.com/earthly/earthly/pull/1699/files#diff-b3414e5141e314866d254a961baf56400082c346f72c0633d94b3ec745dfd93bR85-R91

variables/parse.go
alexcb: The tricky part is this code would have to run in a container attached to the current llb.State and not within buildkitd.

vladaionescu: Yeah.. we'd have to use a function as the value to replace the variable with.

Maybe it's overkill - no strong opinion either way. Maybe it's better to have our own code for this.

vladaionescu:
BTW, don't forget about `isSafeAsyncBuildArgs` in `interpreter.go`.

CHANGELOG.md
```diff
@@ -46,6 +47,7 @@ All notable changes to [Earthly](https://github.com/earthly/earthly) will be doc
   build args. [#1696](https://github.com/earthly/earthly/issues/1696)
 - `ENTRYPOINT` and `CMD` were not properly expanding args when used in shell mode.
 - A race condition sometimes caused a `Canceled` error to be reported, instead of the real error that caused the build to fail
+- ARG values now support shelling-out in the middle of the string (e.g. `foo$(cat data)` will now work).
```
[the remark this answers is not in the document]
vladaionescu: The docs might need adjusting too, when we release 0.7.

earthfile2llb/converter.go
```diff
@@ -1320,8 +1320,47 @@ func (c *Converter) FinalizeStates(ctx context.Context) (*states.MultiTarget, er
...
+			singlequoted = !singlequoted
+			last = 0
+			continue
+		}
+		if last == '$' && c == '(' && !singlequoted {
+			return true
+		}
+		last = c
+	}
+	return false
```
vladaionescu, line 1363: We should unit-test this function really well.

Do we handle multiple parentheses well? e.g. `(($something$(command)))` or `(${something:$(cat ..)})`

vladaionescu (commented):
Two major thoughts on this (and a few relatively minor additional ones inline):

* The more I think about, the more I get the feeling that it's a bit hacky that the string is interpreted by Earthly entirely... unless there's something like `$(` in part of the string... then the *entire* string ends up being interpreted by the shell. I think it should be much more clear-cut: the string should be interpreted *entirely* by Earthly **always**, except for the part in-between `$(...)`, which is passed on to the shell. I think we anyway need this clarity, especially for this upcoming proposal: https://github.com/earthly/earthly/issues/1575, where we plan on relying on Earthly interpretation much more in the future (and ARGs will no longer be also env vars by default).
* If we introduce this, we *really* need to properly distinguish between `"` and `'`. There's a risk that some strings might trigger the `$(...)` execution logic inadvertently and the user needs to have better control over that.

tests/with-docker/Earthfile
```diff
@@ -1,8 +1,9 @@
-VERSION --parallel-load 0.6
+VERSION --parallel-load --shell-out-anywhere 0.6
```
alexcb, line 1: Is it safe to include `--shell-out-anywhere` here, or should I move the docker tests into the `tests/shell-out-anywhere` directory to ensure with-docker tests run under the old expandArgs code?

earthfile2llb/converter.go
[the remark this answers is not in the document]
alexcb: This has been deleted in favour of using a forked version of buildkit's dockerfile shell lexer, which is now under `util/shell`.

earthfile2llb/interpreter.go
```diff
@@ -376,10 +385,9 @@ func (i *Interpreter) handleForArgs(ctx context.Context, forArgs []string, sl *s
 		return "", nil, i.errorf(sl, "expected IN, got %s", args[1])
 	}
 	variable := args[0]
-	expression := args[2:]
 	runOpts := ConvertRunOpts{
 		CommandName: "FOR",
-		Args:        expression,
+		Args:        []string{strings.Join(args[2:], " ")},
```
vladaionescu, line 390: This feels too arbitrary. The join should only take place in the shell wrappers ideally.

tests/shell-out-anywhere/old.earth
```diff
@@ -0,0 +1,16 @@
...
+
+hasfile:
+    FROM +test
+    RUN touch valid-file
+    SAVE ARTIFACT "valid-$(echo file)" # this should fail
+
+copyfile:
+    FROM alpine:3.15
+    COPY "+hasfile/$(echo dmFsaWQtZmlsZQ== | base64 -d)" . # this too should fail
+    RUN test -f valid-file
```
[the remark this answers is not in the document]
vladaionescu: Maybe also
```
BUILD +something --foo=bar
BUILD +something --foo=$bar
BUILD +something $foo=bar
BUILD +something foo=$(bar)
BUILD +something $(bar)
BUILD +something $bar
```

earthfile2llb/interpreter.go
alexcb: I did some more digging and looks like we'll always get `args[2:]` set as [`./cmd`, `"single arg with space"`, ...]. I was worried about the quotes being stripped at this point, but that's not the case, so I'll back out this change.

earthfile2llb/converter.go
```diff
@@ -626,10 +648,15 @@ func (c *Converter) RunExpression(ctx context.Context, expressionName string, op
...
-	opts.shellWrap = withShellAndEnvVarsOutput(outputFile)
-	opts.WithShell = true // force shell wrapping
+	// Redirect stdout to an output file *before* running any commands.
+	// This should be done rather than appending ">output" to the command, in order to maintain
+	// compatibility with /bin/sh's shelling-out, e.g. consider the command:
+	//   echo "hello $(echo -n one && echo -n two >/dev/null && echo -n three)."
+	// it will produce
+	//   "hello onethree."
+	// however, if we appended ">output", instead earthly would errornously output "hello three."
+	opts.Args = []string{fmt.Sprintf("exec 1<>'%s' && %s", escapeShellSingleQuotes(outputFile), opts.Args[0])}
```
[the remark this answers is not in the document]
alexcb: I initially tried moving the `echo` prepend to occur higher up, because in my mind once we hit a shell wrap, I expect the args to to be a valid command.

I could try adding a new bool such as `isExpression`, or `requiresEcho` to

```
func strWithEnvVarsAndDocker(args []string, envVars []string, withShell, withDebugger, forceDebugger, withDocker bool, exitCodeFile string, isExpression bool) string {
```

which would prepend the `echo`.

alternatively, we could introduce a `command string` arg, which would turn into:

```
func strWithEnvVarsAndDocker(command string, args []string, envVars []string, withShell, withDebugger, forceDebugger, withDocker bool, exitCodeFile string) string {
```

but that might make the cases where it is a valid command more verbose as we would then have to do `strWithEnvVarsAndDocker(args[0], args[1:])` along with `if len(args) == 0 return error` handling.

I'm leaning towards `isExpression`, does that sound reasonable?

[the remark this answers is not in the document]
alexcb: similarly when an output file is passed, i can prepend the `"exec 1<>'%s' && ` bit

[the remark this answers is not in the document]
vladaionescu: Yeah, `isExpression` sounds good. We could maybe also put the args to that into some kind of opt? Perhaps in a different PR.

earthfile2llb/converter.go
```diff
@@ -1328,9 +1353,36 @@ func (c *Converter) FinalizeStates(ctx context.Context) (*states.MultiTarget, er
...
+			return fmt.Sprintf("%s=%s", splitArg[0], expanded), nil
+		}
+
+		return c.varCollection.ExpandOld(word), nil
+	}
+	s := c.varCollection.Expand(word, func(cmd string) (string, error) {
+		runOpts.Args = []string{cmd}
+		return c.RunCommand(ctx, "internal-expand-args", runOpts)
+	})
+	return s, nil
```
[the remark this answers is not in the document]
alexcb: That might be tricky, since this function needs to call `c.RunExpression` and `c.RunCommand` (where `c` is `*Converter`).

I could move some of the logic that's required to maintain the old behaviour into `varCollection.ExpandOld` and pass along either a callback to `RunExpression` or a `*Converter` pointer. I'm not sure if it'll clean things up to be honest.

earthfile2llb/converter.go
```diff
@@ -596,13 +595,36 @@ func (c *Converter) RunExitCode(ctx context.Context, opts ConvertRunOpts) (int,
 
 // RunExpression runs an expression and returns its output. The run is transient - any state created
 // is not used in subsequent commands.
+// e.g. the expression "hello $(cat /data)" gets transformed into /bin/sh -c 'echo "hello $(cat /data)"', before being executed
 func (c *Converter) RunExpression(ctx context.Context, expressionName string, opts ConvertRunOpts) (string, error) {
+	if len(opts.Args) != 1 {
+		panic("RunExpression expects a single Arg (which is evaluated via /bin/sh -c 'echo <arg>'")
+	}
```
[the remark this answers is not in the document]
alexcb: This has been removed now that I created `isExpression` and moved the arg handling back into `shell.go`.

tests/shell-out-anywhere/old.earth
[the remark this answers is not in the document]
alexcb: I expanded the test to cover these cases; for cases such as `BUILD +something $(bar)`, are you suggesting the literal `bar` command (i.e. a command that doesn't exist, or fails)? or more of a `BUILD +something $(echo bar)` command?

One interesting change is that under the new behaviour, if the shell-out command returns a failures, we will bubble up that error (or allow the user to enter an `--interactive` shell). This deviates a bit from the default bash experience. If a user truely wants to continue on failures, then I think we'll want to point out a user can do `ARG key=$(maybe || true)`.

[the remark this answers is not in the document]
vladaionescu: > I expanded the test to cover these cases; for cases such as BUILD +something $(bar), are you suggesting the literal bar command (i.e. a command that doesn't exist, or fails)? or more of a BUILD +something $(echo bar) command?

I was thinking of a non-failure `$(echo bar)`.

Yeah... I think not allowing failure is probably a good idea. Not obviously good. But probably good.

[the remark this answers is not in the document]
alexcb: I have a number of `$(cat ...)` tests under https://github.com/earthly/earthly/pull/1699/files#diff-c25c0f9f0fc0a43cd042718f15696d9dfb71eddbdb4932f3781143c93b1e3d44R62-R82

plus the other case of allow failures in the old version and catching them in the new.

I think we're covered on both sides now :)

earthfile2llb/converter.go
```diff
@@ -596,13 +595,34 @@ func (c *Converter) RunExitCode(ctx context.Context, opts ConvertRunOpts) (int,
 
 // RunExpression runs an expression and returns its output. The run is transient - any state created
 // is not used in subsequent commands.
+// e.g. the expression "hello $(cat /data)" gets transformed into /bin/sh -c 'echo "hello $(cat /data)"', before being executed
```
vladaionescu, line 598: Is this comment still accurate?

earthfile2llb/converter.go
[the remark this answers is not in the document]
vladaionescu: > That might be tricky, since this function needs to call c.RunExpression and c.RunCommand (where c is *Converter).

You can always create a method of Converter that returns a closure. That way the function has access to the converter. I don't think we need to pass in the converter pointer itself.

(The old `processNonConstantBuildArgFunc` did exactly that)

earthfile2llb/interpreter.go
```diff
@@ -1069,12 +1206,16 @@ func (i *Interpreter) handleArg(ctx context.Context, cmd spec.Command) error {
...
+		}
 	}
 
-	if i.local && strings.HasPrefix(value, "$(") {
-		return i.errorf(cmd.SourceLocation, "ARG does not currently support shelling-out in combination with LOCALLY")
-	}
+	// TODO I suspect this now works, but this needs to be tested
+	//if i.local && containsShell(value) {
+	//	return i.errorf(cmd.SourceLocation, "ARG does not currently support shelling-out in combination with LOCALLY")
+	//}
```
vladaionescu, line 1218: Forgotten TODO?

vladaionescu (commented):
Looking good! Just a few small things left...

earthfile2llb/converter.go
alexcb: technically yes, but that that occurs in shell.go, so i'll remove this comment.

earthfile2llb/converter.go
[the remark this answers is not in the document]
alexcb: I restored `processNonConstantBuildArgFunc` and was able to delete the majority of this code -- now we'll use `processNonConstantBuildArgFunc` to maintain backwards compatibility (both here and in other parts of the converter, as we did before). When we have `--shell-out-anywhere` set, we'll process the shell expansion here all in a single-shot.

variables/collection.go
```diff
@@ -131,8 +132,9 @@ func (c *Collection) SortedOverridingVariables() []string {
 	return c.overriding().SortedAny()
 }
 
-// Expand expands variables within the given word.
-func (c *Collection) Expand(word string) string {
+// ExpandOld expands variables within the given word, it does not peform shelling-out.
+// it will eventually be removed when the ShellOutAnywhere feature is fully-adopted
+func (c *Collection) ExpandOld(word string) string {
```
vladaionescu, line 137: Nit: I noticed that `ExpandOld` is exactly the same as Expand, but with `shellOut` set to `nil`. Perhaps this could be removed altogether and just set nil from the caller?

vladaionescu (approved):
One (not very important) nit.

Looks great! :shipit:

variables/collection.go
alexcb: it's actually quite different: `dfShell.NewLex('\\')` vs `shell.NewLex('\\')` (the original buildkit code vs our forked version).

It's theoretically possible to write an interface for them, but I think that's overkill.

vladaionescu: :+1:

PR #1706, opened 2022-03-02
expand FROM example to show subdirs and remote targets

alexcb:

### DIFF
--- a/docs/earthfile/earthfile.md
+++ b/docs/earthfile/earthfile.md
@@ -48,7 +48,14 @@ Each recipe contains a series of commands, which are defined below. For an intro
 
 #### Description
 
-The `FROM` command initializes a new build environment and sets the base image for subsequent instructions. It works similarly to the classical [Dockerfile `FROM` instruction](https://docs.docker.com/engine/reference/builder/#from), but it has the added ability to use another target's image as the base image for the build. For example: `FROM +another-target`.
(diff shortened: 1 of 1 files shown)


docs/earthfile/earthfile.md
```diff
@@ -48,7 +48,7 @@ Each recipe contains a series of commands, which are defined below. For an intro
 
 #### Description
 
-The `FROM` command initializes a new build environment and sets the base image for subsequent instructions. It works similarly to the classical [Dockerfile `FROM` instruction](https://docs.docker.com/engine/reference/builder/#from), but it has the added ability to use another target's image as the base image for the build. For example: `FROM +another-target`.
+The `FROM` command initializes a new build environment and sets the base image for subsequent instructions. It works similarly to the classical [Dockerfile `FROM` instruction](https://docs.docker.com/engine/reference/builder/#from), but it has the added ability to use another target's image as the base image for the build. For example: `FROM +another-target`, `FROM ./subdirectory+my-target`, or even `FROM github.com/some/other-repo+other-target`.
```
brandonSc, line 51:
```suggestion
The `FROM` command initializes a new build environment and sets the base image for subsequent instructions. It works similarly to the classical [Dockerfile `FROM` instruction](https://docs.docker.com/engine/reference/builder/#from), but it has the added ability to use another target's image as the base image for the build. 

Earthly supports loading from in the following locations
* Local Targets - `FROM +my-target`
* Relative Targets - `FROM +./subdirectory+my-target`
* Remote Targets from a Public or [Private](https://docs.earthly.dev/docs/guides/auth) Git repository - `FROM github.com/example/project+my-target`

brandonSc: I'm not sure if I wrote that the best way, but maybe explicitly defining them with names like that, and including a link to understand how to do it from a private repo

brandonSc (approved):
🥳

PR #1707, opened 2022-03-02
docs: clarify how auto-auth mode is used

alexcb:
This makes our docs easier to quickly read for users who are having
troubles using ssh-agent and auto-mode.

The goal is to have the majority of our users use ssh-based auth via
the zero-config auto mode.

### DIFF
--- a/docs/guides/auth.md
+++ b/docs/guides/auth.md
@@ -20,11 +20,25 @@ There are two possible ways to pass Git authentication to Earthly builds:
 * Via SSH agent socket (for SSH-based authentication)
 * Via username-password (usually for HTTPS Git URLs)
 
+#### Auto authentication
+
+Earthly defaults to an `auto` authentication mode, where ssh-based authentication is automatically attempted, and falls back to https-based cloning.
+
+{% hint style='info' %}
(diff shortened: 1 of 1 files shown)


docs/guides/auth.md
```diff
@@ -20,11 +20,25 @@ There are two possible ways to pass Git authentication to Earthly builds:
 * Via SSH agent socket (for SSH-based authentication)
 * Via username-password (usually for HTTPS Git URLs)
 
+#### Auto authentication
+
+Earthly defaults to an `auto` authentication mode, where ssh-based authentication is automatically attempted, and falls back to https-based cloning.
+
+{% hint style='info' %}
+If you are having troubles accessing a private repository and want to use ssh-based authentication, first make sure `ssh-agent` is running and the `SSH_AUTH_SOCK`
```
vladaionescu, line 28:
```suggestion
If you are having trouble accessing a private repository and want to use ssh-based authentication, first make sure `ssh-agent` is running and the `SSH_AUTH_SOCK`
```

vladaionescu (approved):

brandonSc:
Might also need to run `earthly config git "{github.com: {auth: ssh, user: git}}"` - or at least I think I did

PR #1717, opened 2022-03-03
Don't try SSH keys for log sharing

brandonSc:
A check for `WhoAmI()` triggers SSH key checking. This change only tries log upload if the user is actively logged in.


alexcb (approved):
looks good.

Is there a way we can add an integration test for this?

cmd/earthly/main.go
```diff
@@ -2791,21 +2789,17 @@ func (app *earthlyApp) actionBuildImp(c *cli.Context, flagArgs, nonFlagArgs []st
...
+				var id string
+				id, uploadErr = cc.UploadLog(logPath)
 				app.console.Printf("Share your build log with this link: %s\n", id)
 			}()
 		} else {
 			// If you are not logged in, then advertise the service, since they probably turned it on to try it.
 			defer func() { // Defer this to keep log upload code together
-				if errors.Is(whoAmIErr, cloud.ErrUnauthorized) {
-					app.console.Printf("Share your logs with an Earthly account (experimental)! Register for one at https://ci.earthly.dev.")
+				if err != nil && !errors.Is(uploadErr, cloud.ErrUnauthorized) {
```
alexcb, line 2799: is `err` intentional?
 
```suggestion
				if uploadErr != nil && !errors.Is(uploadErr, cloud.ErrUnauthorized) {
```

brandonSc: Good eye. I refactored it last minute and make a mistake

brandonSc:
> looks good.
> 
> Is there a way we can add an integration test for this?

Do we have any integration tests in earthly core that call the server (or a mock server?)

alexcb:
> > looks good.
> > Is there a way we can add an integration test for this?
> 
> Do we have any integration tests in earthly core that call the server (or a mock server?)

not yet :)

Here's a very hacky idea:

```Earthfile
server:
    HOST api.earthly.dev 127.0.0.1
    RUN \
    echo -e "#!/bin/sh\\ntouch /i-was-run" > /run-this && chmod +x /run-this && \
    nc -l -p 443 -e /run-this & \
    export webserverpid="$!" && \
    earthly +this-will-fail || true && \
    if [ -f /i-was-run ]; then echo "earthly should not have attempted a connection" && exit 1; fi
    kill -9 $webserverpid || true
```

brandonSc:
@alexcb I'm initially having some trouble getting a test like that working, 

But it got me thinking that we probably want a larger test case in earthly with a proper mock server - e.g. we don't have a test for the success case when logs _are_ uploaded, we may want more test cases for other earthly commands that hit the server.

Also thinking ... we do have those integration tests that use earthly core to hit the endpoints in the cloud repo, but I suppose those are only useful for versions of earthly that have at least been merged to main, not really suited to custom work on a branch of earthly core

Issue #1724, opened 2022-03-04
Duplicate run when using `IF`


vladaionescu:
```Earthfile
VERSION 0.6

FROM busybox:latest

test-img:
    RUN echo hello

base-test:
    IF [ "true" = "true" ]
        FROM earthly/dind:alpine
        WITH DOCKER --load test-img=+test-img
            RUN true
        END
    END
```

`earthly -P --no-cache  +base-test` will output:

```
      busybox:latest | --> Load metadata linux/amd64
               +base | --> FROM busybox:latest
               +base | [██████████] 100% resolve docker.io/library/busybox:latest@sha256:afcc7f1ac1b49db317a7196c902e61c6c3c4607d63599ee1a82d702d249a0ccb
          +base-test | --> IF [ "true" = "true" ]
 earthly/dind:alpine | --> Load metadata linux/amd64
           +test-img | --> RUN echo hello
           +test-img | hello                              <-------- first run
           +test-img | hello                              <-------- duplicate run
          +base-test | --> FROM earthly/dind:alpine
          +base-test | [██████████] 100% resolve docker.io/earthly/dind:alpine@sha256:34080f82b7ab55b1808aa6eddf8bca4cba8db63b9baa86fe4246ffd663a98576
          +base-test | --> WITH DOCKER (install deps)
          +base-test | /usr/local/bin/dockerd
             context | transferred 1 file(s) for context /tmp/earthly-docker-load3034316646 (784 kB, 1 file/dir stats)
          +base-test | --> WITH DOCKER RUN --privileged true
          +base-test | Loading images...
          +base-test | Loaded image: test-img:latest
          +base-test | ...done
              output | --> exporting outputs
```

vladaionescu:
I've noticed that moving the `FROM earthly/dind:alpine` just outside the `IF` the problem goes away

Issue #1268, opened 2021-10-07
Very slow image import/export during WITH DOCKER


konradmalik:
Hey,
Recently I started to use Earthly for some of my projects.
So far, the majority of stuff works great, but the biggest drawback seems to be performance.
Everything runs a little bit slower, which is expected and completely fine. However, the worst impact that we see is for `docker-compose` based integration tests.

Specifically, in the example below:
```
             context | transferred 1 file(s) for context /tmp/earthly-docker-load290768814 (72 MB, 1 file/dir stats)
             context | transferred 1 file(s) for context /tmp/earthly-docker-load850151696 (564 MB, 1 file/dir stats)
             context | transferred 1 file(s) for context /tmp/earthly-docker-load498375883 (623 MB, 1 file/dir stats)
             context | transferred 1 file(s) for context /tmp/earthly-docker-load533876533 (1.3 GB, 1 file/dir stats)
               +test | --> WITH DOCKER RUN --privileged docker run --network=pider-tests --entrypoint="" scraper ./wait-for-it.sh azurite:10000 -- poetry run python -m pytest --numprocesses $PYTEST_PARALLELISM --max-worker-restart 0 --verbose --cov=scraper tests
               +test | Loading images...
               +test | Loaded image: scraper:latest
               +test | Loaded image: mcr.microsoft.com/azure-storage/azurite:3.14.2
               +test | Loaded image: registry.gitlab.com/pider/detectron:0.0.6
               +test | Loaded image: registry.gitlab.com/pider/tohu:0.0.6
               +test | ...done
```
I have one image that is built via earthly and three others that need to be pulled and used in the docker-compose file I provided (dependencies for the integration test).
The problem is, that the import/export of all those images takes much longer (2, 3 times. Depending on the machine it may be 5-10 mins) than the running of the tests itself.
"Standard" approach with plain docker-compose is much, much faster.

This is the command I run for the reference:
```
    WITH DOCKER \
        --compose tests.yaml \
        --load scraper=+tests
        RUN docker run --network=pider-tests --entrypoint="" scraper \
            ./wait-for-it.sh azurite:10000 -- poetry run python -m \
            pytest --numprocesses $PYTEST_PARALLELISM --max-worker-restart 0 --verbose tests/integration
    END
```

Problems I noticed:
- seems like all the images are first loaded into the earthly buildkit context and then transferred (loaded) into the internal dind instance
- maybe the above could be optimized somehow, perhaps done in a one-step instead of two?
[...]

PR #1725, opened 2022-03-05
Parallel conversion for `WITH DOCKER --load`

vladaionescu:
Re https://github.com/earthly/earthly/issues/1268. This addresses point 2 in this comment: https://github.com/earthly/earthly/issues/1268#issuecomment-938261849.

This PR takes the multiple `--load`s and `--pull`s in a `WITH DOCKER` and builds the necessary image tars in parallel. This includes converting the referenced targets in parallel, but also outputting the tars in parallel.

To enable this feature use `VERSION --parallel-load 0.6`.

Separately, this also fixes https://github.com/earthly/earthly/issues/1724. No feature flag is needed to benefit from this.


alexcb (approved):
big changes! good to see `buildMainMulti` wasn't deleted :sweat_smile:

PR #1732, opened 2022-03-08
Introduce new `DOCKERHUB_MIRROR_INSECURE` variable for testing

rrjjvv:
When set to `true` (`false` by default), a registry stanza is added that specifies that mirror `DOCKERHUB_MIRROR` is insecure.

The intended use-case is to support running a local caching/pass-through registry for use specifically for running tests and avoiding rate-limits (in the case of DockerHub) and/or reducing costly high-latency traffic (such as a corporate registry over VPN).

This setting is applied to the 'outer' buildkitd instance as well as the 'inner' instance.  The 'inner' instance now respects `DOCKERHUB_MIRROR` as well.


Earthfile
```diff
@@ -366,11 +367,20 @@ earthly-integration-test-base:
...
+        ELSE
+            ARG _MIRROR_CONFIG=""
+        END
         ENV GLOBAL_CONFIG="{disable_analytics: true, local_registry_host: 'tcp://127.0.0.1:8371', conversion_parallelism: 5, buildkit_additional_config: '[registry.\"docker.io\"]
 
-                           mirrors = [\"$DOCKERHUB_MIRROR\"]'}"
+                           mirrors = [\"$DOCKERHUB_MIRROR\"]
+                           $_MIRROR_CONFIG'}"
         ENV EARTHLY_ADDITIONAL_BUILDKIT_CONFIG="[registry.\"docker.io\"]
-                    mirrors = [\"registry-1.docker.io.mirror.corp.earthly.dev\"]"
```
rrjjvv, line 373: I was hesitant to change this, given that `DOCKERHUB_MIRROR` was introduced just a few lines up (suggestion this was left untouched on purpose).  But I also couldn't see any obvious reason why that specific value was required (unless there's some test, possibly of the release process itself, that requires it).

Issue #1723, opened 2022-03-04
Request: allow specifying cache size as percentage (vs. size-based `cache_size_mb`)


rrjjvv:
I'm overloading this request with some other feedback/questions which might warrant their own issues, but figured I'd start simple.  But the fundamental request is as stated: it would be great to have something like `cache_size_pct` as a companion to `cache_size_mb`.  Or overload `cache_size_mb` to accept a percentage, if that was preferred.

My own machines, as well as build machines I manage, are of different sizes.  I would prefer to have one static config file with a static percentage (which adapts to the host at startup).  Dynamically generated config files (with calculated sizes) aren't impossible, but isn't ideal.

Background
---
I had a build target that was taking a very long time to execute on back-to-back executions, when theoretically everything should have been cached.  It seemed to also be the same sub-target that was building from scratch, and for the life of me, I could not figure out what was causing a cache bust.  It then occurred to me it might not be a cache bust per-se, but a cache expiration/eviction.  And it was.

I was always aware of the `cache_size_mb` setting, but the language says
> If you would like to limit the cache size more aggressively, you can specify a different limit 

To me, that implied that the default value was often too large for people, so it hadn't originally occurred to me that the opposite could occur (where the default is too small).  After some experimentation and strolling through source code, I found that (on linux) the default is [10% of the filesystem containing the cache](https://github.com/moby/buildkit/blob/master/cmd/buildkitd/config/gcpolicy_unix.go#L16).  If my laptop had everything on the root filesystem (which might be the common case with end-users), I'd have a healthy 50GB cache.  But `/var` (thus docker) is on its own filesystem, leaving me with a puny 4GB cache.  The smaller targets all fit in cache just fine, but the larger aggregate targets were exceeding that, and very quickly.  After completion, the target uses 8GB of cache, but grows up to 10GB.  After specifying a larger size, my problematic target went from about 30 minutes down to 6 minutes.

Specifying a custom setting results in [your template](https://github.com/earthly/earthly/blob/main/buildkitd/entrypoint.sh#L136-L138) getting rendered, which results in a policy like
```
GC Policy rule#0:
        All:            false
        Filters:        type==source.local type==source.git.checkout
        Keep Bytes:     1.5GB
GC Policy rule#1:
        All:            true
        Keep Bytes:     15GB
```
[...]

PR #1733, opened 2022-03-08
Introduce `DOCKERHUB_MIRROR_INSECURE` environment variable for testing

rrjjvv:
When set to `true` (`false` by default), a registry stanza is added that specifies that mirror `DOCKERHUB_MIRROR` is insecure.

The intended use-case is to support running a local caching/pass-through registry for use specifically for running tests and avoiding rate-limits (in the case of DockerHub) and/or reducing costly high-latency traffic (such as a corporate registry over VPN).

This setting is applied to the 'outer' buildkitd instance as well as the 'inner' instance.  The 'inner' instance now respects `DOCKERHUB_MIRROR` as well.


rrjjvv:
Even though that failure is 'obviously' related to my changes, let me know if it actually is.  Based on what I can see from those logs, I can't make sense out of why there's a failure to auth with hub (when my change is, in spirit, helping to do the exact opposite).

Thee only clue I have to offer from where I sit is that this would appear to relate to my third item from https://github.com/earthly/earthly/issues/1723#issuecomment-1063675365.  The line above the failure has `http.request.useragent="docker/20.10.11+azure-3 <snip>`, which likely means it's the the docker client in the dind image (or an auto-installed client).  If it was buildkitd (which is what my PR impacts), it would be like `http.request.useragent="containerd/1.6.0-rc.1+unknown"` (taken from the logs of my local registry).

Let me know what I missed/overlooked!

alexcb:
I was able to test this locally, and am still able to run our regular testing workflow.

Maybe we should update https://github.com/earthly/earthly/blob/main/CONTRIBUTING.md#running-tests to include a new section `## running tests with local caching/pass-through registry` where we provide an example of starting up the pull-through-cache as well as a `.env` file to set?

rrjjvv:
Yes, I had also planned on submitting/proposing an addendum to the docs if you guys were amenable to this change (as well as the one that I have not yet pushed).  When running the test suite (esp. for the PR you folks approved  this morning), I have to manually merge this branch, as well as my yet-to-be-pushed branch *together* to not get blocked (and then manually back them out prior to pushing any new code).  To the best of my knowledge though, this PR should be able to work in isolation from the other code, assuming you aren't subject to blocks.

So in my local setup, I have a `docker-compose.yml` that reads:
```
version: "2"
services:
  registry:
    image: registry:2
    environment:
      - REGISTRY_PROXY_REMOTEURL=https://<my corporate mirror>
    restart: always
    ports:
      - "5005:5000"
    volumes:
      - registry_data:/var/lib/registry
volumes:
  registry_data:
```
where `<my corporate mirror>` is a full-fledged caching mirror for Dockerhub (and others), but is available only over our VPN, which isn't super fast.

So that local, insecure registry is bound to port 5005 (for me).  There might be a more elegant way to wire up the networking so the explicit IPs aren't needed, but to get things working, I just used my primary interface's IP to ensure d-in-d, b-in-b, e-in-e, etc. could all reach it.  In my case,
```
$ ip addr show wlp1s0 | grep 'inet ' | awk '{print $2}' | cut -d/ -f1
192.168.1.32
```
With that, I reconfigured my docker daemon:
```
$ sudo cat /etc/docker/daemon.json | jq '."registry-mirrors"'
[
  "http://192.168.1.32:5005"
]
```
Added that to existing `DOCKERHUB_MIRROR`, as well as my proposed `DOCKERHUB_MIRROR_INSECURE` variable:
```
# (earthly project root directory)
$ cat .env
DOCKERHUB_AUTH=false
DOCKERHUB_MIRROR=192.168.1.32:5005
DOCKERHUB_MIRROR_INSECURE=true
```

Everything above is for this PR specifically.  There is one additional local change I make to trigger my unpublished code (which should *not* be needed for this PR, but sharing so you can see the whole picture):
```
$ env | grep -i mirror
EARTHLY_DOCKERD_REGISTRY_MIRROR=http://192.168.1.32:5005
```
and that environment variable is used to populate `registry-mirrors` in the [internal `dockerd` configuration](https://github.com/earthly/earthly/blob/main/buildkitd/dockerd-wrapper.sh#L74-L86), analogous to the host's config.

So that's far from a one-liner, but it's a pretty straight-forward recipe.  The IP is really the main eye-sore (and could be problematic for users without static or dynamic-but-stable addresses), but maybe you have some network tricks to remove that need.

rrjjvv:
And to be clear,
> I was able to test this locally, and am still able to run our regular testing workflow.

Are you saying the changes themselves are good, but the build itself is broken/flaky for unrelated reasons?  Or the changes are good, but something specific to the workflow isn't taking kindly to it (and either my change or the build have to be adapted to each other)?

If you're giving me an (unofficial and tentative) thumbs up, I'd be happy to take a run at updating the docs with what I wrote above.  But if there's something that needs fixing (from me), I'm not clear what that is...

alexcb:
Thank you very much for that details explanation!

The PR on it's own seems fine -- I suspect the two failures were unrelated to your changes (especially given that this code works for me locally). I just re-triggered the builds to see if the failures were related to flaky-network issues.

I'm happy to merge this if we get a :+1: from @dchw

alexcb (approved):

rrjjvv:
I just created #1745 (which is the `EARTHLY_DOCKERD_REGISTRY_MIRROR` part of my instructions from above).

dchw (approved):

alexcb:
@rrjjvv I took a deeper look at this today.

In order for me to get it working against a self-hosted mirror I had to do a minor change under https://github.com/earthly/earthly/pull/1765 

I also built upon your instructions on getting it working, and have a very rough draft of my debugging/setup steps, which I will work into a developer's guide for running our tests against a self hosted mirror.

Below is a super rough draft of my guide -- it's mostly notes for myself which I need to iterate on before looking for feedback.

------

These notes assume you are running the registry on a different server (192.168.0.80 in the examples), first we will setup the registry/mirror/cache on that server

## Connect to the other-server

    ssh 192.168.0.80

and create a configu:

```
cat >> ~/.docker-registry-config.yml <<EOF
version: 0.1
log:
  fields:
    service: registry
storage:
  cache:
    blobdescriptor: inmemory
  filesystem:
    rootdirectory: /var/lib/registry
http:
  addr: :5000
  headers:
    X-Content-Type-Options: [nosniff]
health:
  storagedriver:
    enabled: true
    interval: 10s
    threshold: 3
proxy:
  remoteurl: https://registry-1.docker.io
  username: [username]
  password: [password]
EOF
```

Make sure you update the `[username]` and `[password]` fields.

Next start up the docker registry:

    docker run --rm --network host --name docker-registry -v $HOME/.docker-registry-config.yml:/root/config.yml registry.hub.docker.com/library/registry:2 registry serve /root/config.yml

## Configuring your workstation to use the new registry

Before we get started, we will perform an optional step to block access to docker on our workstation.
(In practice if you are running the registry on the same machine as your workstation, this would not work, as the registry needs access to docker.)

To block access, edit `/etc/hosts` and add:

    0.0.0.0 index.docker.io auth.docker.io registry-1.docker.io dseasb33srnrn.cloudfront.net production.cloudflare.docker.com

    ( these host entries are based on https://support.sonatype.com/hc/en-us/articles/115015442847-Whitelisting-Docker-Hub-Hosts-for-Firewalls-and-HTTP-Proxy-Servers )

When you perform a pull, such as:

    docker pull alpine:3.15

You should get:

    Error response from daemon: Get "https://registry-1.docker.io/v2/": dial tcp 0.0.0.0:443: connect: connection refused

Next, configure your instance of docker to use the registry we setup on the `other-server`, edit /etc/docker/daemon.json, and add:

    {
      "registry-mirrors" : ["http://192.168.0.80:5000"]
    }

    (substitute other-host with the hostname, or ip where the registry is hosted)

Then restart docker:

    sudo service docker restart

Then try another pull:

    docker pull alpine:3.15

which should now work:

    3.15: Pulling from library/alpine
    Digest: sha256:d6d0a0eb4d40ef96f2310ead734848b9c819bb97c9d846385c4aca1767186cd4
    Status: Downloaded newer image for alpine:3.15
    docker.io/library/alpine:3.15

If you connect back to your host running the docker-registry container, under the logs you should see a request:

    192.168.0.126 - - [22/Mar/2022:19:10:39 +0000] "HEAD /v2/library/alpine/manifests/3.15 HTTP/1.1" 200 1638 "" "docker/20.10.12 go/go1.16.12 git-commit/459d0df kernel/5.13.0-35-generic os/linux arch/amd64 UpstreamClient(Docker-Client/20.10.12 \\(linux\\))"

## Validating earthly still works:

First, lets try to compile earthly, by running:

    env -i HOME=$HOST PATH=$PATH earthly-v0.6.11 --env-file=<(echo "") --config=$HOME/.earthly/config.yml +for-linux

    note: `env -i HOME=$HOST PATH=$PATH`, and `--env-file=<(echo "")` is used to ensure no other env variables are passed

    TODO, FIXME: when running with `env -i ...` the `--config` default value fails and it doesn't correctly parse/load the config file (which contains buildkit_additional_config)

This fails with:

    golang:1.17-alpine3.14 | --> Load metadata linux/amd64
    golang:1.17-alpine3.14 | WARN: (Load metadata linux/amd64) failed to authorize: failed to fetch anonymous token: Get "https://auth.docker.io/token?scope=repository%3Alibrary%2Fgolang%3Apull&service=registry.docker.io": dial tcp 0.0.0.0:443: connect: connection refused
    Shareable link: https://ci.earthly.dev/logs?logId=21420bec-554a-4223-9d2a-3d2e7cf19ca1
    Error: build target: build main: failed to solve: failed to fetch anonymous token: Get "https://auth.docker.io/token?scope=repository%3Alibrary%2Fgolang%3Apull&service=registry.docker.io": dial tcp 0.0.0.0:443: connect: connection refused

To fix this, we must add to the buildkit.toml config that is contained within the earthly-buildkit container, to 
do this, we must edit ~/.earthly/config.yml to include:

```
  global:
    buildkit_additional_config: |
      [registry."docker.io"]
        mirrors = ["192.168.0.80:5000"]
      [registry."192.168.0.80:5000"]
        http = true
        insecure = true
```

note: the docs under https://github.com/moby/buildkit/blob/master/docs/buildkitd.toml.md make it look like you can simply define a single version as:

```
  [registry."docker.io"]
    mirrors = ["192.168.0.80:5000"]
    http = true
    insecure = true
```

but that fails with:

    failed to do request: Head "https://192.168.0.80:5000/v2/library/golang/manifests/1.17-alpine3.14?ns=docker.io": http: server gave HTTP response to HTTPS client

## Running the earthly tests with this mirror

First we'll re-bootstrap with the newly build earthly binary:

    env -i HOME=$HOST PATH=$PATH ./build/linux/amd64/earthly --env-file=<(echo "") --config=$HOME/.earthly/config.yml bootstrap

Next, we'll (optionally) overwrite the /etc/hosts *inside* the container, by running:

    docker exec -ti earthly-buildkitd /bin/sh -c 'echo "0.0.0.0 index.docker.io auth.docker.io registry-1.docker.io dseasb33srnrn.cloudfront.net production.cloudflare.docker.com" > /etc/hosts'

Then we can validate we have blocked docker by running:

    docker exec -ti earthly-buildkitd wget registry-1.docker.io

which should return an error:

    Connecting to registry-1.docker.io (0.0.0.0:80)
    wget: can't connect to remote host (0.0.0.0): Connection refused

Then let's proceed with the test:

    env -i HOME=$HOST PATH=$PATH ./build/linux/amd64/earthly --env-file=<(echo "") --config=$HOME/.earthly/config.yml -P ./tests+all --DOCKERHUB_AUTH=false --DOCKERHUB_MIRROR=192.168.0.80:5000 --DOCKERHUB_MIRROR_INSECURE=true

The above should pass once my https://github.com/earthly/earthly/pull/1765 is improved and merged

rrjjvv:
> In order for me to get it working against a self-hosted mirror I had to do a minor change under #1765

For a split second I thought I understood the problem, but I'm not so sure.  The best I can gather is that there are actually two intertwined issues?  The first being that `GLOBAL_CONFIG` is a syntatically correct JSON *file* (thus parseable as valid YAML), but could not be translated into a valid YAML doc from a newline-containing string by earthly itself?  And the second issue that despite being a valid YAML file,  that the embedded TOML (for buildkit) was malformed?  If I'm on the right track, I can speculate on why it _appeared_ to work.  But I'll be disappointed if (for example) my work for #1745 wasn't actually needed and was just a workaround for this flaw.  Either way, no need to elaborate for my sake, but do let me know if it's something I should have caught.

> first we will setup the registry/mirror/cache on that server

Clearly I had made to effort to optimize my original config (as this was all a means to an end for me).  I had noticed that my registry would exhibit heavy CPU/IO at (unexpected) times.  I'll adapt the relevant pieces of your config to my setup and report back anything of note.

> To block access, edit `/etc/hosts` and add:

I had tried accomplishing the same thing with `iptables` rules, but to no avail.  I hadn't thought of doing something that simple (despite having tried using `HOST` in an effort to test this in an automated way).

Once you think your docs are 'complete', I'd be happy to run your directions from the ground up (I have another machine at my disposal which won't impact my daily work)!

alexcb:
> my work for https://github.com/earthly/earthly/pull/1745 wasn't actually needed

It most definitely was needed, I don't see how the insecure settings could have been propagated otherwise -- it helped a lot having all those values propigated throughout all the tests.

I think the yaml newline chomping issue prevented some earthly-in-earthly cases to simply ignore the mirror. Once I added the fake `/etc/hosts` entries, it helped track it down. (I too tried some firewall rules at first, but quickly failed and found [the hosts idea](https://stackoverflow.com/a/59652133) ).

> Once you think your docs are 'complete', I'd be happy to run your directions from the ground up (I have another machine at my disposal which won't impact my daily work)!

:raised_hands:

alexcb:
@rrjjvv I created a new section on our contributing guide: https://github.com/earthly/earthly/blob/main/CONTRIBUTING.md#running-tests-with-an-insecure-pull-through-cache

It provides a link to a newly-created example under: https://docs.earthly.dev/ci-integration/pull-through-cache#insecure-docker-hub-cache-example

If you have a change to have a quick read over, I would appreciate any feedback.

rrjjvv:
I'll have to run through the actual steps later (possibly tomorrow) and I'll let you know how plug-n-play they were.

In terms of the raw content, I think they're fine as-is, if I have a feel for the target audience (and their technical expertise) of each page.  But maybe I don't.  So I'll do my best to unlearn everything from the past couple weeks and jot down my thought process.  You can decide if they're in line with a "typical" user or not.

From the contribution page:
---
> one must also configure the embedded version of Earthly to use the cache via build-args

_Me: I assume I don't **have** to use build-args.  Above it mentions using an `.env` file for similar settings, so I'll try that too._

[Obviously this is true.]

From the public docs:
---
> Note that you'll need to replace [username] and [password] with your dockerhub credentials.

_Me: Obviously I don't **have** to provide credentials; if I had an account, I wouldn't be on this page!_

[Obviously true.  Whether you land on that page because you need to avoid limits/costs, improve performance, or both dictates whether you have creds (or not), whether you provide them (or not), and whether it's for dockerhub (or other)]

> Next, start the registry container with:

_Me: Sigh... I guess I'll create a `docker-compose.yml` myself.  Most sites will at least give a download link if they don't include it in the page!_

[The minute I find myself with a command that takes more than simple arguments, but maybe that's just me.]

_Me: Even though they give me a copy/paste command, I see there's no sign of a volume for the cached data.  I better add one since performance will be terrible if it wasn't declared a `VOLUME`, and even if it was, I'll still lose it via their `--rm`_

[Out of all my thoughts, this is the closest to an actual suggestion; your target audience might be expected to catch that, but would be trivial to add.]

> To configure docker to use this cache as a mirror, edit the /etc/docker/daemon.json file, and add:

_Me: Hmmm. on my workstation?  On the registry server?  Or both?_

[For me, the "workstation" and "registry server" are one and the same.  I did have to stop and think that one through.]

> If you want to verify the cache is working, you can block access to dockerhub on your workstation by adding

_Me: Makes sense, but `dseasb33srnrn.cloudfront.net` sounds transient/dynamic; does this truly cover everything?_

[Honest question.  I believe in your notes you had had a link, but I believe that was from Sonatype, not docker themselves.]

---
Again, I think what you have is awesome like it is.  I'm confident that if this was in place two weeks ago, I'd be up and running in no time at all.  But I don't know if I'm a typical user.

Side note: from the contribution page (existing content, not part of your changes):
> You may need to adjust the docker login command in the earthly-integration-test-base: target by removing the Earthly repository and adjusting for your login credentials provider.

This isn't true anymore :-)

PR #1734, opened 2022-03-08
Ability to set cache size limit as percentage of space (#1723)

rrjjvv:
Fixes https://github.com/earthly/earthly/issues/1723


buildkitd/entrypoint.sh
```diff
@@ -136,6 +141,25 @@ CACHE_SETTINGS=
 if [ "$CACHE_SIZE_MB" -gt "0" ]; then
     CACHE_SETTINGS="$(envsubst </etc/buildkitd.cache.template)"
 fi
+
+if [ "$CACHE_SIZE_PCT" -gt "0" ]; then
+    # %b -> "Total data blocks"
+    # %S -> "Fundamental block size"
+    # -f $EARTHLY_TMP_DIR -> filesystem where directory resides, usually a volume in docker's root directory
+    CALCULATED_CACHE_MB="$(stat -c "%b * %S * ${CACHE_SIZE_PCT} / 100 / 1024 / 1024" -f "$EARTHLY_TMP_DIR" | bc)"
+    echo "${CACHE_SIZE_PCT}% of target filesystem: ${CALCULATED_CACHE_MB}"
```
alexcb, line 150:
```suggestion
    echo "cache size set to ${CALCULATED_CACHE_MB} MB (${CACHE_SIZE_PCT}% of filesystem)"
```

docs/earthly-config/earthly-config.md
```diff
@@ -58,6 +58,11 @@ earthly config global.cache_size_mb 20000
 
 Specifies the total size of the BuildKit cache, in MB. The BuildKit daemon uses this setting to configure automatic garbage collection of old cache. A value of 0 causes the size to be adaptive depending on how much space is available on your system. The default is 0.
 
+### cache_size_pct
```
alexcb, line 61: Could you open up a seperate PR against the `next` branch for the docs update?

Then when we do a release, we'll then merge `next` into `main`.

This is to prevent our docs from getting ahead of our currently released version of earthly.

alexcb (changes requested):
Looks great, just two very minor suggestions.

docs/earthly-config/earthly-config.md
rrjjvv: Makes perfect sense.  Will do that in a little bit.

rrjjvv: Done in #1737

cmd/earthly/main.go
```diff
@@ -512,11 +512,18 @@ func newEarthlyApp(ctx context.Context, console conslogging.ConsoleLogger) *eart
...
 		&cli.IntFlag{
 			Name:        "buildkit-cache-size-mb",
-			Value:       10000,
+			Value:       0,
 			EnvVars:     []string{"EARTHLY_BUILDKIT_CACHE_SIZE_MB"},
 			Usage:       "The total size of the buildkit cache, in MB",
 			Destination: &app.buildkitdSettings.CacheSizeMb,
 		},
+		&cli.IntFlag{
+			Name:        "buildkit-cache-size-pct",
```
alexcb, line 521: Sorry, just another issue I noticed:

I checked out your branch and tried running

```
./earthly +for-linux
./build/linux/amd64/earthly --buildkit-cache-size-pct=1 +for-linux
```

It looks like this flag doesn't get passed through:

```
$ docker exec earthly-buildkitd env | grep CACHE_SIZE
CACHE_SIZE_MB=0
CACHE_SIZE_PCT=0    <----- expected this to be 1
```

Since we have deprecated the `--buildkit-cache-size-mb` flag, I think we can skip defining this cli flag here entirely.

rrjjvv: Ugh, my bad for not explicitly testing that the CLI flag was actually *used* (and not only that it showed up in `-h`, etc.).  And agree it shouldn't be a CLI option (and is only present due to my newbie copy/paste/adapt as I figure out how things work).  Hopefully it's as simple as just removing that chunk (and any resulting cleanup).

rrjjvv: Should be addressed.  I left  `--buildkit-cache-size-mb` alone, since it didn't seem like you were asking (explicitly or implicitly) for it to be removed.

alexcb: Something still seems off with how `--buildkit-cache-size-mb` is handled.

under v0.6.10, it works:

```
earthly-v0.6.10 --buildkit-cache-size-mb=12345 +for-linux
docker exec earthly-buildkitd env | grep CACHE_SIZE_MB
```

and it shows that v0.6.10 does pass through this env:

```
CACHE_SIZE_MB=12345
```

however with the latest changes

```
git checkout 75119ae
./earthly +for-linux
./build/linux/amd64/earthly --buildkit-cache-size-mb=2345 +for-linux
docker exec earthly-buildkitd env | grep CACHE_SIZE_MB
```

I get

```
CACHE_SIZE_MB=0
```

rrjjvv: Do you have a value set in your config file as well?  Though what you're describing seems to be the opposite of my comment, from my commit message,
> This does slightly change the semantics around the deprecated flag, but hopefully for the better.  Other usages of this "flag trumps config" behavior have been removed over time as well, further supporting that the current behavior isn't intentional, just overlooked/forgotten.

But it might also be related to me setting the default to 0 (where it was 10GB) in that same commit.  I thought I was doing "the right thing" by changing the default value to what the docs say, and shifting the code around to have all the setting in the same place.  I'd be happy to just revert that commit entirely, which should remove any doubt as to how that value is used (and does not change how my desired code works at all).  Or if you think it was the right thing to do, but done wrongly, I can dive into it.  I want to respectful of your time/effort (as this was intended to be just a 'cleanup' change).

(And I actually suspect that if I was to revert that commit, and if you were to _not_ have that set in your config, if you'd still find your CLI value not being used... that you'd find it always set at 10GB instead of 0.)

alexcb: This was without any `~/.earthly/config.yml` file.

After talking it over with the rest of the earthly-team, let's just remove support for the `--buildkit-cache-size-mb` flag

i.e.

delete this:
```
&cli.IntFlag{
			Name:        "buildkit-cache-size-mb",
			...
}
```

since we've had the deprecation warning in for over a year now.

rrjjvv: Will do.  Sorry for the ruckus.  (Or maybe "you're welcome".)

alexcb: i'll gladly take the opportunity to full remove that deprecated code :)

rrjjvv: Glad I'm not the only that enjoys deleting code more than writing it!

I want to do one more full test before pushing, but in the meantime, one of the earlier proposed changes looks funny in a larger context:
```
cache size set to 41550 MB (85% of filesystem)
CACHE_SIZE_MB=30000
```
That first line originally said something to the effect of "the calculated value would be X".  If you have a proposed message that removes this contradiction let me know (otherwise I'll put something in).

alexcb: maybe something like this?

```
        if [ "$CALCULATED_CACHE_MB" -lt "$CACHE_SIZE_MB" ]; then
            CACHE_SIZE_MB="$CALCULATED_CACHE_MB"
            echo "clamping cache size to $CALCULATED_CACHE_MB (${CACHE_SIZE_PCT}% of filesystem)"
        else
            echo "clamping cache size to fixed size of $CACHE_SIZE_MB"
        fi
```

I also noticed that `CACHE_SETTINGS="$(envsubst </etc/buildkitd.cache.template)"` is copy/pasted in multiple locations, can you move that to a single line after the various CACHE sizes get set.

rrjjvv: I'm currently planning on removing that first line entirely and merging it into the "lesser of" message: `Lesser of specified size (30000 MB) and percentage (41550 MB (85% of filesystem)) will be used for cache size`

rrjjvv: oh, sorry, I hadn't F5'ed the page and didn't see your response prior to my comment.  I wasn't vetoing your response; I just hadn't seen it.  I also didn't like the codepath that resulted in `envsubst` being called twice.  That ultimately stems from abusing `CACHE_SIZE_MB` as both an input and the output (and mutating it).  But I was aiming to be unobtrusive as possible for my first PR.

If you're game for another round of edits, I can implement your proposed change as well as cleaning up the doubly-hit `envsubst`.  It'll probably make for slightly longer code, but remove most (maybe all) of the nested if/else blocks.

rrjjvv: Okay, my upcoming push (including both the flag removal and the logic cleanup) will result this:
Percent only:
```
$ grep cache_size ~/.earthly/config.yml
    cache_size_pct: 85
#    cache_size_mb: 30000
$ docker logs earthly-buildkitd 2>&1 | grep -i cache
CACHE_SIZE_MB=41550
$ docker exec earthly-buildkitd buildctl debug workers -v | grep -i -A1 'all.*true'
        All:            true
        Keep Bytes:     41.55GB
```
Size only:
```
$ grep cache_size ~/.earthly/config.yml
#    cache_size_pct: 85
    cache_size_mb: 30000
$ docker logs earthly-buildkitd 2>&1 | grep -i cache
CACHE_SIZE_MB=30000
$ docker exec earthly-buildkitd buildctl debug workers -v | grep -i -A1 'all.*true'
        All:            true
        Keep Bytes:     30GB
```
Nothing:
```
$ grep cache_size ~/.earthly/config.yml
#    cache_size_pct: 85
#    cache_size_mb: 30000
$ docker logs earthly-buildkitd 2>&1 | grep -i cache
CACHE_SIZE_MB=0
$ docker exec earthly-buildkitd buildctl debug workers -v | grep -i -A1 'all.*true'
        All:            true
        Keep Bytes:     5GB
```
(you can see why I'm doing this PR!)
Size wins:
```
$ grep cache_size ~/.earthly/config.yml
    cache_size_pct: 85
    cache_size_mb: 30000
$ docker logs earthly-buildkitd 2>&1 | grep -i cache
clamping cache size to fixed size of 30000 MB
CACHE_SIZE_MB=30000
$ docker exec earthly-buildkitd buildctl debug workers -v | grep -i -A1 'all.*true'
        All:            true
        Keep Bytes:     30GB
```
Percent wins:
```
$ grep cache_size ~/.earthly/config.yml
    cache_size_pct: 85
    cache_size_mb: 60000
$ docker logs earthly-buildkitd 2>&1 | grep -i cache
clamping cache size to 41550 MB (85% of filesystem)
CACHE_SIZE_MB=41550
$ docker exec earthly-buildkitd buildctl debug workers -v | grep -i -A1 'all.*true'
        All:            true
        Keep Bytes:     41.55GB
```

alexcb:
:raised_hands: Thanks for the contribution and various changes along the way.

PR #1737, opened 2022-03-09
Documentation of new cache percentage setting

rrjjvv:
Updates related to #1734

### DIFF
--- a/docs/earthly-config/earthly-config.md
+++ b/docs/earthly-config/earthly-config.md
@@ -58,6 +58,11 @@ earthly config global.cache_size_mb 20000
 
 Specifies the total size of the BuildKit cache, in MB. The BuildKit daemon uses this setting to configure automatic garbage collection of old cache. A value of 0 causes the size to be adaptive depending on how much space is available on your system. The default is 0.
 
+### cache_size_pct
+
+Similar to [`cache_size_mb`](#cache_size_mb) (including the default and treatment of 0), but expressed as a percentage (0-100) of available space.
+This allows for an adaptively-sized cache, but in a controlled and deterministic manner.
+
 ### disable_analytics
 
 When set to true, disables collecting command line analytics; otherwise, earthly will report anonymized analytics for invocation of the earthly command. For more information see the [data collection page](../data-collection/data-collection.md).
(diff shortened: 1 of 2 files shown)


docs/earthly-config/earthly-config.md
```diff
@@ -58,6 +58,11 @@ earthly config global.cache_size_mb 20000
 
 Specifies the total size of the BuildKit cache, in MB. The BuildKit daemon uses this setting to configure automatic garbage collection of old cache. A value of 0 causes the size to be adaptive depending on how much space is available on your system. The default is 0.
 
+### cache_size_pct
+
+Similar to [`cache_size_mb`](#cache_size_mb) (including the default and treatment of 0), but expressed as a percentage (0-100) of available space.
```
alexcb, line 63: What if we change this to:

```
Specifies the total size of the BuildKit cache, as a percentage (0-100) of the total filesystem size.
When used in combination of `cache_size_mb`, the lesser of the two values will be used. This limit is ignored when set to `0`.
```

I think that might be more clear rather than saying `(including the default and treatment of 0)`.

docs/guides/cache.md
```diff
@@ -18,11 +18,12 @@ Earthly cache is persisted in a docker volume called `earthly-cache` on your sys
 
 ## Specifying cache size
 
-The default cache size is adaptable depending on available space on your system. If you would like to limit the cache size more aggressively, you can specify a different limit by modifying the `cache_size_mb` setting in the [configuration](../earthly-config/earthly-config.md). For example:
+The default cache size is adaptable depending on available space on your system. If you would like to limit the cache size more aggressively, you can specify a different limit by modifying the `cache_size_mb` and/or `cache_size_pct` setting in the [configuration](../earthly-config/earthly-config.md). For example:
```
alexcb, line 21:
```suggestion
The default cache size is adaptable depending on available space on your system. If you would like to limit the cache size more aggressively, you can specify a different limit by modifying the `cache_size_mb` and/or `cache_size_pct` settings in the [configuration](../earthly-config/earthly-config.md). For example:
```

docs/earthly-config/earthly-config.md
rrjjvv: Implemented almost verbatim, except s/of/with/ and removing the `0` markup to stay consistent with the rest of the doc.

PR #1745, opened 2022-03-12
New environment variables to control mirrors used by `dockerd`

rrjjvv:
This introduces `EARTHLY_DOCKERD_REGISTRY_MIRROR` and `EARTHLY_DOCKERD_INSECURE_REGISTRY` which allows control over the registries used by the `dockerd` instance used in `WITH DOCKER` commands.

Though buildkit is responsible for the majority of image pulling, docker (in docker) is responsible for pulling in some scenarios.

This allows for comma-seperated lists and will prepend an appropriate protocol if one is not present (to allow a buildkit-destined setting, which lacks protocol, to be used verbatim here).

Though technically and functionally separate, this is a logical continuation of #1733.  Both were done in context of testing, but this one can also be valuable for regular usage.


rrjjvv:
This change works as-is, but there may be a few areas that need feedback and possibly improvement.

First, the "interface".  This started as a personal hack, so this is effectively a magic/hidden setting.  To use it requires setting the proper environment variables when launching earthly.  How visible/accessible should these settings be?  I don't think they deserve CLI flags.  They could deserve to be configuration (file) values, so that mirror information for dockerd and buildkitd are in the same place.  Or should it remain exactly like this, but possibly mentioned in the documentation as an advanced/specialty setting?

Second, I don't know if this is sufficient to stop *all* hub traffic.  I just know it's "good enough" (in tandem with #1733) to allow the tests to run repeatedly with no issues.  It was never my quest to actively squash all hub traffic (only to run the tests), but it would be nice to tackle them since it's now on our minds.

Third, up to this point, I haven't figured out a way to write an automated test for the *intended* use (invoking earthly with those environment variables set).  I _think_ that translates into calling the `+DO_EARTHLY` target with the environment variables set via `pre_command`.  I could not get that to work,  which I suspect is due to `WITH DOCKER`.  I tried incorporating some techniques I found in other tests (such as using `HOST`, spinning up an nginx to intercept traffic, etc.), but I couldn't find a winning combination.  The tests in this PR are the best I could come up with so far.  Let me know if it is sufficient.  Even if they are, I'm still curious how you guys would go about making them better.

In summary, let me know about the "interface".  I kept this as a draft because I don't know if this is the best way, and didn't want to imply that I thought so.

rrjjvv:
I _thought_ someone, somewhere, said that putting a PR in Draft would still get looked at.  Maybe I read (or remembered) incorrectly.  Either way, I'll move this to "ready to review", since it truly is.  I want to ensure that any delay is because you're too busy implementing cool stuff, and not because I goofed.

vladaionescu:
> I thought someone, somewhere, said that putting a PR in Draft would still get looked at.

Ah yes - sorry, it slipped our radar :-) I've now assigned @alexcb to this.

Somewhat unsure if this is a good idea, as normally there shouldn't be any pushing or pulling from within `WITH DOCKER`, if `--pull` is used correctly. Do you have a use-case in mind for these?

rrjjvv:
I agree that *normally* there shouldn't.  The (production) edge-case I could see is if a company had egress to dockerhub totally prohibited.  (My company had considered it, but opted not to, thankfully.)  So for this edge-case, it would be for users that didn't know any better.  But that's all hypothetical.  I just know this change was needed for me to *fully* run the earthly test suite.  My other PR (thanks!) got me most of the way, but there was one test in particular (https://github.com/earthly/earthly/blob/main/Earthfile#L101-L111) that would continue to fail (due to dockerhub rate limits).  After this change, that test could then be run to be completion.

I don't know if that is the *only* test.  Or even if my change works "on accident" (I'd like to think not, since it makes sense).  But the primary reason for the change was to allow the tests to pass (thus allow me to make contributions without blindly relying on CI to catch any errors).

rrjjvv:
Running on a pristine checkout (without changes from this PR, but with my upstream changes from my previously merged PR) does indeed still trigger a dockerhub rate-limit, though it now takes a number of iterations to hit it.  I narrowed it down to two tests: the [`+earthly-script-no-stdout`](https://github.com/earthly/earthly/blob/main/Earthfile#L101-L111) target, and the [front-end tests](https://github.com/earthly/earthly/blob/main/util/containerutil/frontend_test.go) from the unit tests themselves.

So if you don't think this change is worthy for 'production' usage, either philosophically or because it doesn't hold its own weight, then I suspect it doesn't belong just for the sake of tests either.  In that case, I might propose stopping those tests from running when (for example) `DOCKERHUB_AUTH=false`.  I'm just hoping to avoid having to do manual work (whether applying this patch privately or manipulating which tests are run) if I can.

alexcb:
> Running on a pristine checkout (without changes from this PR, but with my upstream changes from my previously merged PR) does indeed still trigger a dockerhub rate-limit

This definitely suggests we still need to do some work on preventing access / redirecting it to a mirror.

I can take a deeper look at how dockerhub is accessed and try to put together a test that validates we can still run the earthly tests without access. I see value in being able to run our tests without having to have a dockerhub login.

alexcb:
> I can take a deeper look

I made a loosely-related note under https://github.com/earthly/earthly/pull/1733#issuecomment-1075758989 in terms of setting up a self-hosted insecure mirror. I found it particularily helpful to ignore dockerhub hosts via `echo "1.2.3.4 index.docker.io auth.docker.io registry-1.docker.io dseasb33srnrn.cloudfront.net production.cloudflare.docker.com" >> /etc/hosts`.

I came across a similar instance where `WITH DOCKER` ended up getting rate-limited (and most interestingly since `/etc/hosts` was only modified on my host and in `earthly-buildkit`, it didn't end up in the `buildkitsandbox` container, so one would have to also perform something similar to https://github.com/earthly/earthly/blob/main/tests/git-ssh-server/setup/Earthfile#L59-L68 (or make use of the new experimental `HOST` keyword).

I created #1765 #1764 and 1763 as a result of this deeper look, once I get those dealt with I'll come back for another deep-dive.

alexcb:
I added the following into https://github.com/earthly/earthly/blob/main/buildkitd/dockerd-wrapper.sh#L88 (just before `dockerd` gets started):

```bash
    cp /etc/hosts /tmp/hosts
    umount /etc/hosts
    cp /tmp/hosts /etc/hosts
    echo "1.2.3.4 index.docker.io auth.docker.io registry-1.docker.io dseasb33srnrn.cloudfront.net production.cloudflare.docker.com" >> /etc/hosts
    echo docker hack!!!!!
```

as a way to prevent docker from being able to pull any images, then ran:

```Earthfile
test:
    FROM earthly/dind:alpine
    WITH DOCKER --pull alpine:3.15
        RUN docker run --rm alpine:3.15 hostname
    END
```

and was able to see `alpine:3.15` was correctly pulled via the cache I had configured via my earthly config (`~/.earthly/config.yml`):

```
global:
    disable_analytics: true
    local_registry_host: tcp://127.0.0.1:8371
    conversion_parallelism: 5
    buildkit_additional_config: |
        [registry."docker.io"]
          mirrors = ["192.168.0.80:5000"]
        [registry."192.168.0.80:5000"]
          http = true
          insecure = true
```

It was only when I removed the `--pull alpine:3.15`, did things fail due to me blocking the `/etc/hosts` entries.

As long as you define all the images to pre-pull via `--pull ...`, it seems like earthly can work as-is without this PR. The only downside is for cases where your image names aren't known a head of time and instead are dynamically generated. I'm not sure if that's a use-case we need to worry about, but perhaps it is?

>  I just know this change was needed for me to fully run the earthly test suite.

In this case, #1765 should hopefully fix the issue that was preventing all earthly-in-earthly instances from correctly using the mirror.

rrjjvv:
> In this case, #1765 should hopefully fix the issue that was preventing all earthly-in-earthly instances from correctly using the mirror.

Yeah, based on the conversation in the other PR (around the bad config), I was starting to come to that same assumption/hope: that the configuration wasn't broken to the point it would throw errors, but that it resulted in a "partial" config.  And that the "missing" part of the config was either discarded (e.g. lax parser) or was technically valid but interpreted/used in a way that didn't have the desired effect.

> cases where your image names aren't known a head of time and instead are dynamically generated. 

I'm assuming you're not talking about something like `--pull alpine:"${version}"` (which I assume is possible, not never tried) or `--pull alpine:$(cat .version)` (maybe something the new 0.6.12 can do?)?

Actually, I think what you're describing might be exactly like https://github.com/earthly/earthly/blob/main/util/containerutil/frontend_test.go#L348; where you are running someone else's code that pulls images.  So not so much that the images *can't* be known and fed in to `--pull`, but more that you don't realize it until you get rate-limited, you notice your build time went through the roof, etc.  In fact, this is reminding me of a Slack conversation where I was trying to figure out the best strategy for dealing with this (if I can find it, I'll post a link).

On one hand, I'm the first person to raise the issue since the project's inception (to my knowledge).  It doesn't appear to be a pressing issue in the real world.  If you did want to address that population of users, then I would say my PR would need to be enhanced (as it's currently exposed via only a secret environment variable).

That said... I only have about five minutes of thought into this, but there might be a more robust, future-proof way to solve the problem: rather than solving it via code (and everything that comes along with it), solve it via making the image itself (easily and safely) extendable.  For *very* rough example:
```
FROM earthly/builkitd
# user knows exactly what they're doing; respect it
COPY daemon.json /etc/docker/daemon.json
# or, 'merge' this partial config with what the wrapper script wants to write
RUN echo '{<custom mirror json>}' > /etc/docker/daemon.json.partial
```
I suspect the common case would actually be
```
FROM earthly/earthly
# same idea as above
```
but I guess that's really no different... it just won't work if you're using remote/non-managed buildkit (where the first comes in).

Something like that isn't without it's own set of issues, but it makes the customization *possible* (in a non-frail way).  You can worry about making it *easy* down the road, if that's ever required. 

---
Either way, it's always awesome to claim (and have) zero external dependencies.  But with my previous accepted changes and all the secondary fixes that have come out of this effort, my immediate issues are effectively solved and we've definitely achieved a "good enough" state.  I'm totally game for moving from "good enough" to "perfection", just don't do it on my account!

alexcb:
I got thinking a bit about how we could create a more flexible way to account for future edge-cases where a user needs to provide some custom code to configure docker (i.e. modify the behaviour of https://github.com/earthly/earthly/blob/main/buildkitd/dockerd-wrapper.sh).

Maybe we could allow a way for a user to provide their own custom version of the file? or even automatically evaluate a command if a user specifies something like `ARG WITH_DOCKER_PRE_SCRIPT=/my/custom/script`?

The script itself is added via a mount rather than existing in `earthly/dind:alpine`, but I did come up with a really hacky proof-of-concept that wraps the `docker` command with a script:

```Earthfile
hack:
    FROM earthly/dind:alpine
    RUN mv /usr/local/bin/dockerd /usr/local/bin/dockerd_
    RUN echo "#!/bin/sh
set -e
echo hack hack hack > /root/hack-log

# block dockerhub access
cp /etc/hosts /tmp/hosts
umount /etc/hosts
cp /tmp/hosts /etc/hosts
echo 1.2.3.4 index.docker.io auth.docker.io registry-1.docker.io dseasb33srnrn.cloudfront.net production.cloudflare.docker.com >> /etc/hosts

echo '
{
  \"registry-mirrors\" : [\"http://192.168.0.80:5000\"]
}' > /etc/docker/daemon.json


# run docker
/usr/local/bin/dockerd_ \"\$@\"
" > /usr/local/bin/dockerd && chmod +x /usr/local/bin/dockerd
    WITH DOCKER
        RUN docker pull alpine:3.15; export ec=$?; cat /root/hack-log; exit $ec
    END
```

> I'm assuming you're not talking about something like --pull alpine:"${version}" (which I assume is possible, not never tried) or --pull alpine:$(cat .version) (maybe something the new 0.6.12 can do?)?

The ([yet-to-be-documented](https://github.com/earthly/earthly/issues/1767)) `VERSION --shell-out-anywhere 0.6` option supports this.

> That said... I only have about five minutes of thought into this, but there might be a more robust, future-proof way to solve the problem: rather than solving it via code (and everything that comes along with it), solve it via making the image itself (easily and safely) extendable. For very rough example:

Building upon your idea, maybe an even simpler approach would be to check if `/etc/docker/daemon.json` already exists, and not overwrite it?

```
if ! [ -f /etc/docker/daemon.json ]; then       <--- a simple if guard
    mkdir -p /etc/docker
    cat <<'EOF' >/etc/docker/daemon.json
{
    "default-address-pools" : [
        {
            "base" : "172.21.0.0/16",
            "size" : 24
        },
        {
            "base" : "172.22.0.0/16",
            "size" : 24
        }
    ]
}
fi
EOF

```

rrjjvv:
I had a more thorough response drafted, but lost it when my machine locked up.  Lucky for you you'll get an abridged version!

The main reason my idea was complex (assuming you're referring to the "snippet" idea) was due to the current customization that's already done.  Since those have reasonable defaults, your customization obviously had a reason behind it.  Whether it's because they make things "better" or are required for the entire system to operate, who knows (without going through code or commit history).

If you make me provide a whole file (with no other option), then there will be some combination of:
1. I'll lose out on "better" behavior, if you change/add a performance-related value
2. You'll have to consider those settings 'contract', making it burdensome for you to make changes without breaking users
3. You'll declare "use at your own risk", and users will have to watch out for failures (they'd probably be lucky if things totally broke, but something like you changing to a `/18` (which an unknown component relies on) while they are still specifying a `/16` would probably result in obtuse bugs)

It's very easy for me, as an end-user with this one use-case in mind, to say "what you care about and what I care about don't conflict".  I know I would quickly change my tune if I was the one implementing the code or supporting users.  I think for the specific case of `dockerd` (given the stability of the current customizations and audience needing/wanting this capability), requiring a full override is sufficient.  But in terms of establishing a pattern for the future, it would be ideal (as an end-user) to not have to provide a full config when I care about only a subset.

*Ideally* it would follow the same concept that distro/package maintainers have adopted: there's `/etc/product/conf.txt`, which is intended to be off-limits for editing, but then it sources from `/etc/product/conf.d/`, where you drop in what you care about (though it may not be a valid config on its own).  It does nothing to stop technically or logically invalid configurations, but everyone gets to declare what they care about (and *only* what they care about).

It just occurred to me that there's a parallel scenario regarding buildkit garbage collection (first item in https://github.com/earthly/earthly/issues/1723#issuecomment-1063675365, and described in more detail in the body).

In any event, regardless of what it ends up looking like, I do think dealing with niche settings at the file level, rather than what my PR does, is far superior and future-proof.  I'm a big proponent of "make easy things easy, make hard things possible".  Or in this context, "make typical things easy, make atypical things possible".  This is a great step in that direction!

rrjjvv:
Hopefully this doesn't violate any unwritten rules, but I'll close this out to cut down on noise.  Given that #1780 (merged) essentially supersedes this, not to mention that this PR now shows conflicts, I see no value in keeping it around.

PR #1750, opened 2022-03-17
docs: workaround gitbook bug

alexcb:
Rather than get a link to https://docs.earthly.dev/docs/guides/
gitbook is linking to 	https://github.com/earthly/earthly/blob/main/docs/earthfile/guides/README.md
(which doesn't exist); use a tinyurl to point to the correct location
instead (as suggested by gitbook support).


vladaionescu (approved):
🤦

PR #1756, opened 2022-03-18
New platform functionality

vladaionescu:
### Changed

- A more obvious error is printed if `WITH DOCKER` starts non-natively. This is not supported and it wasn't obvious before.

### Added

- A new experimental mode in which `--platform` works. To enable these features in your builds, set `VERSION --new-platform 0.6`:
  - There is now a distinction between **user** platform and **native** platform. The user platform is the platform of the user running Earthly, while the native platform is the platform of the build worker (these can be different when using a remote buildkit)
  - New platform shorthands are provided: `--platform=native`, `--platform=user`.
  - New builtin args are available: `NATIVEPLATFORM`, `NATIVEOS`, `NATIVEARCH`, `NATIVEVARIANT` (these are the equivalent of the `USER*` and `TARGET*` platform args).
  - When no platform is provided, earthly will default to the **native** platform
  - Additionally, earthly now default to native platform for internal operations too (copy operations, git clones etc)
  - Earthly now allows changing the platform in the middle of a target (`FROM --platform` is not a contradiction anymore). There is a distinction between the "input" platform of a target (the platform the caller passes in) vs the "output" platform (the platform that ends up being the final platform of the image). These can be different if the caller passes `BUILD --platform=something +target`, but the target starts with `FROM --platform=otherthing ...`.


tests/platform/Earthfile
```diff
@@ -0,0 +1,137 @@
...
+    ARG tests="regular/linux/arm/v7 regular/linux/arm64 regular/linux/amd64 
+        native user default
+        override/linux/arm64 override/linux/amd64
+        default_override/linux/arm64 default_override/linux/amd64
+        copy_override/linux/arm64 copy_override/linux/amd64
+        copy_other_set/linux/arm64 copy_other_set/linux/amd64
+        copy/linux/arm64 copy/linux/amd64
+        copy_native1 copy_native2"
+    FOR plat IN $tests
+        RUN /bin/sh -c "set -exo pipefail ; \
```
alexcb, line 18: that's a lot of escaping :sweat_smile:

tests/platform/Earthfile
```diff
@@ -0,0 +1,137 @@
...
+            outdir=./out/$plat ; \
+            test -d \$outdir && \
+            cat \$outdir/uname-m && \
+            echo plat=\$plat && \
```
alexcb, line 22: should this one not be escaped? (since it's `FOR plat IN $tests`)?
```suggestion
            echo plat=$plat && \
```

alexcb: (or otherwise, should the above `outdir=./out/$plat ; \` be `outdir=./out/\$plat ; \`?

alexcb (approved):

tests/platform/Earthfile
vladaionescu: It actually works both ways in this case (it's just a different shell doing the expansion) :D

I've made this consistent now.

PR #1761, opened 2022-03-22
update scrub-https-credentials test user

alexcb:
This test doesn't require the username to exist, this
PR adds a note to reflect that.


vladaionescu (approved):
LoL - I actually own some Kallax's

PR #1763, opened 2022-03-22
prevent docker pull in +earthly-script-no-stdout

alexcb:
This prevents an explicit pull to docker.io/earthly/earthlybinaries:prerelease
in the +earthly-script-no-stdout target which can lead to rate-limiting.

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -104,9 +104,12 @@ earthly-script-no-stdout:
     FROM earthly/dind:alpine
     RUN apk add --no-cache --update bash
     COPY earthly .earthly_version_flag_overrides .
-    WITH DOCKER --pull earthly/earthlybinaries
-        RUN ./earthly --version > earthly-version-output
-    END
+
+    # This script performs an explicit "docker pull earthlybinaries:prerelease" which can cause rate-limiting
+    # to work-around this, we will copy an earthly binary in, and disable auto-updating (and therefore don't require a WITH DOCKER)
+    COPY +for-linux/earthly /root/.earthly/earthly-prerelease
+    RUN EARTHLY_DISABLE_AUTO_UPDATE=true ./earthly --version > earthly-version-output
+
     RUN test "$(cat earthly-version-output | wc -l)" = "1"
     RUN grep '^earthly version.*$' earthly-version-output # only --version info should go to stdout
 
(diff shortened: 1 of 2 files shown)


Earthfile
```diff
@@ -104,9 +104,17 @@ earthly-script-no-stdout:
...
     RUN apk add --no-cache --update bash
     COPY earthly .earthly_version_flag_overrides .
+
+    # TODO this script performs an explicit "docker pull earthlybinaries:prerelease" which can cause rate-limiting
+    # to work-around this, we will copy an earthly binary in, and update the last-check state files
+    COPY +for-linux/earthly /root/.earthly/earthly-prerelease
+    RUN echo "$(date +%s)" >/tmp/last-earthly-prerelease-check
+    RUN echo "$(md5sum .earthly_version_flag_overrides | awk '{print $1}')" > /tmp/last-earthly-flag-hash
+
     WITH DOCKER --pull earthly/earthlybinaries
```
vladaionescu, line 114: BTW shouldn't this be

```suggestion
    WITH DOCKER --pull earthly/earthlybinaries:prerelease
```

Earthfile
```diff
@@ -104,9 +104,17 @@ earthly-script-no-stdout:
     FROM earthly/dind:alpine
     RUN apk add --no-cache --update bash
     COPY earthly .earthly_version_flag_overrides .
+
+    # TODO this script performs an explicit "docker pull earthlybinaries:prerelease" which can cause rate-limiting
+    # to work-around this, we will copy an earthly binary in, and update the last-check state files
+    COPY +for-linux/earthly /root/.earthly/earthly-prerelease
+    RUN echo "$(date +%s)" >/tmp/last-earthly-prerelease-check
+    RUN echo "$(md5sum .earthly_version_flag_overrides | awk '{print $1}')" > /tmp/last-earthly-flag-hash
```
vladaionescu, line 112: Feels a bit hacky - could we add a flag in `./earthly` that disables auto-updates instead?

Earthfile
alexcb: since we're skipping the part that pulls via docker, we can remove the entire `WITH docker`.

Earthfile
alexcb: That's a good call, I added in a `EARTHLY_DISABLE_AUTO_UPDATE=true` to disable it. It seems more robust than having to pop flags (and not have to worry about it showing up in the middle of the regular earthly flags).

Issue #1764, opened 2022-03-22
`earthly config global` doesn't keep newlines


alexcb:
consider:

```
echo "$GLOBAL_CONFIG"
{disable_analytics: true, local_registry_host: 'tcp://127.0.0.1:8371', conversion_parallelism: 5, buildkit_additional_config: '[registry."docker.io"]
  mirrors = ["192.168.0.80:5000"]
[registry."192.168.0.80:5000"]
  http = true
  insecure = true'}
```

```
> foo.yml
earthly --config foo.yml config global "$GLOBAL_CONFIG"
```

then run `cat foo.yml`, to see:

```
global:
    disable_analytics: true
    local_registry_host: tcp://127.0.0.1:8371
    conversion_parallelism: 5
    buildkit_additional_config: '[registry."docker.io"] mirrors = ["192.168.0.80:5000"] [registry."192.168.0.80:5000"] http = true insecure = true'
```

instead it should be

```
global:
    disable_analytics: true
    local_registry_host: tcp://127.0.0.1:8371
    conversion_parallelism: 5
    buildkit_additional_config: '[registry."docker.io"]
  mirrors = ["192.168.0.80:5000"]
[registry."192.168.0.80:5000"]
  http = true
  insecure = true'
```

alexcb:
Here's a one-liner to reproduce it:
```
export foo="{disable_analytics: true, local_registry_host: 'tcp://127.0.0.1:8371', conversion_parallelism: 5, buildkit_additional_config: '[registry."docker.io"]
  mirrors = ["192.168.0.80:5000"]
[registry."192.168.0.80:5000"]
  http = true
  insecure = true'}" && > /tmp/foo.yml && ./build/linux/amd64/earthly --config=/tmp/foo.yml config global "$foo"
```

alexcb:
This ain't no bug -- it's a (yaml) feature! :see_no_evil: 

Accoring to the yaml spec
```yaml
{"key": "one
two"}
```

handles the newline by converting it into a space. And multiple newlines are converted into a single newline.
See https://yaml.org/spec/1.2.2/#single-quoted-style (and in particular `Example 7.9 Single Quoted Lines` ).

It's interesting to note that JSON doesn't even allow strings that contain a newline, e.g. the following is invalid:

```json
{"key": "invalid
value"}
```

Therefore if you want to include a literal newline, you need an **extra** newline:

```yaml
{'key': 'one

two'}
```

which will produce

```
one

two
```

[...]

PR #1765, opened 2022-03-22
fix tests with insecure dockerhub mirror

alexcb:
This works around https://github.com/earthly/earthly/issues/1764
and fixes some indentation issues which were preventing running our
tests against a self-hosted insecure pullthrough cache.

With this PR, I can now run our tests against a self-hosted cache:

   env -i HOME=$HOST PATH=$PATH ./build/linux/amd64/earthly --env-file=<(echo "") --config=$HOME/.earthly/config.yml -P ./tests+all --DOCKERHUB_AUTH=false --DOCKERHUB_MIRROR=192.168.0.80:5000 --DOCKERHUB_MIRROR_INSECURE=true

Note that `env -i HOME=$HOST PATH=$PATH`, and `--env-file=<(echo "")`
are used to ensure no additional cache environment variables are
accidentally fed into earthly.

### DIFF
--- a/earthly-entrypoint.sh
+++ b/earthly-entrypoint.sh
@@ -9,7 +9,7 @@ if [ ! -f "$earthly_config" ]; then
 
   # Apply global configuration
   if [ -n "$GLOBAL_CONFIG" ]; then
-    earthly --config $earthly_config config global "$GLOBAL_CONFIG"
+    earthly --config "$earthly_config" config global "$GLOBAL_CONFIG"
   fi
 
   # Apply git configuration
--- a/Earthfile
+++ b/Earthfile
@@ -1,4 +1,4 @@
-VERSION 0.6
+VERSION --shell-out-anywhere 0.6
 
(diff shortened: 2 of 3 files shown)


earthly-entrypoint.sh
```diff
@@ -9,7 +9,13 @@ if [ ! -f "$earthly_config" ]; then
 
   # Apply global configuration
   if [ -n "$GLOBAL_CONFIG" ]; then
-    earthly --config $earthly_config config global "$GLOBAL_CONFIG"
+    if [ "$( echo "$GLOBAL_CONFIG" | wc -l)" -gt 1 ]; then
+      # TODO FIXME once this issue is fixed, we can revert this hack
```
alexcb, line 13: probably best to just fix https://github.com/earthly/earthly/issues/1764 first rather than commit this work-around.

this was here just to show that it can work once this issue is fixed.

vladaionescu (approved):

PR #1775, opened 2022-03-25
Warn when encountering an Earthfile with no VERSION specified

vladaionescu:
Screen Shot 2022-03-24 at 8 32 07 PM

### DIFF
--- a/ast/parser/Earthfile
+++ b/ast/parser/Earthfile
@@ -1,3 +1,5 @@
+VERSION 0.6
+
 FROM alpine:3.15
 ARG TARGETARCH
 IF [ "$TARGETARCH" = "arm64" ]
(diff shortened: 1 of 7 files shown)


ast/parser/Earthfile
```diff
@@ -1,3 +1,5 @@
+VERSION 0.6
```
alexcb, line 1: looks like it worked :smile:

alexcb (approved):

buildcontext/local.go
```diff
@@ -52,18 +55,37 @@ func (lr *localResolver) resolveLocal(ctx context.Context, gwClient gwclient.Cli
...
+			ftrs, err = parseFeatures(buildFilePath, featureFlagOverrides, ref.GetLocalPath(), lr.console)
+			if err != nil {
+				return nil, err
+			}
+		}
+		return &buildFile{
+			path: buildFilePath,
+			ftrs: ftrs,
+		}, nil
+	})
```
alexcb, line 83: looks like there's a missing:
```
		if err != nil {
			return nil, err
		}
```
here?

vladaionescu: ineffassign FTW

PR #1780, opened 2022-03-25
Keep existing daemon.json data

alexcb:
If /etc/docker/daemon.json already exists, rather than overwrite
it, we should simply update it with the new network IP ranges.

This, for example, allows a user to define an Earthfile which
configures the WITH DOCKER instance to use a different mirror:

    test:
        FROM earthly/dind:alpine
        RUN mkdir -p /etc/docker && echo '{"registry-mirrors" : ["http://192.168.0.80:5000"]}' > /etc/docker/daemon.json
        WITH DOCKER
            RUN --no-cache \
               cat /etc/docker/daemon.json && \
               docker run --rm alpine:3.15 hostname
        END

### DIFF
--- a/buildkitd/docker-auto-install.sh
+++ b/buildkitd/docker-auto-install.sh
@@ -20,6 +20,14 @@ detect_docker_compose() {
     return "$has_dc"
 }
 
+detect_jq() {
+    set +e
+    command -v jq
+    has_jq="$?"
+    set -e
+    return "$has_jq"
+}
+
 print_debug() {
     set +u
     if [ "$EARTHLY_DEBUG" = "true" ] ; then
@@ -78,10 +86,18 @@ install_dockerd() {
     esac
 }
 
+apt_update_done="false"
+apt_get_update() {
+    if [ "$apt_update_done" != "true" ]; then
+        apt-get update
+        apt_update_done=true
+    fi
+}
+
 install_dockerd_debian_like() {
     export DEBIAN_FRONTEND=noninteractive
     apt-get remove -y docker docker-engine docker.io containerd runc || true
-    apt-get update
+    apt_get_update
     apt-get install -y \
         apt-transport-https \
         ca-certificates \
@@ -93,7 +109,7 @@ install_dockerd_debian_like() {
         "deb [arch=$(dpkg --print-architecture)] https://download.docker.com/linux/$distro \
         $(lsb_release -cs) \
         stable"
-    apt-get update
+    apt-get update # dont use apt_get_update since we must update the newly added apt repo
     apt-get install -y docker-ce docker-ce-cli containerd.io
 }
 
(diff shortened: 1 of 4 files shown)


alexcb:
If we decide to go this route, we'll need to update `earthly/dind` to include `jq`.

Initially I tried using sh+sed; however it becomes very tricky to handle a case where `default-address-pools` already contains entries.

buildkitd/docker-auto-install.sh
```diff
@@ -110,10 +118,33 @@ install_dockerd_amazon() {
...
+
 if [ "$(id -u)" != 0 ]; then
     echo "Warning: Docker-in-Earthly needs to be run as root user"
 fi
 
+if ! detect_jq; then
+    echo "jq is missing. Attempting to install automatically."
+    install_jq
+fi
+
```
vladaionescu, line 147: I think we need to update this script too then: https://github.com/earthly/lib/blob/main/install-dind.sh.

buildkitd/docker-auto-install.sh
```diff
@@ -110,10 +118,33 @@ install_dockerd_amazon() {
...
+            apk add --update --no-cache jq
+            ;;
+
+        amzn)
+            yum install jq
+            ;;
+
+        *)
+            export DEBIAN_FRONTEND=noninteractive
+            apt-get update
```
vladaionescu, line 133: Does this cause a duplicate `apt-get update`? One for dockerd and one for jq?

rrjjvv:
I was going to post this in the 'other' PR, but happened to see this PR.  If I may be so bold, I have a suggestion.  It's not *needed*, but it'll add clarity, now that the config is no longer a black-box.  Unless there's a non-obvious reason why it should be this way, I'd recommend converting `--bip` https://github.com/earthly/earthly/blob/422f019ebc32855c7b642e7d750ca7fafb471427/buildkitd/dockerd-wrapper.sh#L92
to equivalent JSON (rather than splitting the config between file and flag), especially since it's clearly related to the address-pool settings.

alexcb:
You found my half-ready PR before I was able to link it to our other conversation :sweat_smile: 

> I'd recommend converting --bip

That's a great idea! I hadn't clued into the fact that there were equivalents there -- it's much more tidy this way.

rrjjvv:
And I was going to commend you on being so fast that you implemented my perfect-world scenario before you even read about it!

buildkitd/docker-auto-install.sh
alexcb: that's a good point.

How does a test like

```
if [ -f /var/lib/apt/periodic/update-success-stamp ] && [ -z "$(find /var/lib/apt/periodic/update-success-stamp -mmin +50)" ]; then
    echo no update is needed
fi
```

look for detecting when we can skip it.

vladaionescu: Yeah, that could work - another idea could be to leave a file around the first time we run apt, and skip if we see the file.

alexcb: I was able to use a variable in the sh script to do this.

vladaionescu (approved):

PR #1784, opened 2022-03-28
Display cursor on same line as prompt

alexcb:
When prompting the user to pick a password, the cursor should be
displayed on the same line as the prompt:

for example:

    pick a password: <cursor>

Previously the prompt looked like:

    pick a password
    <cursor>


alexcb:
Before:
image

After:
image

This mimics the behaviour of sudo:
image

vladaionescu (approved):

PR #1788, opened 2022-03-29
jq is now required when using `WITH DOCKER`

alexcb:
jq is now required when using `WITH DOCKER` (dependency was introduced
in https://github.com/earthly/earthly/commit/88f94c5ed342a0162c202f5ce666d8350bf3fcc4), it is installed
automatically via `docker-auto-install.sh`.

dind-ubuntu already installs dependencies with this script, we should
use it for the alpine version as well to be consistent. Note that the
underlying docker:dind already provides dockerd, so only docker-compose
and jq will be installed.

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -419,7 +419,8 @@ dind:
 
 dind-alpine:
     FROM docker:dind
-    RUN apk add --update --no-cache docker-compose
+    COPY ./buildkitd/docker-auto-install.sh /usr/local/bin/docker-auto-install.sh
+    RUN docker-auto-install.sh
     ARG EARTHLY_TARGET_TAG_DOCKER
     ARG DIND_ALPINE_TAG=alpine-$EARTHLY_TARGET_TAG_DOCKER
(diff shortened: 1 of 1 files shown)


Earthfile
```diff
@@ -419,7 +419,7 @@ dind:
 
 dind-alpine:
     FROM docker:dind
-    RUN apk add --update --no-cache docker-compose
+    RUN apk add --update --no-cache docker-compose jq
```
vladaionescu, line 422: Could we use alpine + the docker-auto-install.sh script, just like `dind-ubuntu` target below?

It would help with minimizing differences.

alexcb: I don't see why not -- the `docker:dind` image will already have `dockerd` installed, so that portion of the code should be skipped. It's only the ubuntu version which performs an `apt-get remove -y docker docker-engine docker.io containerd runc || true`, where as the alpine version doesn't do any of that.

I'll switch it over.

vladaionescu (approved):

PR #1790, opened 2022-03-29
set EARTHLY_GIT_SHORT_HASH on remote targets

alexcb:
This fixes a bug where EARTHLY_GIT_SHORT_HASH was not set when building
a remote target -- this is due to the codepath being different for
locally referenced targets and remote targets.

This fix introduces a test to validate EARTHLY_GIT_SHORT_HASH and
EARTHLY_GIT_HASH are correctly set in both cases.


vladaionescu (approved):
🙌

Issue #1785, opened 2022-03-28
Shell out anywhere safety in the context of conversion parallelism


vladaionescu:
The new shell-out-anywhere feature introduces a bug when used in the context of conversion parallelism: if any `BUILD` command includes `$(...)` as part of any of its args, then the `$(...)` part may be executed too early.

Conversion parallelism works by looking ahead the AST for any `BUILD` commands that could be spun off in parallel. There is currently a check in place to verify that the passed-in *build args* are not dynamic args (which couldn't be ran in parallel, because we'd have to wait for the execution to reach that command first, so that the `$(...)` is ran on top of the current layer). But now that any flag can potentially be the result of a shell-out command (not just build args), we would have to check for all flags/options of the `BUILD` command (e.g. `--platform=$(...)`), to make sure that they are not formed dynamically. This check is currently missing.

PR #1791, opened 2022-03-29
prevent expanding args during async BUILDs

alexcb:
This prevents expandArgs from executing when handling BUILDs during
the async phase.

This fixes https://github.com/earthly/earthly/issues/1785

### DIFF
--- a/earthfile2llb/interpreter.go
+++ b/earthfile2llb/interpreter.go
@@ -1659,7 +1659,7 @@ func (i *Interpreter) expandArgs(ctx context.Context, word string, keepPlusEscap
 
 	ret, err := i.converter.ExpandArgs(ctx, runOpts, escapeSlashPlus(word), !async)
 	if err != nil {
-		if errors.Is(err, errShellOutNotPermitted) {
+		if async && errors.Is(err, errShellOutNotPermitted) {
 			return "", errCannotAsync
 		}
 		return "", err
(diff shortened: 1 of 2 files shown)


earthfile2llb/interpreter.go
```diff
@@ -1657,8 +1657,11 @@ func (i *Interpreter) expandArgs(ctx context.Context, word string, keepPlusEscap
 		WithShell:   true,
 	}
 
-	ret, err := i.converter.ExpandArgs(ctx, runOpts, escapeSlashPlus(word))
+	ret, err := i.converter.ExpandArgs(ctx, runOpts, escapeSlashPlus(word), !async)
 	if err != nil {
+		if errors.Is(err, errShellOutNotPermitted) {
```
vladaionescu, line 1662: Slightly more future-proof...

```suggestion
		if async && errors.Is(err, errShellOutNotPermitted) {
```

earthfile2llb/interpreter.go
```diff
@@ -952,11 +952,11 @@ func (i *Interpreter) handleBuild(ctx context.Context, cmd spec.Command, async b
 	if i.local && !(isSafeAsyncBuildArgsKVStyle(opts.BuildArgs) && isSafeAsyncBuildArgs(args[1:])) {
```
vladaionescu, line 952: I wonder if this kind of check is still necessary now. Perhaps it's only necessary when the shell-out-anywhere feature flag is turned off, right?

vladaionescu (approved):
:+1:

earthfile2llb/interpreter.go
alexcb: That's correct -- when we enable `--shell-out-anywhere` for all and make it impossible for the other case, these checks can be removed.

and in the case of the `i.local`, we might even be able to pass a callback into the `c.varCollection.Expand` in order to support shelling-out.

PR #1801, opened 2022-04-01
WIP: dotEnv might contain secrets, dont print them

alexcb:
Initial work on keeping .env variables as a seperate scope
as they might contain secrets, and should never be printed.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -64,6 +64,7 @@ type Opt struct {
 	SaveInlineCache        bool
 	ImageResolveMode       llb.ResolveMode
 	CleanCollection        *cleanup.Collection
+	DotEnvVars             *variables.Scope
 	OverridingVars         *variables.Scope
 	BuildContextProvider   *provider.BuildContextProvider
 	GitLookup              *buildcontext.GitLookup
@@ -158,7 +159,7 @@ func (b *Builder) convertAndBuild(ctx context.Context, target domain.Target, opt
 		}
 		var err error
 		if !b.builtMain {
-			opt := earthfile2llb.ConvertOpt{
+			convertOpt := earthfile2llb.ConvertOpt{
 				GwClient:             gwClient,
 				Resolver:             b.resolver,
 				ImageResolveMode:     b.opt.ImageResolveMode,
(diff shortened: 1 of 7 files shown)


alexcb:
@vladaionescu -- wanting to get your thoughts on the initial direction I'm trying here.

Currently we merge .env and `--build-arg` values under main via https://github.com/earthly/earthly/blob/1555dad58a10180e1ae9d58a866580d27cf0a13f/cmd/earthly/main.go#L2934

which causes us to loose the source which defined them.

Instead I am thinking about keeping them separate:

```
	DotEnvVars             *variables.Scope   // potentially contains secrets
	OverridingVars         *variables.Scope // save to print these values anywhere
```

As a result I'll need to call `opt.OverridingVars.GetAny(name)` and fall back to `opt.DotEnvVars.GetAny(name)` if it's not found in the first one. I'm leaning towards creating some function to make it easier to do this chained-lookup but wanted to get your thoughts on it before continuing.

vladaionescu:
Feels a bit complex... but maybe necessary?

One problem source is that we combine secrets with build args in the .env file. That seems to have been a mistake, because we want to treat secrets differently. So one solution could be to split up the .env file into secrets and build args, and not make any var both at the same time. Not sure what that might look like, and it also has the disadvantage of breaking compatibility.

Another idea might be to obfuscate any secret used as part of the build on output. This is something we'll anyway need to do at some point, for safety.

And yeah, this draft PR is another fair take on the problem. Though I feel that obfuscating all secrets might be a more general way to solve the problem.

alexcb:
> One problem source is that we combine secrets with build args in the .env file. That seems to have been a mistake, because we want to treat secrets differently. So one solution could be to split up the .env file into secrets and build args, and not make any var both at the same time. Not sure what that might look like, and it also has the disadvantage of breaking compatibility.

It sounds like tackling this problem at the source is the better idea. I think it might be worthwhile to break it up to allow specifying: i) earthly flag values, e.g. `EARTHLY_ALLOW_PRIVILEGED=true`, `EARTHLY_VOLUME_NAME=my-vol-name`, ii) build args, and iii) secrets.

## Option 1

Keep a single `.env` file and add section headers, for example:

```
# EARTHLY_FLAGS
EARTHLY_ALLOW_PRIVILEGED=true

# ARGS
MY_ARG="this is not a secret, and can be used in layer hashing"

# SECRETS
MY_TOKEN="keep-this-secret"
```

Pro:
 - easier for users to migrate from existing free-for-all by simply adding headers to an existing file
Con:
 - makes it impossible for users to check in ARGS or FLAGS into a git repo, as it's an all or nothing approach
 - Unclear what to do with values that are stored under a different header
 - abuses the use of `#` in terms of commenting, perhaps we would want a different character, `@`, `%`, etc.

## Option 2

Split these into 3 separate files:
 - `flag.env`
 - `args.env`
 - `secrets.env`

Pro:
 - individual files can be saved to git independently
Con:
 - when sorted alphabetically, files are not grouped together. (Maybe call them `.earthly.flags`, `.earthly.args`, `.earthly.secrets`? A downside is we no longer have a `.env` suffix.)
 - harder migration from `.env` file (has the similar issue of what to do with values from `.env` that don't fall under a heading).

I am leaning to "Option 2" due to the fact we can support checking in some of these files to git, but not the secrets.

alexcb:
closed in favour of #1811

PR #1802, opened 2022-04-02
`SAVE IMAGE --no-manifest-list`

vladaionescu:
To enable this experimental feature you must use `VERSION --use-no-manifest-list 0.6`.

This PR implements support for `SAVE IMAGE --no-manifest-list`. This option disables creating a multi-platform manifest list for the image, even if the image is created with a non-default platform. This allows the user to create non-native images (e.g. amd64 image on an M1 laptop) that are still compatible with AWS lambda (AWS lambda does not support multi-platform images).

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -208,12 +208,20 @@ func (b *Builder) convertAndBuild(ctx context.Context, target domain.Target, opt
 			res.AddMeta(fmt.Sprintf("%s/final-artifact", refPrefix), []byte("true"))
 		}
 
-		isMultiPlatform := make(map[string]bool) // DockerTag -> bool
+		isMultiPlatform := make(map[string]bool)    // DockerTag -> bool
+		noManifestListImgs := make(map[string]bool) // DockerTag -> bool
 		for _, sts := range mts.All() {
 			if sts.PlatformResolver.Current() != platutil.DefaultPlatform {
 				for _, saveImage := range b.targetPhaseImages(sts) {
 					if saveImage.DockerTag != "" && saveImage.DoSave {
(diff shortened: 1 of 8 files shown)


flupke:
Tested and working perfectly :+1:

builder/builder.go
```diff
@@ -208,12 +208,20 @@ func (b *Builder) convertAndBuild(ctx context.Context, target domain.Target, opt
...
-		isMultiPlatform := make(map[string]bool) // DockerTag -> bool
+		isMultiPlatform := make(map[string]bool)    // DockerTag -> bool
+		noManifestListImgs := make(map[string]bool) // DockerTag -> bool
 		for _, sts := range mts.All() {
 			if sts.PlatformResolver.Current() != platutil.DefaultPlatform {
 				for _, saveImage := range b.targetPhaseImages(sts) {
 					if saveImage.DockerTag != "" && saveImage.DoSave {
-						isMultiPlatform[saveImage.DockerTag] = true
+						if saveImage.NoManifestList {
+							noManifestListImgs[saveImage.DockerTag] = true
```
alexcb, line 218: took me a couple minutes to figure out this works by omitting the entry to `isMultiPlatform`, but makes sense.

alexcb (approved):

builder/builder.go
vladaionescu: Yes, exactly

alexcb:
This feature will solve issues such as 

```
The image manifest or layer media type for the source image XXXXXXXXXXX.dkr.ecr.us-east-1.amazonaws.com/XXXXXXXXXXX:latest is not supported.
```

PR #1806, opened 2022-04-04
CACHE command to not share cached data among targets

### LINKED ISSUE #1805: Inconsistent behaviors between CACHE command and --mount=type=cache
For --mount=type=cache earthly only shares cache mounts between repeated builds of the same target, and only if the build args are the same between invocations.

However, the CACHE command seems to abandon this rule. All targets with the same CACHE commands share cached contents.

Is this expected? IMO it's better to match the behavior of --mount=type=cache.

To reproduce:

```
VERSION --use-cache-command 0.6

FROM alpine

COPY random .

a:
    CACHE /some-cache
    RUN ls /some-cache
    RUN echo before; cat /some-cache/a || true
    RUN date > /some-cache/a
    RUN echo after; cat /some-cache/a || true

b:
    CACHE /some-cache
    RUN ls /some-cache
    RUN echo before; cat /some-cache/b || true
    RUN date > /some-cache/b
    RUN echo after; cat /some-cache/b || true

c:
    BUILD +a
    BUILD +b

```

Run this several times:

```
date > random; earthly +c
```

wxdao:
Resolves #1805 .

### DIFF
--- a/tests/cache-cmd.earth
+++ b/tests/cache-cmd.earth
@@ -48,4 +48,4 @@ test-no-sharing-prev:
 test-no-sharing:
     COPY +test-no-sharing-prev/nothing . # Only to make sure +test-no-sharing-prev runs fitst
     CACHE /no-sharing
-    RUN test -f /no-sharing/should-not-see-me; test $? = 1
+    RUN ! test -f /no-sharing/should-not-see-me
--- a/tests/cache-cmd.earth
+++ b/tests/cache-cmd.earth
@@ -37,3 +37,15 @@ test-save-artifact:
     SAVE ARTIFACT ./my/artifacts AS LOCAL ./artifacts1
     RUN echo "artifact 2" >> ./my/artifacts/2
     SAVE ARTIFACT ./my/artifacts AS LOCAL ./artifacts2
+
+test-no-sharing-prev:
+    CACHE /no-sharing
+    RUN touch /no-sharing/should-not-see-me
+    # A hack to let +test-no-sharing have something to copy.
+    RUN touch nothing
+    SAVE ARTIFACT nothing
+
+test-no-sharing:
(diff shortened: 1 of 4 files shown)


vladaionescu (commented):
This looks good, @wxdao - thank you for the contribution!

Would you be able to extend the integration tests to verify that this bug has been fixed?

The current integration test is here: https://github.com/earthly/earthly/blob/main/tests/Earthfile#L969 and the Eartfhile used for that is this one: https://github.com/earthly/earthly/blob/main/tests/cache-cmd.earth

brandonSc:
Thanks for the fix @wxdao - looks good to me as well. 

I compiled from your fork to try it locally, and everything seems to work as expected.

wxdao:
Good to hear @brandonSc !

Just added a test to verify the fix. Please have a look @vladaionescu .

vladaionescu (approved):

wxdao:
@vladaionescu Looks like my last rebase attempt to address merge conflicts confused GitHub and it won't run the PR tests. Would you please approve them again?

vladaionescu (approved):

brandonSc (approved):

tests/cache-cmd.earth
```diff
@@ -37,3 +37,15 @@ test-save-artifact:
...
+    CACHE /no-sharing
+    RUN touch /no-sharing/should-not-see-me
+    # A hack to let +test-no-sharing have something to copy.
+    RUN touch nothing
+    SAVE ARTIFACT nothing
+
+test-no-sharing:
+    COPY +test-no-sharing-prev/nothing . # Only to make sure +test-no-sharing-prev runs fitst
+    CACHE /no-sharing
+    RUN test -f /no-sharing/should-not-see-me; test $? = 1
```
alexcb, line 51:
```suggestion
    RUN ! test -f /no-sharing/should-not-see-me
```

PR #1810, opened 2022-04-04
chore: fix typo

Dkendal:


alexcb:
That shuold be easier to read :grin: Thanks for the fix.

Issue #1808, opened 2022-04-04
Allow a user to supply a custom `secret_provider` program to lookup secrets


alexcb:
## The current state of `.env`

Earthly currently provides support for an `.env` file which is used to augment environment variables prior to parsing the earthly command line arguments. For example one can run earthly under verbose mode by using `earthly --verbose ...`, `export EARTHLY_VERBOSE=true && earthly ...`, or by storing `EARTHLY_VERBOSE=true` in `.env`.

The `.env` file is also used to storing build args that are automatically provided to build targets. For example consider the target

```Earthfile
foo:
    ARG abc
    RUN echo "abc is '$abc'"
```

One can pass a value for `abc` via `earthly +foo --abc=123`, or by setting `abc=321` in an `.env` file and by running `earthly +foo` without any additional command line arguments.

The `.env` file can also store secrets in it, For example:

```Earthfile
foo:
    RUN --secret password echo "my password is '$password'"
```

which can be passed via `earthly --secret password=keepitsecret +foo` or by storing `password=keepitsecret` in an `.env` and by just running `earthly +foo`.

### Problems

 - The `.env` file doesn't allow users to specify if a key is a) an earthly command line flag, b) a build-arg, or c) a secret. This leads to problems like #1736 
 - Including secrets in a `.env` file means it is not possible for users to commit their `.env` file. (Even though there are mixed-opinions on committing `.env` files to git, earthly should be flexible and allow users to make that decision on their own)
 - (edge-case) It is not possible to use a build-arg and secret which have the same name.
 
#### Edge-case example
 ```Earthfile
foo:
    ARG token
    RUN --secret secrettoken=+secrets/token echo "secrettoken is $secrettoken; token is $token"
```
different values can be supplied with: `earthly --secret token=abc +foo --token=123`; however, only a single key can be supplied in an `.env`.

## Proposal
Create two new files: `.secrets` and `.args`, which will be formatted the same as `.env`. The files will be used to store secrets and build-args respectively.

Two additional cli flags will be created: `--earthlyargs`, and `--earthlysecrets`, which could allow users to specify different files.

To maintain backwards compatibility, earthly will parse the `.env` looking for [command line flags](https://github.com/earthly/earthly/blob/92da4c8a4a49b72d897af1f4b0ad948d0e6d5ec8/cmd/earthly/main.go#L365-L630), if any additional keys are found, earthly will display a warning and allow them to be used as a build-args or secrets.

### Credential Helpers

[...]

PR #1811, opened 2022-04-04
split build-args and secrets into seperate files

alexcb:
.env should only contain earthly cli flag env values; build-args and
secrets should be stored in .earthlyargs and .earthlysecrets
respectively.

### DIFF
--- a/cmd/earthly/main.go
+++ b/cmd/earthly/main.go
@@ -89,8 +89,9 @@ const (
 	defaultEnvFile = ".env"
 	envFileFlag    = "env-file"
 
-	defaultSecretsFile = ".earthlysecrets"
-	defaultArgsFile    = ".earthlyargs"
+	hideEarthlySecretsAndArgs = true
+	defaultSecretsFile        = ".earthlysecrets"
+	defaultArgsFile           = ".earthlyargs"
 )
 
 type earthlyApp struct {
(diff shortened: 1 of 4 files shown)


alexcb:
This implements the #1808 (without support for credential helpers)

cmd/earthly/main.go
```diff
@@ -627,6 +632,20 @@ func newEarthlyApp(ctx context.Context, console conslogging.ConsoleLogger) *eart
 			Value:       defaultEnvFile,
 			Destination: &app.envFile,
 		},
+		&cli.StringFlag{
+			Name:        "secrets-file",
```
vladaionescu, line 636: This is very close to the already existing `--secret-file`. I wonder if we can think of a variation that isn't as close.

cmd/earthly/main.go
```diff
@@ -627,6 +632,20 @@ func newEarthlyApp(ctx context.Context, console conslogging.ConsoleLogger) *eart
...
+			Value:       defaultSecretsFile,
+			Destination: &app.secretsFile,
+		},
+		&cli.StringFlag{
+			Name:        "args-file",
+			EnvVars:     []string{"EARTHLY_ARGS_FILE"},
+			Usage:       "Use build args from this file",
+			Value:       defaultArgsFile,
+			Destination: &app.argsFile,
+		},
```
vladaionescu, line 648: Should these be hidden & experimental until Earthly v0.7?

cmd/earthly/main.go
```diff
@@ -2831,15 +2850,40 @@ func (app *earthlyApp) actionBuildImp(c *cli.Context, flagArgs, nonFlagArgs []st
...
+				return errors.Errorf("secret %s found in both %s and %s", key, app.envFile, app.secretsFile)
+			}
+			argsMap[key] = val
+			secretsStringMap[key] = val
 		}
 	}
+	if len(nonEarthlyEnvs) > 0 {
+		sort.Strings(nonEarthlyEnvs)
+		keys := strings.Join(nonEarthlyEnvs, ", ")
+		app.console.Warnf("Warning: build-args or secrets (%s) should be defined in .earthlyargs or .earthlysecrets; this will become mandatory in a future version of earthly.", keys)
```
vladaionescu, line 2883: If we decide to GA this later, we'd need to take out this message and re-add it later (e.g. in v0.7).

cmd/earthly/main.go
alexcb: that's a real good point. perhaps renaming them to `--earthlysecrets` (and `--earthlyargs`) would be more clear?

cmd/earthly/main.go
alexcb: Possibly? it might make sense since introducing the warning in a patch version does seem a bit odd. I'll hide them (and the warning too).

cmd/earthly/main.go
alexcb: I added a `hideEarthlySecretsAndArgs = true` const to tie them all together.

cmd/earthly/main.go
vladaionescu: Yeah, that makes sense I think.. although it's less consistent with kebab case being used in other args. But I think that it's better this way.

vladaionescu (approved):

PR #1814, opened 2022-04-05
docs: .earthlyargs and .earthlysecrets

alexcb:


vladaionescu:
I believe this will go through some more changing, so I've removed myself as a reviewer for now to avoid the notifications (feel free to re-add me later).

PR #1815, opened 2022-04-05
Better file missing error message

adamgordonbell:
I might need some pointers on this. I am trying to improve an error message.

Old Error message for the file does not exist:
```
Error: build target: build main: failed to solve: failed to compute cache key: failed to calculate checksum of ref 4z7jcxhngtwp6n799l897lhff::54sya1iypr5uovhhi03tce12r: "/bla": not found
```

New error:

```
Error: File not found "/bla"
```

The actual error for `file not found` is in checksum.go and private so the only way I could figure to get the context out was via text matching. Is there a better way?  Ideally, I'd like to actually get the line number of the Earthfile out as well, so the error would be:
```
Earthfile:58  "/bla" not found`
```

### DIFF
--- a/cmd/earthly/main.go
+++ b/cmd/earthly/main.go
@@ -1566,6 +1566,10 @@ func (app *earthlyApp) run(ctx context.Context, args []string) int {
 				"Check your git auth settings.\n" +
 					"Did you ssh-add today? Need to configure ~/.earthly/config.yml?\n" +
 					"For more information see https://docs.earthly.dev/guides/auth\n")
+		} else if strings.Contains(err.Error(), "failed to compute cache key") && strings.Contains(err.Error(), ": not found") {
+			var re = regexp.MustCompile(`".*?"`)
+			var path = re.FindString(err.Error())
+			app.console.Warnf("Error: File not found %v\n", path)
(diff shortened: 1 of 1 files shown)


cmd/earthly/main.go
```diff
@@ -1566,6 +1566,10 @@ func (app *earthlyApp) run(ctx context.Context, args []string) int {
 				"Check your git auth settings.\n" +
 					"Did you ssh-add today? Need to configure ~/.earthly/config.yml?\n" +
 					"For more information see https://docs.earthly.dev/guides/auth\n")
+		} else if strings.Contains(err.Error(), "failed to compute cache key") && strings.Contains(err.Error(), ": not found") {
+			var re = regexp.MustCompile(`".*?"`)
```
alexcb, line 1570: This regex feels a bit fragile. Would something like https://go.dev/play/p/Y57YqC0U-0B be better?

adamgordonbell:
Added ticket for line number idea #1816

cmd/earthly/main.go
adamgordonbell: Addressed. Now logging full original error if regex doesn't match.

alexcb (approved):

PR #1820, opened 2022-04-06
refactor secretprovider to use chain of stores

alexcb:
This refactors the secretprovider to accept an arbitraty chain of stores
which could potentially return the secret. This will make it easier to
insert new secret stores in the future.

### DIFF
--- a/util/llbutil/secretprovider/secrets.go
+++ b/util/llbutil/secretprovider/secrets.go
@@ -2,22 +2,16 @@ package secretprovider
 
 import (
 	"context"
-	"fmt"
 	"strings"
 
-	"github.com/earthly/earthly/cloud"
 	"github.com/moby/buildkit/session"
 	"github.com/moby/buildkit/session/secrets"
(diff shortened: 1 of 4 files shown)


util/llbutil/secretprovider/secrets.go
```diff
@@ -2,76 +2,58 @@ package secretprovider
...
 
-	return &secrets.GetSecretResponse{
-		Data: dt,
-	}, nil
+	return nil, errors.WithStack(errors.Wrapf(secrets.ErrNotFound, "unable to lookup secret %s", secretName))
 }
 
 // New returns a new secrets provider
-func New(client cloud.Client, overrides map[string][]byte) session.Attachable {
+func New(stores ...secrets.SecretStore) session.Attachable {
```
vladaionescu, line 55: I would perhaps add more context in the comment about which store takes precedence over another.

vladaionescu (approved):

Issue #1817, opened 2022-04-06
Add `--chmod` and `--chown` COPY flags 


dszakallas:
BuildKit has a `--chmod` and `--chown` COPY flag for changing file permissions and owner without adding a new a layer. See https://github.com/moby/moby/issues/34819.

Could we have these in Earthly?

adamgordonbell:
PR raised for this here: https://github.com/earthly/earthly/pull/1821

vladaionescu:
Note that `--chown` has already been implemented.

PR #1821, opened 2022-04-06
Copy Chmod

adamgordonbell:
Adding `--chmod=` to COPY command.

Fixes https://github.com/earthly/earthly/issues/1817


earthfile2llb/interpreter.go
```diff
@@ -716,6 +717,15 @@ func (i *Interpreter) handleCopy(ctx context.Context, cmd spec.Command) error {
 	if err != nil {
 		return i.wrapError(err, cmd.SourceLocation, "failed to expand COPY chown: %v", opts.Chown)
 	}
+	expandedMode, err := i.expandArgs(ctx, opts.Chmod, false, false)
+	if err != nil {
+		return i.wrapError(err, cmd.SourceLocation, "failed to expand COPY chmod: %v", opts.Platform)
+	}
+	mode, err := strconv.ParseUint(expandedMode, 10, 32)
```
vladaionescu, line 724: I think we need base 8 here?

adamgordonbell: yep, you were right.

vladaionescu (approved):
Looks good - but the tests are failing due to a new-line ending or something. You can run `earthly +lint` locally.

vladaionescu (commented):
Oh one more thing.. this is a new feature, we should enable it with a feature flag only for now.

You'll need to add a new feature flag here: https://github.com/earthly/earthly/blob/main/features/features.go#L37. (e.g. `use-chmod`) and wire it up correctly. This will only enable the feature on `VERSION --use-chmod 0.6` (and in the future in `VERSION 0.7`).

vladaionescu:
Oh and an entry to the changelog would be much appreciated! We usually label everything that we add as "experimental" to start with.

tests/copy.earth
```diff
@@ -192,3 +193,24 @@ copy-dir-containing-invalid-symlink:
     RUN mkdir /data
     RUN echo hi > /data/hello.txt
     RUN cat /symlinks/symlink | grep -w hi
+
+copy-chmod:
+    RUN echo "no chmod - -rw-r-r-"
+    COPY in/root .
+    RUN echo "./root file-mode=$(stat -c %a ./root)"
+    RUN if [[ "$(stat -c %a ./root)" != "644" ]]; then exit 1; fi
```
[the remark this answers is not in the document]
adamgordonbell: way better ...

alexcb (approved):

PR #1822, opened 2022-04-06
secret provider

alexcb:
allows users to write a custom secret provider


alexcb:
This is a rough work in progress.

One can now supply a command that is called to provide a secret: 

```yaml
global:
    secret_provider: earthly-secret-provider
```

In this case `earthly-secret-provider` is on my `PATH` (under `~/bin/earthly-secret-provider`), 

```
#!/bin/bash
set -e
echo >&2 "earthly-secret-provider called for $1; pwd=$(pwd)"
env >&2
if [ "$1" = "password" ]; then
    echo -n abc123
    exit 0
fi
echo >&2 "no secret found for $1"
exit 2 # continue searching under .env for old secrets
```

which will print the following debug information to stderr:

```
earthly-secret-provider called for password; pwd=/home/alex/gh/earthly/earthly
PWD=/home/alex/gh/earthly/earthly
EARTHLY_PROJECT_GIT_PATH=
SHLVL=1
EARTHLY_PROJECT_LOCAL_PATH=/home/alex/broken/secrets-example
PATH=/home/alex/.local/bin:/home/alex/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:/usr/local/go/bin:/home/alex/.fzf/bin
_=/usr/bin/env
```

So when we run:

```Earthfile
VERSION 0.6
FROM alpine:latest

test:
  RUN --no-cache --secret password echo my password is $password.
```

We will get:

```
/h/a/b/secrets-example+test | my password is abc123.
```

The custom secret-provider script can look at `EARTHLY_PROJECT_LOCAL_PATH=/home/alex/broken/secrets-example` or `EARTHLY_PROJECT_GIT_PATH=` in addition to `$1` (for the `--secret` key ID).

util/llbutil/secretprovider/meta.go
```diff
@@ -0,0 +1,29 @@
...
+	localProject  string
+	remoteProject string
+}
+
+var secretIDs map[string]secretID
+
+// NewSecretID is used to store secretID metadata and returns a uuid
+// which can be used to lookup this metadata
+// This idirection is required due to earthly not being able to (easily) pass this data through buildkit
+func NewSecretID(name, localProject, remoteProject string) string {
```
alexcb, line 18: @vladaionescu any thoughts on this approach -- creating a UUID which is passed to buildkit, then using that UUID to get back to the secretID metadata which contains both the secret name, and the local or remote project path.

vladaionescu: Perhaps I wouldn't include the local project as part of the API here. I would only include the following:
* Detect the Earthly "project" - we don't have this notion today though so this can be left empty or not included for now
* Get the secret ID that the user provides (e.g. from `RUN --secret FOO=+secrets/<my-secret-id>`)

Using these two, we can create an ID that we use with Buildkit (this could be simply "<project>/<my-secret-id>", rather than a UUID). This constructed ID will end up being consistent across runs, in case caching is an issue. But even if it isn't, I think it's simpler rather than doing the UUID trick.

alexcb: > we don't have this notion today though so this can be left empty or not included for now

I don't see how this could work for a user who has two different projects such as:

```
~/project-a/Earthfile
RUN --secret FOO=+secrets/<my-secret-id>

~/project-a/.env
my-secret-id=password-a

~/project-b/Earthfile
RUN --secret FOO=+secrets/<my-secret-id>

~/project-b/.env
my-secret-id=password-b
```

Currently, for a user to work on project-a, one must: `cd ~/project-a && earthly +target` (or alternatively run `earthly --env-file=~/project-a/.env +target`).

Without having the `project` feature available, a user looses the ability to supply different passwords depending on the location of the Earthfile.

> This constructed ID will end up being consistent across runs, in case caching is an issue

Good catch on this point; unfortunately we're hitting this. Supplying a hash instead of UUID half-fixes the issue, but if a user moves the local path of a project, that will bust the cache :disappointed: 

Maybe we'll have to enforce global-level secrets via the custom secret provider until we complete the project name feature.

alexcb: UPDATE: we'll eventually provide a `$EARHTLY_PROJECT_NAME` as an environment variable to the user's `secret_provider` script.

util/llbutil/secretprovider/cmd.go
```diff
@@ -0,0 +1,55 @@
...
+// GetSecret gets a secret from the map store
+func (c *cmdStore) GetSecret(ctx context.Context, id string) ([]byte, error) {
+	if c.cmd == "" {
+		return nil, secrets.ErrNotFound
+	}
+	if id == common.DebuggerSettingsSecretsKey {
+		// the interactive debugger passes config values by abusing secrets,
+		// we must not call the user's secret provider in this case.
+		return nil, secrets.ErrNotFound
+	}
```
vladaionescu, line 35: Good catch :D

vladaionescu (approved):

PR #1825, opened 2022-04-07
docs for secret_provider

alexcb:

### DIFF
--- a/docs/earthly-config/earthly-config.md
+++ b/docs/earthly-config/earthly-config.md
@@ -63,6 +63,53 @@ Specifies the total size of the BuildKit cache, in MB. The BuildKit daemon uses 
 Specifies the total size of the BuildKit cache, as a percentage (0-100) of the total filesystem size.
 When used in combination with `cache_size_mb`, the lesser of the two values will be used. This limit is ignored when set to 0.
 
+### secret_provider (experimental)
+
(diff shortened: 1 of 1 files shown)


docs/earthly-config/earthly-config.md
```diff
@@ -63,6 +63,53 @@ Specifies the total size of the BuildKit cache, in MB. The BuildKit daemon uses
...
+
+{% hint style='info' %}
+##### Note
+
+All stdout data will be used as the secret value, including whitespace (and newlines).
+You may want to use `echo -n` to prevent returning a newline.
+
+Any data sent to stderr will be displayed on the earthly console, this makes it possible
+to insert commands such as `echo >&2 "here is some debug text"` without affecting the contents
+of the secret.
```
vladaionescu, line 110: Maybe this should only be displayed on `--verbose`? Helps keep the Earthly output clean.

vladaionescu (approved):

docs/earthly-config/earthly-config.md
alexcb: interesting idea, right now we just set `cmd.Stderr = os.Stderr` in all cases, we could put a `if verbose` around that. Might seem a bit too magical for a user that's trying to debug their custom secret script. I suspect if it's too noisy a user could just comment out their debug code. I'm inclined to leave it as-is and wait for a user to request it.

PR #1839, opened 2022-04-13
COPY into CACHE doesn't work

alexcb:
Currently the following code is accepted by earthly:

    target:
        CACHE /my-cache
        COPY local-file /my-cache/

However it doesn't correctly copy the file into the cache.
We will fix it in a follow up PR when time permits; however,
for the time-being it's better to explicitly fail if such a COPY
is attempted.

### DIFF
--- a/earthfile2llb/converter.go
+++ b/earthfile2llb/converter.go
@@ -490,6 +490,13 @@ func (c *Converter) CopyClassical(ctx context.Context, srcs []string, dest strin
 		srcState = c.buildContextFactory.Construct()
 	}
 
+	// TODO add support for caches
+	for path := range c.persistentCacheDirs {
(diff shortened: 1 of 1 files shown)


alexcb:
5d6fb3bbc65c527ffc341258d91df609beb1c355 contains a potential fix; however the test for it is not working, and I need to pivot to other tasks at the moment.

vladaionescu (commented):
This will catch only some cases. For example the following case would not be caught:

* mount is `/some/mount`
* COPY dest is `/some`
* COPY src includes `./mount` among other things (effectively causing a merge of the mounted dir with the copied data)

alexcb:
since this doesn't have any tests, nor does it reference an issue (and nobody has any :+1: on it), we'll close this since it's imcomplete.

PR #1841, opened 2022-04-13
unmask cgroups for podman

alexcb:
podman masks off `/sys/fs/cgroup` as read-only by default, which prevents earthly from managing cgroup v2 limits.

### DIFF
--- a/util/containerutil/podman.go
+++ b/util/containerutil/podman.go
@@ -25,7 +25,7 @@ func NewPodmanShellFrontend(ctx context.Context, cfg *FrontendConfig) (Container
 	fe := &podmanShellFrontend{
 		shellFrontend: &shellFrontend{
 			binaryName:              "podman",
-			runCompatibilityArgs:    make([]string, 0),
+			runCompatibilityArgs:    []string{"--security-opt", "unmask=/sys/fs/cgroup"},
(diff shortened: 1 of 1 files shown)


util/containerutil/podman.go
```diff
@@ -25,7 +25,7 @@ func NewPodmanShellFrontend(ctx context.Context, cfg *FrontendConfig) (Container
 	fe := &podmanShellFrontend{
 		shellFrontend: &shellFrontend{
 			binaryName:              "podman",
-			runCompatibilityArgs:    make([]string, 0),
```
dchw, line 28: Does this work or become a noop on Cgroups v1? Does this work on rootless (which is default `podman` setup)? Since its hardcoded for all podman run invocations I would also check if this is valid for older podman versions.

Also can we add a comment for why we need this to the argument here?

dchw (approved):

util/containerutil/podman.go
alexcb: under cgroups v1, the following works fine:

```
podman run --security-opt unmask=/sys/fs/cgroup alpine /bin/sh -c 'mkdir /sys/fs/cgroup/testpodman2; echo $?'
```

(returns `0`).

Plus earthly should detect cgroups v1, so it won't even attempt the mkdir.

PR #1192, opened 2021-08-26
Reference the setup-earthly github actions

alexcb:
Update the GHA example to include an example of using the
earthly/actions/setup-earthly action.

Signed-off-by: Alex Couture-Beil <alex@earthly.dev>

PR #1845, opened 2022-04-18
Update example workflow in GitHub Actions integration guide

mattt:
This PR updates the [GitHub Actions guide](https://docs.earthly.dev/ci-integration/vendor-specific-guides/gh-actions-integration) with the following changes:

- Use [earthly/actions-setup](https://github.com/earthly/actions-setup) instead of installing Earthly manually (thanks for this, by the way — this is a big quality of life improvement)
- Remove `earthly --version` step
- Use latest version (v3) of [actions/checkout](https://github.com/actions/checkout)
- Some more opinionated changes that I can back out of if you prefer:
  - Don't set Docker Hub credentials as environment variables
  - Pass Docker Hub token through standard input

### DIFF
--- a/docs/ci-integration/guides/gh-actions-integration.md
+++ b/docs/ci-integration/guides/gh-actions-integration.md
@@ -20,7 +20,7 @@ jobs:
     env:
       FORCE_COLOR: 1
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
     - name: Put back the git branch into git (Earthly uses it for tagging)
       run: |
         branch=""
--- a/docs/ci-integration/guides/gh-actions-integration.md
+++ b/docs/ci-integration/guides/gh-actions-integration.md
@@ -34,8 +34,10 @@ jobs:
         git checkout -b "$branch" || true
     - name: Docker Login
       run: docker login --username "$DOCKERHUB_USERNAME" --password "$DOCKERHUB_TOKEN"
-    - name: Download latest earthly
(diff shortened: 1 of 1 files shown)


docs/ci-integration/guides/gh-actions-integration.md
```diff
@@ -18,11 +18,9 @@ jobs:
   build:
     runs-on: ubuntu-latest
     env:
-      DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
-      DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
       FORCE_COLOR: 1
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
     - name: Put back the git branch into git (Earthly uses it for tagging)
```
mattt, line 24: Could we instead do this with the [actions/checkout](https://github.com/actions/checkout#checkout-a-different-branch)?

alexcb:
We've had #1192 open for quite some time, and we've been dogfooding it in our own tests (e.g. https://github.com/earthly/earthly/blob/main/.github/workflows/ci.yml#L20 )

However, we also received https://github.com/earthly/actions-setup/pull/2 (which initially slipped through the cracks); which we're leaning towards merging (after a few adjustments). Once that's merged in I would like to dogfood that for a few days first, then we could ultimately merge this PR to point our users to it as the recommended way to use earthly with GHA.

> Don't set Docker Hub credentials as environment variables

This seems like a great change :tada:

docs/ci-integration/guides/gh-actions-integration.md
alexcb: definitely, I updated that in #1856

PR #1850, opened 2022-04-21
Cleaner checkout in Github Action integration example

basilenouvellet:


alexcb (approved):
I wasn't familiar `ref` did this, but it looks like it does https://github.com/actions/checkout#checkout-a-different-branch

alexcb:
I just opened https://github.com/earthly/earthly/pull/1851 to make this change in our own tests, if everything goes well we'll merge this PR to update the docs. Thanks for pointing this out.

alexcb:
It turns out there's a slight different between using `ref` and the current `git checkout -b` version, I've opened #1853 to explain the reasoning.

basilenouvellet:
understood, thanks for the explanation

PR #1851, opened 2022-04-22
use checkout actions ref

alexcb:
It is possible to instruct GHA to checkout a different branch:
https://github.com/actions/checkout#checkout-a-different-branch

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -715,22 +715,17 @@ jobs:
     steps:
       - uses: earthly/actions-setup@v1.0.1
       - uses: actions/checkout@v2
-      -
-        name: Set up QEMU
+        with:
+          # By default, only the latest commit is checked out.
+          # Earthly uses the branch name for tagging,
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -715,22 +715,17 @@ jobs:
     steps:
       - uses: earthly/actions-setup@v1.0.1
       - uses: actions/checkout@v2
-      -
-        name: Set up QEMU
+        with:
+          # By default, only the latest commit is checked out.
+          # Earthly uses the branch name for tagging,
+          # so we need to explicitly check out the branch.
+          ref: ${{ github.head_ref || github.ref_name }}
```
vladaionescu, line 722: The only case where this is slightly not accurate is if someone wants to re-run the CI on an older commit of that branch. The code we had before would still respect the commit hash, but it would reinsert the branch as pointing to that (possibly older) hash.

alexcb: this makes sense; here's a docs update to mention this: #1853

PR #1858, opened 2022-04-27
Use embedded Docker registry for WITH DOCKER commands.

mikejholly:
Makes use of the custom BuildKit embedded Docker registry to improve the performance of `WITH DOCKER ... RUN`. Incidentally, this also eliminates unnecessary round-trips of image tars back to the client in the case of a remote buildkit.

Meant to address #1268.

* Adds a new feature flag called `--use-registry-for-with-docker` which can be used to trigger the feature.
* Supports both Tar file and registry approaches (Tar can be deprecated and removed later).
* Moves image solver code out into separate structs in an effort to better encapsulate that code.
* If enabled, Earthly asks BuidKit to push the image to the local registry, then waits for the push to complete.
* Once complete, the temporary image names are made available to the `WITH DOCKER` context and retagged.
* Some complexity was required because of how the images are meant to be short-lived (noted in the below comments).

Example output:
image


earthfile2llb/withdockerrun.go
```diff
@@ -164,13 +176,51 @@ func (wdr *withDockerRun) Run(ctx context.Context, args []string, opt WithDocker
...
+				l.closer()
+			}
+		}()
+		// Sort the local registry pull list, to make the operation consistent.
+		sort.Slice(wdr.localRegLoads, func(i, j int) bool {
+			return wdr.localRegLoads[i].imgName < wdr.localRegLoads[j].imgName
+		})
+		for _, l := range wdr.localRegLoads {
+			errChan := l.errChan
+			wdr.c.opt.ErrorGroup.Go(func() error {
```
mikejholly, line 196: I'm using the preexisting `ErrorGroup` here since it's already handling output. These errors don't need to be acted on in this context since a failed build will force a failure for the overall command. But there may be a better way to handle and/or present this error.

mikejholly, line 196: I'm using the preexisting `ErrorGroup` here since it's already handling output. These errors don't need to be acted on in this context since a failed build will force a failure for the overall command. But there may be a better way to handle and/or present this error.

buildkitd/dockerd-wrapper.sh
```diff
@@ -159,27 +161,39 @@ stop_dockerd() {
...
         echo "...done"
     fi
 }
 
+load_registry_images() {
+    if [ -n "$EARTHLY_DOCKER_LOAD_REGISTRY" ]; then
+        echo "Loading registry images..."
+        for img in $EARTHLY_DOCKER_LOAD_REGISTRY; do
+            user_tag=$(echo -n "$img" | cut -d'/' -f2)
+            with_reg="172.30.0.1:8371/$img"
```
mikejholly, line 179: Should we use a variable for this IP?

mikejholly, line 179: Should we use a variable for this IP?

vladaionescu (commented):
This is looking really good - I only have pretty minor comments.

Here's an additional possible TODO list:

* Make the solve more efficient - now that we use the Build API and the earthly exporter, we can output multiple images as part of a single build (we just need to use a different image index for each) - so we could have a build per each entire `WITH DOCKER`, and not a build per each individual load. This could be done in a separate PR though.
* We should figure out that `context.TODO()` issue - though that could be a separate PR too, given that this is feature-flagged.
* Add some tests. The easiest thing could be to take some of the existing `WITH DOCKER` tests and also run them with the feature flag turned on. That should allow us to test both the tar version and the local registry version. See `tests/with-docker*` directories.
* Add an entry in the CHANGELOG under `Unreleased`

builder/image_solver.go
```diff
@@ -0,0 +1,288 @@
...
+		res.AddMeta(fmt.Sprintf("%s/%s", refPrefix, exptypes.ExporterImageConfigKey), imgJSON)
+		res.AddMeta(fmt.Sprintf("%s/image.name", refPrefix), []byte(dockerTag))
+		res.AddMeta(fmt.Sprintf("%s/image-index", refPrefix), []byte("0"))
+
+		return res, nil
+	}
+
+	// TODO: figure out why the below context is being canceled.
+	go func() {
+		_, err := s.bkClient.Build(context.TODO(), *solveOpt, "", bf, nil)
```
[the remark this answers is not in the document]
mikejholly: Good question. I'll take a look.

states/builderfun.go
```diff
@@ -4,5 +4,14 @@ import (
...
+// DockerTarImageSolver can create a Docker image and make it available as a tar
+// file.
+type DockerTarImageSolver interface {
+	SolveImage(ctx context.Context, mts *MultiTarget, dockerTag string, outFile string, printOutput bool) error
+}
+
+// DockerImageSolver can create a Docker image for the WITH DOCKER command using
+// the embedded BuildKit registry.
+type DockerImageSolver interface {
+	SolveImage(ctx context.Context, mts *MultiTarget, dockerTag string) (chan string, func(), chan error, error)
```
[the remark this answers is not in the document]
mikejholly: I added a new type to help limit the return values and clarify.

mikejholly:
@vladaionescu: 

> Make the solve more efficient - now that we use the Build API and the earthly exporter, we can output multiple images as part of a single build (we just need to use a different image index for each) - so we could have a build per each entire WITH DOCKER, and not a build per each individual load. This could be done in a separate PR though.

Let's create another issue/PR for that if that's ok. I think this change resolves some trouble for users even though there's more room for improvement.

> We should figure out that context.TODO() issue - though that could be a separate PR too, given that this is feature-flagged.

Yep, I'll look at that now.

> Add some tests. The easiest thing could be to take some of the existing WITH DOCKER tests and also run them with the feature flag turned on. That should allow us to test both the tar version and the local registry version. See tests/with-docker* directories.

Sounds good.

> Add an entry in the CHANGELOG under Unreleased

Will do!

tests/with-docker-registry/Earthfile
```diff
@@ -15,12 +18,61 @@ a-test-image:
...
+        --INDEX=5
+
+docker-load-multi-test:
+    WITH DOCKER \
+        --load=(+another-test-image --INDEX=1) \
+        --load=(+another-test-image --INDEX=2) \
+        --load=(+another-test-image --INDEX=3) \
+        --load=(+another-test-image --INDEX=4) \
+        --load=(+another-test-image --INDEX=5)
+        RUN docker run --rm another-test-img:i1 && \
```
alexcb, line 73: Should we extend this to `docker run --rm another-test-img:i1 | grep "hello another test img 1"` to validate the correct image is getting run? which could catch cases where either the entrypoing config is missed, or the `--index=*` entries are somehow clobbered?

alexcb:
Looking good, typically we would add `,use-registry-for-with-docker` to the `.earthly_version_flag_overrides` file; however, this would cause the existing `./tests/with-docker` tests to run with the new feature (and cause us to loose code-coverage of the old feature.

I was going to suggest updating `with-docker/Earthfile` to include `VERSION --use-registry-for-with-docker=false 0.6`; however according to https://github.com/earthly/earthly/blob/1e36ff5a12bd3b2c6a9b748bb4bef80b66333aa1/buildcontext/parsefeatures.go#L29 that would have no effect and the overridden feature would trump it.

As a result, I think we could address that in a follow up PR, where we allow an explicit `=false` to take final precedence.

Alternatively we could add the override, then create a new test under https://github.com/earthly/earthly/blob/main/.github/workflows/ci.yml which removes the override via `sed` before explicitly running the `./tests/with-docker+all` target.

tests/with-docker-registry/Earthfile
mikejholly: Sure sounds good.

builder/image_solver.go
```diff
@@ -0,0 +1,288 @@
...
+
+type localRegistryImageSolver struct {
+	bkClient        *client.Client
+	sm              *outmon.SolverMonitor
+	attachables     []session.Attachable
+	enttlmnts       []entitlements.Entitlement
+	cacheImports    *states.CacheImports
+	cacheExport     string
+	maxCacheExport  string
+	saveInlineCache bool
```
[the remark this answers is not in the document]
mikejholly: Good to know. Removed.

alexcb (approved):
nice work!

PR #1861, opened 2022-04-29
Add config options, headers

dchw:


cloud/client.go
```diff
@@ -1382,3 +1383,7 @@ func (c *client) UploadLog(pathOnDisk string) (string, error) {
 
 	return fmt.Sprintf(uploadBundleResponse.ViewURL), nil
 }
+
+func (c *client) GetAuthToken() string {
+	return c.authToken
```
alexcb, line 1388: should this force a refresh if it's expired?

alexcb (commented):
looks good, just one minor question.

cloud/client.go
dchw: We only ever call it right after newing up the client, which IIRC refreshes the token if needed.

cloud/client.go
```diff
@@ -1385,5 +1385,6 @@ func (c *client) UploadLog(pathOnDisk string) (string, error) {
 }
 
 func (c *client) GetAuthToken() string {
+	c.Authenticate() // Ensure the current token is valid
```
dchw, line 1388: This will get a new token and write it to disk if needed.

alexcb: Does this need error handling?

alexcb (approved):

PR #1868, opened 2022-05-04
Fixed typo: tagets => targets

arialdomartini:


brandonSc (approved):
Thanks for the fix 🙌

alexcb:
@arialdomartini even though it's only a typo fix, we would appreciate it if you agree to the CLA. Without that, this PR will cause our CLAassisant to raise an alert, which leads to extra admin work on our end. Thanks!

PR #1876, opened 2022-05-06
More efficient WITH DOCKER image solving

mikejholly:
This PR expands on #1858 by combining all `WITH DOCKER` image builds into a single BuildKit build operation. The code paths for the different image-solving methods (tar and registry) were becoming rather convoluted so I decided to split the code into 2 separate structs with some shared elements. There is some minor duplication, but I figured it would be fine for now as we plan to remove the tar solver in the future anyhow. One thing to point out is that the new solving option doesn't respect the `--parallel-load` feature as it seems less applicable given the now single build operation.

### DIFF
--- a/builder/image_solver.go
+++ b/builder/image_solver.go
@@ -208,31 +208,19 @@ func (m *multiImageSolver) SolveImages(ctx context.Context, imageDefs []*states.
 	buildFn := func(childCtx context.Context, gwClient gwclient.Client) (*gwclient.Result, error) {
 		res := gwclient.NewResult()
 
+		eg, childCtx := errgroup.WithContext(childCtx)
+
 		for i, imageDef := range imageDefs {
-			var saveImage = imageDef.MTS.Final.LastSaveImage()
-			imgJSON, err := json.Marshal(saveImage.Image)
-			if err != nil {
-				return nil, errors.Wrap(err, "image json marshal")
-			}
+			idx := i
+			def := imageDef
+			eg.Go(func() error {
+				return m.addRefToResult(childCtx, gwClient, res, def, idx)
+			})
+		}
 
-			if !strings.Contains(imageDef.ImageName, ":") {
-				imageDef.ImageName += ":latest"
-			}
-
-			ref, err := llbutil.StateToRef(childCtx, gwClient, saveImage.State, true, imageDef.MTS.Final.PlatformResolver, m.cacheImports.AsMap())
-			if err != nil {
-				return nil, errors.Wrap(err, "initial state to ref conversion")
-			}
-
-			refKey := fmt.Sprintf("image-%s", imageDef.ImageName)
-			refPrefix := fmt.Sprintf("ref/%s", refKey)
-			res.AddRef(refKey, ref)
-
-			localRegPullID := fmt.Sprintf("sess-%s/%s", gwClient.BuildOpts().SessionID, imageDef.ImageName)
-			res.AddMeta(fmt.Sprintf("%s/export-image-local-registry", refPrefix), []byte(localRegPullID))
-			res.AddMeta(fmt.Sprintf("%s/%s", refPrefix, exptypes.ExporterImageConfigKey), imgJSON)
-			res.AddMeta(fmt.Sprintf("%s/image.name", refPrefix), []byte(imageDef.ImageName))
-			res.AddMeta(fmt.Sprintf("%s/image-index", refPrefix), []byte(strconv.Itoa(i)))
+		err := eg.Wait()
+		if err != nil {
+			return nil, err
 		}
 
 		return res, nil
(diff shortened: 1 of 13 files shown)


builder/image_solver.go
```diff
@@ -227,33 +205,35 @@ func (s *localRegistryImageSolver) SolveImage(ctx context.Context, mts *states.M
...
-		ref, err := llbutil.StateToRef(childCtx, gwClient, saveImage.State, true, mts.Final.PlatformResolver, s.cacheImports.AsMap())
-		if err != nil {
-			return nil, errors.Wrap(err, "initial state to ref conversion")
-		}
+			if !strings.Contains(imageDef.ImageName, ":") {
+				imageDef.ImageName += ":latest"
+			}
 
-		res := gwclient.NewResult()
+			ref, err := llbutil.StateToRef(childCtx, gwClient, saveImage.State, true, imageDef.MTS.Final.PlatformResolver, m.cacheImports.AsMap())
```
alexcb, line 222: Would calling this in parallel (for each `imageDef`) speed things up?

alexcb: actually... probably not, as we're not forcing a solve via `ref.ReadFile/Dir/Stat` (since we just return the ref via https://github.com/earthly/earthly/blob/71ea6dcca8b4590b7ee4338f78a8c2646ccc3e49/util/llbutil/statetoref.go#L44 )

ignore this :)

alexcb (approved):

builder/image_solver.go
mikejholly: Hmm. I suppose it still does a `Solve` request for each image in `StateToRef`. I can try making this concurrent to see if there's a noticeable performance change.

mikejholly: It doesn't seem to account for much (~300ms) but it doesn't seem to be hurting either.

alexcb: the initial call to `Solve` just gets it queued up in buildkit's solver/scheduler, and the returned ref will point to a state that might not be ready (solved) yet. so it makes sense that it's only 300ms (since it's just a tcp-round trip to kick off the build), it's not until we try to read from the ref which is when things actually block. but if it speeds it up and doesn't break, sounds good :)

earthfile2llb/withdockerrun.go
```diff
@@ -525,48 +389,337 @@ func (wdr *withDockerRun) solveImageAsTar(ctx context.Context, mts *states.Multi
...
+
+func platformIncompatMsg(platr *platutil.Resolver) string {
+	currentPlatStr := platr.Materialize(platr.Current()).String()
+	nativePlatStr := platr.Materialize(platutil.NativePlatform).String()
+	return "running WITH DOCKER as a non-native CPU architecture. This is not supported.\n" +
+		fmt.Sprintf("Current platform: %s\n", currentPlatStr) +
+		fmt.Sprintf("Native platform of the worker: %s\n", nativePlatStr) +
+		"Try using\n\n\tFROM --platform=native earthly/dind:alpine\n\ninstead.\n" +
+		"You may still --load and --pull images of a different platform.\n"
+}
```
vladaionescu, line 574: Nit: This function should probably sit lower in the file since it's common.

earthfile2llb/withdockerrun.go
```diff
@@ -62,17 +62,14 @@ type WithDockerOpt struct {
 	ComposeServices []string
 }
 
-type withDockerRun struct {
+type withDockerRunTar struct {
```
vladaionescu, line 65: This file is a bit hard to follow due to the two very similar, but distinct implementations in here. Mind splitting this into two distinct files?

vladaionescu (changes requested):
I believe that the new implementation drops some important parallelism unintentionally. The parallel load flag is still relevant in the new world, as it should control whether the conversion process should take place in parallel or not (this takes place before the solving). This is necessary, to keep the build fast even before the solving starts executing.

In summary, we need to keep this kind of logic in the new implementation too:

```golang
if wdr.enableParallel {
		err = wdr.c.BuildAsync(ctx, depTarget.String(), opt.Platform, opt.AllowPrivileged, opt.BuildArgs, loadCmd, afterFun, wdr.sem)
		if err != nil {
			return nil, err
		}
	} else {
		mts, err := wdr.c.buildTarget(ctx, depTarget.String(), opt.Platform, opt.AllowPrivileged, opt.BuildArgs, false, loadCmd)
		if err != nil {
			return nil, err
		}
		err = afterFun(ctx, mts)
		if err != nil {
			return nil, err
		}
	}
```

The key here is using `BuildAsync`. This will require the promises, I think, and the semaphore logic. Wonder if it's easier to keep the implementation as a single one for both tar and registry versions.

earthfile2llb/withdockerrunreg.go
```diff
@@ -0,0 +1,272 @@
...
+	"github.com/moby/buildkit/client/llb"
+	"github.com/pkg/errors"
+)
+
+type withDockerRunRegistry struct {
+	*withDockerRunBase
+	c *Converter
+
+	enableParallel bool
+	sem            semutil.Semaphore
```
vladaionescu, line 20: It doesn't look like this semaphore is ever initialized. This should be very similar to the tar version:

```go
	// This semaphore ensures that there is at least one thread allowed to progress,
	// even if parallelism is completely starved.
	w.sem = semutil.NewMultiSem(w.c.opt.Parallelism, semutil.NewWeighted(1))
```

The reason this needs to have at least one thread is that all the parallelism assumes that the main thread can always make progress and never needs to wait for a secondary thread itself. In the case of the use of `BuildAsync` here, however, we are breaking that assumption - we are waiting for all the promises to finish before progressing work on the main thread. So to avoid a dead-lock, we ensure that at least one secondary thread (started via `BuildAsync`) can always make progress.

builder/image_solver.go
```diff
@@ -289,12 +269,38 @@ func (s *localRegistryImageSolver) SolveImage(ctx context.Context, mts *states.M
...
+	refPrefix := fmt.Sprintf("ref/%s", refKey)
+	res.AddRef(refKey, ref)
+
+	localRegPullID := fmt.Sprintf("sess-%s/%s", gwClient.BuildOpts().SessionID, imageDef.ImageName)
+	res.AddMeta(fmt.Sprintf("%s/export-image-local-registry", refPrefix), []byte(localRegPullID))
+	res.AddMeta(fmt.Sprintf("%s/%s", refPrefix, exptypes.ExporterImageConfigKey), imgJSON)
+	res.AddMeta(fmt.Sprintf("%s/image.name", refPrefix), []byte(imageDef.ImageName))
+	res.AddMeta(fmt.Sprintf("%s/image-index", refPrefix), []byte(strconv.Itoa(idx)))
+
+	return nil
```
vladaionescu, line 305: I like the encapsulation of this logic in a nice little function.

earthfile2llb/withdockerrun.go
```diff
@@ -566,7 +414,91 @@ func makeWithDockerdWrapFun(dindID string, tarPaths []string, localRegImages []s
...
+		llb.Args(args),
+		llb.WithCustomNamef("%sWITH DOCKER (install deps)", w.c.vertexPrefix(false, false, false)),
+	}
+	w.c.mts.Final.MainState = w.c.mts.Final.MainState.Run(runOpts...).Root()
+	return nil
+}
+
+type withDockerRunBase struct {
+	c *Converter
+}
```
vladaionescu, line 438: Perhaps this should be in its own file too to help with following along.

vladaionescu (commented):
This looks good - one last important thing regarding the semaphore. Left a comment...

earthfile2llb/withdockerrunreg.go
mikejholly: Thanks. I must have missed that. I moved it to both constructors.

earthfile2llb/withdockerrun.go
mikejholly: Good call. Updated.

builder/image_solver.go
mikejholly: Thanks!

PR #1884, opened 2022-05-11
Pin dind version as 20.10.15 is failing

mikejholly:
The 20.10.15 Docker image seems to have introduced a bug/issue. Let's pin our version for now until it's fixed.

```
Repeating the output of the command that caused the failure
./t/dind-auto-install+test *failed* | BASE_IMAGE=../..+dind-alpine
./t/dind-auto-install+test *failed* | --> WITH DOCKER RUN --privileged true
./t/dind-auto-install+test *failed* | Loading images...
./t/dind-auto-install+test *failed* | Loaded image: hello-world:latest
./t/dind-auto-install+test *failed* | ...done
./t/dind-auto-install+test *failed* | Network _default  Creating
./t/dind-auto-install+test *failed* | Network _default  Created
./t/dind-auto-install+test *failed* | Container -hello-1  Creating
./t/dind-auto-install+test *failed* | Error response from daemon: Invalid container name (-hello-1), only [a-zA-Z0-9][a-zA-Z0-9_.-] are allowed
./t/dind-auto-install+test *failed* | ERROR: Command exited with non-zero code: WITH DOCKER RUN --privileged true
```

PR #1885, opened 2022-05-11
Add WORKDIR to dind test for Compose image name issue

mikejholly:
Docker Compose v2 is referencing the directory name when choosing a container name. If we run at `/` the name is empty and causes an error. This change uses `WORKDIR` to resolve that problem.

Related: https://github.com/earthly/earthly/pull/1884

### DIFF
--- a/tests/dind-auto-install/Earthfile
+++ b/tests/dind-auto-install/Earthfile
@@ -2,8 +2,8 @@ VERSION 0.6
 
 all:
     BUILD +test \
-        # Pin the version here as 20.10.15 seems to have introduced a bug.
         --BASE_IMAGE=docker:20.10.14-dind \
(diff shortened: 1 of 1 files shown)


tests/dind-auto-install/Earthfile
```diff
@@ -2,8 +2,8 @@ VERSION 0.6
 
 all:
     BUILD +test \
-        # Pin the version here as 20.10.15 seems to have introduced a bug.
         --BASE_IMAGE=docker:20.10.14-dind \
+        --BASE_IMAGE=docker:20.10.15-dind \
```
vladaionescu, line 6: It might be good for us to catch any future incompatibility. I would maybe keep the unpinned version here.

vladaionescu (approved):

PR #1888, opened 2022-05-12
initial WAIT/END code

alexcb:
This introduces a new WAIT / END block which allows users to ensure RUN
commands and SAVE IMAGE commands occur before continuing past the scope
of the block.

This new functionality is not complete -- SAVE IMAGE commands are only
pushed to external docker registries and are not yet exported locally.


alexcb:
This requires https://github.com/earthly/buildkit/pull/77 to be merged first.

earthfile2llb/converter.go
```diff
@@ -930,20 +965,27 @@ func (c *Converter) SaveImage(ctx context.Context, imageNames []string, pushImag
...
+				CacheHint:           cacheHint,
+				HasPushDependencies: false,
+				ForceSave:           c.opt.ForceSaveImage,
+				CheckDuplicate:      c.ftrs.CheckDuplicateImages,
+				NoManifestList:      noManifestList,
+				Platform:            c.platr.Materialize(c.platr.Current()),
+			}
+
+			if c.ftrs.WaitBlock {
+				shouldPush := pushImages && !c.mts.Final.Target.IsRemote() && si.DockerTag != "" && c.opt.DoSaves
```
vladaionescu, line 983: What's the reasoning for `!c.mts.Final.Target.IsRemote()`?

earthfile2llb/converter.go
```diff
@@ -1536,6 +1579,14 @@ func (c *Converter) internalRun(ctx context.Context, opts ConvertRunOpts) (pllb.
 		opts.shellWrap = withShellAndEnvVars
 	}
 
+	if c.ftrs.WaitBlock {
+		c.opt.Console.Warnf("WAIT/END code is super-experimental and incomplete -- it should currently be avoided")
```
vladaionescu, line 1583: Heh... "super-experimental" is maybe too much. Just "experimental" is enough? Perhaps we can say that it should be avoided for anything other than toy projects.

```suggestion
		c.opt.Console.Warnf("WAIT/END code is super-experimental and incomplete -- it should currently be avoided")
```

earthfile2llb/earthfile2llb.go
```diff
@@ -162,6 +168,17 @@ func Earthfile2LLB(ctx context.Context, target domain.Target, opt ConvertOpt, in
...
+			panic("waitBlock should not be set during initial call")
+		}
+		opt.waitBlock = newWaitBlock()
+
+		defer func() {
+			err := opt.waitBlock.wait(ctx) // ensure final waitblock completes
+			if retErr == nil {
+				retErr = err
+			}
+		}()
```
vladaionescu, line 181: Feels like we could decouple the boolean `initialCall` from `opt.waitBlock` by simply checking `if opt.waitBlock == nil { ... do initialization logic ... }`.

(Similar to`if opt.SolveCache == nil`, `if opt.Visited == nil`, `if opt.MetaResolver == nil`, `if opt.ErrorGroup == nil` above)

earthfile2llb/interpreter.go
```diff
@@ -403,6 +405,24 @@ func (i *Interpreter) handleForArgs(ctx context.Context, forArgs []string, sl *s
...
+	if err != nil {
+		return err
+	}
+
+	err = i.handleBlock(ctx, waitStmt.Body)
+	if err != nil {
+		return err
+	}
+
+	return i.converter.popWaitBlock(ctx)
```
vladaionescu, line 423: Nit: We should probably make `PushWaitBlock` and `PopWaitBlock` public, since they are being called from the interpreter. I know that they are in the same package - it just helps to clarify what the external API of the converter is.

earthfile2llb/wait_block.go
```diff
@@ -0,0 +1,194 @@
...
+			}
+
+			metadata[refPrefix+"/platform"] = []byte(platformStr)
+		}
+	}
+
+	if len(w.converters) == 0 {
+		panic("saveImagesWaitItem should never have been created with zero converters")
+	}
+	gatewayClient := w.converters[0].opt.GwClient // could be any converter's gwClient (they should app be the same)
```
vladaionescu, line 106:
```suggestion
	gatewayClient := w.converters[0].opt.GwClient // could be any converter's gwClient (they should be the same)
```

earthfile2llb/wait_block.go
```diff
@@ -0,0 +1,194 @@
...
+	"github.com/moby/buildkit/exporter/containerimage/exptypes"
+	"github.com/moby/buildkit/frontend/gateway/client"
+	gwclient "github.com/moby/buildkit/frontend/gateway/client"
+	"github.com/pkg/errors"
+)
+
+type saveImagesWaitItem struct {
+	converters []*Converter
+	images     []states.SaveImage
+	push       []bool
```
vladaionescu, line 21: Make a struct out of these so that there is a single slice?

```golang
type saveImagesWaitItem struct {
	images []saveImage
}

type saveImage struct {
	converter  *Converter
	image      states.SaveImage
	push       bool
}
```

earthfile2llb/wait_block.go
```diff
@@ -0,0 +1,194 @@
...
+					return errors.Errorf(
+						"image %s is defined multiple times for the same platform (%s)",
+						saveImage.DockerTag, platformImgName)
+				}
+				platformImgNames[platformImgName] = true
+			}
+
+			metadata[refPrefix+"/platform"] = []byte(platformStr)
+		}
+	}
```
vladaionescu, line 101: The above should be abstracted somehow so that we reuse the code in both the builder and here... Too much brittle duplication.

earthfile2llb/wait_block.go
```diff
@@ -0,0 +1,194 @@
...
+
+func (wb *waitBlock) wait(ctx context.Context) error {
+	wb.mu.Lock()
+	defer wb.mu.Unlock()
+	for _, item := range wb.items {
+		err := item.wait(ctx)
+		if err != nil {
+			return err
+		}
+	}
```
vladaionescu, line 192: I think these items could be fired off in parallel - otherwise, the waiting is relatively slow as this only progress one by one.

There is an `ErrorGroup` in the converter opts which we could use. And also, we should probably use a semaphore to limit parallelism. This kind of semaphore could be appropriate here too: https://github.com/earthly/earthly/blob/main/earthfile2llb/withdockerrun.go#L72-L74.

earthfile2llb/wait_block.go
```diff
@@ -0,0 +1,194 @@
...
+}
+
+type runCommandWaitItem struct {
+	c   *Converter
+	cmd *pllb.State
+}
+
+type waitItem interface {
+	wait(context.Context) error
+}
```
vladaionescu, line 31: If I understand correctly only `RUN` and `SAVE IMAGE` commands cause waiting within the wait block. I wonder if `FROM` or `GIT CLONE` or other commands with external interaction could benefit from waiting too in some cases...

states/states.go
```diff
@@ -263,6 +263,8 @@ type SaveImage struct {
 	// list (usually used for multi-platform setups). This means that the image
 	// can only be a single-platform image.
 	NoManifestList bool
+
+	Platform platutil.Platform
```
vladaionescu, line 267: This feels incorrect - IIRC we should never have bare platforms here because we can't distinguish between null platform that resolves to the native platform vs null platform that resolves to something else. We should always use a `PlatformResolver` instead of a bare platform in these cases. IIRC we already have a `PlatformResolver` in the `states.SingleTarget` struct that contains the `SaveImage`...

Though now `SaveImage` could be loose, only part of a wait block...

I'm now a bit worried about what happens if we attempt to save an image without having run `FinalizeStates` for the target that created it. Note how the `PlatformResolver` gets assigned at this point: https://github.com/earthly/earthly/blob/main/earthfile2llb/converter.go#L1375.

I feel that without `FinalizeStates` there will be many edge cases where things will not work as expected.

Could we instead not wait on `SAVE IMAGE` itself, but rather on things like `BUILD` statements (and of course, waiting on `BUILD` will trigger the save image exports if any)? Then we could ensure that `FinalizeStates` has run for the target being invoked (and then we could use the `PlatformResolver` from the `mts.Final`).

vladaionescu (commented):
I think this looks in the right direction overall, but there are a few more important considerations that I commented about below.

(Another nit: this function also needs special handling for the wait block: https://github.com/earthly/earthly/blob/main/earthfile2llb/interpreter.go#L114. At the very least add a TODO kinda like we did for `WITH`?)

builder/builder.go
```diff
@@ -267,7 +267,6 @@ func (b *Builder) convertAndBuild(ctx context.Context, target domain.Target, opt
 
 					refKey := fmt.Sprintf("image-%d", imageIndex)
 					refPrefix := fmt.Sprintf("ref/%s", refKey)
-					imageIndex++
```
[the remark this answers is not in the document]
alexcb: It's been moved to https://github.com/earthly/earthly/pull/1888/files/60e70ad703aaffcea865c0ff0df03633487d5d08#diff-d371e466d16c4d92dff29b2806e847c619149cd6de147f1cb9ae7e9511d5da7fR289

which lines up with immediately after we `res.AddRef(refKey, ref)`.

This fixes the issue of having `localRegPullID := fmt.Sprintf("sess-%s/sp:img%d", gwClient.BuildOpts().SessionID, imageIndex)` using a different string value.

earthfile2llb/converter.go
alexcb: it came from https://github.com/earthly/earthly/blob/main/builder/builder.go#L242

I'm honestly not sure why we would want this, I'm happy to remove it.

earthfile2llb/wait_block.go
alexcb: That would break the use-case for this feature: ensuring all `SAVE IMAGE --push` commands finish before performing a `RUN ./deploy` command.

earthfile2llb/converter.go
vladaionescu: I think this was needed to support the old 0.5 pushing logic, which did things differently when the target was remote. I believe we can remove it in the new code.

Now that I look at the original logic that you linked, it doesn't look right in that case either... It should only depend on being remote or not in 0.5. 🤔

states/states.go
alexcb: > Could we instead not wait on SAVE IMAGE itself, but rather on things like BUILD statements

This might not work for cases where there is no `BUILD`, e.g.

```
VERSION 0.6
FROM alpine
RUN echo generate-data > output
SAVE IMAGE --push alexcb132/testing:tag
RUN --push echo do this after
```

Which should be translated to:
```
VERSION --wait-block 0.6
FROM alpine
WAIT
    RUN echo generate-data > output
    SAVE IMAGE --push alexcb132/testing:tag
END
RUN --push echo do this after
```

Given it's a pattern we already support, it would be nice if we could continue to support it.

As we discussed on Friday, I'll first try including a bool to detect if it's a single image for the current worker's platform (for cases no `--platform` was given), or a pseudo-multiplatform image which only has a single platform defined (for cases `--platform` was passed in).

earthfile2llb/wait_block.go
alexcb: We should wait for these in a case such as

```
WAIT
    GIT CLONE ...
END
```

but for a case such as:

```
foo:
    FROM alpine
    GIT CLONE ...

WAIT
    BUILD +foo
END
```

I need to add in a wait under the finalizeState (of `+foo`), to add in the last command; however i'm currently fighting with a 

```
time="2022-05-17T00:15:47Z" level=error msg="/moby.buildkit.v1.Control/Solve returned error: rpc error: code = Canceled desc = context canceled\n"
```

error which is preventing me from pushing my latest attempt to implement this.

earthfile2llb/wait_block.go
alexcb: What are your thoughts on 6d0cc0e89913036406d0d9d4f68d7a4927e90b5a ? It manages to only define the applicable metadata tags in a single location.

Perhaps we could expand this helper for additional earthlyoutput exporter options in a follow up PR?

vladaionescu: Yeah, sounds good!

earthfile2llb/converter.go
```diff
@@ -930,20 +971,30 @@ func (c *Converter) SaveImage(ctx context.Context, imageNames []string, pushImag
...
-					DockerTag:           imageName,
-					Push:                pushImages,
-					InsecurePush:        insecurePush,
-					CacheHint:           cacheHint,
-					HasPushDependencies: false,
-					ForceSave:           c.opt.ForceSaveImage,
-					CheckDuplicate:      c.ftrs.CheckDuplicateImages,
-					NoManifestList:      noManifestList,
-				})
+			//singlePlatform := sts.PlatformResolver.Current() == platutil.DefaultPlatform || noManifestList
```
vladaionescu, line 974: Forgotten debug logic?

earthfile2llb/wait_block.go
```diff
@@ -0,0 +1,190 @@
...
+	errGroup, ctx := serrgroup.WithContext(ctx)
+
+	for _, item := range wb.items {
+		cmdItem, ok := item.(*runCommandWaitItem)
+		if !ok {
+			continue
+		}
+		errGroup.Go(func() error {
+			return cmdItem.c.forceExecution(ctx, *cmdItem.cmd, cmdItem.c.platr)
+		})
```
vladaionescu, line 187: We should limit parallelism based on the conversion parallelism semaphore perhaps?

If we do that, we should use a semaphore like this one: https://github.com/earthly/earthly/blob/main/earthfile2llb/withdockerrun.go#L72-L74. This ensures that at least one thread can make progress, since the assumption in the interpreter / converter parallelism code is that the main thread can always make progress.

earthfile2llb/wait_block.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+	return errGroup.Wait()
+}
```
vladaionescu, line 190: Thinking about this some more... I think calling `forceExecution` on targets other than the current one can result in buggy behavior. Here's why...

For the most part, so far, we've assumed that every converter instance can manage its own execution flow and that nothing external can trigger execution unless the converter has finished processing entirely (`FinalizeStates` has been called). It feels that the wait logic in here breaks these assumptions and it's really hard to tell whether there will be things that will no longer work when that happens. And even if there aren't the execution flow becomes much harder to reason about, which could open us up in the future to bugs we cannot explain or debug.

I would do the following:

* if a target is part of a `WAIT` block in its entirety (e.g. via `BUILD +target`, `FROM +target` within a `WAIT`), I would use only `BuildAsync` + the callback for the waiting. That is the main way we wait for targets to complete today in other parts of our code (e.g. in `WITH DOCKER` code).
* if a target is part of a `WAIT` block only partially (basically the target that contains the `WAIT` itself), then we can perform `forceExecution` for the waiting (since now we are within the same converter thread, which means we have ownership over the control of the execution of the converter).

I think this also has the benefit that it keeps the control flow of the converter more manageable, and easier to reason about in our general design.

util/saveimageutil/saveimageutil.go
```diff
@@ -0,0 +1,61 @@
...
+		addRef:  func(k string, ref gwclient.Reference) { refs[k] = ref },
+		addMeta: func(k string, v []byte) { metadata[k] = v },
+	}
+}
+
+// ImageSaver exposes a common interface for adding references and metadata to either buildkit's Build or Export functions
+type ImageSaver struct {
+	addRef  func(string, gwclient.Reference)
+	addMeta func(string, []byte)
+}
```
vladaionescu, line 34: Ideas for slightly improved name (s.t. the exporting of artifacts is possible in the future too):

* `RefsBuilder`
* `ResultBuilder`
* `GatewayResultBuilder`
* `GatewayExporterBuilder`

vladaionescu (commented):
Looks better! I still have some comments below...

Also, I'm still not sure about exporting an image before calling `FinalizeStates` - wondering if that could be brittle and whether we should even allow the use case

```Earthly
WAIT
  SAVE IMAGE ...
END
RUN ...
```

when it could be replaced by simply

```Earthly
WAIT
  BUILD ...
END
RUN ...
```

But maybe there are good reasons to still allow it?

builder/builder.go
```diff
@@ -310,28 +306,19 @@ func (b *Builder) convertAndBuild(ctx context.Context, target domain.Target, opt
...
-						refKey := fmt.Sprintf("image-%d", imageIndex)
-						refPrefix := fmt.Sprintf("ref/%s", refKey)
-						imageIndex++
-
-						res.AddMeta(fmt.Sprintf("%s/image.name", refPrefix), []byte(saveImage.DockerTag))
-						res.AddMeta(fmt.Sprintf("%s/platform", refPrefix), []byte(platformStr))
-						res.AddMeta(fmt.Sprintf("%s/export-image-push", refPrefix), []byte("true"))
-						if saveImage.InsecurePush {
-							res.AddMeta(fmt.Sprintf("%s/insecure-push", refPrefix), []byte("true"))
+						_, err = gwCrafter.AddPushImageEntry(ref, imageIndex, saveImage.DockerTag, shouldPush, saveImage.InsecurePush, saveImage.Image, []byte(platformStr))
```
vladaionescu, line 309: The image index could be maintained in the crafter perhaps? This should simplify some things potentially.

... But then it would have to return the index as an ID and we'd have to keep track of it separately outside of the crafetr. Maybe it's ok the way it is then. (Feel free to ignore this comment)

earthfile2llb/interpreter.go
```diff
@@ -134,6 +134,11 @@ func (i *Interpreter) handleBlockParallel(ctx context.Context, b spec.Block, sta
 			case "DOCKER":
 				// TODO
 			}
+		} else if stmt.If != nil || stmt.For != nil || stmt.Wait != nil {
```
vladaionescu, line 137: I wonder if we should also have an `else -> error` case, so that we catch any future statement types that should be handled here.

earthfile2llb/wait_block.go
```diff
@@ -0,0 +1,187 @@
...
+func (wb *waitBlock) waitCommands(ctx context.Context) error {
+	errGroup, ctx := serrgroup.WithContext(ctx)
+
+	for _, item := range wb.items {
+		cmdItem, ok := item.(*runCommandWaitItem)
+		if !ok {
+			continue
+		}
+		errGroup.Go(func() error {
+			return cmdItem.c.forceExecutionWithSemaphore(ctx, *cmdItem.cmd, cmdItem.c.platr)
```
vladaionescu, line 183: I believe that we need to use that special semaphore that I pointed out previously over here.

https://github.com/earthly/earthly/blob/main/earthfile2llb/withdockerrun.go#L72-L74

earthfile2llb/wait_block.go
```diff
@@ -0,0 +1,187 @@
...
+type saveImageWaitItem struct {
+	c    *Converter
+	si   states.SaveImage
+	push bool
+}
+
+type runCommandWaitItem struct {
+	c   *Converter
+	cmd *pllb.State
+}
```
vladaionescu, line 27: I wonder if this could be simplified even further - rather than add every run command to the wait block, what if we simply force execution of the current target's state at the end of every WAIT block?

util/gatewaycrafter/gatewaycrafter.go
```diff
@@ -0,0 +1,86 @@
...
+type GatewayCrafter interface {
+	AddPushImageEntry(ref gwclient.Reference, refID int, imageName string, shouldPush, insecurePush bool, imageConfig *image.Image, platformStr []byte) (string, error)
+}
+
+// MapBasedGatewayCrafter extends the GatewayCrafter but provides an additional method for
+// fetching the Ref and metadata maps which must be passed to the gwclient.Export method.
+type MapBasedGatewayCrafter interface {
+	GatewayCrafter
+	GetRefsAndMetadata() (map[string]gwclient.Reference, map[string][]byte)
+}
```
vladaionescu, line 24: Feels like this could be simplified.

Instead of defining `gatewayCrafterBase` as a set of functions, we could simply make it

```go
type GatewayCrafeter struct {
  refs map[string]gwclient.Reference{}
  meta map[string][]byte
}
```

Then there would only have to be a single implementation (no interfaces). It just that the caller decides to either extract the information contained within as a `*gwclient.Result` or as refs+meta.

```go
func (c *GatewayCrafter) AsGatewayClientResult() (*gwclient.Result) {
  res := gwclient.NewResult()
  for k, v := range c.refs {
     res.AddRef(k, v)
  }
  for k, v := range c.meta {
     res.AddMeta(k, v)
  }
  return res
}
```

```go
func (c *GatewayCrafter) AsRefsAndMeta() (...) {
  return c.refs, c.meta
}
```

The constructor can then be simply

```go
func NewGatewayCrafter() *GatewayCrafter {
  return &GatewayCrafter{
    meta: make(...),
    refs: make(...),
  }
}
```

util/platutil/platform.go
```diff
@@ -22,6 +22,11 @@ type Platform struct {
 	p      *specs.Platform
 }
 
+// IsPlatformDefined returns true when the platform was explicitly set
+func IsPlatformDefined(p Platform) bool {
+	return p.p != nil || p.user || p.native
```
vladaionescu, line 27: This doesn't seem right. I believe that in other places in the code we perform this check by doing `p == DefaultPlatform`.

builder/builder.go
```diff
@@ -185,6 +186,7 @@ func (b *Builder) convertAndBuild(ctx context.Context, target domain.Target, opt
 			}
 		}
 		res := gwclient.NewResult()
+		gwCrafter := gatewaycrafter.NewGatewayCrafterForBuild(res)
```
vladaionescu, line 189: New name is cool ;-)

earthfile2llb/converter.go
```diff
@@ -1379,6 +1429,10 @@ func (c *Converter) FinalizeStates(ctx context.Context) (*states.MultiTarget, er
 		c.mts.Final.SetDoSaves()
 	}
 
+	if c.ftrs.WaitBlock {
+		c.waitBlock().addCommand(&c.mts.Final.MainState, c)
```
vladaionescu, line 1433: Should we rename this to `addState` perhaps?

util/platutil/platform.go
alexcb: ah ha: https://github.com/earthly/earthly/blob/f50cd370e79d3f6b37b416c6020f1285201e0c06/builder/builder.go#L212

that looks better.

util/gatewaycrafter/gatewaycrafter.go
alexcb: I don't think this will work because the builder/builder.go code still requires that part of the results object be modified directly; we need to either pass in a reference (like we already do), or simply make a call to `AsRefsAndMeta()`, then merge that data into the existing results.

For example here's the builder code:

```
		res := gwclient.NewResult()
		gwCrafter := gatewaycrafter.NewGatewayCrafterForBuild(res)
		if !b.builtMain {
			ref, err := b.stateToRef(childCtx, gwClient, mts.Final.MainState, mts.Final.PlatformResolver)
			if err != nil {
				return nil, err
			}
			res.AddRef("main", ref) <--- This adds directly to the results
		}
```

the part way down the function, we have:

```
					// For push.
					if shouldPush {
						_, err = gwCrafter.AddPushImageEntry(ref, imageIndex, saveImage.DockerTag, shouldPush, saveImage.InsecurePush, saveImage.Image, []byte(platformStr))
						if err != nil {
							return nil, err
						}
						imageIndex++
					}
```

we could change the function to simply export `AsRefsAndMeta()`, where we then iterate over the refs and metadata and merge it back into res, e.g.

```
res := gwclient.NewResult()

for .... {
... code that works directly on res
....and code that calls gwCrafter.AddPushImageEntry(....)
}

...
// somewhere after the loop

refsToMerge, metadataToMerge := gwCrafter.AsRefsAndMeta()
for k, v := range refsToMerge {
    res.AddRef(k, v)
}
for k, v := range metadataToMerge {
    res.AddMeta(k, v)
}
```

vladaionescu: We could perhaps expose `AddRef` in the crafter for that situation?

vladaionescu: Maybe both `AddRef` and `AddMeta`, at least until we port all types of exports for the crafter to handle elegantly.

alexcb: That could work!

builder/builder.go
alexcb: Ultimately that's the direction I want to head in as this code gets moved into the wait_block.go code that's called by popWaitBlock. However that should be left to a follow up PR (to limit the size of this one).

earthfile2llb/wait_block.go
alexcb: This has been changed to only wait during finalizeState, or when a new Build is triggered.

earthfile2llb/wait_block.go
vladaionescu: Hey @alexcb - I think this was forgotten?

vladaionescu (approved):

PR #1902, opened 2022-05-26
Avoid panic in osutil

cupen:
It's a bug of `osutil`.
So maybe we could just avoid it, and wait for this pr to be merged.
https://github.com/wille/osutil/pull/10

### DIFF
--- a/cmd/earthly/main.go
+++ b/cmd/earthly/main.go
@@ -311,7 +311,17 @@ func getVersionPlatform() string {
 }
 
 func getPlatform() string {
-	return fmt.Sprintf("%s/%s; %s", runtime.GOOS, runtime.GOARCH, osutil.GetDisplay())
+	showOSInfo := func() (info string) {
+		defer func() {
+			if err := recover(); err != nil {
+				// skipped
+				info = "unknown"
+				return
+			}
+		}()
+		return osutil.GetDisplay()
+	}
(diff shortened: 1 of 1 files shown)


cmd/earthly/main.go
```diff
@@ -311,7 +311,17 @@ func getVersionPlatform() string {
 }
 
 func getPlatform() string {
-	return fmt.Sprintf("%s/%s; %s", runtime.GOOS, runtime.GOARCH, osutil.GetDisplay())
+	showOSInfo := func() (info string) {
```
alexcb, line 314: Nice work-around! It's even more impressive that you contributed a fix for it upstream.

Would you be able to add a comment here mentioning something like:

```
// Work-around for windows panics; this can be removed once https://github.com/wille/osutil/pull/10 is merged
```

so we can tidy this work-around once the upstream fix is merged.

cupen: Of course yes.  And it's done.

cmd/earthly/main.go
```diff
@@ -311,6 +311,7 @@ func getVersionPlatform() string {
 }
 
 func getPlatform() string {
+	// Work-around for windows panics; this can be removed once https://github.com/wille/osutil/pull/10 is merged
```
cupen, line 314: @alexcb  It's done.

PR #1905, opened 2022-05-30
update buildkit

alexcb:
updated buildkit contains fix for using non-standard git port
when combined without strict host checking.


config/config.go
```diff
@@ -95,6 +95,8 @@ type GitConfig struct {
 	Suffix                string `yaml:"suffix"                       help:"The git repository suffix, like .git."`                                       // .git
 	Auth                  string `yaml:"auth"                         help:"What authentication method do you use? Valid options are: http, https, ssh."` // http, https, ssh
 	User                  string `yaml:"user"                         help:"The username to use when auth is set to git or https."`
+	Port                  int    `yaml:"port"                         help:"The port to connect to when using git; has no effect for http(s)."`
```
alexcb, line 98: I'm thinking we'l keep this as experimental for now until we hear back from some of our users who have required this confirming if it works.

vladaionescu (approved):

PR #1912, opened 2022-06-01
display warning for ARG help

alexcb:
"ARG help" will become a reserved argument name in the future, print a
warning notifying users it will eventually become an error.


vladaionescu (approved):

vladaionescu:
Another thought... perhaps the earthlydoc feature would be simply `earthly help +my-target`, in which case we wouldn't need to reserve `--help` (for `earthly +my-target --help`).

alexcb:
> Another thought... perhaps the earthlydoc feature would be simply `earthly help +my-target`, in which case we wouldn't need to reserve `--help` (for `earthly +my-target --help`).

I like that idea more.

`earthly help` would display both regular cli options as well as targets from the current directory (essentially combining the existing `earthly --help` and `earthly ls`), then we would also make `earthly --help` an alias for `earthly help`.

PR #1920, opened 2022-06-08
Initial Satellites Integration

brandonSc:
WIP - still need to do more integration testing. It's mostly working though and the code may not many changes from this point.


brandonSc:
Converted back to draft. Taking some time to make the improvements we discussed in Google Docs, then will add everyone back for review.

buildkitd/buildkitd.go
```diff
@@ -11,13 +11,14 @@ import (
 	"sync"
 	"time"
 
+	"github.com/moby/buildkit/client"
+	_ "github.com/moby/buildkit/client/connhelper/dockercontainer" // Load "docker-container://" helper.
+	"github.com/pkg/errors"
+
```
brandonSc, line 17: My editor keeps wanting to sort these imports, but it looks like we usually keep them organized like this anyway?

alexcb: I suspect it's just some old unorganised imports -- looks like a good change.

cmd/earthly/main.go
```diff
@@ -1205,9 +1309,30 @@ func (app *earthlyApp) configureSatellite(cc cloud.Client) error {
...
+		app.buildkitdSettings.BuildkitAddress = containerutil.SatelliteAddress
+	}
+
+	app.console.Warnf("") // newline
+	app.console.Warnf("The following feature flags are recommended for use with Satellites and will be auto-enabled:")
+	app.console.Warnf("--new-platform, --use-registry-for-with-docker")
+	app.console.Warnf("") // newline
+
+	if app.featureFlagOverrides == "" {
+		app.featureFlagOverrides = "new-platform,use-registry-for-with-docker"
```
brandonSc, line 1332: This seemed like a clean way to set these feature flags from the central `configureSatellite` function but let me know if there's a more preferred method of doing it.

cloud/client.go
```diff
@@ -972,6 +986,105 @@ func (c *client) WhoAmI() (string, string, bool, error) {
...
+	for _, o := range orgs {
+		if o.Name == orgName {
+			return o.ID, nil
+		}
+	}
+	return "", errors.Errorf("org not found: %s", orgName)
+}
+
+func (c *client) ListSatellites(orgID string) ([]SatelliteInstance, error) {
+	url := fmt.Sprintf("/api/v0/satellites?orgId=%s", orgID)
```
alexcb, line 1029: nit: probably not needed, but makes it more robust:
```suggestion
	url := fmt.Sprintf("/api/v0/satellites?orgId=%s", url.QueryEscape(orgID))
```

(there's a few other spots in this code that could use this too.

cmd/earthly/main.go
```diff
@@ -1053,6 +1058,105 @@ Set up a whole custom git repository for a server called example.com, using a si
 				},
 			},
 		},
+		{
+			Name:    "satellite",
+			Aliases: []string{"satellites"},
+			Usage: "Launch and use a Satellite runner as remote backend for Earthly builds.\n" +
+				"	Satellites can be used to optimize and share cache between multiple builds and users,\n" +
+				"	as well as run builds in native architectures independent of where the Earthly client is invoked.\n" +
+				"	Note: this feature is currently experimental.\n" +
```
alexcb, line 1067: Since it's experimental, should we include a `Hidden:      true` field?

alexcb: our beta testers can add an `export EARTHLY_SHOW_HIDDEN=1` to enable this help text.

cmd/earthly/main.go
alexcb: seems like a reasonable place to set them.

Can you change it to something like:

```
if app.featureFlagOverrides != "" {
    app.featureFlagOverrides += ","
}
app.featureFlagOverrides += "new-platform,use-registry-for-with-docker"
```

so we don't have the flags copy/pasted in both parts of the if/else?

cmd/earthly/main.go
```diff
@@ -3205,6 +3329,257 @@ func (app *earthlyApp) actionListTargets(c *cli.Context) error {
...
+	if err != nil {
+		return errors.Wrap(err, "could not save config")
+	}
+
+	return nil
+}
+
+func (app *earthlyApp) printSatellites(satellites []cloud.SatelliteInstance) {
+	for _, satellite := range satellites {
+		if app.cfg.Satellite.Name == satellite.Name {
```
alexcb, line 3362: what happens if the same satellite name exists in two different orgs?

alexcb (approved):
looks great, just left some minor nits/questions.

vladaionescu (commented):
Some initial findings...

* `inspect` doesn't have a help string
* `inspect` shows empty for the `version` field - is that expected?
* The launch command set a uuid as the `org` in the config. I think that this broke something?
  ```
   1. Init 🚀
  
  Note: the Interactive Debugger, Interactive RUN commands, and Local Registries do not yet work on Earthly Satellites.
  Shareable link: https://ci.staging.earthly.dev/logs?logId=536173e2-4150-4554-b8c2-13bac7c995c0
  Error: build new buildkitd client: could not construct new buildkit client: invalid org provided: org not found: 464f2043-f46c-40b7-82ea-c739ad4e52fb
  ```
* Even when using `select` and thus setting the right org as `test`, I get this error:
  ```
  Error: build new buildkitd client: connect provided buildkit: timeout 1m0s: buildkitd did not make connection available after start with error: failed to list workers: Unavailable: connection error: desc = "transport: Error while dialing dial tcp: lookup satellite.earthly.dev on [2001:558:feed::1]:53: no such host": buildkitd failed to start (in time)
  ```

Additional findings which could be fixed in subsequent PRs
* Launching satellite with a name that already exists results in the following
  ```
  Launching Satellite. This could take a moment...
  retrying http request due to unexpected status code 500: internal error
  retrying http request due to unexpected status code 500: internal error
  retrying http request due to unexpected status code 500: internal error
  retrying http request due to unexpected status code 500: internal error
  retrying http request due to unexpected status code 500: internal error
  retrying http request due to unexpected status code 500: internal error
  ^CReceived signal interrupt. Cleaning up before exiting...
  retrying http request due to unexpected status code 500: internal error
  retrying http request due to unexpected status code 500: internal error
  ^CReceived second signal interrupt. Forcing exit.
  ```
  (Note also that the first ctrl-c did not stop if - presumably because the ctx wasn't propagated from `main.go`?)

cmd/earthly/main.go
```diff
@@ -3205,6 +3329,257 @@ func (app *earthlyApp) actionListTargets(c *cli.Context) error {
...
+	err = cc.DeleteSatellite(app.satelliteName, orgID)
+	if err != nil {
+		return errors.Wrapf(err, "failed to delete satellite %s", app.satelliteName)
+	}
+
+	if app.satelliteName == app.cfg.Satellite.Name {
+		// TODO what strategy do we want to use if you delete your current satellite?
+		if err = app.useSatellite(c, "", ""); err != nil {
+			return errors.Wrapf(err, "failed unselecting satellite")
+		}
```
vladaionescu, line 3486: I would additionally `Printf` that the satellite was unselected as a result of the destruction, for clarity.

cmd/earthly/main.go
```diff
@@ -3205,6 +3329,257 @@ func (app *earthlyApp) actionListTargets(c *cli.Context) error {
...
+func (app *earthlyApp) actionSatelliteUnselect(c *cli.Context) error {
+	app.commandName = "unselect"
+
+	if c.NArg() != 0 {
+		return errors.New("command does not accept any arguments")
+	}
+
+	app.satelliteName = c.Args().Get(0)
+
+	if err := app.useSatellite(c, "", ""); err != nil {
```
vladaionescu, line 3576: Similarly here

cmd/earthly/main.go
```diff
@@ -3205,6 +3329,257 @@ func (app *earthlyApp) actionListTargets(c *cli.Context) error {
...
+
+	return nil
+}
+
+func (app *earthlyApp) printSatellites(satellites []cloud.SatelliteInstance) {
+	for _, satellite := range satellites {
+		if app.cfg.Satellite.Name == satellite.Name {
+			app.console.Printf("* %s", satellite.Name)
+		} else {
+			app.console.Printf("  %s", satellite.Name)
```
vladaionescu, line 3365: Nit: Ideally, we should use `stdout` for these (`app.console` uses `stderr`, I believe).

cmd/earthly/main.go
```diff
@@ -3205,6 +3329,257 @@ func (app *earthlyApp) actionListTargets(c *cli.Context) error {
...
+	if err != nil {
+		return errors.Wrap(err, "failed to create cloud client")
+	}
+
+	orgID, err := app.getSatelliteOrgID(cc)
+	if err != nil {
+		return err
+	}
+
+	app.console.Warnf("Destroying Satellite. This could take a moment...")
```
vladaionescu, line 3476:
```suggestion
	app.console.Printf("Destroying Satellite. This could take a moment...")
```

vladaionescu (commented):
Looking pretty good :+1:

Some additional nits...

cmd/earthly/main.go
brandonSc: Good point, the code should consider that here

cmd/earthly/main.go
brandonSc: Yep that is cleaner 👍

cmd/earthly/main.go
brandonSc: Will do. Didn't realize `app.console.Printf` was also going to stderr.

brandonSc:
@vladaionescu to your question 
> inspect shows empty for the version field - is that expected?
It's currently empty on the server because we haven't implemented code that knows how to set or get the buildkit version. I think we should hardcode that to something on the server now so it doesn't look like a bug.

vladaionescu:
> It's currently empty on the server because we haven't implemented code that knows how to set or get the buildkit version. I think we should hardcode that to something on the server for now so it doesn't look like a bug.

Sounds good - no worries then. I was just double-checking.

brandonSc:
@vladaionescu 
> (Note also that the first ctrl-c did not stop if - presumably because the ctx wasn't propagated from main.go?)

Definitely the context should be propagated through to the cloud client endpoints. I took a stab at implementing it, however, doing it nicely involves a refactoring of all the endpoint calls. I think, in the essence of time, we should tackle that change in a near-future PR.

tests/autocompletion/Earthfile
```diff
@@ -21,6 +21,7 @@ config
 ls 
 org 
 prune 
+satellite
```
alexcb, line 24: This feels wrong: it seems like there's a bug in the autocompletion logic -- if you want to punt on this issue can you create an issue to track it?

alexcb: before you do that can you try changing https://github.com/earthly/earthly/blob/main/autocomplete/complete.go#L344

`subCommands := app.Commands` to `subCommands = getVisibleCommands(app.Commands)`

brandonSc: Yeah I was a bit surprised by the fact it showed up while marked as hidden. Happy to try the fix.

vladaionescu (approved):

PR #1924, opened 2022-06-09
wait-block: initial tests

alexcb:
This introduces tests for the `VERSION --wait-block` feature.
It runs a registry on the localhost's docker, pushes an image to it,
then performs a curl against the regsitry to verify the image was pushed
in the correct order.


alexcb:
image

vladaionescu (approved):
Neat!

PR #1927, opened 2022-06-14
Add docs on how to use earthly/earthly image and remove `SRC_DIR`

vladaionescu:
This PR removes the setting `SRC_DIR` in favor of simply using the current working directory (which also defaults to `/workspace`, just as `SRC_DIR` did before). The user can now override this by using `-w` instead of `-e SRC_DIR` in a typical `docker run` command.

In addition, this also adds a separate page about using the `earthly/earthly` image in general.


dchw (approved):
This works. The point of `SRC_DIR` was to provide a configurable `WORKDIR`, but I was unaware of the docker CLI option to override a specified one.

PR #1928, opened 2022-06-14
fix staging release pipeline

alexcb:
f95c65e4e573f445ee95171fc1ab0465b36d8289 propagated the platform into
the parser, which then caused the parser to attempt to fetch an arm7
version of openJDK which is not supported. ARM7 support was experimental
and was targeting older raspberry pi platforms -- it never worked.

Additionally prevents the prerelease from erroneously pushing a
pre-released version to the latest tag; however the latest tag
is reserved for the latest released version, not pre-released versions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -769,59 +769,6 @@ jobs:
         run: docker logs earthly-buildkitd
         if: ${{ failure() }}
 
-  buildkitd-armv7:
-    name: +all-buildkitd
-    runs-on: ubuntu-latest
-    env:
-      FORCE_COLOR: 1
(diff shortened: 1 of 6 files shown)


vladaionescu (approved):

alexcb:
whoops:
```
[Invalid workflow file: .github/workflows/ci.yml#L998](https://github.com/earthly/earthly/actions/runs/2505432003/workflow)
The workflow is not valid. .github/workflows/ci.yml (Line: 998, Col: 9): Job 'push-prerelease' depends on unknown job 'buildkitd-armv7'.
```

just pushed up a small fix for this.

PR #1938, opened 2022-06-17
Use Satellites in Github Actions

brandonSc:
Added a `satellites.yml` file to Github actions, which copy-and-pastes some of the targets from the main `ci.yaml` script and runs them on the designated `earthly-core` satellite.

This does result in two separate entries in the Github actions tab, however, it does also separate them nicely as the main ci.yaml is still going to be the primary focus for a while and it's quite long as is. We can merge them into one if we prefer though.

Screen Shot 2022-06-21 at 11 16 21 AM


.github/workflows/satellites.yml
```diff
@@ -0,0 +1,181 @@
...
+        run: ./build/linux/amd64/earthly --use-inline-cache +for-linux
+      - name: Set EARTHLY_VERSION_FLAG_OVERRIDES env
+        run: |-
+            set -euo pipefail
+            EARTHLY_VERSION_FLAG_OVERRIDES="$(tr -d '\n' < .earthly_version_flag_overrides)"
+            echo "EARTHLY_VERSION_FLAG_OVERRIDES=$EARTHLY_VERSION_FLAG_OVERRIDES" >> "$GITHUB_ENV"
+      - name: Enable local registry-based exporter
+        run: ./build/linux/amd64/earthly config global.local_registry_host 'tcp://127.0.0.1:8371'
+      - name: Configure Earthly Conversion Parallelism
+        run: ./build/linux/amd64/earthly config global.conversion_parallelism 5
```
vladaionescu, line 52: These are no longer necessary - you can remove them. (In fact, maybe you can remove them everywhere?)

vladaionescu (approved):

.github/workflows/satellites.yml
```diff
@@ -0,0 +1,181 @@
...
+      - name: Execute tests (Earthly Only)
+        run: |-
+          ./build/linux/amd64/earthly --ci -P \
+            --build-arg DOCKERHUB_AUTH=true \
+            --build-arg DOCKERHUB_USER_SECRET=+secrets/earthly-technologies/dockerhub-mirror/user \
+            --build-arg DOCKERHUB_TOKEN_SECRET=+secrets/earthly-technologies/dockerhub-mirror/pass \
+            --build-arg DOCKERHUB_MIRROR=registry-1.docker.io.mirror.corp.earthly.dev \
+          +test
+        if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name == github.repository
+      - name: Execute tests (Fork)
```
alexcb, line 62: will satellites work in forked branches?

.github/workflows/satellites.yml
```diff
@@ -0,0 +1,181 @@
...
+        uses: docker/setup-qemu-action@v1
+        with:
+          image: tonistiigi/binfmt:latest
+          platforms: all
+      - uses: actions/checkout@v3
+      - name: Docker mirror login (Earthly Only)
+        run: docker login registry-1.docker.io.mirror.corp.earthly.dev --username "${{ secrets.DOCKERHUB_MIRROR_USERNAME }}" --password "${{ secrets.DOCKERHUB_MIRROR_PASSWORD }}"
+        if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name == github.repository
+      - name: Configure Satellites
+        run: earthly satellite select earthly-core
```
alexcb, line 33: I suspect we'll want a `if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name == github.repository` here to avoid this from running in forks?

Currently if this runs in a fork, that user shouldn't have access to `secrets.EARTHLY_TOKEN`, and then this will fail.

brandonSc: FYI I tested this from a fork - these just run on a local buildkit instead and pass

Issue #1278, opened 2021-10-08
Not able to run docker-in-docker from npm script


bjaanes:
I have an npm test that uses testcontainers-node to automatically spin up and tear down during tests (so there are no manual dependent steps to just run integration tests locally).

I have created a minimum reproducible example here: 
https://github.com/bjaanes/earthly-dind-reproduced-issue 

All the necessary code is in that repo, but I will copy some of the code for context in this issue:

Earthfile:
```
FROM earthly/dind:ubuntu

this-works:
    WITH DOCKER --pull hello-world
        RUN docker run hello-world
    END

this-does-not:
    COPY ./ ./
    RUN apt-get update -yq \
        && apt-get -yq install curl gnupg ca-certificates gcc g++ make \
        && curl -L https://deb.nodesource.com/setup_16.x | bash \
        && apt-get update -yq \
        && apt-get install -yq nodejs
    RUN npm i
    WITH DOCKER
        RUN npm test
    END
```

`WITH DOCKER` works just fine with the "hello-world" example, but this particular case does not. 
So this produces no errors:
```shell
$ earthly -P +this-works
```
But this one does:
```shell
$ earthly -P +this-does-not
```

I get the following error:
```
(HTTP code 500) server error - OCI runtime create failed: container_linux.go:380: starting container process caused: process_linux.go:545: container init caused: process_linux.go:508: setting cgroup config for procHooks process caused: failed to write "a *:* rwm": write /sys/fs/cgroup/devices/docker/665963e797d624839a294a5023f68d4784ecbaf5b567323b91fea367d0de2467/devices.allow: operation not permitted: unknown
```

bjaanes:
```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
  buildx: Build with BuildKit (Docker Inc., v0.6.3-docker)
  scan: Docker Scan (Docker Inc., v0.8.0)

Server:
 Containers: 4
  Running: 2
  Paused: 0
  Stopped: 2
 Images: 87
 Server Version: 20.10.9
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
[...]

PR #1940, opened 2022-06-17
update s3-related release scripts with MFA

alexcb:
This introduces a new assume-developer-role.sh script which can be used
to assume the developer role. It must be called in each RUN command via
an `eval $(assume-developer-role.sh)` call.

The assume-developer-role.sh can only be called once every few seconds,
otherwise it will re-use the same TOTP token code, which AWS denies, as
a result one can call `eval $(assume-developer-role.sh N)`, where N is
the next Nth token produced in the sequence.

### DIFF
--- a/release/apt-repo/Earthfile
+++ b/release/apt-repo/Earthfile
@@ -1,3 +1,4 @@
+VERSION 0.6
 FROM alpine:3.15
 
 deps:
@@ -39,16 +40,9 @@ deb-arm64:
         +deb/* ./
     SAVE ARTIFACT *.deb
 
-deb-arm7:
-    COPY \
-        --build-arg  EARTHLY_PLATFORM=arm7 \
-        +deb/* ./
-    SAVE ARTIFACT *.deb
-
 deb-all:
     COPY +deb-amd64/*.deb .
     COPY +deb-arm64/*.deb .
-    COPY +deb-arm7/*.deb .
     SAVE ARTIFACT ./*.deb
 
 # If for any reason you need to generate a new PGP key, it can be done with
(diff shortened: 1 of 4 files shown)


alexcb:
This additionally removes arm7 from the repo since we are no longer building arm7 (see #1278 1928)

release/common-repo/assume-developer-role.sh
```diff
@@ -0,0 +1,33 @@
...
+
+AWS_ACCESS_KEY_ID=$(jq -r '.Credentials.AccessKeyId' "$role_path")
+AWS_SECRET_ACCESS_KEY=$(jq -r '.Credentials.SecretAccessKey' "$role_path")
+AWS_SESSION_TOKEN=$(jq -r '.Credentials.SessionToken' "$role_path")
+
+rm "$role_path"
+
+echo "export AWS_ACCESS_KEY_ID=\"$AWS_ACCESS_KEY_ID\";"
+echo "export AWS_SECRET_ACCESS_KEY=\"$AWS_SECRET_ACCESS_KEY\";"
+echo "export AWS_SESSION_TOKEN=\"$AWS_SESSION_TOKEN\";"
```
dchw, line 33: This assumes you already signed in right?

release/apt-repo/Earthfile
```diff
@@ -73,28 +67,14 @@ Expire-Date: 0
...
-
 download:
-    FROM +aws
+    FROM ../common-repo+aws
     ARG --required S3_BUCKET
-    RUN env | grep dev
     RUN --no-cache \
+        --secret MFA_ARN=+secrets/user/earthly-technologies/aws/mfa-arn \
+        --secret MFA_KEY=+secrets/user/earthly-technologies/aws/mfa-key \
         --mount type=secret,id=+secrets/user/earthly-technologies/aws/credentials,target=/root/.aws/credentials \
```
dchw, line 76: These credentials should _also_ require an MFA session? Are you MFA locally and re-setting your credential files?

dchw (approved):
Approved because it already works I am guessing? You could also re-use the script we save/generate in our infra tools via `+mfa`, if you want

release/yum-repo/Earthfile
```diff
@@ -135,12 +122,13 @@ upload:
     ARG --required S3_BUCKET
     RUN --push \
         --mount type=secret,id=+secrets/earthly-technologies/release/keys/earthly-public.pgp,target=/release-key/earthly-public.pgp \
+        --secret MFA_ARN=+secrets/user/earthly-technologies/aws/mfa-arn \
+        --secret MFA_KEY=+secrets/user/earthly-technologies/aws/mfa-key \
         --mount type=secret,id=+secrets/user/earthly-technologies/aws/credentials,target=/root/.aws/credentials \
+        eval $(assume-developer-role 1) && \
```
dchw, line 128: Why do we need to re-issue the assumed role credentials? They should be valid for an hour?

release/common-repo/assume-developer-role.sh
alexcb: This entire script signs you in, which occurs above on `aws sts assume-role`, if it fails the `set -e` will catch it.

I'm not sure if that answers your question or not?

release/yum-repo/Earthfile
alexcb: I deliberately avoid keeping credentials in any layers, each `RUN` command must re-run `eval $(assume-developer-role 1)` to get fresh credentials.

release/apt-repo/Earthfile
alexcb: `/root/.aws/credentials` never gets overwritten, nor does it contain any MFA-related session data.

`aws sts assume-role` is only called once, and if I'm not mistaken, uses both the user's regular `credentials`, which contain `aws_access_key_id` and `aws_secret_access_key`, and the TOTP token value, which is passed to `--token-code`.

release/common-repo/assume-developer-role.sh
dchw: You need to MFA to sign-in, and again to assume a role? I only see one MFA request so I was curious

release/yum-repo/Earthfile
dchw: Fair enough; I wonder how many open sessions one is allowed though?

> I deliberately avoid keeping credentials in any layers

Out of curiosity, how is it different than keeping regular credentials on your machine when you aren't deploying the target image?

Issue #1939, opened 2022-06-17
Registry `WITH DOCKER` loading causes hang when no image needs to be loaded


vladaionescu:
The new `VERSION --use-registry-for-with-docker 0.6` feature hangs when no image needs to be loaded:

```
broken:
    FROM earthly/dind:alpine
    WITH DOCKER
        RUN echo hello
    END
```

vladaionescu:
Workaround:

```
broken:
    FROM earthly/dind:alpine
    WITH DOCKER --pull busybox
        RUN echo hello
    END
```

PR #1942, opened 2022-06-18
Fix registry with-docker loading when no images to load

vladaionescu:
Fixes https://github.com/earthly/earthly/issues/1939

I've also included some changes related to channel size and `select <-ctx.Done()` as I noticed that the hang in #1939 wasn't resolved by a single ctrl-c.


mikejholly:
LGTM

PR #1946, opened 2022-06-20
Add TLS configuration for satellites

dchw:
Depends on: https://github.com/earthly/buildkit/pull/79

The current buildkit TLS client configuration seems to assume that the remote buildkit is configured to handle the TLS connection; especially for an mTLS type configuration.

This, in conjunction with the mentioned buildkit PR, allows just regular old TLS.

This could also be broken out into its own `client.ClientOpt`, but it seemed like it was more code on the `buildkit` side, especially when integrating it with the existing TLS config options. Less code over there means fewer future merge conflicts.

### DIFF
--- a/buildkitd/buildkitd.go
+++ b/buildkitd/buildkitd.go
@@ -761,7 +761,7 @@ func addRequiredOpts(settings Settings, opts ...client.ClientOpt) ([]client.Clie
 			"satellite_name", settings.SatelliteName,
 			"satellite_org", settings.SatelliteOrgID,
 			"satellite_token", settings.SatelliteToken),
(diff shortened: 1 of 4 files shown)


buildkitd/buildkitd.go
```diff
@@ -561,6 +561,7 @@ func checkConnection(ctx context.Context, address string, opts ...client.ClientO
 		if err != nil {
 			connErrMu.Lock()
 			connErr = err
+			fmt.Println(err.Error())
```
brandonSc, line 564: Deliberate or leftover debugging? The err handling code above looks similar but doesn't do `fmt.Println`

alexcb (approved):

PR #1947, opened 2022-06-21
Longer timeout for requests to delete or launch a Satellite

brandonSc:
I was frequently getting `unexpected EOF` error messages trying to delete satellites. The satellite delete would still work behind the scenes, but the request would get disconnected early after about 1 minute. 

I'm mostly sure the change in this PR fixes it, but to be honest, I'm still not sure which component is the one causing the disconnect. The request is hung up on a few places when it times out: `earthly-core -> pipelines-gateway -> pipelines-grpc -> builder-grpc` and I haven't been able to find any error messages on any of the components (other than the one I pasted above from earthly-core). 

However, the header I'm setting here is a [standard protocol](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md) for controlling gRPC request timeouts and is supported by [gRPC-Gateway](https://github.com/grpc-ecosystem/grpc-gateway/blob/6d2b64e3a9edc3d206345280e594703a7d4c5543/runtime/context.go#L39-L41). 

I deleted a few satellites with this fix and haven't been able to get the error again.

I was also getting timed-out around the 1 minute mark before having this header, but now it seems to wait much longer without printing the error, some screenshots below of the requests with a timer:

Before:
Screen Shot 2022-06-21 at 11 00 40 AM

After:
Screen Shot 2022-06-21 at 11 03 48 AM

Long story short: I'm actually not 100% sure this is the entire fix, but the evidence seems to show it works.

### DIFF
--- a/cloud/client.go
+++ b/cloud/client.go
@@ -35,7 +35,10 @@ var ErrUnauthorized = errors.New("unauthorized")
 // ErrNoAuthorizedPublicKeys occurs when no authorized public keys are found
 var ErrNoAuthorizedPublicKeys = errors.New("no authorized public keys found")
 
-const tokenExpiryLayout = "2006-01-02 15:04:05.999999999 -0700 MST"
+const (
+	tokenExpiryLayout    = "2006-01-02 15:04:05.999999999 -0700 MST"
+	satelliteMgmtTimeout = "5M" // 5 minute timeout when launching or deleting a Satellite
+)
 
 // OrgDetail contains an organization and details
 type OrgDetail struct {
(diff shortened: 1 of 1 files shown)


cloud/client.go
```diff
@@ -993,7 +993,8 @@ func (c *client) LaunchSatellite(name, orgID string) (*SatelliteInstance, error)
 		Name:     name,
 		Platform: "linux/amd64", // TODO support arm64 as well
 	}
-	status, body, err := c.doCall("POST", "/api/v0/satellites", withAuth(), withJSONBody(&req))
+	status, body, err := c.doCall("POST", "/api/v0/satellites",
+		withAuth(), withHeader("Grpc-Timeout", "5M"), withJSONBody(&req))
```
alexcb, line 997: can you move the `"5M"` to a variable under the client struct, e.g. `satelliteMgmtTimeout`, that way we can update this value in a single location if we need to change it.

```suggestion
		withAuth(), withHeader("Grpc-Timeout", c.satelliteMgmtTimeout), withJSONBody(&req))
```

or do you foresee us having to configure these independently?

brandonSc: I can do that. In theory, they could be different since the delete is a bit longer than the launch. But I think it's just simpler to leave them the same for now.

brandonSc: Preferred setting it as a constant especially since we were already using a constant for a similar type of thing

PR #1949, opened 2022-06-21
release.sh: fix if statement

alexcb:


alexcb:
@dchw is working on a large fix that may incorporate this fix; i'll leave this open for now in case this PR becomes invalidated.

PR #1951, opened 2022-06-21
Clarify that satellite is automatically selected after launch

vladaionescu:


cmd/earthly/main.go
```diff
@@ -3395,6 +3395,7 @@ func (app *earthlyApp) actionSatelliteLaunch(c *cli.Context) error {
 	}
 
 	app.printSatellites(satellites, orgID)
```
brandonSc, line 3397: Maybe we don't need to print the satellite list then. That was to show that your new satellite was selected, but seems it wasn't as obvious as an explicit message

PR #1953, opened 2022-06-22
Fix AMI Release

dchw:
* AMI Release now works in the script (added this missing `$` :disappointed: ).
* Changed up the targets in the Earthfile to work with the new MFA on our AWS account.
* Small Terraform fixes to the build component.

I hand-ran just these targets to build v0.6.14-17, inclusive.

### DIFF
--- a/release/ami/Earthfile
+++ b/release/ami/Earthfile
@@ -1,60 +1,51 @@
 VERSION 0.6
 
 ami-base:
-    FROM --platform=linux/amd64 ubuntu:22.04
+    FROM --platform=linux/amd64 ../common-repo+aws
 
     DO +ADD_APT_REQS
     DO +ADD_YQ
-    DO +ADD_AWSCLI
     DO +ADD_TERRAFORM
 
     WORKDIR ami
-
-GET_VERSIONS:
-    COMMAND
-
-    RUN --no-cache \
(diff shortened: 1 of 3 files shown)


release/ami/Earthfile
```diff
@@ -1,94 +1,80 @@
...
+        terraform workspace select $ENVIRONMENT && \
+        terraform refresh > /dev/null && \ # Do not print all variables to the logs. Terraform is one noisy fella
+        export PIPELINE_ARN=$(terraform output -json | yq .pipelines.value.[env\(RELEASE_TAG\)].arn) && \
         aws imagebuilder start-image-pipeline-execution --image-pipeline-arn $PIPELINE_ARN
 
 ADD_APT_REQS:
     COMMAND
-    RUN apt update && \
-        apt install -y \
+    RUN yum update && \
```
alexcb, line 76: if this is changed to `yum`, then should the command be changed from `ADD_APT_REQS` to `ADD_YUM_REQS`?

alexcb (approved):
just a minor nit, looks good otherwise.

release/ami/Earthfile
dchw: You are right! Ill fix that since it automerged :(

PR #1958, opened 2022-06-22
move earthly-entrypoint.sh to seperate cgroup

alexcb:
if the earthly is run via the earthly/earthly docker image, it
will first run earthly-entrypoint.sh as the root process (with PID=1),
which will then start up the buildkitd/entrypoint.sh script as a sub
process. That subprocess will then setup earthly and buildkit cgroups
but will fail with `sh: write error: Resource busy` while attempting to
enable pids via `echo +pids > /sys/fs/cgroup/cgroup.subtree_control`.

That fails because `/sys/fs/cgroup/cgroup.procs` contains both the
current process **and** it's parent process. The solution is to ensure
the parent process is in a different cgroup.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to [Earthly](https://github.com/earthly/earthly) will be doc
 
 ## Unreleased
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


vladaionescu (approved):
Add release notes entry too?

BTW should we document anything about this behavior?

PR #1963, opened 2022-06-24
Confirm that the user wants to, before asking the specific SSH key

vladaionescu:
CC @dchw 

This came up in a call with a user. They were confused about why the SSH key was necessary. They did not realize it could be used as a form of login. This PR adds a prompt, to clarify the intention of how an SSH key would be used (and to give the user another chance to skip this step).

### DIFF
--- a/cmd/earthly/main.go
+++ b/cmd/earthly/main.go
@@ -2227,32 +2227,38 @@ func (app *earthlyApp) actionRegister(c *cli.Context) error {
 			rawAccept = "n"
 		}
 		accept := strings.ToLower(rawAccept)[0]
-
-		interactiveAccept = accept == 'y'
+		interactiveAccept = (accept == 'y')
 	}
 	termsConditionsPrivacy := app.termsConditionsPrivacy || interactiveAccept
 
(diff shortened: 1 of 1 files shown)


cmd/earthly/main.go
```diff
@@ -2227,32 +2227,38 @@ func (app *earthlyApp) actionRegister(c *cli.Context) error {
...
-			if i < 0 || i > len(publicKeys) {
-				return errors.Errorf("invalid key number")
+			rawIsRegisterSSHKey := promptInput("Would you like to register an SSH key to be used as a form of login? [Y/n]: ")
+			if rawIsRegisterSSHKey == "" {
+				rawIsRegisterSSHKey = "y"
 			}
-			if i > 0 {
-				publicKey = publicKeys[i-1].String()
+			isRegisterSSHKey := (strings.ToLower(rawIsRegisterSSHKey)[0] == 'y')
+			if isRegisterSSHKey {
```
vladaionescu, line 2242: This is the only actual change

brandonSc (approved):

PR #1966, opened 2022-06-27
include binary build source in version

alexcb:
homebrew builds will be passing `-X main.BuiltBy=homebrew`, which will
make it easier to pinpoint issues that may differ between homebrew and official-binary builds.


alexcb:
related to upcoming change in https://github.com/Homebrew/homebrew-core/pull/104083

PR #1968, opened 2022-06-28
Add `--satellite` flag to be used with builds. 

vladaionescu:
This PR enables the following usages:

* `earthly --sat=something +build` (run this specific build on a satellite)
* `earthly --no-sat +build` (run this specific build locally)

This PR also, unhides the `earthly satellite` subcommand.

### DIFF
--- a/cmd/earthly/main.go
+++ b/cmd/earthly/main.go
@@ -1298,6 +1298,9 @@ func (app *earthlyApp) configureSatellite(c *cli.Context, cc cloud.Client) error
 	if c.IsSet("buildkit-host") && c.IsSet("satellite") {
 		return errors.New("cannot specify both buildkit-host and satellite")
 	}
+	if c.IsSet("satellite") && app.noSatellite {
+		return errors.New("cannot specify both no-satellite and satellite")
+	}
 	if !app.isUsingSatellite(c) || cc == nil {
 		// If the app is not using a cloud client, or the command doesn't interact with the cloud (prune, bootstrap)
 		// then pretend its all good and use your regular configuration.
--- a/cmd/earthly/main.go
+++ b/cmd/earthly/main.go
@@ -612,7 +612,7 @@ func newEarthlyApp(ctx context.Context, console conslogging.ConsoleLogger) *eart
 		&cli.StringFlag{
 			Name:        "server",
 			Value:       "https://api.earthly.dev",
-			EnvVars:     []string{"EARTHLY_SERVER"},
+			EnvVars:     []string{"EARTHLY_SERVER_ADDRESS"},
 			Usage:       "API server override for dev purposes",
 			Destination: &app.apiServer,
 			Hidden:      true, // Internal.
--- a/cmd/earthly/main.go
+++ b/cmd/earthly/main.go
@@ -197,15 +197,15 @@ func main() {
 	startTime := time.Now()
 	ctx := context.Background()
 	ctx, cancel := context.WithCancel(ctx)
-	c := make(chan os.Signal, 1)
-	signal.Notify(c, syscall.SIGINT, syscall.SIGTERM)
+	sigChan := make(chan os.Signal, 1)
+	signal.Notify(sigChan, syscall.SIGINT, syscall.SIGTERM)
 	defer func() {
-		signal.Stop(c)
+		signal.Stop(sigChan)
 		cancel()
 	}()
 	go func() {
 		receivedSignal := false
-		for sig := range c {
+		for sig := range sigChan {
 			cancel()
 			if receivedSignal {
 				// This is the second time we have received a signal. Quit immediately.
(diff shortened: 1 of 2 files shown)


cmd/earthly/main.go
```diff
@@ -594,9 +618,9 @@ func newEarthlyApp(ctx context.Context, console conslogging.ConsoleLogger) *eart
 			Hidden:      true, // Internal.
 		},
 		&cli.StringFlag{
-			Name:        "satellite",
+			Name:        "satellite-address",
 			Value:       containerutil.SatelliteAddress,
-			EnvVars:     []string{"EARTHLY_SATELLITE"},
+			EnvVars:     []string{"EARTHLY_SATELLITE_ADDRESS"},
```
dchw, line 623: If we want `_ADDRESS` here, can we add it to `EARTHLY_SERVER` too?

cmd/earthly/main.go
```diff
@@ -1271,24 +1294,31 @@ func (app *earthlyApp) before(context *cli.Context) error {
...
 
 	// Set up extra settings needed for buildkit RPC metadata
-	app.satelliteName = app.cfg.Satellite.Name
-	app.satelliteOrg = app.cfg.Satellite.Org
+	if app.satelliteName == "" {
+		app.satelliteName = app.cfg.Satellite.Name
+	}
+	if app.satelliteOrg == "" {
+		app.satelliteOrg = app.cfg.Satellite.Org
+	}
```
dchw, line 1316: When would we encounter this new condition? It shouldn't be possible to land here without those being set? And if we do, and they are empty, setting them yields the same result?

cmd/earthly/main.go
```diff
@@ -1316,12 +1346,19 @@ func (app *earthlyApp) configureSatellite(cc cloud.Client) error {
...
-	return len(app.cfg.Satellite.Name) > 0
+func (app *earthlyApp) isUsingSatellite(c *cli.Context) bool {
+	if app.noSatellite {
+		return false
+	}
+	if c.IsSet("buildkit-host") {
+		// buildkit-host takes precedence
+		return false
+	}
+	return len(app.cfg.Satellite.Name) > 0 || app.satelliteName != ""
```
dchw, line 1357: Shouldn't the signal here be both org and name, now that we've extended it into that tuple to specify a satellite name?

cmd/earthly/main.go
```diff
@@ -1316,12 +1346,19 @@ func (app *earthlyApp) configureSatellite(cc cloud.Client) error {
...
 }
 
 func (app *earthlyApp) GetBuildkitClient(c *cli.Context, cc cloud.Client) (*client.Client, error) {
-	err := app.configureSatellite(cc)
+	err := app.configureSatellite(c, cc)
```
dchw, line 1361: Honestly, `c` and `cc` is unreadable for me :( But I may be in the minority

dchw (approved):
If it works, LGTM

cmd/earthly/main.go
alexcb: i was going to suggest `ctx`, but it's not the same `context.Context` as one would expect, so i'm fine with `c` in this case.

dchw: `cliCtx` and `cloudClient` :D

cmd/earthly/main.go
vladaionescu: Org isn't strictly necessary - for example, if a user is only part of one org they can specify only the satellite name and we'll default to that org automatically.

alexcb (commented):
it's not obvious if an error will be (or should be) displayed if a user specifies both `--satellite` and `--no-satellite`.

vladaionescu:
> it's not obvious if an error will be (or should be) displayed if a user specifies both --satellite and --no-satellite.

Good point

cmd/earthly/main.go
dchw: Cool. Initial thoughts was that this was for guards.

Man, we need to centralize/clean up this config override stuff and initialization stuff so its not spread out like this...

cmd/earthly/main.go
dchw: It is? Shouldnt that be accounted for here if org is implicit like that? What if they use `--sat` with no org or login?

brandonSc: nit: check for empty-string the same way on both conditions?

```
return app.cfg.Satellite.Name != ""  || app.satelliteName != ""
```

brandonSc (approved):
Good quality of life improvement when using Satellites 👍

cmd/earthly/main.go
vladaionescu: So this is the error shown when an org is not specified and cannot be deduced:

Screen Shot 2022-06-28 at 12 00 44 PM

The implicitness happens in `getSatelliteOrgID`, which is called a little lower in the same function.

With this implementation, it is allowed to skip specifying the org in the following two cases:

1. User is part of only a single org
2. A satellite is selected, in which case we use the same org of the selected satellite

cmd/earthly/main.go
vladaionescu: Done.

Issue #1899, opened 2022-05-25
Authorization not working with `--use-registry-for-with-docker`


ingwarsw:
When using `WITH DOCKER` together with `--use-registry-for-with-docker` authorisation is not working in that "internal" buildkit pull.

We are using Google artifact registry and without `--use-registry-for-with-docker` it works but with it we have
```
unauthorized: You don't have the needed permissions to perform this operation, and you may have invalid credentials. To authenticate your request, follow the steps in: https://cloud.google.com/container-registry/docs/advanced-authentication
```
I even tried to use hack ..
```
test-db-image:
    FROM gcr.io/db-image:master
    SAVE IMAGE something-random

it-tests:
    WITH DOCKER \
            --compose docker-compose.yml \
            --load gcr.io/db-image:master=+test-db-image
         RUN ...
    END
```

But it still seems to load in inside buildkit registry..

PR #1969, opened 2022-06-28
Fix retagging when using `WITH DOCKER` registry-based loading

vladaionescu:
Fixes https://github.com/earthly/earthly/issues/1899

### DIFF
--- a/ast/tests/with-docker.ast.json
+++ b/ast/tests/with-docker.ast.json
@@ -64,6 +64,22 @@
         {
           "command": {
             "args": [
+              "+docker-pull-test-long"
+            ],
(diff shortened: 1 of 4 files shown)


mikejholly (approved):
LGTM

ingwarsw:
@vladaionescu You should not add test with `WITH DOCKER --compose docker-compose.yml`?
Thats how I found it..

vladaionescu:
This should cover the compose issue you were seeing too. Behind the scenes, a similar pull was taking place as the one used with compose. The tests I added simply narrow down to the actual issue underneath.

PR #1976, opened 2022-06-30
Adds new client methods for project, secret, & org invite endpoints.

mikejholly:
This PR adds support for the new projects, secrets, and org endpoints in Earthly Cloud. These will act as the basis for the forthcoming command changes.

### DIFF
--- a/cloud/client.go
+++ b/cloud/client.go
@@ -26,6 +26,7 @@ import (
 	"github.com/golang/protobuf/jsonpb"
 	"github.com/golang/protobuf/proto"
 	"github.com/golang/protobuf/ptypes"
+	"github.com/google/uuid"
 	"github.com/pkg/errors"
 	"golang.org/x/crypto/ssh/agent"
 )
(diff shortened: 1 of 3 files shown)


cloud/client.go
```diff
@@ -1111,6 +1165,385 @@ func (c *client) DeleteSatellite(ctx context.Context, name, orgID string) error
...
+			CreatedAt:  secret.CreatedAt.AsTime(),
+			ModifiedAt: secret.ModifiedAt.AsTime(),
+		})
+	}
+
+	return secrets, nil
+}
+
+// SetSecret adds or updates the given path and secret combination.
+func (c *client) SetSecret(ctx context.Context, path string, secret []byte) error {
```
alexcb, line 1400: What's the difference between this new function and the existing one we have for setting secrets: https://github.com/earthly/earthly/blob/42839f02736cc1e486628915afef9b1304e78adb/cloud/client.go#L714-L730

cloud/client.go
```diff
@@ -1111,6 +1165,385 @@ func (c *client) DeleteSatellite(ctx context.Context, name, orgID string) error
...
+
+// AddProjectMember adds a new member to the project by email or user ID.
+func (c *client) AddProjectMember(ctx context.Context, orgName, name, idOrEmail, permission string) error {
+	u := fmt.Sprintf("/api/v0/projects/%s/%s", orgName, name)
+
+	req := &secretsapi.AddProjectMemberRequest{
+		Permission: permission,
+	}
+
+	if userID, err := uuid.Parse(idOrEmail); err == nil && userID != uuid.Nil {
```
alexcb, line 1290: Is it possible to move this `idOrEmail` logic into a separate function, and then extend all the project functions (e.g. `UpdateProjectMember`, `ListProjectMembers`, etc), to all accept an email or ID?

alexcb (commented):
looking good, just a couple questions.

cloud/client.go
mikejholly: So @vladaionescu asked me to keep the old code around and add the new methods/commands in parallel until we're ready to cut over. The new `SetSecret` method is just for the new project-based secret update endpoint. Perhaps some sort of prefix would help clarify, but I don't think the older code will be around for too much longer.

cloud/client.go
mikejholly: I think that's only used for that one endpoint/method. The other ones assume a userID, which is returned from the member list endpoint.

vladaionescu (commented):
It's a bit difficult to tell which methods are new and which ones are the ones that will be deprecated. I would split this file up in the following way (or something along these lines):

* `oldclient.go` - keep all the old APIs here
* `org.go` - org-related methods
* `project.go` - project-related methods
* `secret.go` - secret-related methods
* `account.go` - login, account registration and keys/tokens management methods
* `common.go` - anything generic that doesn't fit in any of the above

I think this would make this API more manageable as it's already growing out of proportion for a single file.

vladaionescu (approved):

PR #1977, opened 2022-07-01
Split up Cloud client

mikejholly:
This PR splits the huge `client.go` file into sub-files by functionality.

* `client.go` contains the base client and interface
* `auth.go` contains credential and auth code
* `account.go` contains all `/account` methods
* `project.go` contains all `/projects` methods
* `analytics.go` contains all `/analytics` methods
* `log.go` contains all `/logs` methods
* `org.go` contains all `/organizations` methods
* `org_new.go` contains all new org endpoint code
* `satellite.go` contains all `/satellites` methods
* `secret.go` contains all `/secrets` methods
* `secret_new.go` contains all new secret endpoint code

I used `_new.go` files since it's not yet 100% clear which methods will be deprecated. Some methods just require minor changes (which will be copied/modified into `_new.go` files). But, with this, at least we can keep the code cleanly delineated until we remove the old methods.


vladaionescu (approved):
You might need to add godoc. But this looks good :+1:

PR #1982, opened 2022-07-04
implement WAIT/END SAVE IMAGE logic

alexcb:
This implements exporting images to the local docker instance;
previously only `SAVE IMAGE --push` logic was implemented.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -206,6 +206,12 @@ jobs:
       EARTHLY_INSTALL_ID: "earthly-githubactions"
     steps:
       - uses: earthly/actions-setup@v1.0.1
+      - name: Set up QEMU
+        id: qemu
+        uses: docker/setup-qemu-action@v1
+        with:
+          image: tonistiigi/binfmt:latest
+          platforms: all
       - uses: actions/checkout@v3
       - uses: actions/setup-go@v1
         with:
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -92,11 +92,10 @@ type BuildOpt struct {
 
 // Builder executes Earthly builds.
 type Builder struct {
-	s           *solver
-	opt         Opt
-	resolver    *buildcontext.Resolver
-	builtMain   bool
-	pullPingMap *gatewaycrafter.PullPingMap
+	s         *solver
+	opt       Opt
+	resolver  *buildcontext.Resolver
+	builtMain bool
 
 	outDirOnce sync.Once
 	outDir     string
@@ -115,9 +114,8 @@ func NewBuilder(ctx context.Context, opt Opt) (*Builder, error) {
 			enttlmnts:       opt.Enttlmnts,
 			saveInlineCache: opt.SaveInlineCache,
 		},
-		opt:         opt,
-		pullPingMap: gatewaycrafter.NewPullPingMap(),
-		resolver:    nil, // initialized below
+		opt:      opt,
+		resolver: nil, // initialized below
 	}
 	b.resolver = buildcontext.NewResolver(opt.SessionID, opt.CleanCollection, opt.GitLookup, opt.Console, opt.FeatureFlagOverrides)
 	return b, nil
@@ -140,6 +138,7 @@ func (b *Builder) convertAndBuild(ctx context.Context, target domain.Target, opt
 		manifestLists         = make(map[string][]manifest) // parent image -> child images
 		platformImgNames      = make(map[string]bool)       // ensure that these are unique
 		singPlatImgNames      = make(map[string]bool)       // ensure that these are unique
+		pullPingMap           = gatewaycrafter.NewPullPingMap()
 	)
 	var (
 		depIndex   = 0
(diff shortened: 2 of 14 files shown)


.github/workflows/ci.yml
```diff
@@ -206,6 +206,12 @@ jobs:
       EARTHLY_INSTALL_ID: "earthly-githubactions"
     steps:
       - uses: earthly/actions-setup@v1.0.1
+      - name: Set up QEMU
```
alexcb, line 209: qemu is required for `save-multi-platform-image/test.sh`

earthfile2llb/wait_block.go
```diff
@@ -147,13 +149,47 @@ func (wb *waitBlock) saveImages(ctx context.Context) error {
...
+		if item.localExport {
+			if isMultiPlatform[item.si.DockerTag] {
+				// local docker instance does not support multi-platform images, so we must create a new entry and set it to the platformImgName
+				refPrefix, err := gwCrafter.AddPushImageEntry(ref, refID, platformImgName, false, false, item.si.Image, nil)
+				if err != nil {
+					return err
+				}
+				if item.c.opt.UseLocalRegistry {
+					localRegPullID := pullPingMap.Insert(sessionID, platformImgName)
+					gwCrafter.AddMeta(fmt.Sprintf("%s/export-image-local-registry", refPrefix), []byte(localRegPullID))
```
alexcb, line 178: this logic is duplicated between builder.go and here; I can spend some more time looking into moving it into the gatewaycrafter if you think that's worthwhile.

vladaionescu: I think we should eliminate the duplication soon - but it doesn't have to be in this very PR.

BTW, also ideally nothing adds meta entries or refs to the crafter directly - that part is abstracted away by the crafter. The crafter just offers a way to add images and artifacts (+ settings about how to export them) and deals with what metadata needs to be set accordingly within the implementation, without exposing it.

util/gatewaycrafter/pullpingmap.go
```diff
@@ -0,0 +1,38 @@
...
+func (ppm *PullPingMap) Get(k string) (string, bool) {
+	ppm.m.Lock()
+	defer ppm.m.Unlock()
+	v, ok := ppm.localImages[k]
+	return v, ok
+}
+
+// Insert creates a new entry for the value under sessionID/<uuid>
+func (ppm *PullPingMap) Insert(sessionID, v string) string {
+	k := fmt.Sprintf("sess-%s/pullping:%s", sessionID, uuid.New().String())
```
vladaionescu, line 33: I wonder if we should make this something a little more transparent than UUID. It could help with debuggability.

builder/builder.go
```diff
@@ -114,8 +115,9 @@ func NewBuilder(ctx context.Context, opt Opt) (*Builder, error) {
 			enttlmnts:       opt.Enttlmnts,
 			saveInlineCache: opt.SaveInlineCache,
 		},
-		opt:      opt,
-		resolver: nil, // initialized below
+		opt:         opt,
+		pullPingMap: gatewaycrafter.NewPullPingMap(),
```
vladaionescu, line 119: Could this be one per-build (basically, initialized in `convertAndBuild`) rather than just just global for the entire `Builder`?

vladaionescu: Since the map is so tied conceptually to outputting via the local registry, I wonder if it would be nicer to just tuck this map within the gateway crafter object itself.

The global `onPull` function could be provided during the construction of the crafter (which currently takes place in `convertAndBuild`).

vladaionescu (commented):
Looking good so far! I left some thoughts in the comments...

builder/builder.go
alexcb: > Could this be one per-build (basically, initialized in convertAndBuild) rather than just just global for the entire Builder?

That could work!

> Since the map is so tied conceptually to outputting via the local registry, I wonder if it would be nicer to just tuck this map within the gateway crafter object itself.

This is harder because multiple gatewaycrafter instances are created per session (one for each wait/end block, plus for the final return of the build function), where as the pullpingmap is only created a single time due to there only being a single onPull function created once per session.

If instead there was a way to tell buildkit to call back a different onPull function, we could reference the per-export specific instance of the gatewaycrafter in
https://github.com/earthly/earthly/blob/1c5ebbd5e7d2bcd710352dc51536af988cdc3028/earthfile2llb/wait_block.go#L122
```
        otherOnPullFunc := func(...) {
            gwCrafter.PullPingMap.Get(...)
            ...etc...
        }
	refs, metadata := gwCrafter.GetRefsAndMetadata()
	err := gatewayClient.Export(ctx, gwclient.ExportRequest{
		Refs:     refs,
		Metadata: metadata,
	},
	otherOnPullFunc)
```

but this would require more work in buildkit to allow us to essentially pass a `void*` pointer that the per-session onPull function could use to dispatch back to this otherOnPullFunc.

util/gatewaycrafter/pullpingmap.go
alexcb: possibly, but we need to ensure it's unique.

Adding a reference to the WAIT/END block (e.g. Earthfile and line number), plus an index alone wouldn't guarentee uniqueness since it could be executed multiple times with different build-args.

Any thoughts on an easy-to-implement solution?

earthfile2llb/wait_block.go
alexcb: That's the long-term vision indeed, eventually I want to remove the refPrefix it returns.

builder/builder.go
vladaionescu: Ah gotcha - so the gateway crafter is one per export, whereas the map and the onPull function is one per convertAndBuild call.

util/gatewaycrafter/pullpingmap.go
vladaionescu: Maybe it can be based on the name of the final image, but concatenated with some uniqueness, like the uuid, or an incrementing index.

vladaionescu (approved):

PR #1983, opened 2022-07-04
New Cloud Project commands

mikejholly:
This PR includes the command definitions and callbacks for interacting with Cloud Projects and their members.

Included:
* Create a new project
* Remove a project
* List projects
* Update a project
* List project members
* Add a project member
* Remove a member
* Update a member's permission

Questions:
* Do we want to print success messages for commands that don't otherwise expect output (project create, etc.)?

### DIFF
--- a/cmd/earthly/main.go
+++ b/cmd/earthly/main.go
@@ -1194,6 +1194,7 @@ Set up a whole custom git repository for a server called example.com, using a si
 					Subcommands: []*cli.Command{
 						{
 							Name:        "ls",
+							Aliases:     []string{"list"},
 							Usage:       "List all projects that belong to the specified organization",
 							Description: "List all projects that belong to the specified organization",
 							UsageText:   "earthly project --org <organization-name> ls",
@@ -1201,17 +1202,18 @@ Set up a whole custom git repository for a server called example.com, using a si
 						},
 						{
 							Name:        "rm",
+							Aliases:     []string{"remove", "delete"},
 							Usage:       "Remove an existing project from the organization",
 							Description: "Remove an existing project from the organization",
 							UsageText:   "earthly project --org <organization-name> rm <project-name>",
-							Action:      app.actionProjectList,
+							Action:      app.actionProjectRemove,
 						},
 						{
 							Name:        "create",
 							Usage:       "Create a new project in the specified organization",
 							Description: "Create a new project in the specified organization",
 							UsageText:   "earthly project --org <organization-name> create <project-name>",
-							Action:      app.actionProjectList,
+							Action:      app.actionProjectCreate,
 						},
 						{
 							Name:        "member",
@@ -1223,28 +1225,30 @@ Set up a whole custom git repository for a server called example.com, using a si
 									Usage:       "Add a new member to the specified project",
 									Description: "Add a new member to the specified project",
 									UsageText:   "earthly project --org <organization-name> member add <project-name> <user-id-or-email> <permission>",
-									Action:      app.actionProjectList,
+									Action:      app.actionProjectMemberAdd,
 								},
 								{
 									Name:        "rm",
+									Aliases:     []string{"remove", "delete"},
 									Usage:       "Remove a member from the specified project",
 									Description: "Remove a member from the specified project",
 									UsageText:   "earthly project --org <organization-name> member rm <project-name> <user-id>",
-									Action:      app.actionProjectList,
+									Action:      app.actionProjectMemberRemove,
 								},
 								{
 									Name:        "ls",
+									Aliases:     []string{"list"},
 									Usage:       "List all members in the specified project",
 									Description: "List all members in the specified project",
 									UsageText:   "earthly project --org <organization-name> member ls <project-name>",
-									Action:      app.actionProjectList,
+									Action:      app.actionProjectMemberList,
 								},
 								{
 									Name:        "update",
 									Usage:       "Update the project member's permission",
 									Description: "Update the project member's permission",
 									UsageText:   "earthly project --org <organization-name> member update <project-name> <user-id> <permission>",
-									Action:      app.actionProjectList,
+									Action:      app.actionProjectMemberUpdate,
 								},
 							},
 						},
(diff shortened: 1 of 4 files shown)


cmd/earthly/main.go
```diff
@@ -1176,6 +1176,86 @@ Set up a whole custom git repository for a server called example.com, using a si
...
+							Required: false,
+						},
+					},
+					Subcommands: []*cli.Command{
+						{
+							Name:        "ls",
+							Aliases:     []string{"list"},
+							Usage:       "List all projects that belong to the specified organization",
+							Description: "List all projects that belong to the specified organization",
+							UsageText:   "earthly project --org <organization-name> ls",
```
alexcb, line 1200: should we give a hint that org is optional for most users? e.g.
```suggestion
							UsageText:   "earthly project [--org <organization-name>] ls",
```

cmd/earthly/main.go
```diff
@@ -3664,6 +3744,203 @@ func (app *earthlyApp) actionSatelliteUnselect(cliCtx *cli.Context) error {
...
+		return errors.Wrap(err, "failed to create cloud client")
+	}
+
+	members, err := cloudClient.ListProjectMembers(cliCtx.Context, orgName, projectName)
+	if err != nil {
+		return errors.Wrap(err, "failed to list project members")
+	}
+
+	for _, m := range members {
+		app.console.Printf("%s\t%s\t%s\n", m.UserID, m.UserEmail, m.Permission)
```
alexcb, line 3841: can you use the `tabwriter` package? see `actionOrgListPermissions` for an example of how it is used.

cmd/earthly/main.go
```diff
@@ -1176,6 +1176,86 @@ Set up a whole custom git repository for a server called example.com, using a si
...
+									Description: "Add a new member to the specified project",
+									UsageText:   "earthly project --org <organization-name> member add <project-name> <user-id-or-email> <permission>",
+									Action:      app.actionProjectMemberAdd,
+								},
+								{
+									Name:        "rm",
+									Aliases:     []string{"remove", "delete"},
+									Usage:       "Remove a member from the specified project",
+									Description: "Remove a member from the specified project",
+									UsageText:   "earthly project --org <organization-name> member rm <project-name> <user-id>",
```
alexcb, line 1235: it still feels a bit odd that we can't use an `email` here. Can we create a ticket/issue/task to implement it down the road?

alexcb (commented):
looking good.

I noticed several places where we do `val := cliCtx.Args().Get(n)`, but don't seem to be doing any error handling for cases where a user gave us extra parameters. I think it would be good to error out for cases where we expect `n` but got `n+1` args rather than silently ignoring the extra args.

cmd/earthly/main.go
mikejholly: Can do, although I didn't realize the org would be optional at this point. I figured we would be able to set it with the `PROJECT` command but not until that's ready. Is there a way to set the project without this flag in the current CLI version?

cmd/earthly/main.go
mikejholly: SGTM

cmd/earthly/main.go
mikejholly: We can normalize these commands and the endpoints to use email if that's favorable. I feel a little apprehensive about using email addresses in place of user IDs or names as emails are usually treated as semi-private (PII, I believe), but we're already using them elsewhere so I'm not sure if that's a concern.

mikejholly:
> error handling for cases where a user gave us extra parameters

Sure, I don't mind adding checks for arg length.

cmd/earthly/main.go
brandonSc: We could follow the convention we have on the `satellite` commands - org only needs to be specified if you're a member of more than one. Also we have been using an `EARTHLY_ORG` env for those commands as an alternative to specifying `--org`.

Another thought, we I don't think the UsageText has the square brackets like that to show it's optional for `satellite` commands but it could be nice to update it there as well

cmd/earthly/main.go
```diff
@@ -1176,6 +1176,86 @@ Set up a whole custom git repository for a server called example.com, using a si
...
+							Name:        "ls",
+							Aliases:     []string{"list"},
+							Usage:       "List all projects that belong to the specified organization",
+							Description: "List all projects that belong to the specified organization",
+							UsageText:   "earthly project --org <organization-name> ls",
+							Action:      app.actionProjectList,
+						},
+						{
+							Name:        "rm",
+							Aliases:     []string{"remove", "delete"},
```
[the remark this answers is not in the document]
mikejholly: I saw them for `ls`/`list` so I just tossed them in. I'm fine to remove these aliases if they're deemed superfluous. Also curious what @vladaionescu and @alexcb think.

cmd/earthly/main.go
mikejholly: That seems like a decent heuristic. I can add support for that now and see where things settle with the `PROJECT` command (and other options).

cmd/earthly/main.go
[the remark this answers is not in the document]
brandonSc: We could `remove` and `delete` as aliases to the other `rm` commands as well if we think they are convenient for the user

cmd/earthly/main.go
alexcb: ok, lets punt on it. Maybe we'll need a notion of usernames if emails-are-PII becomes an issue.

cmd/earthly/main.go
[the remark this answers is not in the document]
alexcb: Aliases seem fine, but we should add them everywhere for consistency. Maybe just remove them for this PR and add them in everywhere if we ever get feedback from a user? I have no strong opinion rather than being consistent.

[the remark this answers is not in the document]
vladaionescu: I think less is more in cases like these.

Maybe even for `ls` I would only keep one option, unless it's for backwards compatibility reasons, or for ergonomics reasons. Otherwise, it's just added noise.

cmd/earthly/main.go
```diff
@@ -1176,6 +1176,82 @@ Set up a whole custom git repository for a server called example.com, using a si
...
+						},
+						{
+							Name:        "create",
+							Usage:       "Create a new project in the specified organization",
+							Description: "Create a new project in the specified organization",
+							UsageText:   "earthly project [--org <organization-name>] create <project-name>",
+							Action:      app.actionProjectCreate,
+						},
+						{
+							Name:        "member",
```
vladaionescu, line 1217: For some reason, the help text for `member` doesn't show up...

Screen Shot 2022-07-06 at 2 29 48 PM
Screen Shot 2022-07-06 at 2 32 04 PM

cmd/earthly/main.go
```diff
@@ -1176,6 +1176,82 @@ Set up a whole custom git repository for a server called example.com, using a si
...
+							Description: "Manage project members",
+							UsageText:   "earthly project member (ls|rm|add|update)",
+							Subcommands: []*cli.Command{
+								{
+									Name:        "add",
+									Usage:       "Add a new member to the specified project",
+									Description: "Add a new member to the specified project",
+									UsageText:   "earthly project [--org <organization-name>] member add <project-name> <user-id-or-email> <permission>",
+									Action:      app.actionProjectMemberAdd,
+								},
```
vladaionescu, line 1227: When not providing the right arguments, it would be nice to say what arguments are expected, rather than just the number of them.

Screen Shot 2022-07-06 at 2 33 18 PM

vladaionescu (commented):
The commands don't work for me - is that expected?

```
> ./build/darwin/arm64/earthly preview project --org earthly-technologies ls
Error: failed to list projects: failed to list projects: {"code":5, "message":"Not Found", "details":[]}
```

mikejholly:
@vladaionescu yes, [this Traefik routing change](https://github.com/earthly/cloud/pull/169) needs to be deployed before the commands will work. That should happen tomorrow.

vladaionescu (approved):

cmd/earthly/main.go
mikejholly: Fixed. A description field was missing.

cmd/earthly/main.go
mikejholly: Sure, sounds good.

PR #1985, opened 2022-07-06
New org email invite command

mikejholly:
This PR contains the new "preview" command for the email org invite feature.

```
earthly org invite <email> --name <recipient-name> --org <organization-name> --permission <permission> --message <message>
```

I made `name`, `permission`, and `message` optional arguments, and `permission` will default to `read`. I'm not yet very accustomed to our command format conventions.

### DIFF
--- a/cmd/earthly/main.go
+++ b/cmd/earthly/main.go
@@ -2229,37 +2229,50 @@ func (app *earthlyApp) actionOrgInviteEmail(cliCtx *cli.Context) error {
 		return errors.New("invalid number of arguments provided")
 	}
 
-	userEmail := cliCtx.Args().Get(0)
-	if !strings.Contains(userEmail, "@") {
-		return errors.New("invalid email address")
-	}
-
-	orgName := cliCtx.String("org")
-	if orgName == "" {
-		return errors.New("invalid org name")
-	}
-
-	permission := cliCtx.String("permission")
-	if permission == "" {
-		permission = "read"
-	}
-
 	cloudClient, err := cloud.NewClient(app.apiServer, app.sshAuthSock, app.authToken, app.console.Warnf)
 	if err != nil {
 		return errors.Wrap(err, "failed to create cloud client")
 	}
 
-	invite := &cloud.OrgInvitation{
-		Email:      userEmail,
-		Permission: permission,
-		OrgName:    orgName,
+	orgName, err := projectOrgName(cliCtx, cloudClient)
+	if err != nil {
+		return err
 	}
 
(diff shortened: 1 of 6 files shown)


brandonSc:
I wonder if this would be better as an interactive command since there are lot of options?

Something like:
```
> earthly org invite homer@simpson.con
Enter a name (Empty for no name):
> Homer Simpson
Give write permissions? (y/N): 
... etc
```

mikejholly:
Good idea. I was thinking that as well but wasn't sure if we preferred the explicit arg approach. Thoughts @vladaionescu & @alexcb?

alexcb:
I like the idea of an interactive option, as long as we still allow user's a mechanism to script against the cli.

An additional idea: should we create a confirmation prompt (and allow a bypass when a flag such as `--yes` or `--force` is passed).

alexcb (approved):
code looks good, but perhaps an interactive mode would lead to a better experience.

vladaionescu:
I'll take a look at this later today - I want to play with it hands-on to get a feel for the experience.

mikejholly:
Here's some more detail on the proposed invite flow: https://github.com/earthly/cloud/pull/152

cmd/earthly/main.go
```diff
@@ -1179,9 +1179,48 @@ Set up a whole custom git repository for a server called example.com, using a si
...
+					Name:        "org",
+					Aliases:     []string{"orgs"},
+					Usage:       "Earthly organization administration *experimental*",
+					Description: "Earthly organization administration *experimental*",
+					Subcommands: []*cli.Command{
+						{
+							Name:        "invite",
+							Usage:       "Invite accounts to your organization",
+							Description: "Invite accounts to your organization",
+							UsageText:   "earthly org invite <email> --name <recipient-name> --org <organization-name> --permission <permission> --message <message>",
```
vladaionescu, line 1192: In `earthly preview project` the `--org` flag is passed to the `project` command, whereas in `earthly preview org invite`, the `--org` flag is passed to the `invite` command. Should we make this more consistent?

mikejholly: I was actually just revising this along with the default org logic.

vladaionescu (approved):

mikejholly:
Ok, here's an example of the "interactive" mode. The command will ask for any/all of the missing arguments.

```
➜  earthly git:(mh/org-invite-email) ✗ ./build/linux/amd64/earthly preview org --org mikeorg invite mike@earthly.dev
New user's name: Mike  
New user's permission: read
Message to user: Come join my team!
```

PR #1988, opened 2022-07-07
New project-based secret commands

mikejholly:
This PR adds support for the new project-based secret commands. The baseline secret commands are pretty much equivalent to the previous versions but they make use of the new project-based secret endpoints and include some temporary flags/envs that specify the relevant org and project. It's still not 100% worked out, but the `PROJECT` command will likely replace those flags.

I'll also be replacing user IDs with email addresses in some of the new endpoints and commands in a followup PR: https://linear.app/earthly-technologies/issue/ET-226/standardize-use-of-emails-for-user-identifier

Copied secret commands:

```
earthly preview secret (ls|get|set|rm)
```

New commands:

```
earthly preview secret permission ls <path>
earthly preview secret permission rm <path> <user-id>
earthly preview secret permission set <path> <user-id> <permission>
```

### DIFF
--- a/cmd/earthly/root_cmds.go
+++ b/cmd/earthly/root_cmds.go
@@ -270,6 +270,27 @@ Set up a whole custom git repository for a server called example.com, using a si
 					},
 					Subcommands: app.projectCmds(),
 				},
+				{
+					Name:        "secret",
+					Aliases:     []string{"secrets"},
+					Description: "Manage cloud secrets *experimental*",
+					Usage:       "Manage cloud secrets *experimental*",
+					Flags: []cli.Flag{
+						&cli.StringFlag{
+							Name:     "org",
+							EnvVars:  []string{"EARTHLY_ORG"},
+							Usage:    "The organization to which the project belongs.",
+							Required: true,
+						},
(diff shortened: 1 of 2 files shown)


cmd/earthly/main.go
```diff
@@ -1219,6 +1219,100 @@ Set up a whole custom git repository for a server called example.com, using a si
...
+							Required: true,
+						},
+					},
+					Subcommands: []*cli.Command{
+						{
+							Name:  "set",
+							Usage: "Stores a secret in the secrets store",
+							UsageText: "earthly [options] secret set <path> <value>\n" +
+								"   earthly [options] secret set --file <local-path> <path>\n" +
+								"   earthly [options] secret set --file <local-path> <path>",
```
alexcb, line 1247: should one of these be an example of `--stdin` instead? (currently there's two examples for `--file` which appear to be the same).

alexcb (approved):

cmd/earthly/main.go
mikejholly: Oh, good point. I actually copied this from the existing commands, so I'll fix it in both places.

PR #1989, opened 2022-07-07
Improved Error Handling in Satellites Commands

brandonSc:
* There was a bug where if a request would retry due to an error, it would result in an "unauthorized" error - this was due to the request headers mutating between retries
* Better handle on user hitting ctrl-c when waiting for a satellite to launch or destroy
* Also improved some errors on the backend

### DIFF
--- a/cloud/client.go
+++ b/cloud/client.go
@@ -17,11 +17,12 @@ import (
 	"golang.org/x/crypto/ssh/agent"
 )
 
-// ErrUnauthorized occurs when a user is unauthorized to access a resource
-var ErrUnauthorized = errors.New("unauthorized")
-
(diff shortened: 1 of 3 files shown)


cloud/client.go
```diff
@@ -181,6 +182,13 @@ func (c *client) doCall(ctx context.Context, method, url string, opts ...request
 	for attempt := 0; attempt < maxAttempt; attempt++ {
 		status, body, err = c.doCallImp(ctx, r, method, url, opts...)
 
+		if err != nil && strings.Contains(err.Error(), "context canceled") {
```
alexcb, line 185: nice fix! I've seen this crop up a couple of times which resulted in me having to hit ctrl-c a second time.

PR #1990, opened 2022-07-07
add EARTHLY_PUSH arg

alexcb:
This adds an EARTHLY_PUSH builtin arg which is set to "true" or "false"
which mimics the behaviour of a RUN --push command.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ All notable changes to [Earthly](https://github.com/earthly/earthly) will be doc
 ### Added
 
 - Adding support for saving artifact from `--interactive-keep`. [#1980](https://github.com/earthly/earthly/issues/1980)
+- New `EARTHLY_PUSH` builtin arg, which is set to `true` when earthly is run with the `--push` flag, and the argument
+  is referenced under the direct target, or a target which is indirectly referenced via a `BUILD` command; otherwise
+  it will be set to `false`. The value mimics when a `RUN --push` command is executed. This feature must be enabled with
+  `VERSION --wait-block 0.6`.
 
 ## v0.6.19 - 2022-06-29
 
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -178,6 +178,8 @@ func (b *Builder) convertAndBuild(ctx context.Context, target domain.Target, opt
 				NoCache:              b.opt.NoCache,
 				ContainerFrontend:    b.opt.ContainerFrontend,
 				UseLocalRegistry:     (b.opt.LocalRegistryAddr != ""),
+				DoSaves:              !opt.NoOutput,
+				DoPushes:             opt.Push,
 			}
 			mts, err = earthfile2llb.Earthfile2LLB(childCtx, target, opt, true)
 			if err != nil {
@@ -211,8 +213,8 @@ func (b *Builder) convertAndBuild(ctx context.Context, target domain.Target, opt
 				continue
 			}
 			for _, saveImage := range b.targetPhaseImages(sts) {
-				doSave := (sts.GetDoSaves() || saveImage.ForceSave)
-				if saveImage.DockerTag != "" && doSave {
+				doSaveOrPush := (sts.GetDoSaves() || sts.GetDoPushes() || saveImage.ForceSave)
+				if saveImage.DockerTag != "" && doSaveOrPush {
 					if saveImage.NoManifestList {
 						noManifestListImgs[saveImage.DockerTag] = true
 					} else {
@@ -226,7 +228,7 @@ func (b *Builder) convertAndBuild(ctx context.Context, target domain.Target, opt
 		}
 
 		for _, sts := range mts.All() {
-			hasRunPush := (sts.GetDoSaves() && sts.RunPush.HasState)
+			hasRunPush := (sts.GetDoPushes() && sts.RunPush.HasState)
 			if (sts.HasDangling && !b.opt.UseFakeDep) || (b.builtMain && hasRunPush) {
 				depRef, err := b.stateToRef(childCtx, gwClient, b.targetPhaseState(sts), sts.PlatformResolver)
 				if err != nil {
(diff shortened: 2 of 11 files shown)


builder/builder.go
```diff
@@ -498,7 +502,8 @@ func (b *Builder) convertAndBuild(ctx context.Context, target domain.Target, opt
 			console := b.opt.Console.WithPrefixAndSalt(sts.Target.String(), sts.ID)
 			for _, saveImage := range sts.SaveImages {
 				doSave := (sts.GetDoSaves() || saveImage.ForceSave)
-				shouldPush := opt.Push && saveImage.Push && !sts.Target.IsRemote() && saveImage.DockerTag != "" && doSave
+				doPush := (sts.GetDoPushes() || saveImage.ForceSave)
```
vladaionescu, line 505: What is `ForceSave` used for?

earthfile2llb/earthfile2llb.go
```diff
@@ -77,10 +77,12 @@ type ConvertOpt struct {
 	// AllowPrivileged is used to allow (or prevent) any "RUN --privileged" or RUNs under a LOCALLY target to be executed,
 	// when set to false, it prevents other referenced remote targets from requesting elevated privileges
 	AllowPrivileged bool
-	// DoSaves is used to control when SAVE ARTIFACT AS LOCAL calls will actually output the artifacts locally
-	// this is to differentiate between calling a target that saves an artifact directly vs using a FROM which indirectly
-	// calls a target which saves an artifact as a side effect.
+	// DoSaves controls when SAVE ARTIFACT AS LOCAL, and SAVE IMAGE (to the local docker instance) calls are executed
+	// When a SAVE IMAGE --push is encountered, the image is still pushed to the remote registry, but is not exported to the local docker instance
```
vladaionescu, line 81:
```suggestion
	// When a SAVE IMAGE --push is encountered, the image may still pushed to the remote registry (as long as DoPushes=true), but is not exported to the local docker instance
```

variables/builtin.go
```diff
@@ -22,10 +23,11 @@ type DefaultArgs struct {
 }
 
 // BuiltinArgs returns a scope containing the builtin args.
-func BuiltinArgs(target domain.Target, platr *platutil.Resolver, gitMeta *gitutil.GitMetadata, defaultArgs DefaultArgs, ftrs *features.Features) *Scope {
+func BuiltinArgs(target domain.Target, platr *platutil.Resolver, gitMeta *gitutil.GitMetadata, defaultArgs DefaultArgs, ftrs *features.Features, push bool) *Scope {
 	ret := NewScope()
 	ret.AddInactive(arg.EarthlyTarget, target.StringCanonical())
 	ret.AddInactive(arg.EarthlyTargetProject, target.ProjectCanonical())
+	ret.AddInactive(arg.EarthlyPush, fmt.Sprintf("%t", push))
```
vladaionescu, line 30: We should feature-flag the presence of this builtin arg - mainly because someone could be writing an Earthfile that depends on this and could be sharing with a colleague that has an older Earthly - which would result in different behaviors.

builder/builder.go
alexcb: Good point, I think I might be able to drop it.

It looks like it's just for older version:
```
	// ForceSaveImage is used to force all SAVE IMAGE commands are executed regardless of if they are
	// for a local or remote target; this is to support the legacy behaviour that was first introduced in earthly (up to 0.5)
	// When this is set to false, SAVE IMAGE commands are only executed when DoSaves is true.
```

```
# No Version
pi:
    FROM ubuntu:latest
    RUN apt-get update && apt-get install -y bc
    ARG --required iterations
    RUN seq -f '4/%g' 1 2 $iterations | paste -sd-+ | bc -l > /root/pi
    CMD cat /root/pi
    SAVE ARTIFACT /root/pi
    SAVE IMAGE --push alexcb132/pi:$iterations

indirect:
    FROM +pi
    RUN cat /root/pi
```

running this via: `/home/alex/gh/earthly/earthly/build/linux/amd64/earthly --push +indirect --iterations=11` still results in a push even once I have removed `ForceSave`.

I'll do a few more tests before submitting an update with this removed.

variables/builtin.go
alexcb: Do you think it's sufficient to use the `--wait-block` feature for this, or a separate one?
Technically `wait-block` has already been released, so a user on v0.6.19 could think they are using it, but would need a newer version. However since we say it's experimental, and I print out an extra warning in v0.6.19 saying it's not-fully implemented, I think it's safe-enough to re-use it. Thoughts?

vladaionescu: Yeah, we can re-use it. We can reuse any feature flag, in general until it has actually been released as part of a `VERSION` number. We don't guarantee forwards-backward compatibility if someone uses feature flags directly - so we're good.

builder/builder.go
alexcb: I double checked that

```
test:
    BUILD github.com/alexcb/earthly-push-tests:poorly-hidden-branch+pi --iterations=100
```

doesn't propagate a push, so we should be safe to remove `|| ForceSave` from the condition.

CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ All notable changes to [Earthly](https://github.com/earthly/earthly) will be doc
 ### Added
 
 - Adding support for saving artifact from `--interactive-keep`. [#1980](https://github.com/earthly/earthly/issues/1980)
+- New `EARTHLY_PUSH` builtin arg, which is set to `true` when earthly is run with the `--push` flag, and the argument
+  is referenced under the direct target, or a target which is indirectly referenced via a `BUILD` command; otherwise
+  it will be set to `false`. The value mimics when a `RUN --push` command is executed.
```
vladaionescu, line 16: Add mention that this is behind a feature flag?

earthfile2llb/earthfile2llb.go
```diff
@@ -77,10 +77,13 @@ type ConvertOpt struct {
 	// AllowPrivileged is used to allow (or prevent) any "RUN --privileged" or RUNs under a LOCALLY target to be executed,
 	// when set to false, it prevents other referenced remote targets from requesting elevated privileges
 	AllowPrivileged bool
-	// DoSaves is used to control when SAVE ARTIFACT AS LOCAL calls will actually output the artifacts locally
-	// this is to differentiate between calling a target that saves an artifact directly vs using a FROM which indirectly
-	// calls a target which saves an artifact as a side effect.
+	// DoSaves controls when SAVE ARTIFACT AS LOCAL, and SAVE IMAGE (to the local docker instance) calls are executed
+	// When a SAVE IMAGE --push is encountered, the image may still pushed to the remote registry (as long as DoPushes=true), but is
```
vladaionescu, line 81: (Sorry, this was my error :-) )

```suggestion
	// When a SAVE IMAGE --push is encountered, the image may still be pushed to the remote registry (as long as DoPushes=true), but is
```

vladaionescu (approved):

Issue #1991, opened 2022-07-08
Sometimes failures are reported as `context.Canceled` rather than the actual root cause


vladaionescu:
Have been seeing this lately - I think it's a race condition in propagating the error status somewhere in our code.

PR #2001, opened 2022-07-08
Fix errors frequently being reported as context cancelled

vladaionescu:
Fixes https://github.com/earthly/earthly/issues/1991


alexcb (approved):
nice find!

PR #2012, opened 2022-07-13
Add support for PROJECT statement

mikejholly:
This PR adds lexing, parsing, and interpreter support for the new `PROJECT` statement. The value of the `PROJECT` statement should include the appropriate organization and project in the following format: `PROJECT <organization/project>`.


earthfile2llb/interpreter.go
```diff
@@ -1587,6 +1589,27 @@ func (i *Interpreter) handleImport(ctx context.Context, cmd spec.Command) error
...
 }
 
+func (i *Interpreter) handleProject(ctx context.Context, cmd spec.Command) error {
+	projectVal, err := i.expandArgs(ctx, cmd.Args[0], false, false)
+	if err != nil {
+		return i.wrapError(err, cmd.SourceLocation, "failed to expand PROJECT %s", cmd.Args[0])
+	}
+
+	if !strings.Contains(projectVal, "/") {
+		return errors.New("format for PROJECT statement should be: <organization>/<project>")
```
vladaionescu, line 1599: We should use `i.errorf` for this. It adds source location metadata to the error.

earthfile2llb/interpreter.go
```diff
@@ -1587,6 +1589,27 @@ func (i *Interpreter) handleImport(ctx context.Context, cmd spec.Command) error
...
+	}
+
+	parts := strings.Split(projectVal, "/")
+	if len(parts) != 2 {
+		return errors.New("unexpected format for PROJECT statement")
```
vladaionescu, line 1604: Same here

earthfile2llb/interpreter.go
```diff
@@ -1587,6 +1589,27 @@ func (i *Interpreter) handleImport(ctx context.Context, cmd spec.Command) error
...
+	}
+
+	i.converter.org = parts[0]
+	i.converter.project = parts[1]
```
vladaionescu, line 1608: We try not to use private members of the converter from here - mainly to help create separation. I think we should just have a `Project` method in the converter that sets these. (We typically name the methods in converter by the name of the commands they implement)

examples/project/Earthfile
```diff
@@ -0,0 +1,7 @@
+VERSION 0.6
+PROJECT foo/bar
+
+FROM alpine:3.15
+
+hello:
+    RUN echo "Hello World"
```
vladaionescu, line 7: I think that this is not a strong enough example of how to use the `PROJECT` keyword. I wouldn't add this in, to keep our examples clean.

vladaionescu (commented):
Looks good overall :+1:

Just some nits...

examples/project/Earthfile
mikejholly: SGTM. Perhaps I'll revisit for the new secret fetching and feature flag.

earthfile2llb/interpreter.go
```diff
@@ -1587,6 +1589,26 @@ func (i *Interpreter) handleImport(ctx context.Context, cmd spec.Command) error
 	return nil
 }
 
+func (i *Interpreter) handleProject(ctx context.Context, cmd spec.Command) error {
+	projectVal, err := i.expandArgs(ctx, cmd.Args[0], false, false)
+	if err != nil {
+		return i.wrapError(err, cmd.SourceLocation, "failed to expand PROJECT %s", cmd.Args[0])
+	}
+
+	if !strings.Contains(projectVal, "/") {
```
alexcb, line 1598: nit: this check is unnessisary as it's covered below in the `if len(parts) != 2` check.

I prefer the error message used here though, maybe just simply do:

```
	parts := strings.Split(projectVal, "/")
	if len(parts) != 2 {
		return i.errorf(cmd.SourceLocation, "unexpected format for PROJECT statement, should be: <organization>/<project>")
	}
```

alexcb (commented):
looks good, i'll leave it up to Vlad for the final +:1:

vladaionescu (approved):

earthfile2llb/interpreter.go
mikejholly: SGTM

alexcb (approved):

PR #2016, opened 2022-07-14
Add GC stats

vladaionescu:
Depends on https://github.com/earthly/buildkit/pull/83

This PR adds the following:

* GC stats as part of `earthly debug buildkit-workers` command
* Print a select few GC statistics on startup, if buildkit is remote (or satellite)
* Print warnings on startup if GC is either slow right now, or has been slow historically
* Print out GC info if GC is ongoing right now during `ongoing` output
* Print a warning if the GC is either slow right now, or has been slow historically during `ongoing` output


brandonSc (approved):
I think this will be super helpful!

PR #2017, opened 2022-07-14
Fix use of cache in WITH DOCKER

vladaionescu:
Running simple tests like `./tests/with-docker-registry+docker-pull-test` was not being cached correctly for two reasons:

* The dindID (part of docker data root dir path) contained a sessionID, which meant that it was different each time, and that was passed into the LLB
* The list of intermediate images to be loaded via registry also contained the session ID in the image name

For the first problem, we went back to using a target input hash, and have implemented an algorithm to detect clash of directories, in case two dockerd's try to use the same dind dir. In case of clash, we simply use another directory.

For the second problem, we pass along the list of intermediate images via a secret, to prevent busting the cache (the values of secrets do not contribute to the cache hash).

### DIFF
--- a/buildkitd/dockerd-wrapper.sh
+++ b/buildkitd/dockerd-wrapper.sh
@@ -26,6 +26,11 @@ execute() {
         echo "EARTHLY_DOCKERD_DATA_ROOT not set"
         exit 1
     fi
+    mkdir -p "$EARTHLY_DOCKERD_DATA_ROOT"
+
+    if ! find_free_subdir; then
+        exit 1
+    fi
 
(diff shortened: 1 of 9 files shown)


buildkitd/dockerd-wrapper.sh
```diff
@@ -71,7 +76,30 @@ execute() {
     return "$exit_code"
 }
 
+find_free_subdir() {
```
mikejholly, line 79: Out of curiosity, why use (and test for the existence of) 1-100 instead of using a temp name? I suppose the former limits garbage to clean up. Is that the idea?

mikejholly (approved):
LGTM with a question.

Issue #2015, opened 2022-07-13
Image gets built twice when loaded in `WITH DOCKER` and `--use-registry-for-with-docker` is used


vladaionescu:
```Earthfile
VERSION --use-registry-for-with-docker 0.6
FROM earthly/dind:alpine

image:
  FROM ubuntu:20.04
  RUN apt-get update && apt-get install -y curl
  RUN curl -sL https://deb.nodesource.com/setup_16.x | bash
  RUN apt-get install -y nodejs
  ENV RECORD_REPLAY_API_KEY
  CMD ["/bin/bash", "script.sh"]
  SAVE IMAGE replay-cli-bug

run:
  ARG RECORD_REPLAY_API_KEY
  WITH DOCKER --load replay-cli-bug:latest=+image
    RUN docker run -e RECORD_REPLAY_API_KEY=$RECORD_REPLAY_API_KEY replay-cli-bug
  END
```

Example run https://ci.earthly.dev/logs?logId=e9196dec-f26e-4839-b959-0fbe5e4580d5 (notice line 26 where is says cached, but then it continues to run it later on anyway)

vladaionescu:
Some initial investigation reveals no clear culprit:

* Only one STS is created for the target `+image` - Earthly builds the LLB only once
* There are two solves being triggered for that LLB:
  * `WITH DOCKER` code triggers `forceExecution` for `+image` during `BuildAsync` - this takes place as part of the main build session
  * Then, the `WITH DOCKER` code triggers a `SolveImage` using the `multiImageSolver` for `+image` to export it into `dockerd` of `WITH DOCKER` via the registry - this takes place as part of a secondary build session
* This is not getting triggered without `--use-registry-for-with-docker`

For some reason, Buildkit doesn't notice that these two solves are in fact of the exact same build and should not repeat it. Given that this is a large layer, could it be possible that the GC comes in and sweeps it in-between, thus causing a re-run? It's hard to believe that this edge case is so consistent across so many systems.

The upcoming work on unifying `WITH DOCKER` loads into a single session as the main build might take care of this on its own. We can either wait for that or do some more investigation.

PR #2021, opened 2022-07-15
Fix duplicate execution in `--use-registry-for-with-docker`

vladaionescu:
Fixes https://github.com/earthly/earthly/issues/2015


mikejholly (approved):
LGTM

PR #2023, opened 2022-07-15
Mh/use project secrets

mikejholly:


earthfile2llb/converter.go
```diff
@@ -1645,16 +1647,21 @@ func (c *Converter) internalRun(ctx context.Context, opts ConvertRunOpts) (pllb.
 	runOpts = append(runOpts, llb.WithCustomNamef("%s%s", c.vertexPrefix(opts.Locally, isInteractive, false), commandStr))
 
 	var extraEnvVars []string
+
+	if c.ftrs.UseProjectSecrets && (c.varCollection.Org() == "" || c.varCollection.Project() == "") {
+		return pllb.State{}, errors.New("PROJECT statement is required with --use-project-secrets flag")
+	}
```
vladaionescu, line 1653: I think this shouldn't be strictly required.

In the new implementation, it should be possible to use secrets only locally. When using secrets locally only, the user should not have to create an Earthly account and add the `PROJECT` declaration.

earthfile2llb/converter.go
```diff
@@ -1773,6 +1780,23 @@ func (c *Converter) internalRun(ctx context.Context, opts ConvertRunOpts) (pllb.
...
+func (c *Converter) secretID(name string) string {
+	v := url.Values{}
+	v.Set("name", name)
+	if c.ftrs.UseProjectSecrets {
+		v.Set("v", "1")
+		v.Set("org", c.varCollection.Org())
+		v.Set("project", c.varCollection.Project())
+	} else {
+		v.Set("v", "0")
+	}
```
vladaionescu, line 1796: The order of the query parameters is guaranteed to be consistent, I imagine? This is important for LLB/cache consistency.

earthfile2llb/runmount.go
```diff
@@ -25,7 +23,7 @@ func parseMounts(mounts []string, target domain.Target, ti dedup.TargetInput, ca
 	return runOpts, nil
 }
 
-func parseMount(mount string, target domain.Target, ti dedup.TargetInput, cacheContext pllb.State, platr *platutil.Resolver) ([]llb.RunOption, error) {
+func (c *Converter) parseMount(mount string) ([]llb.RunOption, error) {
```
vladaionescu, line 26: I guess this was necessary... makes sense.

util/llbutil/secretprovider/cloud.go
```diff
@@ -23,12 +26,63 @@ func NewCloudStore(client cloud.Client) secrets.SecretStore {
...
+				match = sec
+				break
+			}
+		}
+		if match == nil {
+			return nil, secrets.ErrNotFound
+		}
+		data = []byte(match.Value)
+	default:
+		return nil, errors.New("invalid secret ID format")
```
vladaionescu, line 84: I'm just realizing that `FROM DOCKERFILE` might have its own secret IDs that will not abide by these rules. I wonder if we should assume they are local-only secrets in such cases?

util/llbutil/secretprovider/cloud.go
```diff
@@ -23,12 +26,63 @@ func NewCloudStore(client cloud.Client) secrets.SecretStore {
...
+	}
+
+	switch q.Get("v") {
+	case "0": // Legacy secret ID format includes the name only
+
+		// This format requires the secret to be prefixed with either <org-name>
+		// or 'user'.
+		if !strings.Contains(name, "/") {
+			return nil, secrets.ErrNotFound
+		}
```
vladaionescu, line 49: I think that this is not right. I think we allowed `/` to not be present for local-only secrets. I could be wrong though? @alexcb to comment?

util/llbutil/secretprovider/cmd.go
```diff
@@ -28,12 +30,17 @@ func (c *cmdStore) GetSecret(ctx context.Context, id string) ([]byte, error) {
...
+		return nil, errors.New("invalid secret ID format")
+	}
+	name := q.Get("name")
+	if name == common.DebuggerSettingsSecretsKey {
 		// the interactive debugger passes config values by abusing secrets,
 		// we must not call the user's secret provider in this case.
 		return nil, secrets.ErrNotFound
 	}
-	cmd := exec.CommandContext(ctx, "/bin/sh", "-c", c.cmd+" "+shellescape.Quote(id))
+	cmd := exec.CommandContext(ctx, "/bin/sh", "-c", c.cmd+" "+shellescape.Quote(name))
```
vladaionescu, line 43: Wondering if org ID and project ID should be exposed as env vars for this? Thoughts @alexcb?

vladaionescu (commented):
Looks good so far! I left some comments below - I think we need to look at local secrets (the ones passed in via `earthly --secret` and its friends) a bit more.

We should also add some tests - I realize that most of the interaction would be with the cloud so that testing would probably be done somewhere else. But the local secrets use-case should be tested now - for both the old version (if we don't already have some) and the new version.

We should also let @alexcb take a look at this since a lot of this intersected with areas he worked in.

earthfile2llb/converter.go
```diff
@@ -1368,8 +1367,9 @@ func (c *Converter) Host(ctx context.Context, hostname string, ip net.IP) error
 
 // Project handles a "PROJECT" command in base target.
 func (c *Converter) Project(ctx context.Context, org, project string) error {
-	c.org = org
-	c.project = project
+	c.varCollection.SetOrg(org)
+	c.varCollection.SetProject(project)
```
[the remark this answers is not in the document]
mikejholly: If `PROJECT` is used without the flag, the old-style secrets should continue to work as expected. It'll just set the org and project without impacting anything else, I believe.

util/llbutil/secretprovider/cloud.go
mikejholly: This code is used to check whether secrets exist in the cloud repository. It shouldn't impact local secrets. Old-style secret names need to be formatted like `/<org>/<secret` if I'm not mistaken.

vladaionescu: Oh ok - yeah, I think we simply use a completely separate secret provider for the local secrets, right?

Conversely, would the user be able to override a cloud secret via a locally passed in secret (without crafting this query param format on their own) in the new implementation?

earthfile2llb/converter.go
mikejholly: I don't think we'll want to force the use of `PROJECT` once `--use-project-secrets` is no longer an optional feature flag. My thinking here was that we should probably enforce `PROJECT` if the feature flag is set for the time being. Once it's no longer optional, we can remove that check. I don't mind removing this check now if that's preferred.

util/llbutil/secretprovider/cloud.go
mikejholly: Yep, it's in `map_store.go`.

Yes, definitely. This change should support all existing behaviors around local/env secrets. The only real difference is that one needs to use `PROJECT` when using the new project-based cloud secrets option.

mikejholly: I believe. That was my intention. I'll add some tests for that case.

alexcb: > I think that this is not right. I think we allowed `/` to not be present for local-only secrets. I could be wrong though? @alexcb to comment?

This check is expected here (as Mike pointed out this code already existed).

util/llbutil/secretprovider/cloud.go
alexcb: I think that's a fair compromise.

Maybe we could implement a way to remap secrets (down the road, if someone requests it) via`FROM DOCKERFILE --secret dockerfilesecret=earthlyprojectsecret` or some such syntax.

util/llbutil/secretprovider/map_store.go
```diff
@@ -15,7 +17,15 @@ func NewMapStore(m map[string][]byte) secrets.SecretStore {
...
-	v, ok := m[id]
+	q, err := url.ParseQuery(id)
+	if err != nil {
+		return nil, errors.New("failed to parse secret ID")
+	}
+
+	if q.Get("name") == "" {
+		return nil, errors.New("name parameter not found")
+	}
+	v, ok := m[q.Get("name")]
```
alexcb, line 28: should org or project be referenced here too?

How would a user deal with two Earthfiles referencing the same secret name; however one Earthfile defines a project, but the other one doesn't -- i.e. one is supplied locally, but the other one is cloud-based.

earthfile2llb/converter.go
```diff
@@ -1672,7 +1678,7 @@ func (c *Converter) internalRun(ctx context.Context, opts ConvertRunOpts) (pllb.
 	if !opts.Locally {
 		// Debugger.
 		secretOpts := []llb.SecretOption{
-			llb.SecretID(common.DebuggerSettingsSecretsKey),
+			llb.SecretID(c.secretID(common.DebuggerSettingsSecretsKey)),
```
alexcb, line 1681: Something feels a bit odd with wrapping this var to potentially include an org and project -- only to then ignore it.

Maybe we should have a `c.secretIDForceLocal()`, or even `c.secretIDHack()`.

This is a nit-pick comment, and eventually the debugger should be reworked to avoid using ports and then this hack will go away. Ultimately if the tests are passing, I'm ok with this change :)

tests/Earthfile
```diff
@@ -300,6 +300,25 @@ secrets-test:
         --should_fail=true \
         --output_contains="unable to lookup secret SECRET3: not found"
 
+project-secrets-test:
```
alexcb, line 303: Where is this target being called from?

where is `project-secrets.earth`, did you forget a `git add`?

alexcb (commented):
This is looking like it's in the right direction.

~~I would like to see a test where we mix multiple Earthfiles where one doesn't have a `PROJECT` and another one does, and try to override them via the `+RUN_EARTHLY`'s `--extra_args` option.~~ this is not needed as it's intentially that we won't support such a use case.

util/llbutil/secretprovider/map_store.go
alexcb: update: after our chat I understand this is by design, and it's not possible to independently override one.

tests/Earthfile
mikejholly: Fixed. Thanks.

alexcb: I'm confused, I don't see a `BUILD +project-secrets-test` anywhere. Where did this get fixed?

util/llbutil/secretprovider/cmd.go
alexcb: update: ignore what I said about `buildtins`, instead I should have said via the `cmd.Env` below.

earthfile2llb/converter.go
```diff
@@ -1773,10 +1779,28 @@ func (c *Converter) internalRun(ctx context.Context, opts ConvertRunOpts) (pllb.
...
+		v.Set("v", "0")
+	}
+	return v.Encode()
+}
+
 func (c *Converter) parseSecretFlag(secretKeyValue string) (secretID string, envVar string, err error) {
 	parts := strings.SplitN(secretKeyValue, "=", 2)
 	if len(parts) == 2 {
 		if strings.HasPrefix(parts[1], "+secrets/") {
+			c.opt.Console.Printf("Deprecation: the '+secrets/' prefix is not required and support for it will be removed in an upcoming release")
```
alexcb, line 1803: can you update the CHANGELOG to mention that the prefix is now optional and will eventually result in an error in a future release

tests/secret-provider-config/test-secret-provider
```diff
@@ -33,4 +33,9 @@ if [ "$1" = "binarysecret" ]; then
     exit 0
 fi
 
+if [ "$1" = "projectsecrets" ]; then
```
alexcb, line 36: nice test :)

tests/secret-provider-config/test.earth
```diff
@@ -1,5 +1,5 @@
-FROM alpine
+FROM alpine
```
alexcb, line 1: nit: is this whitespace change intended?

alexcb (approved):

earthfile2llb/converter.go
mikejholly: Done.

PR #2025, opened 2022-07-15
Fix `WITH DOCKER` registry caching

vladaionescu:
Depends on https://github.com/earthly/buildkit/pull/85

This adds the digests of the images being loaded into `WITH DOCKER` into the LLB, such that if any of them change, the cache gets busted automatically.

@alexcb - note that there are changes in the PullPing callback. I hope it doesn't impact the stuff you're working on!


alexcb (approved):
most interesting! Fortunately I haven't had to touch this area in my current safe artifact related branch.

PR #2032, opened 2022-07-19
docs split options into global and build sections

alexcb:
This change reflects how earthly now splits options under global and
build commands.
This additionally documents the --config, --output, --auth-token, and
--verbose options.


vladaionescu (approved):
Yay!

PR #2034, opened 2022-07-20
Use Two Satellites and Re-Enable PR Tests

brandonSc:


vladaionescu (commented):
Some replacements are missing --no-output. Is that intentional?

BTW, can you also add --strict? --ci actually includes --strict too.

PR #2038, opened 2022-07-20
Allow buildkit to be profiled on :6061 when using `EARTHLY_PROFILE`.

dchw:
This should allow us to profile any problematic builds both client and daemon-side.


buildkitd/buildkitd.pprof.template
```diff
@@ -0,0 +1 @@
+  debugAddress = "0.0.0.0:6060"
```
alexcb, line 1: nit: missing newline.

alexcb (approved):

buildkitd/buildkitd.pprof.template
dchw: I can add it. It just s p r e a d s out the config file in our logs :(

alexcb: maybe `head -c -1 buildkitd.pprof.template` could be used instead?

but that makes things brittle too :(

PR #2039, opened 2022-07-20
Set grpc-timeout Header when using Satellites

brandonSc:
Apparently, it's a [standard field](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md) for controlling timeouts on RPC calls.  

Not sure how much, if at all, this will improve usage on slow internet but it's worth a shot.


mikejholly (approved):

alexcb:
I would be cautious about introducing such a large timeout here that can't be configured. We might trade in one issue for another issue -- e.g. some users would rather fail-fast than keep trying if their network drops to 9600baud.

We need to do more vigorous testing using `tc` to figure out which timeouts should be modified.

brandonSc:
@alexcb if this works, we can move it to the ~/.earthly/config instead?

brandonSc:
Closing since the header didn't actually help with buildkit timeouts.

PR #2043, opened 2022-07-21
Different Behavior for --ci flag When Using Satellites

brandonSc:
The commonly used `--ci` flag is an alias that enables inline-cache related features, however, those flags are usually not useful and can sometimes cause performance issues when using Satellites.

This PR alters the behaviour of --ci when using Satellites so it will not enable inline-cache features.

Screen Shot 2022-07-21 at 3 52 53 PM


cmd/earthly/buildkit.go
```diff
@@ -79,9 +79,21 @@ func (app *earthlyApp) configureSatellite(cliCtx *cli.Context, cloudClient cloud
...
 	app.console.Warnf("The following feature flags are recommended for use with Satellites and will be auto-enabled:")
-	app.console.Warnf("--new-platform, --use-registry-for-with-docker")
+	app.console.Warnf("  --new-platform, --use-registry-for-with-docker")
 	app.console.Warnf("") // newline
 
+	if app.ci {
+		app.console.Warnf("When using Satellites, the --ci flag is an alias for:")
+		app.console.Warnf("  --strict --no-output when running earthly in target mode.")
+		app.console.Warnf("  --strict when running earthly in artifact or image mode.")
+		app.console.Warnf("") // newline
```
vladaionescu, line 89:
```suggestion
		app.console.Warnf("Please note that --use-inline-cache and --save-inline-cache are currently disabled when using Satellites and --ci")
		app.console.Warnf("") // newline
```

This wording is a bit easier to understand

cmd/earthly/buildkit.go
```diff
@@ -79,9 +79,21 @@ func (app *earthlyApp) configureSatellite(cliCtx *cli.Context, cloudClient cloud
...
+		if !app.imageMode && !app.artifactMode {
+			app.noOutput = true
+		}
+		app.strict = true
+		app.ci = false
```
vladaionescu, line 94: Most of this kind of logic takes place in `actionBuildImp` under `build_cmd.go`. I would move this logic over there, to keep things in a single spot and prevent confusion.

(To detect whether you're on a satellite in there you could use this existing variable, which will cover both remote buildkit and satellites `isLocal := containerutil.IsLocal(app.buildkitdSettings.BuildkitAddress)`)

vladaionescu (commented):
Some comments below...

Mind also changing

```
app.console.Warnf("Note: the interactive debugger, interactive RUN commands, and fast output via embedded registry do not yet work on Earthly Satellites.")
```

to

```
app.console.Warnf("Note: the interactive debugger, interactive RUN commands, and inline caching do not yet work on Earthly Satellites.")
```

cmd/earthly/build_cmd.go
```diff
@@ -199,12 +199,29 @@ func (app *earthlyApp) actionBuildImp(cliCtx *cli.Context, flagArgs, nonFlagArgs
...
+
+	isLocal := containerutil.IsLocal(app.buildkitdSettings.BuildkitAddress)
+
+	if !isLocal && app.ci {
+		app.console.Warnf("Please note that --use-inline-cache and --save-inline-cache are currently disabled when using --ci on Satellites or remote Buildkit.")
+		app.console.Warnf("") // newline
+		if !app.imageMode && !app.artifactMode {
+			app.noOutput = true
+		}
+		app.strict = true
```
vladaionescu, line 215: Nit: `noOutput` and `strict` should already be true given the logic in `actionBuild`

vladaionescu (approved):

alexcb (approved):

vladaionescu (approved):

vladaionescu (approved):

PR #2046, opened 2022-07-21
display detailed progress on receving files

alexcb:
When running under `--verbose` mode, display individual files (and
filesize) being copied from containers onto the host.

### DIFF
--- a/buildcontext/provider/provider.go
+++ b/buildcontext/provider/provider.go
@@ -5,14 +5,13 @@ package provider
 
 import (
 	"os"
-	"path"
 	"strings"
 	"sync"
 	"time"
 
 	"github.com/earthly/earthly/conslogging"
+	"github.com/earthly/earthly/util/fsutilprogress"
 
-	"github.com/dustin/go-humanize"
 	"github.com/moby/buildkit/session"
 	"github.com/moby/buildkit/session/filesync"
 	"github.com/pkg/errors"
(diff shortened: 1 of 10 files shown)


buildcontext/provider/provider.go
```diff
@@ -187,8 +156,8 @@ func (bcp *BuildContextProvider) handle(method string, stream grpc.ServerStream)
 		IncludePatterns:   includes,
 		FollowPaths:       followPaths,
 		Map:               dir.Map,
-		VerboseProgressCB: verboseProgressCB,
-	}), progress, verboseProgressCB)
+		VerboseProgressCB: progressCB.Verbose,
```
alexcb, line 159: this is weird: fsutils has both an opt and an arg.

my very quick look at this leads me to this the opt version is ignored, but I just left it as-is for now.

alexcb:
this depends on https://github.com/earthly/buildkit/pull/87

builder/solver.go
```diff
@@ -85,6 +87,11 @@ func (s *solver) newSolveOptMulti(ctx context.Context, eg *errgroup.Group, onIma
 		cacheExports = append(cacheExports, newInlineCacheOpt())
 	}
 
+	verboseProgressConsole := console.WithPrefixAndSalt("output",
+		"local context .", // TODO this salt must be the same as the salt used in SolverMonitor.processStatus
+	)
```
vladaionescu, line 92: BTW, I don't think this has to use the same salt - because the prefix is anyway different (context vs output). I would just use `"output", "output"` for now.

vladaionescu (approved):

builder/solver.go
alexcb: When I did that I ended up with different coloured prefixes, which is what led me to use the `WithPrefixAndSalt` function.

vladaionescu: Different from what? Can you say more?

alexcb: e.g. one was orange, the other was blue.

util/fsutilprogress/progress.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+func (s *progressCallback) Verbose(relPath string, status fsutil.VerboseProgressStatus, numBytes int) {
+	s.mutex.Lock()
+	defer s.mutex.Unlock()
+	fullPath := path.Join(s.pathPrefix, relPath)
+	switch status {
+	case fsutil.StatusStat:
+		s.numStats++
+		//s.console.VerbosePrintf("sent file stat for %s\n", fullPath) ignored as it is too verbose. TODO add different verbose levels to support ExtraVerbosePrintf
```
vladaionescu, line 52: Is this leftover intended?

vladaionescu (approved):

util/fsutilprogress/progress.go
alexcb: it's useful for super-verbose debugging during development.

vladaionescu (approved):

PR #2050, opened 2022-07-22
New client methods for org members

mikejholly:
This PR adds new subcommands for organization members (listing, updating, and removing). Adding members requires using the `invite` command (already committed). The code also includes extensions to the Cloud client that are used by the new commands.

New commands:

```
$ earthly preview orgs --org mikeorg members ls
alex@earthly.dev  read
mike@earthly.dev  admin
mike@earthly.dev  write

$ earthly preview orgs --org mikeorg members rm mike@earthly.dev
Member 'mike@earthly.dev' removed successfully

$ earthly preview orgs --org mikeorg members update mike@earthly.dev read
Member 'mike@earthly.dev' updated successfully
```


go.mod
```diff
@@ -42,7 +42,6 @@ require (
 replace (
 	github.com/docker/docker => github.com/docker/docker v20.10.3-0.20220414164044-61404de7df1a+incompatible
 	github.com/jessevdk/go-flags => github.com/alexcb/go-flags v0.0.0-20210722203016-f11d7ecb5ee5
-	github.com/moby/buildkit => github.com/earthly/buildkit v0.0.1-0.20220721222420-7a6f9e1ab2a3
-	github.com/tonistiigi/fsutil => github.com/earthly/fsutil v0.0.0-20220719234708-392da7eebeb5
-  github.com/earthly/cloud-api => ../cloud-api
+	github.com/moby/buildkit => github.com/earthly/buildkit v0.0.1-0.20220719221101-47e36164897b
```
alexcb: this is unexpected. Given all of our changes to buildkit and fsutil. I would avoid changing this unless it's related to this PR.

alexcb (commented):
Looks good, but if we can wait for Vlad to give it a +1, that would be best since it's somewhat new-feature-related

go.mod
mikejholly: I think you may be looking at an old commit. I reverted that here: https://github.com/earthly/earthly/pull/2050/commits/3313a5141b1e5e54a33c7680f774d59537bcf7ef

cmd/earthly/org_cmds.go
```diff
@@ -56,6 +56,37 @@ func (app *earthlyApp) orgCmds() []*cli.Command {
 
 func (app *earthlyApp) orgCmdsPreview() []*cli.Command {
 	return []*cli.Command{
+		{
+			Name:        "members",
+			Aliases:     []string{"member"},
```
vladaionescu, line 61: Nit: For consistency.

```suggestion
			Name:        "member",
			Aliases:     []string{"members"},
```

cmd/earthly/org_cmds.go
```diff
@@ -56,6 +56,37 @@ func (app *earthlyApp) orgCmds() []*cli.Command {
...
+				{
+					Name:        "ls",
+					Aliases:     []string{"list"},
+					Usage:       "List organization members and their permission level",
+					Description: "List organization members and their permission level",
+					UsageText:   "earthly org [--org organization] members ls",
+					Action:      app.actionOrgMemberList,
+				},
+				{
+					Name:        "update",
```
vladaionescu, line 75: Should this perhaps be called `update-permission`?

I'm thinking that there will be additional things to update about a member in the future and it might be a bit difficult if the user had to specify everything about the member on the commandline, when they probably want to update only a thing or two.

Another (possibly better) alternative might be to use a flag: `earthly update --permission <permission> <email>` - this way only the properties mentioned via flags are updated.

mikejholly: Could do. I was thinking that there may be more fields we may need to support--as you mentioned. I think I prefer the flag approach. Good idea.

cmd/earthly/org_cmds.go
```diff
@@ -265,3 +302,103 @@ func (app *earthlyApp) actionOrgRevoke(cliCtx *cli.Context) error {
...
+	if err != nil {
+		return err
+	}
+
+	userEmail := cliCtx.Args().Get(0)
+	if userEmail == "" {
+		return errors.New("member email required")
+	}
+
+	permission := cliCtx.String("permission")
```
vladaionescu, line 359: Mind using a variable here like we do for everything else?

mikejholly: Sure. That big arg struct is bothering me a bit :P

vladaionescu (approved):

cmd/earthly/org_cmds.go
vladaionescu: Yeah... we should categorize flags a bit better soon...

PR #2056, opened 2022-07-28
debug level logging

alexcb:
Adds new debug level for logging, and enables per-file stat debug
logging when transfering files.


mikejholly (approved):
Nice addition. I was thinking it would be neat to also output HTTP ops when debug level is enabled.

alexcb:
> output HTTP ops when debug level is enabled.

That's a great idea.

PR #2058, opened 2022-07-29
Add simple org invite accept command

mikejholly:
This PR adds a simple invitation accept command that can be used to complete the invitation flow from the CLI.

Here's an example:
```
$ earthly org invite accept <invite-code>
```

Perhaps we can expand the command later such that codes aren't required.

```
$ earthly org invite accept

You have a few invites. Please select one to accept:
  other-org
> cool-org
  less-cool-org
<return>
You've been added to "cool-org" as a "read-only" member.
```


vladaionescu (approved):
This is pretty cool

PR #2060, opened 2022-07-29
fix wait/end race condition

alexcb:
This fixes a race condition where the deferred waitgroup function was
referencing a variable that was modified by the for loop.


tests/wait-block/no-cache/Earthfile
```diff
@@ -0,0 +1,21 @@
...
+    RUN --no-cache echo YnJhdm8K | base64 -d && sleep 1
+
+test:
+    WAIT
+        BUILD +a
+    END
+    WAIT
+        BUILD +b
+    END
+    # TODO: this should cause a second build of +a, but doesn't
```
mikejholly, line 18: Needed?

earthfile2llb/wait_block.go
```diff
@@ -232,16 +232,17 @@ func (wb *waitBlock) waitStates(ctx context.Context) error {
...
-			rel, err := sem.Acquire(ctx, 1)
-			if err != nil {
-				return errors.Wrapf(err, "acquiring parallelism semaphore during waitStates for %s", item.c.target.String())
-			}
-			defer rel()
-
-			return item.c.forceExecution(ctx, *item.state, item.c.platr)
-		})
+	for _, itemItr := range stateItems {
+		func(item *stateWaitItem) {
```
mikejholly, line 236: I think we can also use the `item := item` idiom to avoid the extra func wrapper.

https://riptutorial.com/go/example/3897/using-closures-with-goroutines-in-a-loop

alexcb: I considered that, but thought the extra function wrapper made it more explicit (and unlikely for someone to remove in the future).

alexcb: at the same time, a comment saying `item := item // required due to delayed func` might be sufficient in that case. I'll do that.

mikejholly: Cool. I don't have a strong opinion. Both approaches aren't exactly intuitive. The `item := item` one is a little tidier but perhaps less obvious.

tests/wait-block/no-cache/Earthfile
alexcb: I need to fix this in a much-bigger separate WAIT-END PR.

PR #2061, opened 2022-07-30
account password reset

alexcb:
This adds two new calls:

    earthly account reset --email <email>

which sends a request to the server for a password reset token to be
send to the supplied email.

and:

    earthly account reset --email <email> --token <token>

which then prompts for a new password, which is then sent to the server,
to perform a password reset.

This additionally fixes the password prompt to prevent displaying a
newline before the cursor that accepts a hidden password on stdin (this
mimics the behaviour of the sudo command).

### DIFF
--- a/cmd/earthly/account_cmds.go
+++ b/cmd/earthly/account_cmds.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"os/signal"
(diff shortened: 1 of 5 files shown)


cmd/earthly/account_cmds.go
```diff
@@ -295,6 +303,25 @@ func (app *earthlyApp) actionAccountRegister(cliCtx *cli.Context) error {
 	return nil
 }
 
+func promptPassword() (string, error) {
+	fmt.Printf("pick a password: ")
```
mikejholly, line 307: Should we capitalize these? Asking because some other prompts are capitalized.

mikejholly (approved):

cmd/earthly/account_cmds.go
alexcb: good question!

sudo, for instance, uses lower case:

```
sudo whoami
[sudo] password for alex: 
```

where as 'adduser` (under ubuntu), uses upper case:

```
docker run --rm -ti ubuntu /bin/sh -c 'adduser testuser'
Adding user `testuser' ...
Adding new group `testuser' (1000) ...
Adding new user `testuser' (1000) with group `testuser' ...
Creating home directory `/home/testuser' ...
Copying files from `/etc/skel' ...
New password: 
```

so much for looking at standard cli practices for inspiration :laughing: 

I'll capitalise them.

PR #2066, opened 2022-08-02
docs: GitLab CI example uses Earthly's image

DeveloperC286:
Refactoring the GitLab CI example to stop using a generic image and then installing Earthly to just using Earthly's image.


DeveloperC286:
Some example GitLab repositories using this method.

https://gitlab.com/DeveloperC/image-steganography
https://gitlab.com/DeveloperC/investment-charges-comparison

alexcb:
Thanks for the suggestion.

I'm cautious about changing the entrypoint, because earthly requires https://github.com/earthly/earthly/blob/main/earthly-entrypoint.sh#L4-L12 to be executed in the case a cgroups v2 host is detected.

Any thoughts on how we can use the image, but still maintaining the ability to have the entrypoint run?

alexcb:
there is already an escape-hatch near the bottom of the entrypoint that has:

```
if [ -n "$EARTHLY_EXEC_CMD" ]; then
    export earthly_config
    exec "$EARTHLY_EXEC_CMD"
    exit 1 # this should never be reached
fi
```

maybe this could be used somehow?

or maybe we need a new similar escape hatch such as:

```
# allow escape-hatch for users who want to run other commands before 
if [ -n "$EARTHLY_SHELL_ENTRYPOINT" ]; then
    exec "$EARTHLY_SHELL_ENTRYPOINT" -c "$@"
fi
```

DeveloperC286:
Good point about not overriding the entrypoint, this could skip some important setup etc.

Using the env variable `EARTHLY_EXEC_CMD` we can just export that as `/bin/sh`.

E.g.

```
services:
  - docker:dind

variables:
  DOCKER_HOST: tcp://docker:2375
  EARTHLY_EXEC_CMD: "/bin/sh"

image:
  name: earthly/earthly:v0.1659490371.77891517

...
```

Once we start using a version of the Earthly Docker image which includes this commit

https://github.com/earthly/earthly/commit/26c9e91768605bb724576c271cab51c23a1d8673

Which removes changing into a directory unless `SRC_DIR` is set. 
Because before this commit it defaults to changing into the directory `/workspace` which does not exist on GitLab CI.

Here is an example on GitLab CI of not overriding the entrypoint and using `EARTHLY_EXEC_CMD` instead.

https://gitlab.com/DeveloperC/investment-charges-comparison/-/merge_requests/20 

Am I good to update GitLab CI example to this method?
Should we wait to update this example until that commit is included in a release?

alexcb:
It looks like the PR that was linked is failing with

```
/usr/bin/earthly-entrypoint.sh: cd: line 66: can't cd to /workspace: No such file or directory
```

This issue can be bypassed by setting `SRC_DIR="/"` (or other location)

however when I try running the image directly with a different command, e.g.

```
docker run --rm -ti -e EARTHLY_EXEC_CMD=/bin/sh -e DOCKER_HOST=unix:///var/run/docker.sock -v /var/run/docker.sock:/var/run/docker.sock -e SRC_DIR="/" --privileged earthly/earthly hostname
```

hostname is never executed because `$@` is never used in the entrypoint.

as is, I don't think we can update the CI docs to use this because it doesn't work (see my comment regarding `or maybe we need a new similar escape hatch such as`).

I think we can get there, but we'll need some help with an updated entrypoint and a working example in gitlab ci first.

DeveloperC286:
> It looks like the PR that was linked is failing with
> 
> ```
> /usr/bin/earthly-entrypoint.sh: cd: line 66: can't cd to /workspace: No such file or directory
> ```

Yes because I updated the pull request to try the latest image 

https://gitlab.com/DeveloperC/investment-charges-comparison/-/merge_requests/20/diffs?diff_id=460291906&start_sha=5eb9b66042fa429ec091216953f69dc1d19bcd01

and it at the time the images used by the pipeline did not include the commit https://github.com/earthly/earthly/commit/26c9e91768605bb724576c271cab51c23a1d8673.

> This issue can be bypassed by setting `SRC_DIR="/"` (or other location)

This is not needed when you use an image which includes the changes https://github.com/earthly/earthly/commit/26c9e91768605bb724576c271cab51c23a1d8673.

E.g. see the pipeline for the commit using the image `v0.6.22`.

https://gitlab.com/DeveloperC/investment-charges-comparison/-/commit/98e44c5fc85b0ef1af6070414ff099d19d9d8f33

> (on a side note, setting it to `SRC_DIR=""` didn't seem to work, but ideally this environment variable shouldn't be used as it's going to be removed down the road; however there appears to be a bug causing it to be overriden to `workspace` when set to the empty string)

This was the default behaviour prior to https://github.com/earthly/earthly/commit/26c9e91768605bb724576c271cab51c23a1d8673 not a bug, you must have been using an image which did not include this change.

> however when I try running the image directly with a different command, e.g.
> 
> ```
> docker run --rm -ti -e EARTHLY_EXEC_CMD=/bin/sh -e DOCKER_HOST=unix:///var/run/docker.sock -v /var/run/docker.sock:/var/run/docker.sock -e SRC_DIR="/" --privileged earthly/earthly hostname
> ```
> 
> hostname is never executed because `$@` is never used in the entrypoint.

It appears GitLab CI starts the Docker image and then gives you access to the shell after, it does not pass the script commands via `${@}`. See the working example here https://gitlab.com/DeveloperC/investment-charges-comparison/-/commit/98e44c5fc85b0ef1af6070414ff099d19d9d8f33

> as is, I don't think we can update the CI docs to use this because it doesn't work (see my comment regarding `or maybe we need a new similar escape hatch such as`).
> 
> I think we can get there, but we'll need some help with an updated entrypoint and a working example in gitlab ci first.

Sorry about the confusion with me testing on the pipeline, now the change is baked into a released image are we good to update the example?

Because in my opinion this pull request makes the example a lot simpler.

DeveloperC286:
@alexcb Any update on if this is an improvement and will likely be accepted if I rebase it?

alexcb:
@DeveloperC286 sorry for the delay (life has been busy of late).

Do you have a public gitlab repo with this config -- I'm most curious to see working; otherwise I can spin one up to give it a test before accepting the change.

DeveloperC286:
@alexcb No worries 👍 

Sure here are a number of GitLab repositories with this method merged into main and being used.

Python
- https://gitlab.com/DeveloperC/image-steganography/-/blob/master/.gitlab-ci.yml
- https://gitlab.com/DeveloperC/investment-charges-comparison/-/blob/master/.gitlab-ci.yml

Go
- https://gitlab.com/DeveloperC/git-changed/-/blob/main/.gitlab-ci.yml
- https://gitlab.com/DeveloperC/git-manager/-/blob/main/.gitlab-ci.yml

C
- https://gitlab.com/DeveloperC/nasm-x86-shellcode-generator/-/blob/master/.gitlab-ci.yml

alexcb:
thanks for the links @DeveloperC286 :raised_hands: 

I added a much smaller `before_script` section to perform the bootstrap and docker login -- how does this look?

DeveloperC286:
@alexcb 

Just curious why is the `docker login` required?

and what does `earthly bootstrap` do that is not covered by the entry script?

alexcb:
> @alexcb
> 
> Just curious why is the `docker login` required?

It's not strictly-required*; however it helps with avoiding docker hub rate limiting issues.

*unless one is pushing to the docker hub registry.

> 
> and what does `earthly bootstrap` do that is not covered by the entry script?

This simply starts up the earthly-buildkit container -- it could be helpful to explicitly start it in a separate step, so if there's any failure it's clear that it's in the init stage, rather than the user's test step.

DeveloperC286:
> > @alexcb
> > Just curious why is the `docker login` required?
> 
> It's not strictly-required*; however it helps with avoiding docker hub rate limiting issues.
> 
> *unless one is pushing to the docker hub registry.
> 
> > and what does `earthly bootstrap` do that is not covered by the entry script?
> 
> This simply starts up the earthly-buildkit container -- it could be helpful to explicitly start it in a separate step, so if there's any failure it's clear that it's in the init stage, rather than the user's test step.

Thank you for explaining 👍  

The changes look good to me and is an improvement 👍

Issue #1483, opened 2021-12-08
Can't provide secrets to FROM DOCKERFILE


glenonn:
I have an existing Dockerfile that I'd like to use earthly with, but it requires a secret and I don't know how to provide that secret.

Let's say I have the following Dockerfile:
```
FROM busybox:latest

RUN --mount=type=secret,id=my_secret,target=/mount.txt \
  cp /mount.txt /secret.txt
```

To build this I'd run
`docker build . --secret id=my_secret,env=A_SECRET_IN_MY_ENV`

I want to write some earthly code that looks like
`FROM DOCKERFILE --secret id=my_secret,env=A_SECRET_IN_MY_ENV .`
but --secret is not supported. What alternatives do I have?

vladaionescu:
This isn't yet supported. It should be relatively easy to add support for it though. We'd just need to figure out the syntax / semantics so that it follows similar conventions as other places in Earthly where you can specify secrets.

alexcb:
based on our current implementation, when we call `dockerfile2llb.Dockerfile2LLB`, it returns an llb state that we reference in the build graph, it already has access to our secrets attachable.

I tried:
```
FROM busybox:latest
RUN --mount=type=secret,id=+secrets/key,target=/mount.txt cp /mount.txt /leaked-secret.txt
```
and referenced it with
```
test:
    FROM DOCKERFILE .
    RUN cat /leaked-secret.txt
```

and got this interesting error:

```
             stage-0 | --> RUN --mount=type=secret,id=+secrets/key,target=/mount.txt cp /mount.txt /leaked-secret.txt
             stage-0 | WARN: (RUN --mount=type=secret,id=+secrets/key,target=/mount.txt cp /mount.txt /leaked-secret.txt) failed to lookup secret "/+secrets/key" from secrets server: failed to get secret: unauthorized
Error: build target: build main: bkClient.Build: failed to solve: failed to lookup secret "/+secrets/key" from secrets server: failed to get secret: unauthorized
```

Under the hood, when earthly sees two slashes in the secret path, it attempts to reach out to earthly's cloud secrets service. 

So I tried this:

```
earthly secrets set /user/key-in-dockerfile this-is-in-the-cloud
```

```
FROM busybox:latest
RUN --mount=type=secret,id=user/key-in-dockerfile,target=/mount.txt cp /mount.txt /leaked-secret.txt
```

```
test:
    FROM DOCKERFILE .
```

and it worked:

```
[...]

PR #2070, opened 2022-08-03
Added support for passing secrets to `FROM DOCKERFILE` command

sandeshlmore:
Fixes: #1483

This PR allows user to pass secrets to `FROM DOCKERFILE`  earthly command/stage. secret can be pass to `FROM DOCKERFILE` with flag ```--secret```.

#### Dockerfile
```
FROM busybox:latest

RUN --mount=type=secret,id=TEST_SECRET_1,target=/mount.txt cp /mount.txt /TEST_ENV_SECRET_1.txt
RUN --mount=type=secret,id=TEST_SECRET_2,target=/mount.txt cp /mount.txt /TEST_ENV_SECRET_2.txt
```
#### Earthfile
```
test:
    FROM DOCKERFILE --secret TEST_SECRET_1 --secret TEST_SECRET_2=+secrets/SUPER_SECRET .
    RUN test "$(cat /TEST_ENV_SECRET_1.txt)" = "this is secret from env"
    RUN test "$(cat /TEST_ENV_SECRET_2.txt)" = "foo"

```

#### Set Env Locally
```
export TEST_SECRET_1="this is secret from env"
```

#### Build target test from Earthfile
```
earthly --secret TEST_SECRET_1 --secret  SUPER_SECRET="foo"  --no-cache +test
```

OUTPUT:
```
            internal | --> load metadata for docker.io/library/busybox:latest
             stage-0 | --> FROM docker.io/library/busybox:latest@sha256:ef320ff10026a50cf5f0213d35537ce0041ac1d96e9b7800bafd8bc9eff6c693
             stage-0 | [          ]   0% resolve docker.io/library/busybox:latest@sha256:ef320ff10026a50cf5f0213d35537ce0041ac1d96e9b7800bafd8bc9eff6c693 [██████████] 100% resolve docker.io/library/busybox:latest@sha256:ef320ff10026a50cf5f0213d35537ce0041ac1d96e9b7800bafd8bc9eff6c693
             stage-0 | --> RUN --mount=type=secret,id=TEST_SECRET_1,target=/mount.txt cp /mount.txt /TEST_ENV_SECRET_1.txt
             stage-0 | --> RUN --mount=type=secret,id=TEST_SECRET_2,target=/mount.txt cp /mount.txt /TEST_ENV_SECRET_2.txt
               +test | --> RUN test "$(cat /TEST_ENV_SECRET_1.txt)" = "this is secret from env"
               +test | --> RUN test "$(cat /TEST_ENV_SECRET_2.txt)" = "foo"

```

I have also added a test case under /tests/from-dockerfile-secrets.earth which can be run as 
```
./build/linux/amd64/earthly --no-cache -P ./tests+from-dockerfile-secrets --DOCKERHUB_AUTH=false
```

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -31,6 +31,7 @@ import (
 	"github.com/moby/buildkit/client/llb"
 	gwclient "github.com/moby/buildkit/frontend/gateway/client"
 	"github.com/moby/buildkit/session"
+	"github.com/moby/buildkit/session/secrets"
 	"github.com/moby/buildkit/util/entitlements"
 	reccopy "github.com/otiai10/copy"
 	"github.com/pkg/errors"
@@ -76,6 +77,7 @@ type Opt struct {
 	FeatureFlagOverrides   string
 	ContainerFrontend      containerutil.ContainerFrontend
 	InternalSecretStore    *secretprovider.MutableMapStore
+	SecretProviders        []secrets.SecretStore
 }
 
 // BuildOpt is a collection of build options.
@@ -185,6 +187,7 @@ func (b *Builder) convertAndBuild(ctx context.Context, target domain.Target, opt
 				DoPushes:              opt.Push,
 				PullPingMap:           pullPingMap,
 				InternalSecretStore:   b.opt.InternalSecretStore,
+				SecretProviders:       b.opt.SecretProviders,
 			}
 			mts, err = earthfile2llb.Earthfile2LLB(childCtx, target, opt, true)
 			if err != nil {
--- a/cmd/earthly/build_cmd.go
+++ b/cmd/earthly/build_cmd.go
@@ -31,6 +31,7 @@ import (
 	"github.com/moby/buildkit/session"
 	"github.com/moby/buildkit/session/auth/authprovider"
 	"github.com/moby/buildkit/session/localhost/localhostprovider"
+	"github.com/moby/buildkit/session/secrets"
 	"github.com/moby/buildkit/session/sshforward/sshprovider"
 	"github.com/moby/buildkit/util/entitlements"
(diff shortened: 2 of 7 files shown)


alexcb (changes requested):
Thanks for the submission,

This feels like a departure from how `RUN --secret=...` syntax works.

is it possible to change the supported syntax to work as:

```
FROM DOCKERFILE --secret GITHUB_TOKEN=+secrets/GH_TOKEN ...
```

which would allow a user to make use of an existing secret that was passed in from the command line (e.g. `export GH_TOKEN=secret; earthly --secret GH_TOKEN +sometarget`), then the user has an option of referencing the secret `GH_TOKEN`, but exposing it to the dockerfile as `GITHUB_TOKEN`.

The `--env-file` and `--secret-file` options seem out of place, as they are options for the cli, and do no mirror the way RUN exposes secrets; ultimately I don't think it's appropriate to have the interpreter reading an env-file directly.

sandeshlmore:
thanks @alexcb for providing feedback. 

from the example you gave, i understand that user should be able to refer to secret set along with earthly command run. 

quick question: can user also refer to cloud-secrets(`earthly secrets set` command ) or only secrets set from command-line with --secret flag should be supported here?

alexcb:
> quick question: can user also refer to cloud-secrets(`earthly secrets set` command ) or only secrets set from command-line with --secret flag should be supported here?

ideally yes; however if there's a technical reason why it's not possible, we'll have to settle on the user performing remapping via the `FROM DOCKERFILE --secret Foo=+secrets/my/cloud/secret` syntax.

alexcb (commented):
This PR seems to be bypassing the secrets attachables; is there a technical reason for this?

Issue #2071, opened 2022-08-03
Can't use port on LHS in `WITH DOCKER --load` alongside `--use-registry-for-with-docker`


rrjjvv:
Reproduction should make it clearer:
```Earthfile
VERSION --use-registry-for-with-docker 0.6

wrapped:
    FROM alpine
    SAVE IMAGE wrapped-alpine

load-port:
    FROM earthly/dind:alpine
    WITH DOCKER --load example.com:9999/img=+wrapped
        RUN echo "broke"
    END

load-no-port:
    FROM earthly/dind:alpine
    WITH DOCKER --load example.com/img=+wrapped
        RUN echo "this works"
    END
```

I would expect both to behave identically, but the usage of a port in the alias results in failure.
```
$ earthly --version 
earthly version v0.6.20 ad869c06c884b10f88948b5852ab22b4d7262e20 linux/amd64; Ubuntu 20.04.4 LTS (Focal Fossa)

# works as expected; showing output for comparison sake
$ earthly --no-cache -P +load-no-port
<snip>
       +load-no-port | Loading images from BuildKit via embedded registry...
       +load-no-port | Pulling 172.30.0.1:8371/sess-w317eepv6emwhvng0dx18uphy/example.com/img:latest and retagging as example.com/img:latest
<snip>

# broken; "my" custom port (semicolon) results in a badly formatted name
$ earthly --no-cache -P +load-port
<snip>
         +load-port | Loading images from BuildKit via embedded registry...
          +load-port | Pulling 172.30.0.1:8371/sess-op7ikdxqh6ag10avh9yzdsayd/example.com:9999/img and retagging as example.com:9999/img
          +load-port | invalid reference format
          +load-port | ERROR: Command exited with non-zero code: WITH DOCKER RUN --privileged echo "broke"
<snip>
```

I suspect the fix is an extension of https://github.com/earthly/earthly/commit/bf27e1ce5b7a372fefaba7acf25b48e89d93959f, a very similar (but not quite as trivial).

---

This isn't a blocker by any means; this is specific to `--use-registry-for-with-docker` (still experimental, last I knew).  Usage of a port might sound silly on the surface, but my scenario is in the spirit of #1797 (I am pushing to a corporate registry that listens on a non-standard port).  Re-tagging inside `WITH DOCKER` would be an easy workaround.  My usage of `--use-registry-for-with-docker` in this context was trying to diagnose a separate issue (which I may file separately).

vladaionescu:
[...]

PR #2076, opened 2022-08-03
Fix `WITH DOCKER` loading when image contains port number

vladaionescu:
Fixes https://github.com/earthly/earthly/issues/2071

This PR goes the extra mile to maintain forwards-backwards compatibility given the usage of the feature flag via satellites. Besides the automated test added, I have also manually tested old earthly + new buildkit and new earthly + old buildkit (this particular bug is not fixed in these combinations, but everything else still works).

Minor hidden addition as part of this PR: a way to inhibit buildkit updating and restarting (`earthly --no-buildkit-update`) - this helps test an old buildkit with a new earthly.

### DIFF
--- a/builder/image_solver.go
+++ b/builder/image_solver.go
@@ -198,6 +198,20 @@ func (m *multiImageSolver) SolveImages(ctx context.Context, imageDefs []*states.
 		return ret, nil
 	}
 
+	newInterImgFormat := false
+	info, err := m.bkClient.Info(ctx)
+	if err != nil {
+		// Maybe older buildkit.
+	} else {
+		switch info.BuildkitVersion.Version {
+		case "v0.6.19":
+		case "v0.6.20":
+		default:
+			newInterImgFormat = true
+		}
+	}
+
+	onPullMap := make(map[string]string) // intermediate image name -> final image name
 	// This func is executed when the image create/push process is complete.
 	onPull := func(ctx context.Context, images []string, resp map[string]string) error {
 		if resp == nil {
@@ -205,17 +219,15 @@ func (m *multiImageSolver) SolveImages(ctx context.Context, imageDefs []*states.
 		}
 		results := make([]*states.ImageResult, len(images))
 		for i, img := range images {
-			// Note that this img split by / algo is also repeated in the
-			// dockerd wrapper.
-			imgParts := strings.SplitN(img, "/", 2)
-			if len(imgParts) != 2 {
-				return errors.Errorf(
-					"invalid image name: %s. Was expecting sess prefix", img)
+			finalImageName, ok := onPullMap[img]
+			if !ok {
+				return errors.Errorf("image %s not found in onPullMap", img)
 			}
 			result := &states.ImageResult{
 				IntermediateImageName: img,
-				FinalImageName:        imgParts[1],
+				FinalImageName:        finalImageName,
 				Annotations:           make(map[string]string),
+				NewInterImgFormat:     newInterImgFormat,
 			}
 
 			for k, v := range resp {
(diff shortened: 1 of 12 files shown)


builder/image_solver.go
```diff
@@ -198,24 +198,36 @@ func (m *multiImageSolver) SolveImages(ctx context.Context, imageDefs []*states.
...
+	} else {
+		switch info.BuildkitVersion.Version {
+		case "v0.6.19":
+		case "v0.6.20":
+		default:
+			newInterImgFormat = true
+		}
+	}
+
+	onPullMap := make(map[string]string) // intermediate image name -> final image name
```
alexcb, line 214: is this duplicating https://github.com/earthly/earthly/blob/main/util/gatewaycrafter/pullpingmap.go ?

we have a similar onPull function under builder, where we pull images out of the pullpingmap: https://github.com/earthly/earthly/blob/main/builder/builder.go#L412-L416

vladaionescu: It is very similar but different enough. The intention is the same - to keep track of a mapping between intermediate names vs final (re-tagged) names. However, the images are not exported locally - they are exported into `WITH DOCKER` - so the code you're referencing couldn't be reused in this situation.

vladaionescu:
Note that the satellite failure is expected. This is what this PR is fixing (after a satellite update with the fix in this PR, this failure would not show up anymore).

Screen Shot 2022-08-03 at 4 21 25 PM

buildkitd/dockerd-wrapper.sh
```diff
@@ -193,8 +193,18 @@ load_registry_images() {
     if [ -n "$EARTHLY_DOCKER_LOAD_REGISTRY" ]; then
         echo "Loading images from BuildKit via embedded registry..."
         for img in $EARTHLY_DOCKER_LOAD_REGISTRY; do
-            user_tag=$(printf '%s' "$img" | cut -d'/' -f2-)
-            with_reg="$buildkit_docker_registry/$img"
+            case "$img" in
+                *'|'*)
+                    with_reg="$buildkit_docker_registry"/$(printf '%s' "$img" | cut -d'|' -f1)
+                    user_tag=$(printf '%s' "$img" | cut -d'|' -f2-)
```
alexcb, line 199:
```suggestion
                    user_tag="$(printf '%s' "$img" | cut -d'|' -f2-)"
```

buildkitd/dockerd-wrapper.sh
```diff
@@ -193,8 +193,18 @@ load_registry_images() {
...
+                    ;;
+                *)
+                    # Old format before v0.6.21.
+                    with_reg="$buildkit_docker_registry/$img"
+                    user_tag=$(printf '%s' "$img" | cut -d'/' -f2-)
```
alexcb, line 204:
```suggestion
                    user_tag="$(printf '%s' "$img" | cut -d'/' -f2-)"
```

vladaionescu: I think it doesn't work due to the `"$img"` in the middle - not what's the correct way to quote this (or if it even needs quotes in this case)

alexcb: that surprises me, we have similar bits of shell that use double quotes both around and inside a shell out: https://github.com/earthly/earthly/blob/main/tests/secret-provider-config/test-secret-provider#L25

alexcb: anyways, if it's passing our linter and tests work might be good enough.

alexcb (approved):

buildkitd/dockerd-wrapper.sh
vladaionescu: TIL!

Fixed

PR #2077, opened 2022-08-04
Add satellites documentation

vladaionescu:


alexcb (approved):

brandonSc (approved):
Looks good! I wonder if we should do any comparisons to the shared caching features, perhaps on the docs page for shared caching.

PR #2082, opened 2022-08-05
Use the SHA that has the timeout changes in it.

dchw:
Bump the SHA so we can use the timeout options.


vladaionescu (approved):
Mind updating go.mod too? We keep them in sync typically (even though it probably doesn't matter in this case).

alexcb (changes requested):
did you use the `+update-buildkit` target: https://github.com/earthly/earthly/blob/main/Earthfile#L50 as outlined under https://github.com/earthly/earthly/blob/main/CONTRIBUTING.md#updates-to-buildkit-or-fsutils ?

This should handle both the `buildkit/Earthfile` and `go.{mod,sum}` files.

dchw:
I did run that target, I didnt see those changes though...

alexcb (approved):

alexcb:
> ~I did run that target, I didnt see those changes though...~
> 
> I ran `./buildkit+update-buildkit`. sob

that's some good feedback, I updated the buildkit target name in #2084 to make it less confusing.

Issue #1986, opened 2022-07-06
`EXPOSE` parse problem


adamgordonbell:
The docs for `EXPOSE` say that it mirrors `EXPOSE` in docker and that is not true in at least one specific case.

All the following work in a dockerfile:
```
EXPOSE 8080:8080
EXPOSE 8080
EXPOSE 8080 8080
```
and all result (in my testing) with the following result in the image config:
```
$ docker inspect ktor2 | jq '.[].Config.ExposedPorts'
{
  "8080/tcp": {}
}
```
In Earthly, both of these
```
EXPOSE 8080
EXPOSE 8080 8080
```
give this:
```
docker inspect ktor-docker-sample | jq '.[].Config.ExposedPorts'
{
  "8080": {}
}
```
(note missing `tcp`, I'm unsure if that is an issue)

And :
```
EXPOSE 8080:8080
```
sets: 
```
{
  "8080:8080": {}
}
``` 
which is certainly wrong. 

See ticket #1975

shahpratikr:
@alexcb can you please review PR https://github.com/earthly/earthly/pull/2083?

PR #2083, opened 2022-08-05
Fixed issue-1986: EXPOSE parse problem

shahpratikr:
Fixes: https://github.com/earthly/earthly/issues/1986
Earthfile:
```
VERSION 0.6
FROM alpine:3.15
test:
    EXPOSE 8080:8081
    EXPOSE 8082
    EXPOSE 8083 8084
    SAVE IMAGE earthly-expose:latest
```
Dockerfile:
```
FROM alpine:3.15
EXPOSE 8080:8081
EXPOSE 8082
EXPOSE 8083 8084
```
Docker containers:
```
$ docker ps
CONTAINER ID   IMAGE                                     COMMAND                  CREATED         STATUS         PORTS                                                NAMES
dba016652adb   docker-expose:latest                      "/bin/sh"                5 seconds ago   Up 4 seconds   8081-8084/tcp                                        goofy_mestorf
4300adb49d2f   earthly-expose:latest                     "/bin/sh"                3 seconds ago    Up 2 seconds    8081-8084/tcp                                        friendly_buck
```
`docker inspect` command output on dockerfile image:
```
$ docker inspect dba016652adb | jq '.[].Config.ExposedPorts'
{
  "8081/tcp": {},
  "8082/tcp": {},
  "8083/tcp": {},
  "8084/tcp": {}
}
```
`docker inspect` command output on earthfile image:
```
$ docker inspect 4300adb49d2f | jq '.[].Config.ExposedPorts'
{
  "8081/tcp": {},
  "8082/tcp": {},
  "8083/tcp": {},
  "8084/tcp": {}
}
```

### DIFF
--- a/earthfile2llb/interpreter.go
+++ b/earthfile2llb/interpreter.go
@@ -1132,18 +1132,69 @@ func (i *Interpreter) handleExpose(ctx context.Context, cmd spec.Command) error 
 		return i.errorf(cmd.SourceLocation, "no arguments provided to the EXPOSE command")
 	}
 	ports := getArgsCopy(cmd)
-	for index, port := range ports {
+	var exposedPorts = []string{}
+	var portList = []string{}
+	for _, port := range ports {
 		expandedPort, err := i.expandArgs(ctx, port, false, false)
 		if err != nil {
 			return i.wrapError(err, cmd.SourceLocation, "failed to expand EXPOSE %s", port)
 		}
-		ports[index] = expandedPort
+		if strings.Contains(expandedPort, ":") {
+			expandedPort = strings.SplitN(expandedPort, ":", 2)[1]
+		}
+		if strings.Contains(expandedPort, "-") {
+			portList, err = i.getListOfPorts(expandedPort)
+			if err != nil {
+				return i.wrapError(err, cmd.SourceLocation, "apply EXPOSE")
+			}
+		} else {
+			portList = append(portList, expandedPort)
+		}
+		exposedPorts, err = i.handleSlashInPort(portList)
+		if err != nil {
+			return i.errorf(cmd.SourceLocation, err.Error())
+		}
 	}
-	err := i.converter.Expose(ctx, ports)
+	err := i.converter.Expose(ctx, exposedPorts)
 	if err != nil {
 		return i.wrapError(err, cmd.SourceLocation, "apply EXPOSE")
 	}
 	return nil
+}
+
+func (i *Interpreter) getListOfPorts(expandedPort string) ([]string, error) {
+	portList := []string{}
+	parts := strings.Split(expandedPort, "-")
+	startPort, err := strconv.Atoi(parts[0])
+	if err != nil {
+		return portList, err
+	}
+	endPort, err := strconv.Atoi(parts[1])
+	if err != nil {
+		return portList, err
+	}
+	for startPort <= endPort {
+		portList = append(portList, strconv.Itoa(startPort))
+		startPort += 1
+	}
+	return portList, nil
+}
+
+func (i *Interpreter) handleSlashInPort(portList []string) ([]string, error) {
+	var exposedPorts = []string{}
+	for _, port := range portList {
+		if strings.Contains(port, "/") {
+			protocol := strings.SplitN(port, "/", 2)[1]
+			if protocol != "tcp" && protocol != "udp" {
+				return exposedPorts, errors.New(
+					fmt.Sprintf("failed to parse EXPOSE %s: invalid protocol %s", port, protocol))
+			}
+			exposedPorts = append(exposedPorts, port)
+		} else {
+			exposedPorts = append(exposedPorts, port+"/tcp")
+		}
+	}
+	return exposedPorts, nil
 }
 
 func (i *Interpreter) handleVolume(ctx context.Context, cmd spec.Command) error {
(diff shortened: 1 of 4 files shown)


earthfile2llb/converter.go
```diff
@@ -1140,7 +1140,10 @@ func (c *Converter) Expose(ctx context.Context, ports []string) error {
 	}
 	c.nonSaveCommand()
 	for _, port := range ports {
-		c.mts.Final.MainImage.Config.ExposedPorts[port] = struct{}{}
+		if strings.Contains(port, ":") {
+			port = strings.SplitN(port, ":", 2)[1]
+		}
+		c.mts.Final.MainImage.Config.ExposedPorts[port + "/tcp"] = struct{}{}
```
alexcb, line 1146: assuming it's `tcp` here seems incorrect; is it possible for the user to expose a port for `udp`?

earthfile2llb/converter.go
```diff
@@ -1140,7 +1140,10 @@ func (c *Converter) Expose(ctx context.Context, ports []string) error {
 	}
 	c.nonSaveCommand()
 	for _, port := range ports {
-		c.mts.Final.MainImage.Config.ExposedPorts[port] = struct{}{}
+		if strings.Contains(port, ":") {
+			port = strings.SplitN(port, ":", 2)[1]
```
alexcb, line 1144: why is the first value being ignored?

alexcb (changes requested):
thanks for the contribution, but it seems fragile. can you expand it to be more generic; we should mirror the functionality of https://docs.docker.com/engine/reference/builder/#expose

earthfile2llb/converter.go
shahpratikr: since container will be exposed on 2nd port, we are getting only 2nd port.
for example, if value is 8080:8081, then 8081 gets set as container port

earthfile2llb/converter.go
shahpratikr: fixed

earthfile2llb/converter.go
alexcb: what does the first value represent? It seems odd that we can safely ignore it. Can you add a comment pointing to some documentation which backs up your reasoning.

shahpratikr: first value represents host port. https://docs.docker.com/engine/reference/run/#expose-incoming-ports this provides more detail

alexcb: That's a link to the cli reference; if nothing exists specifically for the dockerfile reference, maybe you can link to some code showing that docker (or buildkit) ignores this value too.

earthfile2llb/converter.go
```diff
@@ -1140,7 +1140,20 @@ func (c *Converter) Expose(ctx context.Context, ports []string) error {
...
 	for _, port := range ports {
-		c.mts.Final.MainImage.Config.ExposedPorts[port] = struct{}{}
+		var exposedPort string
+		if strings.Contains(port, ":") {
+			port = strings.SplitN(port, ":", 2)[1]
+		}
+		if strings.Contains(port, "/") {
+			protocol := strings.SplitN(port, "/", 2)[1]
+			if protocol != "tcp" && protocol != "udp" {
+				return errors.New("Invalid protocol provided")
```
alexcb, line 1150: It's best to move all of this parsing logic into interpreter.go, and then you can make use of 
```
			return i.errorf(err, cmd.SourceLocation, "failed to parse EXPOSE %s: invalid protocol %s", port, protocol)
```

which will then display the corresponding Earthfile source location.

earthfile2llb/converter.go
```diff
@@ -1140,7 +1140,20 @@ func (c *Converter) Expose(ctx context.Context, ports []string) error {
...
+			}
+			exposedPort = port
+		} else {
+			exposedPort = port + "/tcp"
```
alexcb, line 1154: is this what docker does when no protocol is defined?

shahpratikr: yes if no protocol is present tcp is assigned by default

earthfile2llb/converter.go
shahpratikr: https://github.com/earthly/buildkit/blob/earthly-main/vendor/github.com/docker/go-connections/nat/nat.go#L176 this function ignores the first port and sets second port as container port.

earthfile2llb/converter.go
shahpratikr: fixed

earthfile2llb/converter.go
shahpratikr: https://github.com/earthly/buildkit/blob/earthly-main/vendor/github.com/docker/go-connections/nat/nat.go#L176 this function is separating container and host port. docker uses container port to communicate to host. if host port is not given, docker allocates that automatically. hence when we provide <port1>:<port2> it assumes port1 is host port and port2 is container port

shahpratikr: 2 containers I ran with shown dockerfile but ran with -P & -p options to expose all ports in the dockerfile or only particular ones, respectively. Note the exposed ports for both.
Screenshot from 2022-08-09 19-15-27

 Contents of file /proc/sys/net/ipv4/ip_local_port_range on Ubuntu 20.04
Screenshot from 2022-08-09 19-20-37

Docker inspect output for the 2 containers:
Screenshot from 2022-08-09 19-44-31

alexcb:
Thanks for attempting a fix; it turns out there was a simpler solution: #2089

PR #2084, opened 2022-08-05
rename buildkit+update-buildkit target to avoid confusion

alexcb:
This renames the buildkit+update-buildkit target, to avoid confusion
regarding which `+update-buildkit` target users should run when updating
earthly to use a different buildkit version.


vladaionescu (approved):

vladaionescu:
Oh, do we need to change the mention in CONTRIBUTING.md?

alexcb:
> Oh, do we need to change the mention in CONTRIBUTING.md?

nope, because it mentions the `+update-buildkit` target -- the one in the root Earthfile.

alexcb:
in hindsight, maybe it should have been renamed to `update-earthfile` rather than `update-buildkit-earthfile`; however neighter of these conflict with the same named target in the root earthfile.

PR #2085, opened 2022-08-05
WAIT/END support for save artifact as local

alexcb:
This extends the WAIT/END code to support waiting for an artifact to be
saved locally before continuing.

This introduces a new constraint that prevents a user from copying a
file into a context after it has already been output by earthly.
Previously such a workflow was not possible as outputs always happened
at the end of an earthly build (after all copies were performed).

This changes the save artifact as local workflow to use a random
alphanumeric string rather than an index-<number>, due to the fact that
multiple wait/end blocks may now be processed in parallel, which would
lead to conflicting index-<number> entries.

### DIFF
--- a/util/saveartifactlocally/saveartifactlocally.go
+++ b/util/saveartifactlocally/saveartifactlocally.go
@@ -1,0 +1,123 @@
+package saveartifactlocally
+
+import (
+	"context"
+	"os"
+	"path"
+	"path/filepath"
+	"strings"
+
+	"github.com/earthly/earthly/conslogging"
+	"github.com/earthly/earthly/domain"
+
+	reccopy "github.com/otiai10/copy"
+	"github.com/pkg/errors"
+)
+
+// SaveArtifactLocally handles saving artifacts to the local host, and is called from both builder and waitblock
+func SaveArtifactLocally(ctx context.Context, parentConsole conslogging.ConsoleLogger, console *conslogging.BufferedLogger, artifact domain.Artifact, indexOutDir string, destPath string, salt string, printPhases, ifExists bool) error {
+	fromPattern := filepath.Join(indexOutDir, filepath.FromSlash(artifact.Artifact))
+	// Resolve possible wildcards.
+	// TODO: Note that this is not very portable, as the glob is host-platform dependent,
+	//       while the pattern is also guest-platform dependent.
+	fromGlobMatches, err := filepath.Glob(fromPattern)
+	if err != nil {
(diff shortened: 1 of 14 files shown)


alexcb:
I tested:
```
test:
    FROM alpine
    RUN (date && cat bar) > data
    SAVE ARTIFACT data AS LOCAL the-date
```

with both this version, and v0.6.21, and the random alphanumeric string didn't prevent cache from being re-used.

util/saveartifactlocally/saveartifactlocally.go
```diff
@@ -0,0 +1,123 @@
...
+
+func trimFilePathPrefix(prefix string, thePath string, console conslogging.ConsoleLogger) string {
+	ret, err := filepath.Rel(prefix, thePath)
+	if err != nil {
+		console.Warnf("Warning: Could not compute relative path for %s "+
+			"as being relative to %s: %s\n", thePath, prefix, err.Error())
+		return thePath
+	}
+	return ret
+}
```
vladaionescu, line 123: This code was simply moved without modification, right?

vladaionescu (approved):
Looks good to me!

util/saveartifactlocally/saveartifactlocally.go
alexcb: yes (with exception for `parentConsole` now being passed in):

```
$ diff old new
1c1,2
< func (b *Builder) saveArtifactLocally(ctx context.Context, console *conslogging.BufferedLogger, artifact domain.Artifact, indexOutDir string, destPath string, salt string, opt BuildOpt, ifExists bool) error {
---
> // SaveArtifactLocally handles saving artifacts to the local host, and is called from both builder and waitblock
> func SaveArtifactLocally(ctx context.Context, parentConsole conslogging.ConsoleLogger, console *conslogging.BufferedLogger, artifact domain.Artifact, indexOutDir string, destPath string, salt string, printPhases, ifExists bool) error {
80c81
< 		artifactPath := trimFilePathPrefix(indexOutDir, from, b.opt.Console)
---
> 		artifactPath := trimFilePathPrefix(indexOutDir, from, parentConsole)
89,90c90,91
< 		if opt.PrintPhases {
< 			artifactColor := b.opt.Console.WithPrefixAndSalt(artifact.Target.String(), salt).PrefixColor()
---
> 		if printPhases {
> 			artifactColor := parentConsole.WithPrefixAndSalt(artifact.Target.String(), salt).PrefixColor()
```

PR #2093, opened 2022-08-11
Friendlier message on ctrl-c in satellite launch or rm

brandonSc:
The operations can be long, especially the delete, and the user may not want to wait for it. The backend handles context-cancelled and allows the operation to finish gracefully. 

Screen Shot 2022-08-11 at 11 12 13 AM

### DIFF
--- a/cmd/earthly/satellite_cmds.go
+++ b/cmd/earthly/satellite_cmds.go
@@ -162,6 +162,10 @@ func (app *earthlyApp) actionSatelliteLaunch(cliCtx *cli.Context) error {
 	app.console.Printf("Launching Satellite. This could take a moment...\n")
 	err = cloudClient.LaunchSatellite(cliCtx.Context, app.satelliteName, orgID)
 	if err != nil {
+		if errors.Cause(err) == context.Canceled {
+			app.console.Printf("Operation cancelled. Satellite will finish launching in background.\n")
+			return nil
+		}
 		return errors.Wrapf(err, "failed to create satellite %s", app.satelliteName)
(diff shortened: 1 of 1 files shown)


cmd/earthly/satellite_cmds.go
```diff
@@ -162,6 +162,10 @@ func (app *earthlyApp) actionSatelliteLaunch(cliCtx *cli.Context) error {
 	app.console.Printf("Launching Satellite. This could take a moment...\n")
 	err = cloudClient.LaunchSatellite(cliCtx.Context, app.satelliteName, orgID)
 	if err != nil {
+		if errors.Cause(err) == context.Canceled {
+			app.console.Printf("Operation cancelled. Satellite will finish launching in background.\n")
```
alexcb, line 166: The wording here sounds a bit odd, it's almost as if it says "Operation cancelled, Operation will continue (in the background)".

Furthermore, there's no guarantee the satellite will actually launch -- the user could be hung waiting for a TCP handshake before making the launch request.

maybe this would be more accurate:

```suggestion
			app.console.Printf("Exiting due to interrupt signal; Satellite launch will continue (if request was successfully received by the server)\n")
```

brandonSc: Yeah that's true. I'll change the wording. Eventually, we could make the process more robust, perhaps by the server acknowledging that it's started the process. In typical cases this should work though, I almost never wait for the operations to finish as it is 😅

vladaionescu (approved):

PR #2099, opened 2022-08-12
print pass/failure message as final line on test

alexcb:
This displays a final pass/failure on the buildkite test, to make
it obvious when the test passes.

Without this, a user may mistake the output of the fail-test (the test
that validates earthly returns a failure code on errors) as the
buildkite-test failing.

### DIFF
--- a/.buildkite/tests.sh
+++ b/.buildkite/tests.sh
@@ -2,6 +2,7 @@
 set -xeu
 
 function cleanup() {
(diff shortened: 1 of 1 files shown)


alexcb:
I initially did a double-take and had to do a `echo $?` when I came back to my terminal and saw this:
image

vladaionescu (approved):

PR #2103, opened 2022-08-15
Fix WITH DOCKER WAIT/END related bug

alexcb:
This fixes a WAIT/END related bug that occured in combination with WITH DOCKER commands,
where the `--load` flag could not interpret the loaded target's docker
name, due to SAVE IMAGE not being stored in mts.

This fix introduces a new `allowDoSaves` arg to prepBuildTarget, which
WITH DOCKER commands will set to false, to prevent loaded images from
being exported to the local docker instance.

### DIFF
--- a/earthfile2llb/converter.go
+++ b/earthfile2llb/converter.go
@@ -1007,10 +1007,10 @@ func (c *Converter) SaveImage(ctx context.Context, imageNames []string, pushImag
 				shouldPush := pushImages && si.DockerTag != "" && c.opt.DoSaves
 				shouldExportLocally := si.DockerTag != "" && c.opt.DoSaves
 				c.waitBlock().addSaveImage(si, c, shouldPush, shouldExportLocally)
-			} else {
-				c.mts.Final.SaveImages = append(c.mts.Final.SaveImages, si)
-			}
 
+				si.SkipBuilder = true
+			}
+			c.mts.Final.SaveImages = append(c.mts.Final.SaveImages, si)
 		}
 
 		if pushImages && imageName != "" && c.opt.UseInlineCache {
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -277,6 +277,18 @@ jobs:
       - name: Execute wait-block test (Fork Only)
         run: ./tests/wait-block/test.sh --build-arg DOCKERHUB_AUTH=false
         if: github.event_name != 'push' && github.event.pull_request.head.repo.full_name != github.repository
+      - name: Execute with-docker test using wait-block override (Earthly Only)
+        run: |-
+          ./build/linux/amd64/earthly --ci -P --version-flag-overrides="wait-block,use-copy-include-patterns" \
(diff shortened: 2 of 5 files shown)


earthfile2llb/converter.go
```diff
@@ -1535,11 +1535,12 @@ func (c *Converter) prepBuildTarget(ctx context.Context, fullTargetName string,
 		opt.DoSaves = c.opt.DoSaves && !target.IsRemote()   // legacy mode only saves artifacts from local targets
 		opt.DoPushes = c.opt.DoPushes && !target.IsRemote() // legacy mode only saves artifacts from local targets
 	}
+	opt.DoSaves = opt.DoSaves && allowDoSaves
```
vladaionescu, line 1538: Wouldn't this already be false on `WITH DOCKER` by the virtue of `WITH DOCKER` calling `BuildAsync` with the `cmdT` set to `LOAD`?

This condition would evaluate to false anyway, in this situation:

```Go
opt.DoSaves = (cmdT == buildCmd && c.opt.DoSaves && !c.opt.OnlyFinalTargetImages)
```

Perhaps the test you were trying to fix was testing the old implementation, where `ReferencedSaveOnly` was `false`? I think we should try not to maintain that combination.

earthfile2llb/converter.go
```diff
@@ -1007,10 +1007,10 @@ func (c *Converter) SaveImage(ctx context.Context, imageNames []string, pushImag
 				shouldPush := pushImages && si.DockerTag != "" && c.opt.DoSaves
 				shouldExportLocally := si.DockerTag != "" && c.opt.DoSaves
 				c.waitBlock().addSaveImage(si, c, shouldPush, shouldExportLocally)
-			} else {
-				c.mts.Final.SaveImages = append(c.mts.Final.SaveImages, si)
-			}
 
+				si.SkipBuilder = true
```
vladaionescu, line 1011: This makes sense: if the wait block handles the export, the builder shouldn't. Also, this will make it easier to wrap the whole execution in an implied wait block when we get there.

earthfile2llb/converter.go
```diff
@@ -1007,10 +1007,10 @@ func (c *Converter) SaveImage(ctx context.Context, imageNames []string, pushImag
...
+			}
+			c.mts.Final.SaveImages = append(c.mts.Final.SaveImages, si)
```
vladaionescu, line 1013: This makes sense - there are places in our code where we introspect the `mts` for the saved images (e.g. `WITH DOCKER`), so we should always put them there. It shouldn't mean that we always export them though.

earthfile2llb/converter.go
alexcb: > Perhaps the test you were trying to fix was testing the old implementation, where `ReferencedSaveOnly` was `false`? I think we should try not to maintain that combination.

That's likely what happened, i'll try backing this out, and might very well need to expand the overrides to be: `--version-flag-overrides="wait-block,use-copy-include-patterns,referenced-save-only"

vladaionescu (approved):

PR #2108, opened 2022-08-17
Init frontend only when actually needed

vladaionescu:
This helps prevent those spurious warnings on unrelated commands such as `earthly account login`.

This PR also reduces the frontend initialization error output to a simple message "No frontend initialized". This can be expanded with the `--verbose` flag.

### DIFF
--- a/cmd/earthly/buildkit.go
+++ b/cmd/earthly/buildkit.go
@@ -1,6 +1,9 @@
 package main
 
 import (
+	"net/url"
+	"time"
+
 	"github.com/earthly/earthly/buildkitd"
 	"github.com/earthly/earthly/cloud"
 	"github.com/earthly/earthly/util/containerutil"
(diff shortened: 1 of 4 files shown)


cmd/earthly/buildkit.go
```diff
@@ -9,7 +12,88 @@ import (
...
+		return errors.New("invalid overridden MTU size")
+	}
+	app.buildkitdSettings.CniMtu = app.cfg.Global.CniMtu
+
+	if app.cfg.Global.IPTables != "" && app.cfg.Global.IPTables != "iptables-legacy" && app.cfg.Global.IPTables != "iptables-nft" {
+		return errors.New(`invalid overridden iptables name. Valid values are "iptables-legacy" or "iptables-nft"`)
+	}
+	app.buildkitdSettings.IPTables = app.cfg.Global.IPTables
+	return nil
+}
```
vladaionescu, line 88: This code just moved from `before` in `main.go`.

brandonSc (approved):
Tested locally as well seems to do the trick

alexcb (approved):

PR #2110, opened 2022-08-17
Add a comparison between satellites and shared cache

vladaionescu:
[Preview](https://docs.earthly.dev/~/revisions/hvID5h4RR4PUbLmlnoMa/earthly-cloud/satellites)

### DIFF
--- a/docs/cloud/satellites.md
+++ b/docs/cloud/satellites.md
@@ -26,6 +26,17 @@ When using Earthly Satellites, even though the build executes remotely, the foll
 
 ![Satellite workflow](./img/satellite-workflow.png)
 
+Earthly Satellite instances come with their own cache volume. This means that performing builds repeatedly on the same satellite will result in faster builds. This can be especially useful when using Satellites from a sandboxed CI environment, where cache from previous builds would not otherwise be available. Below is a comparison with Earthly's [Shared remote cache features](../guides/shared-cache.md).
+
+| Cache characteristic | Satellite | Shared Cache |
+| --- | --- | --- |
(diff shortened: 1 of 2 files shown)


docs/cloud/satellites.md
```diff
@@ -26,6 +26,17 @@ When using Earthly Satellites, even though the build executes remotely, the foll
 
 ![Satellite workflow](./img/satellite-workflow.png)
 
+Earthly Satellite instances come with their own cache volume. This means that performing builds repeatedly on the same satellite will result in faster builds. This can be especially useful when using Satellites from a sandboxed CI environment, where cache from previous builds would not otherwise be available. Below is a comparison with Earthly's [Shared remote cache features](../guides/shared-cache.md).
+
+| Cache characteristic | Satellite | Shared Cache |
+| --- | --- | --- |
+| Storage location | Satellite | A container registry of your choice |
+| Proximity to compute | ✔️ Same machine | ❌ Performing upload/download is required |
```
alexcb, line 34: ultra nit: is an EBS volume considered the same machine? AWS makes it feel as if it is performance-wise.

alexcb (approved):

brandonSc (approved):
Looks good! Perhaps a call out to this page on the docs for remote cache would help inform existing users of that feature there

docs/cloud/satellites.md
vladaionescu: Oh yeah, true. I would keep it as "implementation detail" and not even mention it, to keep this representation simple.

PR #2112, opened 2022-08-17
Update Buildkit and show new field "Seconds since last run"

brandonSc:


vladaionescu (approved):
Looks good - assuming that the buildkit reference will be updated once the buildkit PR this depends on is merged.

vladaionescu:
Hadn't realized that this had auto-merge on. Mind reverting until we get the buildkit PR sorted?

PR #2113, opened 2022-08-17
Return errors in stub frontend

vladaionescu:
In an environment where Docker was not set up and the user forgot that there would be some local outputs from a satellite build they were getting this error:

```
output | WARN: (exporting outputs) failed to copy to tar: rpc error: code = Unknown desc = io: read/write on closed pipe
2022/08/17 16:41:39 http2: server connection error from localhost: connection error: PROTOCOL_ERROR
```

The issue was that the local outputs could not be loaded into Docker because the frontend was not initialized (Docker isn't available). This PR helps with a better error in such cases.

### DIFF
--- a/util/containerutil/stub.go
+++ b/util/containerutil/stub.go
@@ -13,8 +13,8 @@ type stubFrontend struct {
 	*shellFrontend
 }
 
-// FrontendNotInitializedError is returned when the frontend is not initialized.
-var FrontendNotInitializedError = errors.New("frontend (e.g. docker/podman) not initialized")
+// ErrFrontendNotInitialized is returned when the frontend is not initialized.
+var ErrFrontendNotInitialized = errors.New("frontend (e.g. docker/podman) not initialized")
 
 // NewStubFrontend creates a stubbed frontend. Useful in cases where a frontend could not be detected, but we still need a frontend.
 // Examples include earthly/earthly, or integration tests. It is currently only used as a fallback when docker or other frontends are missing.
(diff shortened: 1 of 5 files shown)


alexcb (commented):
nothing obvious jumps out, but I'll defer to @dchw as he's more familiar with this code.

util/containerutil/stub.go
```diff
@@ -7,12 +7,15 @@ import (
 	"github.com/pkg/errors"
 )
 
-// This is a stub for use in internal testing when its too much effort to provide a legitimate backend.
+// This is a stub for use when a proper frontend is not available.
 // Should never be used IRL.
```
dchw, line 11: If we are going to use it IRL, then remove this comment

vladaionescu: Done

dchw:
I can't remember/find the nuances I remember around it. But I am not sure how this will help? If you do not have Docker or Podman installed, it should already error nicely? At least, thats what the _intent_ of this function is: https://github.com/earthly/earthly/blob/823b7cbd704d282722aa6db7e2f632e999dc8954/util/containerutil/frontend.go#L63-L78


... Ah. Looks like we added in using the stub if it fails, _with some warning already_:
https://github.com/earthly/earthly/blob/823b7cbd704d282722aa6db7e2f632e999dc8954/cmd/earthly/main.go#L415-L423

But I guess thats not convenient at the start of a log, when it will cause failures at the end of the log. Is there some way we could just print (or re-print) our already existing error if you are using the stub, rather than depending on the behavior of the stub? Im worried we are starting to depend on the behavior of an implementation originally designed for testing...

vladaionescu:
Right.. getting that printed at the start of the log is one issue - especially if the log is really long (e.g. thousands+ lines).

The other issue is that the error is very cryptic, because it prints multiple lines, some of which are related to technologies you don't even have installed, and others that contain somewhat misleading words like "exit code" if you do not read carefully.

In any case, the issue the user had in this case was that the error was that the error was printed but it was not seen, and when Earthly tried to output something locally, it crashed with something extremely cryptic.

```
output | WARN: (exporting outputs) failed to copy to tar: rpc error: code = Unknown desc = io: read/write on closed pipe
2022/08/17 16:41:39 http2: server connection error from localhost: connection error: PROTOCOL_ERROR
```

This PR makes this error more obvious - it'll say "frontend not initialized" (a more succint and meaningful message), exactly at the moment when this is being exercised (thus more likely for the user to actually see it).

vladaionescu:
> Im worried we are starting to depend on the behavior of an implementation originally designed for testing...

And yes - it was originally designed for testing, but I see nothing wrong with repurposing consciously as it is already being used in real code. Might as well make it official.

dchw (approved):
You have reluctantly convinced me, but I would still rather we take a more proactive approach - if we have loaded the stub frontend, then we can make better decisions instead of relying on an error result. Maybe thats further down the road in a larger initialization refactor?

vladaionescu:
> we can make better decisions instead of relying on an error result

What do you have in mind here, more specifically, for the future?

PR #2119, opened 2022-08-18
Fix retry on context deadline exceeded

vladaionescu:
This does two things:

* Does not retry on context deadline exceeded
* Does not unwrap context canceled and context deadline exceeded before returning the error from cloud client

### DIFF
--- a/cloud/client.go
+++ b/cloud/client.go
@@ -183,27 +183,33 @@ func (c *client) doCall(ctx context.Context, method, url string, opts ...request
 
 	var status int
 	var body string
-	var err error
+	var callErr error
 	duration := time.Millisecond * 100
 	for attempt := 0; attempt < maxAttempt; attempt++ {
-		status, body, err = c.doCallImp(ctx, r, method, url, opts...)
-
-		if err != nil && strings.Contains(err.Error(), "context canceled") {
+		status, body, callErr = c.doCallImp(ctx, r, method, url, opts...)
+		if !shouldRetry(status, body, callErr, c.warnFunc) {
+			return status, body, callErr
+		}
+		if callErr != nil {
 			// Some operations can be canceled gracefully, so we can signal this to higher-level functions.
 			// Note the actual error caught here is not an instance of context.Canceled,
 			// but is an error message that contains the HTTP call plus the string "context canceled".
-			return status, body, context.Canceled
-		}
-
-		if !shouldRetry(status, body, err, c.warnFunc) {
-			return status, body, err
+			switch {
+			case strings.Contains(callErr.Error(), "context canceled"):
+				return status, body, context.Canceled
+			case strings.Contains(callErr.Error(), "context deadline exceeded"):
+				return status, body, context.DeadlineExceeded
+			default:
+				// Keep going.
+			}
 		}
 
 		if status == http.StatusUnauthorized {
 			if !r.hasAuth || alreadyReAuthed {
 				return status, body, ErrUnauthorized
 			}
-			if err = c.Authenticate(ctx); err != nil {
+			err := c.Authenticate(ctx)
+			if err != nil {
 				return status, body, errors.Wrap(err, "auth credentials not valid")
 			}
 			alreadyReAuthed = true
@@ -217,10 +223,10 @@ func (c *client) doCall(ctx context.Context, method, url string, opts ...request
 		duration *= 2
 	}
 
-	return status, body, err
+	return status, body, callErr
 }
 
-func shouldRetry(status int, body string, err error, warnFunc func(string, ...interface{})) bool {
+func shouldRetry(status int, body string, callErr error, warnFunc func(string, ...interface{})) bool {
 	if status == http.StatusUnauthorized {
 		return true
 	}
@@ -233,13 +239,16 @@ func shouldRetry(status int, body string, err error, warnFunc func(string, ...in
 		}
 		return true
 	}
-	if err != nil {
-		if errors.Cause(err) == ErrNoAuthorizedPublicKeys ||
-			errors.Cause(err) == ErrNoSSHAgent ||
-			strings.Contains(err.Error(), "failed to connect to ssh-agent") {
+	if callErr != nil {
+		causeErr := errors.Cause(callErr)
+		if causeErr == ErrNoAuthorizedPublicKeys ||
+			causeErr == ErrNoSSHAgent ||
+			causeErr == context.Canceled ||
+			causeErr == context.DeadlineExceeded ||
+			strings.Contains(callErr.Error(), "failed to connect to ssh-agent") {
 			return false
 		}
-		warnFunc("retrying http request due to unexpected error %v", err)
+		warnFunc("retrying http request due to unexpected error %v", callErr)
 		return true
 	}
 	return false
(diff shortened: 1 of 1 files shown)


cloud/client.go
```diff
@@ -233,13 +239,16 @@ func shouldRetry(status int, body string, err error, warnFunc func(string, ...in
 		}
 		return true
 	}
-	if err != nil {
-		if errors.Cause(err) == ErrNoAuthorizedPublicKeys ||
-			errors.Cause(err) == ErrNoSSHAgent ||
-			strings.Contains(err.Error(), "failed to connect to ssh-agent") {
+	if callErr != nil {
+		causeErr := errors.Cause(callErr)
```
mikejholly, line 243: Can we use `errors.Is` for these?

cloud/client.go
```diff
@@ -183,27 +183,33 @@ func (c *client) doCall(ctx context.Context, method, url string, opts ...request
...
 			// Some operations can be canceled gracefully, so we can signal this to higher-level functions.
 			// Note the actual error caught here is not an instance of context.Canceled,
 			// but is an error message that contains the HTTP call plus the string "context canceled".
-			return status, body, context.Canceled
-		}
-
-		if !shouldRetry(status, body, err, c.warnFunc) {
-			return status, body, err
+			switch {
+			case strings.Contains(callErr.Error(), "context canceled"):
```
mikejholly, line 198: I'm not quite following this. Are we getting `context canceled` as part of the API's response? If the HTTP client is hitting the cancellation, it should produce a legitimate `context.Canceled`, if I'm not mistaken.

alexcb:
An integration test for this would be helpful; this could be done via:

```
earthly --server=http://127.0.0.1:1 +this-doesnt-exist-but-should-attempt-to-send-analytics
```

Currently (without this PR); this produces:

```
 1. Init 🚀

           buildkitd | ...Done


 2. Build 🔧

retrying http request due to unexpected error Post "http://127.0.0.1:1/api/v0/logs": dial tcp 127.0.0.1:1: connect: connection refused
retrying http request due to unexpected error Post "http://127.0.0.1:1/api/v0/logs": dial tcp 127.0.0.1:1: connect: connection refused
retrying http request due to unexpected error Post "http://127.0.0.1:1/api/v0/logs": dial tcp 127.0.0.1:1: connect: connection refused
retrying http request due to unexpected error Post "http://127.0.0.1:1/api/v0/logs": dial tcp 127.0.0.1:1: connect: connection refused
retrying http request due to unexpected error Post "http://127.0.0.1:1/api/v0/logs": dial tcp 127.0.0.1:1: connect: connection refused
retrying http request due to unexpected error Post "http://127.0.0.1:1/api/v0/logs": dial tcp 127.0.0.1:1: connect: connection refused
retrying http request due to unexpected error Post "http://127.0.0.1:1/api/v0/logs": dial tcp 127.0.0.1:1: connect: connection refused
retrying http request due to unexpected error Post "http://127.0.0.1:1/api/v0/logs": dial tcp 127.0.0.1:1: connect: connection refused
retrying http request due to unexpected error Post "http://127.0.0.1:1/api/v0/logs": dial tcp 127.0.0.1:1: connect: connection refused
retrying http request due to unexpected error Post "http://127.0.0.1:1/api/v0/logs": dial tcp 127.0.0.1:1: connect: connection refused
failed to upload log: Post "http://127.0.0.1:1/api/v0/logs": dial tcp 127.0.0.1:1: connect: connection refused
Error: build target: build main: failed to solve: Earthfile line 0:0 target this-doesnt-exist-but-should-attempt-to-send-analytics not found
in		github.com/earthly/earthly:acb/builder-cleanup+this-doesnt-exist-but-should-attempt-to-send-analytics --DOCKERHUB_AUTH=true --DOCKERHUB_MIRROR=registry-1.docker.io.mirror.corp.earthly.dev --DOCKERHUB_TOKEN_SECRET=+secrets/earthly-technologies/dockerhub-mirror/pass --DOCKERHUB_USER_SECRET=+secrets/earthly-technologies/dockerhub-mirror/user --KEEPIT=secret
retrying http request due to unexpected error Put "http://127.0.0.1:1/analytics": dial tcp 127.0.0.1:1: connect: connection refused
retrying http request due to unexpected error Put "http://127.0.0.1:1/analytics": dial tcp 127.0.0.1:1: connect: connection refused
retrying http request due to unexpected error Put "http://127.0.0.1:1/analytics": dial tcp 127.0.0.1:1: connect: connection refused
retrying http request due to unexpected error Put "http://127.0.0.1:1/analytics": dial tcp 127.0.0.1:1: connect: connection refused
```

cloud/client.go
vladaionescu: I believe that @alexcb might know the history behind looking for the string itself.

brandonSc: I added the string check and left a comment above. Apparently, I wasn't getting an instance of `context.Canceled` but a an error containing that in the string. That's why I caught it here and converted to an actual `context.Canceled`

brandonSc: As to why that is the case, I don't think I ever figured that out

cloud/client.go
brandonSc: +1 - apparently this was [added semi-recently in pkg/errors](https://github.com/pkg/errors/releases/tag/v0.9.0) for compatibility with go1.13+ error

brandonSc:
Nit: one difference in this new code seems to be that that the `context canceled` error gets printed now when you Ctrl-C in satellite launch or rm.
Screenshot 2022-08-18 142101

vladaionescu:
Thanks for catching that @brandonSc - I'll adjust accordingly.

cloud/client.go
```diff
@@ -183,27 +183,33 @@ func (c *client) doCall(ctx context.Context, method, url string, opts ...request
...
 	var body string
-	var err error
+	var callErr error
 	duration := time.Millisecond * 100
 	for attempt := 0; attempt < maxAttempt; attempt++ {
-		status, body, err = c.doCallImp(ctx, r, method, url, opts...)
-
-		if err != nil && strings.Contains(err.Error(), "context canceled") {
+		status, body, callErr = c.doCallImp(ctx, r, method, url, opts...)
+		if !shouldRetry(status, body, callErr, c.warnFunc) {
```
brandonSc, line 190: I think by moving this call further up, we are printing the warning "retrying HTTP request" when the error is context cancelled but not actually an instance of it according to the  `errors.Cause` check that was added in `shouldRetry`.  I pasted a screenshot of this in a different comment

alexcb (commented):
reiterating my previous comment regarding a test for this (apparently github doesn't see that comment as me responding to this pr-feedback-request event)

vladaionescu:
@alexcb, regarding the integration test: that wouldn't work for what this PR is fixing. The output is the same both before and after. It is correct for Earthly to retry on connection refused for both logs and analytics. The only difference though is that after the analytics timeout passes, in the new version it stops trying (`context.DeadlineExceeded` error now causes the retry to stop) and thus no additional output is shown after that.

brandonSc (approved):
🥇

alexcb:
> that wouldn't work for what this PR is fixing. The output is the same both before and after.

It would probably need a more sophisticated stub-server, which seems like more work than is warranted for this PR then.

alexcb (approved):

PR #2120, opened 2022-08-18
refactor builder.go to use wait_end.go

alexcb:
This introduces a new `--global-wait-end` flag which enables an implicit
global WAIT/END block, which the directly referenced target (and all
children) belong to.

This allows a large portion of the builder.go's build-function (bf), to
be skipped in favour of using code in wait_end.go' final `wait()` call.

This flag exists to give us test coverage for when we enable wait-end
blocks by default AND remove support for VERSION 0.6.


vladaionescu (approved):
Looks good to me! Should we wait to merge this after the next release? Or do you think it's low-risk enough to put it in the next one?

alexcb:
> Or do you think it's low-risk enough to put it in the next one?

it's rather low risk, since we're not going to encourage users to set `--global-wait-end`, it's merely there to ensure we have (and maintain) test-coverage for our ultimate switch-over. even when we enable `VERSION --wait-end` by default (i.e. in 0.7), we'll still keep this `--global-wait-end` hidden and turned off for our users.

When we're closer to that point in time, we can set `EARTHLY_GLOBAL_WAIT_END=true` in our `./earthly` script to ensure we dog-food it on our local machines.

PR #2124, opened 2022-08-19
A few tweaks and fixes for project/org commands

mikejholly:
* Language tweaks for new commands
* Success message when accepting an invite
* Fix for checking member update args

### DIFF
--- a/cmd/earthly/org_cmds.go
+++ b/cmd/earthly/org_cmds.go
@@ -374,8 +374,12 @@ func (app *earthlyApp) actionOrgMemberList(cliCtx *cli.Context) error {
 func (app *earthlyApp) actionOrgMemberUpdate(cliCtx *cli.Context) error {
 	app.commandName = "orgMemberUpdate"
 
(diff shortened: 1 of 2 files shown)


cmd/earthly/org_cmds.go
```diff
@@ -371,8 +374,8 @@ func (app *earthlyApp) actionOrgMemberList(cliCtx *cli.Context) error {
 func (app *earthlyApp) actionOrgMemberUpdate(cliCtx *cli.Context) error {
 	app.commandName = "orgMemberUpdate"
 
-	if cliCtx.NArg() != 2 {
-		return errors.New("member email and new permission required")
+	if cliCtx.NArg() != 1 {
```
vladaionescu, line 377: I would check here `< 1` and `> 1` separately, so we can say "too many arguments provided" or similar for the `> 1` case.

vladaionescu (approved):

alexcb (approved):

PR #2125, opened 2022-08-19
Remove user name from invite command

mikejholly:
Requiring a name doesn't necessarily add much and can slow down the invite flow.


alexcb (approved):
nice simplification.

PR #2126, opened 2022-08-19
multi4 test is failing (do not merge)

alexcb:
This can be reproduced with:

    ./earthly +for-linux && export earthly=/home/alex/gh/earthly/earthly/build/linux/amd64/earthly && scripts/tests/export.sh

(or change earthly path to match your machine)

It appears this test was passing due to our CI being amd64 and the
`--platform=linux/amd64` being listed first. When I changed the order,
this test now fails.

### DIFF
--- a/builder/docker.go
+++ b/builder/docker.go
@@ -23,12 +23,31 @@ func loadDockerManifest(ctx context.Context, console conslogging.ConsoleLogger, 
 	if len(children) == 0 {
 		return errors.Errorf("no images in manifest list for %s", parentImageName)
 	}
-	// Check if any child has the platform as the default platform (use the first one if none found).
+
+	// Check if any child has the platform as the default platform
 	defaultChild := 0
+	foundPlatform := false
 	for i, child := range children {
 		if child.platform == platutil.DefaultPlatform {
 			defaultChild = i
+			foundPlatform = true
 			break
+		}
+	}
+	if !foundPlatform {
+		// otherwise, look for a platform matching the user's platform
+		for i, child := range children {
+			if child.platform == platutil.UserPlatform {
+				defaultChild = i
+				foundPlatform = true
+				break
+			}
+		}
+		if !foundPlatform {
+			// fall back to using first defined platform (and display a warning)
+			console.Warnf(
+				"Failed to find default, or user-specific platform (%s) of multi-platform image %s; defaulting to the first platform type: %s\n",
+				platutil.UserPlatform, parentImageName, children[defaultChild].platform)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


scripts/tests/export.sh
```diff
@@ -10,7 +10,7 @@ echo "running tests with $earthly"
 date +%s > /tmp/last-earthly-prerelease-check
 
 # ensure earthly login works (and print out who gets logged in)
-test -n "$EARTHLY_TOKEN"
+#test -n "$EARTHLY_TOKEN"
```
alexcb, line 13: commented out to make it easier to run locally.

scripts/tests/export.sh
```diff
@@ -87,7 +87,7 @@ cat >> Earthfile <<EOF
 VERSION 0.6
 
 multi4:
-    BUILD --platform=linux/amd64 --platform=linux/arm64 --platform=linux/arm/v7 +test4
+    BUILD --platform=linux/arm/v7 --platform=linux/amd64 --platform=linux/arm64 +test4
```
alexcb, line 90: changing the order here breaks the test

builder/docker.go
```diff
@@ -23,14 +23,33 @@ func loadDockerManifest(ctx context.Context, console conslogging.ConsoleLogger,
...
 		if child.platform == platutil.DefaultPlatform {
 			defaultChild = i
+			foundPlatform = true
 			break
 		}
 	}
+	if !foundPlatform {
+		// otherwise, look for a platform matching the user's platform
+		for i, child := range children {
+			if child.platform == platutil.UserPlatform {
```
alexcb, line 40: for the test we compare "linux/amd64" vs "user", which doesn't work.

Is there a way to resolve "user" to the actual platform here?

and similarly, does the defaultPlatform need to be resolved?

alexcb: should we somehow call `if child.platform == c.platr.Materialize(platutil.UserPlatform)` ?

builder/docker.go
```diff
@@ -23,14 +23,33 @@ func loadDockerManifest(ctx context.Context, console conslogging.ConsoleLogger,
...
+				defaultChild = i
+				foundPlatform = true
+				break
+			}
+		}
+		if !foundPlatform {
+			// fall back to using first defined platform (and display a warning)
+			console.Warnf(
+				"Failed to find default, or user-specific platform (%s) of multi-platform image %s; defaulting to the first platform type: %s\n",
+				platutil.UserPlatform, parentImageName, children[defaultChild].platform)
```
alexcb, line 50: This shows up on my system as:

```
Failed to find default, or user-specific platform (user) of multi-platform image earthly-export-test-4:test; defaulting to the first platform type: linux/arm/v7
```

builder/docker.go
vladaionescu, line 40: We generally don't default to the user's platform, but to the native platform. At least with the `--new-platform` flag.

alexcb:
fix is in #2127

PR #2133, opened 2022-08-22
Fix typo in go-monorepo example

steshaw:
The `go-monorepo` example doesn't build because of this small typo.


brandonSc:
That was definitely a typo 😅 
Thanks for the fix!

brandonSc (approved):

PR #2134, opened 2022-08-22
Use New Buildkit Reservation Endpoint

brandonSc:
On every connection to Buildkit, we attempt to call the new `Reserve` endpoint. When Buildkit is running as a Satellite, this will ensure the instance is not put to sleep during the initial connection phase.


brandonSc:
Corresponding buildkit PR https://github.com/earthly/buildkit/pull/94

conslogging/conslogging.go
```diff
@@ -206,7 +206,6 @@ func (cl ConsoleLogger) PrintPhaseHeader(phase string, disabled bool, special st
 	if underlineLength < barWidth {
 		underlineLength = barWidth
 	}
-	cl.errW.Write([]byte("\n"))
```
vladaionescu, line 209: Intended change?

cmd/earthly/debug_cmds.go
```diff
@@ -104,7 +104,6 @@ func (app *earthlyApp) actionDebugBuildkitInfo(cliCtx *cli.Context) error {
 	fmt.Printf("Buildkit revision: %s\n", info.BuildkitVersion.Revision)
 	fmt.Printf("Buildkit package: %s\n", info.BuildkitVersion.Package)
 	fmt.Printf("Num sessions: %d\n", info.NumSessions)
-	fmt.Printf("Seconds idle: %d\n", info.SecondsIdle)
```
vladaionescu, line 107: Intended change?

CONTRIBUTING.md
```diff
@@ -98,6 +98,8 @@ To update earthly's reference to buildkit, you may run `earthly +update-buildkit
 
 Updates to fsutils must first be vendored into buildkit, then updated under `go.mod`; additional docs and scripts exist in the buildkit repo.
 
+Note that for record-keeping purposes, contributors are encouraged to manually sqaush their commits before merging a PR on our buildkit fork, as well as leave a comment showing a corresponding PR in `earthly/earthly` using that buildkit version with passing test cases.
```
vladaionescu, line 101: This change seems unrelated?

buildkitd/buildkitd.go
```diff
@@ -594,6 +594,15 @@ func checkConnection(ctx context.Context, address string, opts ...client.ClientO
 			return
 		}
 		defer bkClient.Close()
+		err = bkClient.Reserve(ctxTimeout, 5*time.Second)
+		if err != nil {
+			s, ok := status.FromError(errors.Cause(err))
+			if ok && s.Code() != codes.Unimplemented {
+				mu.Lock()
+				connErr = err
+				mu.Unlock()
```
vladaionescu, line 603:
```suggestion
			if !ok || s.Code() != codes.Unimplemented {
				mu.Lock()
				connErr = err
				mu.Unlock()
				return
			}
```

CONTRIBUTING.md
brandonSc: Sorry this was a bad squash and merge thing. There are changes in main showing up here. I'll fix it asap.

cmd/earthly/debug_cmds.go
brandonSc: This one was intentional actually because the seconds Idle always shows zero now.  I.e. we reset to zero through the Reserve call as we're connecting to it in order to print the debug info. 

We could change the code so that it doesn't always call the reserve endpoint, but I didn't think that would be worth it.

vladaionescu: Oh yeah, true

brandonSc:
Too many issues with old commits that were squashed in a previous PR. Opening a new PR on a different branch.

PR #2135, opened 2022-08-22
Fixes and improvements for new secrets commands

mikejholly:
A few minor fixes and tweaks for the new projects & secrets commands.

### DIFF
--- a/cmd/earthly/secret_cmds.go
+++ b/cmd/earthly/secret_cmds.go
@@ -117,7 +117,7 @@ func (app *earthlyApp) secretCmdsPreview() []*cli.Command {
 		{
 			Name:      "migrate",
 			Usage:     "Migrate existing secrets into the new project-based structure",
-			UsageText: "earthly [options] secret migrate --org <organization> --project <project> <source-organization>",
+			UsageText: "earthly [options] secret --org <organization> --project <project> migrate <source-organization>",
 			Action:    app.actionSecretsMigrate,
 			Flags: []cli.Flag{
 				&cli.BoolFlag{
@@ -291,7 +291,12 @@ func (app *earthlyApp) actionSecretsListV2(cliCtx *cli.Context) error {
 	}
 
 	for _, secret := range secrets {
-		fmt.Println(secret.Path)
+		display := secret.Path
+		if !strings.HasPrefix(display, "/user") {
+			prefix := fmt.Sprintf("/%s/%s/", cliCtx.String("org"), cliCtx.String("project"))
+			display = strings.Replace(display, prefix, "", 1)
+		}
+		app.console.Printf(display)
 	}
 
 	return nil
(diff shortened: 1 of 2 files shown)


cmd/earthly/secret_cmds.go
```diff
@@ -291,7 +291,12 @@ func (app *earthlyApp) actionSecretsListV2(cliCtx *cli.Context) error {
 	}
 
 	for _, secret := range secrets {
-		fmt.Println(secret.Path)
+		display := secret.Path
+		if !strings.HasPrefix(display, "/user") {
+			prefix := fmt.Sprintf("/%s/%s/", cliCtx.String("org"), cliCtx.String("project"))
```
vladaionescu, line 296: Can you use the variables here, rather than `cliCtx.String(...)`?

cmd/earthly/secret_cmds.go
```diff
@@ -291,7 +291,12 @@ func (app *earthlyApp) actionSecretsListV2(cliCtx *cli.Context) error {
...
+			display = strings.Replace(display, prefix, "", 1)
```
vladaionescu, line 297: Wondering if this can be brittle to future changes? This replacement doesn't actually require for the "prefix" to be at the beginning, yet it yanks it regardless.

What do you think about using `TrimPrefix` instead?

cmd/earthly/secret_cmds.go
mikejholly: Which variables?

cmd/earthly/secret_cmds.go
mikejholly: Good call. I forgot about that function.

cmd/earthly/secret_cmds.go
vladaionescu: That long list that you don't really like :-)

https://github.com/earthly/earthly/blob/main/cmd/earthly/main.go#L65

mikejholly: I was going to PR the `--project` flag (rather than arg) changes separately. Mind if we switch to the `cliFlags` struct there?

vladaionescu: :+1:

vladaionescu (approved):

cmd/earthly/secret_cmds.go
```diff
@@ -291,7 +291,12 @@ func (app *earthlyApp) actionSecretsListV2(cliCtx *cli.Context) error {
 	}
 
 	for _, secret := range secrets {
-		fmt.Println(secret.Path)
+		display := secret.Path
+		if !strings.HasPrefix(display, "/user") {
```
vladaionescu, line 295: Oh one more comment

```suggestion
		if !strings.HasPrefix(display, "/user/") {
```

vladaionescu (changes requested):

cmd/earthly/secret_cmds.go
vladaionescu: (In case somebody has an org name `userblabla`)

vladaionescu (approved):

PR #2136, opened 2022-08-23
Use new Buildkit "Reservation" Endpoint

brandonSc:


buildkitd/buildkitd.go
```diff
@@ -594,6 +594,26 @@ func checkConnection(ctx context.Context, address string, opts ...client.ClientO
...
 		defer bkClient.Close()
+		err = bkClient.Reserve(ctxTimeout, 5*time.Second)
+		if err != nil {
+			s, ok := status.FromError(errors.Cause(err))
+			// Note that if err == codes.Unimplemented,
+			// we assume an older version of buildkit and continue to call ListWorkers.
+			switch {
+			case !ok:
+				mu.Lock()
+				connErr = errors.New("unexpected error from buildkit reserve endpoint")
```
vladaionescu, line 605: I would not hire the original error in this case - we should at least `Wrap` the original error.

buildkitd/buildkitd.go
```diff
@@ -594,6 +594,26 @@ func checkConnection(ctx context.Context, address string, opts ...client.ClientO
...
+				mu.Unlock()
+			case s.Code() == codes.Unavailable:
+				mu.Lock()
+				connErr = errors.New("buildkit is currently shutting down")
```
vladaionescu, line 609: Shutting down or starting up? Why not just use the original `err`?

buildkitd/buildkitd.go
```diff
@@ -594,6 +594,26 @@ func checkConnection(ctx context.Context, address string, opts ...client.ClientO
...
+				mu.Unlock()
+			case s.Code() != codes.Unimplemented:
+				mu.Lock()
+				connErr = err
+				mu.Unlock()
+			}
+		}
```
vladaionescu, line 616: I would maybe only keep a simple `if` statement, given that we want to do pretty much the same thing in each case (pass on the original error)

brandonSc: Reworked as you described. Single `if` statement + return original error

vladaionescu (approved):
Looks good - assuming that the buildkit references are updated to point to a sha in `earthly-main` branch before merging.

PR #2137, opened 2022-08-23
Use project flag for project commands and cliFlags

mikejholly:
A few tweaks for the new project commands.

### DIFF
--- a/cmd/earthly/main.go
+++ b/cmd/earthly/main.go
@@ -133,7 +133,6 @@ type cliFlags struct {
 	containerFrontend         containerutil.ContainerFrontend
 	logSharing                bool
 	satelliteName             string
-	satelliteOrg              string
 	noSatellite               bool
 	userPermission            string
 	noBuildkitUpdate          bool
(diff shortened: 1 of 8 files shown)


cmd/earthly/main.go
```diff
@@ -138,6 +138,8 @@ type cliFlags struct {
 	userPermission            string
 	noBuildkitUpdate          bool
 	globalWaitEnd             bool // for feature-flipping builder.go code removal
+	projectName               string
+	orgName                   string
```
vladaionescu, line 142: There's also `satelliteOrg` - can we merge these two?

cmd/earthly/org_cmds.go
```diff
@@ -266,7 +266,7 @@ func (app *earthlyApp) actionOrgInviteEmail(cliCtx *cli.Context) error {
 		return errors.Wrap(err, "failed to create cloud client")
 	}
 
-	orgName, err := projectOrgName(cliCtx, cloudClient)
+	orgName, err := projectOrgName(cliCtx.Context, app, cloudClient)
```
vladaionescu, line 269: Better to turn `projectOrgName` to a member function of `app`?

vladaionescu (approved):

PR #2138, opened 2022-08-24
Fix +examples2 on Satellites from Forks

brandonSc:
+examples2 was failing because QEMU was not installed. The satellites tests just run local buildkit instead of actually on satellites when they are triggered by a fork.

Example of passing test from a fork https://github.com/earthly/earthly/actions/runs/2920999640

### DIFF
--- a/.github/workflows/satellites.yml
+++ b/.github/workflows/satellites.yml
@@ -108,6 +108,13 @@ jobs:
     steps:
       - uses: earthly/actions-setup@v1.0.1
       - uses: actions/checkout@v3
+      - name: Set up QEMU (Forks Only)
+        id: qemu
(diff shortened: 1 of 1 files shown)


vladaionescu (approved):

vladaionescu:
BTW should we just not run this at all on forks? We anyway have the non-satellite `+examples2` run.

brandonSc:
> BTW should we just not run this at all on forks? We anyway have the non-satellite +examples2 run.

We could also skip the step that actually runs the build. Do you know of a cleaner way to just skip the entire set of tests instead of skipping each individual step with an `if`?

brandonSc:
Will improve the script more in a future PR so that it skips entirely the tests on satellites. For now this at least immediately fixes the broken test.

PR #2143, opened 2022-08-25
Invite tweaks: require the least amount of decisions of the user

vladaionescu:
This makes the "read" permission the default choice if they enter nothing. It also removes the "message" parameter from the flow by default (it can still be passed in via a flag).

### DIFF
--- a/cmd/earthly/org_cmds.go
+++ b/cmd/earthly/org_cmds.go
@@ -297,6 +297,11 @@ func (app *earthlyApp) actionOrgInviteEmail(cliCtx *cli.Context) error {
 		permission = "admin"
 	default:
 	}
+	switch permission {
+	case "read", "write", "admin":
+	default:
+		return fmt.Errorf("invalid permission %s", permission)
+	}
 	invite.Permission = permission
 	invite.Message = app.inviteMessage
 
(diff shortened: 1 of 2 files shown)


cmd/earthly/org_cmds.go
```diff
@@ -281,23 +278,27 @@ func (app *earthlyApp) actionOrgInviteEmail(cliCtx *cli.Context) error {
...
-	if message == "" {
-		message, err = promptInput(cliCtx.Context, "Message to user: ")
-		if err != nil {
-			return errors.Wrap(err, "failed to read message")
-		}
+	permission = strings.ToLower(permission)
+	switch permission {
+	case "":
+		permission = "read"
+	case "r":
```
mikejholly, line 292: Are we expecting the first char or the full word? Based on the prompt text I'd assume the latter.

brandonSc: Perhaps it should take both?

vladaionescu: This simply allows it to take both

mikejholly: I suppose with the `default` it'll pass through the full word. Might be nice to limit/validate, but this is fine.

mikejholly (approved):

cmd/earthly/org_cmds.go
vladaionescu: :+1: Added validation

PR #2152, opened 2022-08-26
Update crypto to address "parse error in message type 27" errors

### LINKED ISSUE #2151: Spurious "parse error in message type 27" errors in build output
With increasing frequency, my builds have started generating annoying (but harmless) errors such as `2022/08/26 13:59:40 agent 27: ssh: parse error in message type 27`.  I finally tracked it down to https://github.com/golang/go/issues/51689, and the increased frequency was directly related to upgrading projects to use newer alpine 3.16-based images (which includes the newer/impacted SSH tools).

I don't know if there are other ways to see the behavior in action, but the commonality is all of my occurrences is usage of `RUN --ssh`.  As such, it would be hard to give a reproducible example (needing a live SSH server).

A snippet of the error in action
```
            +new-ssh | --> RUN git clone ssh://git.example.com:7999/a/test.git
            +new-ssh | Cloning into 'test'...
2022/08/26 14:35:41 agent 27: ssh: parse error in message type 27
```
Of course that's a no-no, but in my _actual_ use case,  it's running `RUN --ssh pip3 install -c constraints.txt -r requirements-prod.txt`, and one of the dependencies is `git+ssh://git.example.com:7999/a/test.git#egg=structlog-ecs-utils`.

rrjjvv:
Fixes #2151 

This resolves errors stemming from https://github.com/golang/go/issues/51689; it just so happens it was fixed a few hours ago.  I didn't think a dependency change of this nature would warrant any new tests, but correct me if I'm wrong.  But this does eliminate the errors after applying this change.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -33,7 +33,7 @@ require (
 	github.com/stretchr/testify v1.7.0
 	github.com/tonistiigi/fsutil v0.0.0-20220510150904-0dbf3a8a7d58
 	github.com/urfave/cli/v2 v2.3.0
-	golang.org/x/crypto v0.0.0-20220315160706-3147a52a75dd
+	golang.org/x/crypto v0.0.0-20220826181053-bd7e27e6170d
 	golang.org/x/sync v0.0.0-20210220032951-036812b2e83c
 	golang.org/x/term v0.0.0-20210927222741-03fcf44c2211
 	google.golang.org/grpc v1.47.0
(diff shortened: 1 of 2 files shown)


rrjjvv:
Also, for my future reference... do you consider it a "rule" to have an Issue filed prior to submitting a PR (at least from external contributors)?  I was going to create this PR only, but at the last minute, thought you might want/require a matching Issue.  If it could/should be omitted for trivial changes like this, I'll gladly comply.

alexcb (approved):

rrjjvv:
Sweet, thanks!  It looks like 4/5 of those failures are the Hub 'EOF' issue mentioned in Slack earlier.  The fifth is a little obtuse to my untrained eye, but blinding assuming/hoping it's the same underlying issue.  I'll assume there's nothing for me to do at this point until those are re-run.

alexcb:
I'll try to get this released this week, thanks again for the fix.

PR #2153, opened 2022-08-29
npm-update-all target

alexcb:
This adds a target which performs an npm update on all examples and
tests. This is to keep packages up to date with security patches.


vladaionescu (approved):
<3

Issue #2142, opened 2022-08-24
earthly/earthly:latest does not point to earthly/earthly:v0.6.22


DeveloperC286:
The image `earthly/earthly:latest` on Docker hub does not point to the image `earthly/earthly:v0.6.22`.
This is a bug because customers believe they are using the latest release when in fact they are using an old version of Earthly.

The image `earthly/earthly:latest` currently points to a two month old version.

https://hub.docker.com/layers/earthly/earthly/earthly/latest/images/sha256-64a547c4b42178370adb7f2b3394fc81e893ebc5843a5c88300f1788c0f492ac?context=explore

alexcb:
Thanks for pointing that out -- I fixed the issue in our release script, and manually updated the current latest tag to point to 0.6.22

```bash
docker rmi earthly/earthly:latest && docker run --rm --entrypoint='' earthly/earthly:latest /usr/bin/earthly --version
Untagged: earthly/earthly:latest
Unable to find image 'earthly/earthly:latest' locally
latest: Pulling from earthly/earthly
Digest: sha256:82ca574edfb868b0a8ad08e0f7d62f81b38380dd9221497fb92400a4008a9503
Status: Downloaded newer image for earthly/earthly:latest
earthly version v0.6.22 1c8bcfbab64b9b52302ef63c897e6b989b88a794 linux/amd64; Alpine Linux 
```

alexcb:
it appears my previous fix needs a bit more work, both versions point to the same git sha that was used for the build; however the latest tag displays "latest" when I would expect it to say "v0.7.4" -- and in fact reference the exact same binary.
```
$ docker run --rm --entrypoint="" earthly/earthly:v0.7.4 earthly --version
earthly version v0.7.4 a1c46ca2b75f21a4948fd53b5464de5a2bd90372 linux/amd64; Alpine Linux 

$ docker run --rm --entrypoint="" earthly/earthly:latest earthly --version
earthly version latest a1c46ca2b75f21a4948fd53b5464de5a2bd90372 linux/amd64; Alpine Linux 
```

alexcb:
v0.7.7 still has this issue:

```
$ docker run --rm --entrypoint="" earthly/earthly:latest earthly --version
earthly version latest 547fb642f99f6b260aa19b00d7ae22eaafb2ba64 linux/amd64; Alpine Linux 
```

instead I would expect to see 

```
[...]

PR #2155, opened 2022-08-29
fix release.sh to update latest docker tag

alexcb:


alexcb:
this should fix #2142 for future releases.

vladaionescu (approved):

PR #2156, opened 2022-08-29
fix wait-end multiplatform exports

alexcb:
This fixes a bug where the default platform was never exported to the
local docker instance.

### DIFF
--- a/earthfile2llb/wait_block.go
+++ b/earthfile2llb/wait_block.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"path/filepath"
+	"strings"
 	"sync"
 
 	"github.com/earthly/earthly/conslogging"
@@ -147,6 +148,10 @@ func (wb *waitBlock) saveImages(ctx context.Context) error {
 
 	gwCrafter := gatewaycrafter.NewGatewayCrafter()
 
+	// these are used to pass manifest data to the onImage function in builder.go; this only applies to non-local-registry exports (e.g. satellites)
+	var tarImagesInWaitBlockRefPrefixes []string
+	var tarImagesInWaitBlock []string
+
 	refID := 0
 	for _, item := range imageWaitItems {
 		sessionID := item.c.opt.GwClient.BuildOpts().SessionID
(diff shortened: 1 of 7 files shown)


earthfile2llb/wait_block.go
```diff
@@ -198,16 +199,19 @@ func (wb *waitBlock) saveImages(ctx context.Context) error {
...
 					return err
 				}
-				if item.c.opt.UseLocalRegistry {
-					localRegPullID := pullPingMap.Insert(sessionID, platformImgName)
-					gwCrafter.AddMeta(fmt.Sprintf("%s/export-image-local-registry", refPrefix), []byte(localRegPullID))
-				} else {
-					gwCrafter.AddMeta(fmt.Sprintf("%s/export-image", refPrefix), []byte("true"))
+				if !item.c.opt.UseLocalRegistry {
+					return errors.Errorf(
+						"the --uselocalregistry feature must also be enabled in order to use multi-platform images in combination with the --wait-block feature")
```
vladaionescu, line 204: Note that in the context of satellites, we still use the old tar-based exporter... I think we ignore this flag in that case.

Should we support tar-based exporting here too for the satellite case?

alexcb: interesting point. I was using the pullping map to "smuggle" the manifest from this code to the builder code -- i'll have to do some testing against a satellite to see what breaks.

vladaionescu: Let me know if you need to pair up - I'm in and out this week with an unpredictable schedule, but I can make myself available if you need me.

alexcb: I had to pass along a reference to the pullpingmap, along with a list of manifest keys (in order to know when all images from the wait block have been exported); it's a bit ugly since the pullpingmap is used for both onPull and onImage now.

maybe it should be renamed to manifestSmuggler, or something more generic?

if you have other ideas on how to achieve this, i'm keen to pair up.

vladaionescu: Do I understand correctly that this is a temporary hack until we move off of the old builder code?

alexcb: unfortunately not all of it -- the `onImage` and `onImageDone` functions are required for the new code.

vladaionescu: Makes sense - this could probably be improved, but it could also be a future PR.

Perhaps the PullPingMap needs to evolve into something more complex, or something.

vladaionescu (approved):

PR #2122, opened 2022-08-19
Remove WAIT/END is experimental warning

alexcb:
It is still experimental; however, the code is now much more complete,
and has a higher degree of test-coverage.

The push-images test was modified due to the --wait-block feature no
longer prevents saving an image after a push, e.g.

    RUN --push ...
    SAVE IMAGE ...

is now valid when run with the `--wait-end` feature, since it roughly
translates into

    ARG EARTHLY_PUSH
    IF [ $EARTHLY_PUSH = "true" ]
        RUN --push ...
    END
    SAVE IMAGE ...

Signed-off-by: Alex Couture-Beil <alex@earthly.dev>

PR #2159, opened 2022-08-30
fix: SAVE IMAGE --push in WAIT/END always happened

alexcb:
This fixes a bug where SAVE IMAGE --push commands always pushed an
image, even when earthly was run without the --push flag.


alexcb:
This bug was found via #2122; once we enable dogfooding, we'll have proper test coverage.

vladaionescu (approved):

PR #2165, opened 2022-09-02
Allow Feature Flags when Launching Satellites

brandonSc:
To enable a satellite with sleep/wake functionality:
```
earthly satellite launch --feature-flags satellite-sleep my-satellite
```

This can be re-used to enable feature flags we add for satellites in the future as well. This is specific, however, to features that affect how the satellite is created (vs other features which could take effect at run-time during a build).


go.sum
```diff
@@ -372,8 +372,8 @@ github.com/dustin/go-humanize v1.0.0 h1:VSnTsYCnlFHaM2/igO1h6X3HA71jcobQuxemgkq4
 github.com/dustin/go-humanize v1.0.0/go.mod h1:HtrtbFcZ19U5GC7JDqmcUSB87Iq5E25KnS6fMYU6eOk=
 github.com/earthly/buildkit v0.0.1-0.20220824153500-64d715da40d6 h1:oZcbJXVZsKLv+ftz3qu6fma/vCp/Z1NZvkKmaxO3l/w=
 github.com/earthly/buildkit v0.0.1-0.20220824153500-64d715da40d6/go.mod h1:Ih6/jh6JTbktkhxRID0gj6dn2GAPI6H9U6xKVxgNqBk=
-github.com/earthly/cloud-api v1.0.1-0.20220819185746-bab168ecbbf9 h1:r24RzVlEmXaJf/tX0smUEUuGYk4bkcn8TqFI5BoqR9g=
-github.com/earthly/cloud-api v1.0.1-0.20220819185746-bab168ecbbf9/go.mod h1:JhlHsW6o8zYa+XsM0nMAevRQtES35KE5R2G8tJALsnI=
+github.com/earthly/cloud-api v1.0.1-0.20220901205940-19f57fdc4d9c h1:bz2bqba5hDZuw3bEzRxgUdLtHBqaC8mO4MTOV1H4K4E=
+github.com/earthly/cloud-api v1.0.1-0.20220901205940-19f57fdc4d9c/go.mod h1:JhlHsW6o8zYa+XsM0nMAevRQtES35KE5R2G8tJALsnI=
```
brandonSc, line 376: I need to update this before merging, after the PR for satellite sleep on the backend is merged.

cmd/earthly/satellite_cmds.go
```diff
@@ -23,6 +24,16 @@ func (app *earthlyApp) satelliteCmds() []*cli.Command {
 			UsageText: "earthly satellite launch <satellite-name>\n" +
 				"	earthly satellite [--org <organization-name>] launch <satellite-name>",
 			Action: app.actionSatelliteLaunch,
+			Flags: []cli.Flag{
+				&cli.StringFlag{
```
vladaionescu, line 28: You can use the `StringSliceFlag` type here. In which case I would call it `feature-flag` (singular).

You use it like this: `--feature-flag feat1 --feature-flag feat2 ...` on the command-line, and comma-separated via the env var.

vladaionescu (approved):

cmd/earthly/satellite_cmds.go
brandonSc: Oh that is cleaner, I'll switch to that

alexcb (approved):

vladaionescu (approved):

PR #2166, opened 2022-09-03
New subcommand to list invite status

mikejholly:
New subcommand that helps users determine the status of previously sent invites.

```
earthly preview orgs --org earthly-technologies invite ls
User Email                     Permission  Created     Accepted
mike@earthly.dev               admin       2022-07-29  No
mikejholly@gmail.com           read        2022-08-22  2022-08-22
mikejholly+test001@gmail.com   read        2022-08-22  2022-08-22
mikejholly+test0002@gmail.com  read        2022-08-24  No
mikejholly+test0003@gmail.com  admin       2022-08-24  No
mikejholly+test0004@gmail.com  read        2022-08-25  2022-08-25
```

### DIFF
--- a/cmd/earthly/org_cmds.go
+++ b/cmd/earthly/org_cmds.go
@@ -122,6 +122,14 @@ func (app *earthlyApp) orgCmdsPreview() []*cli.Command {
 					UsageText:   "earthly org invite accept <invite-code>",
 					Action:      app.actionOrgInviteAccept,
 				},
+				{
+					Name:        "ls",
+					Aliases:     []string{"list"},
+					Usage:       "List all pending and accepted invitations",
+					Description: "List all pending and accepted invitations",
+					UsageText:   "earthly org [--org <organization>] invite ls",
+					Action:      app.actionOrgInviteList,
+				},
 			},
 		},
 	}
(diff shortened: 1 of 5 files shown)


alexcb (commented):
I think it would be useful to also include the functionality for a user to accept one or more of these invites.

currently it's not clear how a user could act on the pending invites.

mikejholly:
Well, we have the invite accept command: `earthly preview org invite accept <code>`. Is that what you mean?

alexcb:
> Well, we have the invite accept command: `earthly preview org invite accept <code>`. Is that what you mean?

How does one get the code?

From a usability point of view, it would be nice to be able to accept an invite without having to go to your email.

alexcb:
let me revisit this -- I was thinking about this from the perspective of the use accepting the invite; not from the perspective of an admin sending out the invite.

Do we have functionality to delete pending non-accepted invites?

alexcb:
I wonder if this should (eventually) show both received and sent invites, for example:

```
earthly preview orgs --org earthly-technologies invite ls
received:
Permission     Created        Code
read           2022-07-29     abc123

sent:
User Email                     Permission  Created     Accepted
mike@earthly.dev               admin       2022-07-29  No
mikejholly@gmail.com           read        2022-08-22  2022-08-22
mikejholly+test001@gmail.com   read        2022-08-22  2022-08-22
mikejholly+test0002@gmail.com  read        2022-08-24  No
mikejholly+test0003@gmail.com  admin       2022-08-24  No
mikejholly+test0004@gmail.com  read        2022-08-25  2022-08-25
```

Something like the above would have made it more clear to me (a user seeing it for the first time). But now that I understand it's only for outgoing invites, it seems good.

cmd/earthly/org_cmds.go
```diff
@@ -122,6 +122,14 @@ func (app *earthlyApp) orgCmdsPreview() []*cli.Command {
 					UsageText:   "earthly org invite accept <invite-code>",
 					Action:      app.actionOrgInviteAccept,
 				},
+				{
+					Name:        "ls",
+					Aliases:     []string{"list"},
+					Usage:       "List all pending and accepted invitations",
```
alexcb, line 128: maybe this needs to be clarified that it's only sent invites:
```suggestion
					Usage:       "List all sent invitations (both pending and accepted)",
```

alexcb (approved):
looks good, just a minor suggestion for usage text.

vladaionescu (approved):

mikejholly:
> not from the perspective of an admin sending out the invite.

Yep, this is meant to give admins an idea of who's been invited and still needs to accept. Personally, I think accepting via email is a good start. Perhaps we can improve the non-email process later.

> Do we have functionality to delete pending non-accepted invites?

No, not at the moment. But it's not a bad idea.

PR #2171, opened 2022-09-07
Improve new command UX with success messages

mikejholly:
Add some helpful success messages for some of the new commands.

### DIFF
--- a/cmd/earthly/org_cmds.go
+++ b/cmd/earthly/org_cmds.go
@@ -246,7 +246,7 @@ func (app *earthlyApp) actionOrgInviteAccept(cliCtx *cli.Context) error {
 		return errors.Wrap(err, "failed to accept invite")
 	}
 
-	app.console.Printf("Invite accepted!\n")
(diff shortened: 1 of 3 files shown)


vladaionescu (approved):

cmd/earthly/org_cmds.go
```diff
@@ -359,6 +359,11 @@ func (app *earthlyApp) actionOrgMemberList(cliCtx *cli.Context) error {
 		return err
 	}
 
+	if len(members) == 0 {
+		app.console.Printf("No members in %s", orgName)
```
alexcb, line 363: the downside with a change like this, is now it's more difficult to use pipes and commands like `xargs`.

mikejholly: I thought about this as well but I think `app.console.Printf` will write to Stderr.

alexcb: oh nice! win-win :)

PR #2175, opened 2022-09-08
Node.js Typescript Example

r-LaForge:
Add a Typescript example with Node.js
include integration test.


examples/typescript-node/Earthfile
```diff
@@ -0,0 +1,45 @@
...
+integration-test:
+    FROM +test-setup
+    COPY +build/dist dist
+    WITH DOCKER --load typescript-node-example:latest=+docker
+        RUN docker run --rm -d -p 8080:8080 typescript-node-example:latest && npm run test:integration
+    END
+
+
+all:
+    BUILD +deps
```
vladaionescu, line 41: `+deps` does not seem necessary to be mentioned here, as it has no output of its own.

vladaionescu (commented):
Looks great!

One nit! And also, I would also add this to the `+examples2` target so that it gets picked up by CI: https://github.com/earthly/earthly/blob/main/Earthfile#L560

examples/typescript-node/Earthfile
r-LaForge: 👍 done

r-LaForge:
@vladaionescu [I have already done that](https://github.com/earthly/earthly/pull/2175/files#diff-2705f6016a64189abd14041e2f3e8dbb829553295d9088e4eb42eccb5e61bf47R579) :)

vladaionescu (approved):

PR #2177, opened 2022-09-12
Satellite select with no name prints current

r-LaForge:
## Description
Now using `earthly sat select` with no name prints the selected satellite and specifies usage.
- [x] Specify the currently selected satellite if not given a name.
- [x] Displaying usage text if no name was given.
- [x] preserve the error exit as well

## Screenshots of error scenarios
**Running Select with no satellite selected**
image

**Running Select with a satellite already selected**
image

**Attempting to set 2 satellite names**
image

@vladaionescu


alexcb (approved):

cmd/earthly/satellite_cmds.go
```diff
@@ -302,7 +304,15 @@ func (app *earthlyApp) actionSatelliteSelect(cliCtx *cli.Context) error {
 	app.commandName = "satelliteSelect"
 
 	if cliCtx.NArg() != 1 {
-		return errors.New("satellite name is required")
+		if app.cfg.Satellite.Name == "" {
+			app.console.Printf("No satellite selected\n\n")
+			app.console.Printf("To select a satellite:\n\t%s\n", selectUsageText)
```
brandonSc, line 309: I like the usage text being printed when the command isn't used properly, however, I wonder if the command should return  non-zero exit code.

I'm picturing someone's CI script where it might be useful to know that this failed here and didn't actually select a satellite before it continued on to other things.

r-LaForge: That's a very good point.. A common case that immediately comes to mind is using an ARG / ENV variable for the select statement and forgetting to set it.

I'll take another look at this later and see about returning a non-zero exit code.

cmd/earthly/satellite_cmds.go
```diff
@@ -302,7 +304,15 @@ func (app *earthlyApp) actionSatelliteSelect(cliCtx *cli.Context) error {
...
+		} else {
+			app.console.Printf("Selected satellite: %s\n\n", app.cfg.Satellite.Name)
+			app.console.Printf("To select a different satellite:\n\t%s\n", selectUsageText)
```
brandonSc, line 312: Is this handling the case where a user is re-selecting the satellite with the same name?

cmd/earthly/satellite_cmds.go
brandonSc: Normally it's the fact that an error was returned instead of nil.

I do really like the idea in general of commands printing the usage text when they aren't used properly. Wonder if there's a nice solution that can be reused in all commands. What do you think @alexcb ?

alexcb: It might be possible to tie in to the `cli` package to access the print usage function (whatever it's called), but I don't know offhand (happy to help if needed).

vladaionescu (approved):

cmd/earthly/satellite_cmds.go
r-LaForge: Not sure if there is a nicer solution to add that as default behaviour.. We could look.
Worst case scenario, we could make a little closure to do just that.
Roughly, something like this:
```
func wrapAction(usageText string, action func(cliCtx *cli.Context) error) func(cliCtx *cli.Context) error {
	return func(cliCtx *cli.Context) error {
		err := action(cliCtx)
		if err != nil {
			fmt.Printf(usageText)
		}
		return err
	}
}
```

and then we'd use it like this: 
```
{
			Name:        "select",
			Aliases:     []string{"s"},
			Usage:       "Choose which satellite to use to build your app",
			Description: "Choose which satellite to use to build your app",
			UsageText:   selectUsageText,
			Action:      wrapAction(selectUsageText, app.actionSatelliteSelect),
},
```

r-LaForge: Now returning an error 1 line below.

brandonSc: I also just noticed the `cli` package has this function, which might do a very similar thing:
```
// ShowCommandHelpAndExit - exits with code after showing help
func ShowCommandHelpAndExit(c *Context, command string, code int) {
```

cmd/earthly/satellite_cmds.go
brandonSc: I ask because it sounds like that what is intended in the PR description:

> Select with a satellite already selected

but the code looks to me like its handling the case where multiple satellites are selected at once. E.g.: 

```
earthly satellite select my-sat-1 my-sat-2
```

Apologies if I'm misunderstood something

cmd/earthly/satellite_cmds.go
brandonSc: That could be a nice solution actually, I feel like the code could just look like this instead:

```golang
	if cliCtx.NArg() != 1 {
		cli.ShowCommandHelpAndExit(cliCtx, app.commandName, 1)
	}
```

Not just here, but in a lot of other commands in earthly where we do similar checks. Probably outside the scope of this PR to fix the other ones though.

cmd/earthly/satellite_cmds.go
r-LaForge: The behaviour of re-selecting a satellite with the same name will remain the same; it's idempotent and the selection will remain the same.

> Select with a satellite already selected

oooooohh sorry if this was confusing. The text output was slightly different depending on if one was selected or not.
If none are selected, it will output `No satellite selected` in the screenshot. If there is one selected it will print out its name `Selected satellite: local-dev` in the screenshot.

cmd/earthly/satellite_cmds.go
r-LaForge: Nice find! I really like that solution.

cmd/earthly/satellite_cmds.go
brandonSc: ah, I get what you mean now 😄  - I totally mis-read that, not sure why I had it in my head that it was for a satellite _with the same name_

cmd/earthly/satellite_cmds.go
r-LaForge: Had to tweak it slightly, but it works nicely!

brandonSc (approved):
🎉

PR #2181, opened 2022-09-13
Support optional +secrets/ prefix

mikejholly:
In a previous change, we added a deprecation warning for the `+secrets/` prefix but didn't actually make a code change to support its optionality. This PR adds support for `NAME=SECRET` syntax without the prefix. This should likely be allowed for all cases as it is not particularly tied to cases that use the `PROJECT` feature.

### DIFF
--- a/earthfile2llb/converter.go
+++ b/earthfile2llb/converter.go
@@ -1826,30 +1826,30 @@ func (c *Converter) secretID(name string) string {
 }
 
 func (c *Converter) parseSecretFlag(secretKeyValue string) (secretID string, envVar string, err error) {
+	if secretKeyValue == "" {
+		// If empty string, don't use (used for optional secrets).
+		// TODO: This should be an actual secret (with an empty value),
+		//       so that the cache works correctly.
+		return "", "", nil
+	}
 	parts := strings.SplitN(secretKeyValue, "=", 2)
 	if len(parts) == 2 {
-		if strings.HasPrefix(parts[1], "+secrets/") {
+		secretID := parts[1]
+		if secretID == "" {
+			// If empty string, don't use (used for optional secrets).
+			// TODO: This should be an actual secret (with an empty value),
+			//       so that the cache works correctly.
+			return "", "", nil
+		}
+		if strings.HasPrefix(secretID, "+secrets/") {
 			c.opt.Console.Printf("Deprecation: the '+secrets/' prefix is not required and support for it will be removed in an upcoming release")
-			secretID := strings.TrimPrefix(parts[1], "+secrets/")
-			return secretID, parts[0], nil
-		} else if parts[1] == "" {
-			// If empty string, don't use (used for optional secrets).
-			// TODO: This should be an actual secret (with an empty value),
-			//       so that the cache works correctly.
-			return "", "", nil
-		} else {
-			return "", "", errors.Errorf("secret definition %s not supported. Format must be either <env-var>=+secrets/<secret-id> or <secret-id>", secretKeyValue)
+			secretID = strings.TrimPrefix(secretID, "+secrets/")
 		}
+		return secretID, parts[0], nil
 	} else if len(parts) == 1 {
-		if secretKeyValue == "" {
-			// If empty string, don't use (used for optional secrets).
-			// TODO: This should be an actual secret (with an empty value),
-			//       so that the cache works correctly.
-			return "", "", nil
-		}
 		return parts[0], parts[0], nil
 	} else {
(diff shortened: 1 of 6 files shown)


alexcb (commented):
Is there a way we can feature flag this? perhaps a `VERSION --optional-secret-prefix`?

Otherwise we could end up with a case where an Earthfile works for one person (using the latest version); where as another person is a few versions behind and gets errors.

However if the `VERSION --optional-secret-prefix` feature is visible on the first line of the Earthfile, the person with the outdated version of earthly will get a much more clear "optiona-secret-prefix feature is unknown" error, which will lead to updating earthly.

mikejholly:
@alexcb that sounds reasonable. My only pushback would be that I think we actually intended to add support for both styles when we added the deprecation warning. As it stands the warning is actually incorrect ("not required").

alexcb:
it seems like we should have initially introduced feature flipping for it, better late than never.

mikejholly:
Do you think we need a feature flag for an "expanded" feature like this that doesn't necessarily impact existing use cases?

earthfile2llb/converter.go
```diff
@@ -1826,30 +1826,30 @@ func (c *Converter) secretID(name string) string {
...
+		if strings.HasPrefix(secretID, "+secrets/") {
+			c.opt.Console.Printf("Deprecation: the '+secrets/' prefix is not required and support for it will be removed in an upcoming release")
+			secretID = strings.TrimPrefix(secretID, "+secrets/")
 		}
+		return secretID, parts[0], nil
+	} else if len(parts) == 1 {
 		return parts[0], parts[0], nil
 	} else {
-		return "", "", errors.Errorf("secret definition %s not supported. Format must be either <env-var>=+secrets/<secret-id> or <secret-id>", secretKeyValue)
+		return "", "", errors.Errorf("secret definition %s not supported. Format must be either <env-var>=+secrets/<secret-id>, <env-var>=<secret-id>, or <secret-id>", secretKeyValue)
```
vladaionescu, line 1852: I would change this error to only talk about the non-deprecated versions.

earthfile2llb/converter.go
```diff
@@ -1826,30 +1826,30 @@ func (c *Converter) secretID(name string) string {
...
-		if secretKeyValue == "" {
-			// If empty string, don't use (used for optional secrets).
-			// TODO: This should be an actual secret (with an empty value),
-			//       so that the cache works correctly.
-			return "", "", nil
+		if strings.HasPrefix(secretID, "+secrets/") {
+			c.opt.Console.Printf("Deprecation: the '+secrets/' prefix is not required and support for it will be removed in an upcoming release")
+			secretID = strings.TrimPrefix(secretID, "+secrets/")
 		}
+		return secretID, parts[0], nil
```
vladaionescu, line 1848: The fact that `+secrets/` is no longer required is a language change. We should put it behind a feature flag so that it goes out with a new `VERSION` number only. To keep things simple we could put it in the same feature flag that enables the `PROJECT` keyword.

In fact, perhaps the deprecation should only show up when the feature flag is set.

vladaionescu (commented):
I think we need a feature flag for this... see thoughts below.

earthfile2llb/converter.go
```diff
@@ -1826,31 +1826,41 @@ func (c *Converter) secretID(name string) string {
...
-		return "", "", errors.Errorf("secret definition %s not supported. Format must be either <env-var>=+secrets/<secret-id> or <secret-id>", secretKeyValue)
+		return secretID, parts[0], nil
+	}
+
+	if strings.HasPrefix(secretID, "+secrets/") {
+		secretID = strings.TrimPrefix(secretID, "+secrets/")
+		return secretID, parts[0], nil
 	}
+
+	return "", "", errors.Errorf("secret definition %s not supported. Format must be either <env-var>=+secrets/<secret-id> or <secret-id>", secretKeyValue)
```
mikejholly, line 1863: Can't be reached if `--use-project-secrets` is set since that's more permissive.

vladaionescu (approved):
Looks good!

earthfile2llb/converter.go
alexcb: should this be changed to:

```suggestion
	return "", "", errors.Errorf("secret definition %s not supported. Format must be either <env-var>=+secrets/<secret-id>", secretKeyValue)
```

mikejholly: Are you sure? That's the error that's been there for a while. And `--secret SECRET_NAME` is still supported.

alexcb: oh sorry, I misread that as `<env-var>=<secret-id>`. Your right.

alexcb (approved):

PR #2184, opened 2022-09-13
PIPELINE and TRIGGER commands

mikejholly:
This PR adds support for the first iteration of the `PIPELINE` command.

* `PIPELINE` must be the first statement in a pipeline target
* A pipeline target can include multiple `TRIGGER` commands
* Valid trigger types are `push`, `pr`, and `manual`
* Pipeline targets will usually be run in the Cloud CI and the commands themselves aren't executed


ast/parser/EarthLexer.g4
```diff
@@ -131,6 +135,8 @@ COMMAND_B: COMMAND -> type(COMMAND), pushMode(COMMAND_ARGS);
 IMPORT_B: IMPORT -> type(IMPORT), pushMode(COMMAND_ARGS);
 CACHE_B: CACHE -> type(CACHE), pushMode(COMMAND_ARGS);
 HOST_B: HOST -> type(HOST), pushMode(COMMAND_ARGS);
+PIPELINE_B: PIPELINE -> type(PIPELINE), pushMode(COMMAND_ARGS);
+TRIGGER_B: TRIGGER -> type(TRIGGER), pushMode(COMMAND_ARGS);
```
vladaionescu, line 139: I think that we don't need these in the `BLOCK` mode. The `BLOCK` mode is only used for the bodies of `IF`, `FOR`, `WAIT` and `WITH DOCKER`.

Then again, it might help with providing a better error message if this check is done in the interpreter instead. Just thinking out loud.

earthfile2llb/converter.go
```diff
@@ -2114,20 +2128,32 @@ func (c *Converter) joinRefs(relRef domain.Reference) (domain.Reference, error)
...
 	}
 
-	switch command {
-	case fromCmd, fromDockerfileCmd, locallyCmd, buildCmd, argCmd, importCmd:
-		return nil
-	default:
-		return errors.New("the first command has to be FROM, FROM DOCKERFILE, LOCALLY, ARG, BUILD or IMPORT")
+	if !c.mts.Final.RanFromLike {
+		switch command {
+		case fromCmd, fromDockerfileCmd, locallyCmd, buildCmd, argCmd, importCmd, projectCmd:
```
vladaionescu, line 2149: I would make `PIPELINE` a from-like command and thus add it to this list? Nothing seems to break if this is not classified as such so far (since ARG and BUILD are allowed without a FROM too), but it feels like the logical thing to do in case of possible future expansion.

earthfile2llb/interpreter.go
```diff
@@ -76,12 +76,31 @@ func (i *Interpreter) Run(ctx context.Context, ef spec.Earthfile) (err error) {
...
+func (i *Interpreter) isPipelineTarget(ctx context.Context, t spec.Target) bool {
+	if len(t.Recipe) == 0 {
+		return false
+	}
+
+	for _, stmt := range t.Recipe {
+		if stmt.Command != nil && stmt.Command.Name == "PIPELINE" {
+			return true
+		}
+	}
```
vladaionescu, line 88: Should we just look at the first command only? I think the intention is for `PIPELINE` to forever remain the first command in the recipe.

earthfile2llb/interpreter.go
```diff
@@ -1631,6 +1652,75 @@ func (i *Interpreter) handleProject(ctx context.Context, cmd spec.Command) error
...
+
+	if len(cmd.Args) > 1 {
+		return i.errorf(cmd.SourceLocation, "invalid number of PIPELINE arguments")
+	}
+
+	for _, arg := range cmd.Args {
+		if arg != "--push" {
+			return i.errorf(cmd.SourceLocation, "--push is the only supported PIPELINE argument")
+		}
+	}
```
vladaionescu, line 1700: We should use arg parsing, via `parseArgs` the way we do it for the other commands, even if we only support one arg. There are some benefits, like code consistency and pretty error output. See https://github.com/earthly/earthly/blob/main/earthfile2llb/flags.go

earthfile2llb/interpreter.go
```diff
@@ -1631,6 +1652,75 @@ func (i *Interpreter) handleProject(ctx context.Context, cmd spec.Command) error
...
+}
+
+func (i *Interpreter) handleTrigger(ctx context.Context, cmd spec.Command) error {
+
+	if len(cmd.Args) < 1 {
+		return i.errorf(cmd.SourceLocation, "TRIGGER requires at least 1 argument")
+	}
+
+	switch cmd.Args[0] {
+	case "manual":
```
vladaionescu, line 1712: Check len args for manual too?

vladaionescu (commented):
Looks good for the most part!

Some small remarks below...

earthfile2llb/interpreter.go
mikejholly: I thought this was a good way to trigger the pipeline code path and later validate the order. That way, the validation code lives in one place.

earthfile2llb/converter.go
mikejholly: Sure, makes sense.

earthfile2llb/interpreter.go
mikejholly: Cool. Sounds good.

earthfile2llb/interpreter.go
vladaionescu: True - that works I guess. The if statement before the `for` seems unnecessary then?

ast/parser/EarthLexer.g4
mikejholly: Sounds reasonable to leave it out for now.

earthfile2llb/converter.go
```diff
@@ -1385,15 +1388,26 @@ func (c *Converter) Host(ctx context.Context, hostname string, ip net.IP) error
...
 	return nil
 }
 
+// Pipeline handles a "PIPELINE" command.
+func (c *Converter) Pipeline(ctx context.Context) error {
+	err := c.checkAllowed(pipelineCmd)
+	if err != nil {
+		return err
+	}
+	c.isPipeline = true
```
vladaionescu, line 1407:
```suggestion
	c.isPipeline = true
	c.mts.Final.RanFromLike = true
```

vladaionescu (approved):
One more nit

PR #2185, opened 2022-09-14
TRY/CATCH: support for saving files

alexcb:
This implements a partial-solution for saving files after a RUN command
failes:

    TRY
        RUN ./test
    FINALLY
        SAVE ARTIFACT junit.xml AS LOCAL junit.xml
    END

Note that this implementation only supports a single RUN command inside a TRY,
and only SAVE ARTIFACT commands within the FINALLY.

This is not a final solution: buildkit will likely need to be modified
to remove the limitations of this implementation.


alexcb:
This has been expanded to support `FINALLY`.

alexcb:
This has been changed to allow:

```
TRY
   ...
FINALLY
   ...
END
```

using `CATCH` now causes an error -- we'll re-evaluate supporting `CATCH` once we figure out how to allow re-throwing or swallowing errors.

cmd/debugger/main.go
```diff
@@ -84,6 +84,45 @@ func populateShellHistory(cmd string) error {
...
+	}
+
+	err = common.WriteDataPacket(conn, len(b), b)
+	if err != nil {
+		return err
+	}
+
+	// send end of file packet
+	return common.WriteDataPacket(conn, 0x00, nil)
+}
```
vladaionescu, line 124: Wondering if we should do anything special about backwards compatibility. If we're working against a satellite that is on an older version, the debugger will be of that older version. Maybe we don't have to do anything though, since this is experimental for now.

cmd/earthly/build_cmd.go
```diff
@@ -472,3 +460,52 @@ func (app *earthlyApp) actionBuildImp(cliCtx *cli.Context, flagArgs, nonFlagArgs
...
+func getTryCatchSaveFileHandler(localArtifactWhiteList *gatewaycrafter.LocalArtifactWhiteList) func(ctx context.Context, conn io.ReadWriteCloser) error {
+	return func(ctx context.Context, conn io.ReadWriteCloser) error {
+		// version
+		n, _, err := debuggercommon.ReadDataPacket(conn)
+		if err != nil {
+			return err
+		}
+		if n != 1 {
+			return fmt.Errorf("unexpected version %d", n)
+		}
```
vladaionescu, line 473: This is nice

debugger/common/settings.go
```diff
@@ -6,10 +6,22 @@ const DebuggerSettingsSecretsKey = "earthly_debugger_settings"
 // DebuggerDefaultSocketPath is the default socket to connect to (path is inside the container)
 const DebuggerDefaultSocketPath = "/var/run/earthly_interactive"
 
+// DefaultSaveFileSocketPath is the default socket to connect to when sending back files (path is inside the container)
+const DefaultSaveFileSocketPath = "/var/run/earthly_save"
```
vladaionescu, line 10: Nit: use this syntax perhpas?

```
const (
  ...
)
```

vladaionescu (approved):
Very minor remarks - looks good as far as I can tell!

cmd/debugger/main.go
alexcb: so far we only have a single version:

```
	// send a protocol version
	err = common.WriteDataPacket(conn, 0x01, nil)
	if err != nil {
		return err
	}
```

Then once we expand the version, the client would populate a prioritised list of accepted protocols in `DebuggerSettings` (defined under https://github.com/earthly/earthly/blob/main/debugger/common/settings.go#L10 )

PR #2186, opened 2022-09-14
bugfix: interactive debugger upgrade warning

alexcb:
A warning that a newer version of buildkit was required was displayed when earthly was run in non-interactive mode.

This additionally removes the initial note stating that interactive mode is not supported by satellites (and will only display a warning when the `--interactive` flag is set and the user is using an outdated version of buildkit)


r-LaForge (approved):

cmd/earthly/buildkit.go
```diff
@@ -139,8 +139,7 @@ func (app *earthlyApp) configureSatellite(cliCtx *cli.Context, cloudClient cloud
 		return nil
 	}
 
-	// When using a satellite, interactive and local do not work; as they are not SSL nor routable yet.
-	app.console.Warnf("Note: the interactive debugger, interactive RUN commands, and inline caching do not yet work on Earthly Satellites.")
+	app.console.Warnf("Note: inline caching do not yet work on Earthly Satellites.")
```
brandonSc, line 142:
```suggestion
	app.console.Warnf("Note: inline caching does not yet work on Earthly Satellites.")
```

brandonSc (approved):
Output looks good now on Satellites 👍

PR #2187, opened 2022-09-15
Use gRPC Client Instead of JSON

brandonSc:
I was running into some issues with the JSON-to-protobuf unmarshaller we've been using. In particular with enums, where existing enums can not be added to on the server without breaking old clients versions - i.e. the JSON unmarshaller throws an error if it encounters an enum value it doesn't recognize. Traditional gRPC does not have this limitation.

Unsurprisingly, the code is also much simpler now in gRPC than it was before using REST calls.

Note: this switches the Pipelines endpoints only, as other endpoints don't support gRPC yet. In due time, we can convert the other endpoints too (after we've converted them on the server first).

### DIFF
--- a/cloud/client.go
+++ b/cloud/client.go
@@ -27,6 +27,7 @@ const (
 	satelliteMgmtTimeout = "5M" // 5 minute timeout when launching or deleting a Satellite
 )
 
+// Client contains gRPC and REST endpoints to the Earthly Cloud backend.
 type Client interface {
 	RegisterEmail(ctx context.Context, email string) error
 	CreateAccount(ctx context.Context, email, verificationToken, password, publicKey string, termsConditionsPrivacy bool) error
--- a/cloud/client.go
+++ b/cloud/client.go
@@ -131,7 +131,7 @@ func NewClient(httpAddr, grpcAddr, agentSockPath, authCredsOverride string, warn
 	tlsConfig := credentials.NewTLS(&tls.Config{})
 	ctx := context.Background()
 	retryOpts := []grpc_retry.CallOption{
-		grpc_retry.WithBackoff(grpc_retry.BackoffLinear(100 * time.Millisecond)),
+		grpc_retry.WithBackoff(grpc_retry.BackoffExponential(100 * time.Millisecond)),
 		grpc_retry.WithCodes(codes.Internal, codes.Unavailable),
 	}
 	conn, err := grpc.DialContext(ctx, grpcAddr,
--- a/cloud/client.go
+++ b/cloud/client.go
@@ -1,20 +1,18 @@
 package cloud
 
 import (
-	"bytes"
 	"context"
-	"encoding/json"
-	"fmt"
-	"io"
-	"net/http"
-	"os"
-	"strings"
+	"crypto/tls"
 	"time"
 
+	"github.com/earthly/cloud-api/pipelines"
 	"github.com/golang/protobuf/jsonpb"
-	"github.com/golang/protobuf/proto"
+	grpc_retry "github.com/grpc-ecosystem/go-grpc-middleware/retry"
 	"github.com/pkg/errors"
 	"golang.org/x/crypto/ssh/agent"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/credentials"
 )
 
 var (
@@ -29,7 +27,6 @@ const (
 	satelliteMgmtTimeout = "5M" // 5 minute timeout when launching or deleting a Satellite
 )
 
-// Client provides a client to the shared secrets service
 type Client interface {
 	RegisterEmail(ctx context.Context, email string) error
 	CreateAccount(ctx context.Context, email, verificationToken, password, publicKey string, termsConditionsPrivacy bool) error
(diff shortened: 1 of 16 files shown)


cloud/http.go
```diff
@@ -0,0 +1,239 @@
+package cloud
```
brandonSc, line 1: Not new code in this file, just move here from `client.go`.

cloud/grpc.go
```diff
@@ -0,0 +1,79 @@
...
+				if err != nil {
+					return nil, errors.Wrap(err, "failed re-authenticating")
+				}
+				md, ok := metadata.FromOutgoingContext(ctx)
+				if !ok {
+					return nil, errors.New("could not parse metadata")
+				}
+				md.Delete("authorization")
+				ctx = metadata.NewOutgoingContext(ctx, md)
+				// TODO not sure if newStreamer(...) should be called here instead
```
brandonSc, line 72: I'm going to be adding the first streaming endpoint in my next PR for Satellite Reserve/Wake and intend on testing this properly then.

cmd/earthly/flags.go
```diff
@@ -75,7 +76,15 @@ func (app *earthlyApp) rootFlags() []cli.Flag {
 			Value:       "https://api.earthly.dev",
 			EnvVars:     []string{"EARTHLY_SERVER_ADDRESS"},
 			Usage:       "API server override for dev purposes",
-			Destination: &app.apiServer,
+			Destination: &app.cloudHTTPAddr,
+			Hidden:      true, // Internal.
+		},
+		&cli.StringFlag{
+			Name:        "grpc",
+			Value:       "ci.earthly.dev:443",
```
vladaionescu, line 84: Could we make this the same format as the address above? I know it's internal, but it would help with our own internal consistency.

`https://ci.earthly.dev` (the port would be implied by `https`)

brandonSc: I don't think gRPC likes the protocol being set. It likes the format to be `host:port`. I get an error like this if I try to set it to `https://ci.earthly.dev`. 
```
transport: Error while dialing dial tcp: lookup tcp///ci.earthly.dev: nodename nor servname provided, or not known
```

Otherwise, I would have just re-used the existing server address we have for the other HTTP calls 😞

vladaionescu: Yeah, what I meant was for us to take the `https://ci.earthly.dev` address and turn it into `ci.earthly.dev:443` automatically underneath before passing it to gRPC.

cloud/client.go
```diff
@@ -339,17 +128,19 @@ func NewClient(host, agentSockPath, authCredsOverride string, warnFunc func(stri
...
-	msg := struct {
-		Message string `json:"message"`
-	}{}
-	err := decoder.Decode(&msg)
+	tlsConfig := credentials.NewTLS(&tls.Config{})
+	ctx := context.Background()
+	retryOpts := []grpc_retry.CallOption{
+		grpc_retry.WithBackoff(grpc_retry.BackoffLinear(100 * time.Millisecond)),
+		grpc_retry.WithCodes(codes.Internal, codes.Unavailable),
+	}
```
r-LaForge, line 136: I'm a little worried about these retry options being built into every call.
Unless we can ensure that every call will always be idempotent.
So I think we should either:

 - Leave this off the default dial options and expose these as defaults we purposefully use per RPC. I like this one because it makes it very obvious how the call will behave.
 - Do something struct embedding and implement specific calls on each struct depending on idempotency.

cloud/http.go
r-LaForge: ty

cmd/earthly/flags.go
brandonSc: Right - I actually had it doing something similar at first but then changed it. 

I imagined devs pointing earthly to their remote or local dev environment, and that they might get blocked if the parsing rules I wrote weren't flexible enough for their use case. Leaving the full address configurable is less code and probably more convenient.  Plus eventually we'll prob drop the HTTP address and only have a single one for gRPC anyways.

cloud/client.go
brandonSc: To be honest, I was just trying to roughly replicate the retry logic that was already here when these were HTTP calls. 

Ref: https://github.com/earthly/earthly/blob/3111d90507ab09de8aded484300ea3d7e207739f/cloud/client.go#L162

I assumed the reason these retries were coded in the first place was just in case there was some transient error, and we didn't want to bother the user with it if a simple retry would solve it.

As for idempotency - I don't think we need to worry too much about it. I don't think it's a big deal if these operations are re-run on a failure, and in fact the user would probably just re-run it themselves anyway. If we have an operation that is risky if it's not idempotent (e.g. charging a credit card) we should make sure its safe on the backend (we have the tools necessary with DB rollbacks, request-IDs, etc)

In general, I'd sooner just get rid of this retry logic than let it get too complicated. think it might be nice to have, in the odd time that it actually kicks in and works, but the old custom retry code (at least in my experience) was a bit difficult to maintain and a good source for coding mistakes.

cloud/grpc.go
```diff
@@ -0,0 +1,77 @@
...
+		if time.Now().UTC().After(c.authTokenExpiry) {
+			err := c.Authenticate(ctx)
+			if err != nil {
+				return errors.Wrap(err, "failed refreshing expired token")
+			}
+		}
+		err := invoker(ctx, method, req, reply, cc, opts...)
+		if err != nil {
+			s, ok := status.FromError(err)
+			if ok && s.Code() == codes.Unauthenticated {
```
mikejholly, line 30: Is it possible to bake some of this auth retry logic into the gRPC retrier code?

mikejholly (approved):
Nice work!

cloud/client.go
r-LaForge: Mike, Corey, Brandon, and I had a discussion around this in a Zoom meeting and we came to the  following conclusion:
 - The old retry logic was complex and we're glad to replace it.
 - Risk vs Reward: we believe it's more important to have retries and for the client to always handle that by default rather than adding call options as the third argument to every gRPC call.  i.e.,
 ```
 resp, err := c.pipelines.ListSatellites(c.withAuth(ctx), &pb.ListSatellitesRequest{
		OrgId: orgID,
	}, ...defaultRetryOptions())
```
 - If we have an operation we are worried about in the future we will make an exception then.
 - We will aim to try our best to handle these scenarios on the server side through rate limiting and idempotency.

r-LaForge (approved):
Nice work!
Glad to see us shifting over :)

cloud/grpc.go
brandonSc: I didn't see a clean way. Would be cool if the library supported hooking-in custom functions to run before retrying (and also modify the retry limits from within that func).

PR #2190, opened 2022-09-15
Make ./ast a submodule

mikejholly:


Earthfile
```diff
@@ -26,6 +26,7 @@ deps:
     RUN go install golang.org/x/lint/golint@latest
     RUN go install github.com/gordonklaus/ineffassign@latest
     COPY go.mod go.sum ./
+    COPY --dir ast ./
```
vladaionescu, line 29: Could this just copy the skeleton of `ast` at this stage (e.g. dir containing go.mod and go.sum only) and only later download the rest of the Go files?

Otherwise, the cache reuse experience would be really bad for changes to `ast`.

alexcb (approved):

Issue #1895, opened 2022-05-24
Ability to run Earthly on an M1 computer that uses a remote x86 Docker


vladaionescu:
### Setup

* User has an M1 laptop
* User uses a remote Docker daemon, which is running on an x86

### Problem

User wants to run Earthly locally, but as Earthly attempts to start the buildkit daemon in Docker, it pulls the arm64 version of the image, instead of the x86 version of it. This causes a crash as buildkit cannot be emulated.

### What should happen instead

* Earthly should detect that the Docker daemon is running on x86 (which may be different than the current host's architecture) and thus use an x86 version of the Earthly image.
* Earthly builds run on this kind of setup should default to use platform x86, when no platform is specified (it's possible that the new platform logic behind a feature flag already does this)

vladaionescu:
Possible workaround

* Start an x86 Buildkit daemon via Docker using the instructions here: https://docs.earthly.dev/ci-integration/remote-buildkit
* Point Earthly to that daemon by setting `EARTHLY_BUILDKIT_HOST` env var

dchw:
We actually already expose this information, too! Its in this structure:

https://github.com/earthly/earthly/blob/b0e5a0f191b765a79f60d9b2d5c20652109343d9/util/containerutil/types.go#L53-L62

And you get it (regardless of frontend) with this method on a given frontend:

https://github.com/earthly/earthly/blob/b0e5a0f191b765a79f60d9b2d5c20652109343d9/util/containerutil/frontend.go#L21

Implementations:
* Docker: https://github.com/earthly/earthly/blob/b0e5a0f191b765a79f60d9b2d5c20652109343d9/util/containerutil/docker.go#L77-L112
* Podman: https://github.com/earthly/earthly/blob/b0e5a0f191b765a79f60d9b2d5c20652109343d9/util/containerutil/podman.go#L78-L131

vladaionescu:
Oh great - that's really useful! I'll play with this this weekend.

vladaionescu:
Note that interactive mode might not work immediately in this setup and more tinkering may be necessary. Also, we should test that local registry-based outputs still work fine in this case.

r-LaForge:
Digging into the problem more I think @alexcb is correct.
We are using an[ incorrect assumption](https://github.com/earthly/earthly/blob/90d79633da0859cae5ee9a2170b93a06ce4fc2d6/util/containerutil/shell_shared.go#L302) when running the buildkit image.

[...]

PR #2191, opened 2022-09-16
Fix remote docker host running x86 platform. Fixes https://github.com/earthly/earthly/issues/1895

r-LaForge:
fixes https://github.com/earthly/earthly/issues/1895

Running Earthly from an M1 mac to a remote docker host running x86 was broken as described in the issue.
We need to ensure an appropriate image is pulled that is. compatible with the architecture the docker host itself.

The solution is to use the frontend information we collect via the `Information` call to determine if the current client's platform also works on the server. 

If it is supported, this acts as a sort of preference to the client's platform.
If not, it will default to the server platform to allow the buildkitd container to run.

@alexcb @vladaionescu

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ All notable changes to [Earthly](https://github.com/earthly/earthly) will be doc
   fallback to `/usr/share/zsh/site-functions`.
 
 ### Fixed
-Fixed Earthly failing when using a remote docker host from a machine with an incompatible architecture. Fixes (#1895)[https://github.com/earthly/earthly/issues/1895]
+- Fixed Earthly failing when using a remote docker host from a machine with an incompatible architecture. Fixes (#1895)[https://github.com/earthly/earthly/issues/1895]
 
 ## v0.6.23 - 2022-09-06
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,8 @@ All notable changes to [Earthly](https://github.com/earthly/earthly) will be doc
   fallback to `/usr/share/zsh/site-functions`.
 
 ### Fixed
-- Fixed Earthly failing when using a remote docker host from a machine with an incompatible architecture. Fixes (#1895)[https://github.com/earthly/earthly/issues/1895]
+
+- Fixed Earthly failing when using a remote docker host from a machine with an incompatible architecture. [#1895](https://github.com/earthly/earthly/issues/1895)
 
 ## v0.6.23 - 2022-09-06
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ All notable changes to [Earthly](https://github.com/earthly/earthly) will be doc
 - Bootstraping zsh autocompletion will first attempt to install under `/usr/local/share/zsh/site-functions`, and will now
   fallback to `/usr/share/zsh/site-functions`.
 
+### Fixed
+Fixed Earthly failing when using a remote docker host from a machine with an incompatible architecture. Fixes (#1895)[https://github.com/earthly/earthly/issues/1895]
+
 ## v0.6.23 - 2022-09-06
 
 ### Fixed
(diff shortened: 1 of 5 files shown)


util/containerutil/docker.go
```diff
@@ -31,6 +31,8 @@ func NewDockerShellFrontend(ctx context.Context, cfg *FrontendConfig) (Container
 			globalCompatibilityArgs: make([]string, 0),
 		},
 	}
+	// TODO: Find a cleaner way to pass down information to the shellFrontend
+	fe.FrontendInformation = fe.Information
```
r-LaForge, line 35: I want to revisit this in the work I will do soon on Podman.

util/containerutil/podman.go
```diff
@@ -29,6 +29,8 @@ func NewPodmanShellFrontend(ctx context.Context, cfg *FrontendConfig) (Container
 			globalCompatibilityArgs: make([]string, 0),
 		},
 	}
+	// TODO: Find a cleaner way to pass down information to the shellFrontend
+	fe.FrontendInformation = fe.Information
```
r-LaForge, line 33: I want to revisit this in the work I will do soon on Podman.

util/containerutil/shell_shared.go
```diff
@@ -291,12 +297,12 @@ func (sf *shellFrontend) commandContextOutput(ctx context.Context, args ...strin
...
-	// an actual image which may require downloading.
-	output, _ := sf.commandContextOutput(ctx, "run", "--rm", "--platform", getPlatform(), "scratch")
-	return strings.Contains(output.string(), "Unable to find image")
+func (sf *shellFrontend) supportsPlatform(ctx context.Context, platform string) (bool, error) {
+	frontendInfo, err := sf.FrontendInformation(ctx)
+	if err != nil {
+		return false, errors.Wrapf(err, "failed to get platform information")
+	}
+	return frontendInfo.ServerPlatform == platform, nil
 }
```
r-LaForge, line 306: This function assumed if the error message `Unable to find image` was output it meant the platform was supported. That does not seem to be the case; At least, not anymore.

util/containerutil/shell_shared.go
```diff
@@ -291,12 +297,12 @@ func (sf *shellFrontend) commandContextOutput(ctx context.Context, args ...strin
...
-	// --platform is supported or not. This is faster than attempting to run
-	// an actual image which may require downloading.
-	output, _ := sf.commandContextOutput(ctx, "run", "--rm", "--platform", getPlatform(), "scratch")
-	return strings.Contains(output.string(), "Unable to find image")
+func (sf *shellFrontend) supportsPlatform(ctx context.Context, platform string) (bool, error) {
+	frontendInfo, err := sf.FrontendInformation(ctx)
+	if err != nil {
+		return false, errors.Wrapf(err, "failed to get platform information")
+	}
+	return frontendInfo.ServerPlatform == platform, nil
```
vladaionescu, line 305: I wonder if these should be normalized when they are compared.

You could use `Normalize` from `github.com/containerd/containerd/platforms`. See how it's used here: https://github.com/earthly/earthly/blob/main/util/platutil/platform.go#L32. You can use `Parse` from the same package to turn the string into the platform object. Then normalize, then you could `Format` to turn back into strings and compare the normalized strings.

vladaionescu (commented):
Looks good for the most part!

Not sure how to better pass that information object either. Left a small remark.

vladaionescu:
Oh - can you also add an entry in `CHANGELOG.md`? That will be used for release description.

util/containerutil/shell_shared.go
r-LaForge: done

CHANGELOG.md
```diff
@@ -13,6 +13,9 @@ All notable changes to [Earthly](https://github.com/earthly/earthly) will be doc
 - Bootstraping zsh autocompletion will first attempt to install under `/usr/local/share/zsh/site-functions`, and will now
   fallback to `/usr/share/zsh/site-functions`.
 
+### Fixed
+- Fixed Earthly failing when using a remote docker host from a machine with an incompatible architecture. Fixes (#1895)[https://github.com/earthly/earthly/issues/1895]
```
vladaionescu, line 17: Fix link + consistency.

```suggestion
### Fixed

- Fixed Earthly failing when using a remote docker host from a machine with an incompatible architecture. [#1895](https://github.com/earthly/earthly/issues/1895)
```

vladaionescu (approved):

PR #2200, opened 2022-09-20
Move new org invite and members commands out of preview

mikejholly:
This PR:

* Removes the older `invite` command 
* Moves the new `invite` and `members` subcommands out of `preview`

### DIFF
--- a/cmd/earthly/org_cmds.go
+++ b/cmd/earthly/org_cmds.go
@@ -118,10 +118,6 @@ func (app *earthlyApp) orgCmds() []*cli.Command {
 	}
 }
 
-func (app *earthlyApp) orgCmdsPreview() []*cli.Command {
-	return []*cli.Command{}
-}
-
 func (app *earthlyApp) actionOrgCreate(cliCtx *cli.Context) error {
 	app.commandName = "orgCreate"
 	if cliCtx.NArg() != 1 {
--- a/cmd/earthly/org_cmds.go
+++ b/cmd/earthly/org_cmds.go
@@ -34,66 +34,10 @@ func (app *earthlyApp) orgCmds() []*cli.Command {
 			Action:    app.actionOrgListPermissions,
 		},
 		{
-			Name:      "invite",
-			Usage:     "Invite accounts to your organization",
-			UsageText: "earthly [options] org invite [options] <path> <email> [<email> ...]",
-			Action:    app.actionOrgInvite,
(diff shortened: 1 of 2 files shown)


cmd/earthly/org_cmds.go
```diff
@@ -133,9 +77,51 @@ func (app *earthlyApp) orgCmdsPreview() []*cli.Command {
...
+					Description: "Remove a user from the organization",
+					UsageText:   "earthly org [--org organization] members rm <user-email>",
+					Action:      app.actionOrgMemberRemove,
+				},
+			},
+		},
 	}
 }
 
+func (app *earthlyApp) orgCmdsPreview() []*cli.Command {
```
mikejholly, line 121: Shall I also remove this? I figured we may reuse it at some point in the near future.

vladaionescu: We can always add it back? No strong opinion.

vladaionescu (approved):

cmd/earthly/org_cmds.go
alexcb: it seems trivial enought to add back in if we need it down the road. I would remove it, but it's not a strong opinion.

alexcb (approved):

cmd/earthly/org_cmds.go
```diff
@@ -33,68 +33,12 @@ func (app *earthlyApp) orgCmds() []*cli.Command {
 			UsageText: "earthly [options] org list-permissions <org-name>",
 			Action:    app.actionOrgListPermissions,
 		},
-		{
-			Name:      "invite",
```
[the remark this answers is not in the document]
mikejholly: Perhaps we should remove it once the new secrets commands are in place. `list-permissions` lists "secret path" permissions that may still be useful to some users.

cmd/earthly/org_cmds.go
mikejholly: Cool. I'll remove.

vladaionescu: You know what - I think it's better to keep the `earthly preview invite` command, but just return an error explaining that this has graduated out of preview. It'll help prevent questions from our users.

mikejholly: Sure, that's fine.

mikejholly:
Updated with handy message.
image

PR #2201, opened 2022-09-20
Use New Satellite Reserve Endpoint

brandonSc:
The new [satellite reserve endpoint](https://github.com/earthly/cloud/pull/344) wakes up a satellite if it's asleep, and calls buildkit Reserve().  It's a streaming RPC call, which returns periodic updates until the instance is awake.

Screen Shot 2022-09-21 at 12 16 30 PM

### DIFF
--- a/cloud/satellite.go
+++ b/cloud/satellite.go
@@ -10,14 +10,22 @@ import (
 )
 
 const (
+	// SatelliteStatusOperational indicates an on satellite that is ready to accept connections.
 	SatelliteStatusOperational = "Operational"
-	SatelliteStatusSleep       = "Sleeping"
-	SatelliteStatusStarting    = "Starting"
-	SatelliteStatusCreating    = "Creating"
-	SatelliteStatusFailed      = "Failed"
-	SatelliteStatusDestroying  = "Destroying"
-	SatelliteStatusOffline     = "Offline"
-	SatelliteStatusUnknown     = "Unknown"
+	// SatelliteStatusSleep indicates a satellite that is in a sleep state.
+	SatelliteStatusSleep = "Sleeping"
+	// SatelliteStatusStarting indicates a satellite that is waking from a sleep state.
+	SatelliteStatusStarting = "Starting"
+	// SatelliteStatusCreating indicates a new satellite that is currently being launched.
+	SatelliteStatusCreating = "Creating"
+	// SatelliteStatusFailed indicates a satellite that has crashed and cannot be used.
+	SatelliteStatusFailed = "Failed"
+	// SatelliteStatusDestroying indicates a satellite that is actively being deleted.
+	SatelliteStatusDestroying = "Destroying"
+	// SatelliteStatusOffline indicates a satellite that has been stopped and will not be woken up normally via build.
+	SatelliteStatusOffline = "Offline"
+	// SatelliteStatusUnknown is used when an unexpected satellite status is returned by the server.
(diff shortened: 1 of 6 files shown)


brandonSc:
Satellite tests are failing right now, since the endpoint its using isn't deployed to production

cloud/satellite.go
```diff
@@ -74,21 +95,47 @@ func (c *client) LaunchSatellite(ctx context.Context, name, orgID string, featur
...
+		Name:  name,
+	})
+	if err != nil {
+		return errors.Wrap(err, "failed opening satellite reserve stream")
+	}
+	var update *pb.ReserveSatelliteResponse
+	for {
+		update, err = stream.Recv()
+		if err == io.EOF {
+			return nil
```
alexcb, line 113: if update was ever set to `SatelliteStatusFailed`, it seems like the reserve call failed, and returning `nil` as the error could be confusing.

Can we do an explicit check for this sending it to the `out` channel, and return an error instead?

cmd/earthly/buildkit.go
```diff
@@ -178,6 +179,28 @@ func (app *earthlyApp) configureSatellite(cliCtx *cli.Context, cloudClient cloud
...
+	console := app.console.WithPrefix("satellite")
+	out := make(chan string)
+	var reserveErr error
+	go func() { reserveErr = cloudClient.ReserveSatellite(cliCtx.Context, app.satelliteName, orgID, out) }()
+	for status := range out {
+		switch status {
+		case cloud.SatelliteStatusSleep:
+			console.Printf("%s is waking up. Please wait...", app.satelliteName)
+		case cloud.SatelliteStatusStarting:
+			console.VerbosePrintf("...Still waking up...")
```
alexcb, line 193: very cool!

Maybe we can be silly and do something like:

```
wakeupText := []string{
    "tracking orbit",
    "adjusting course",
    "modifying stasis chamber temperature",
    ...
   }
console.VerbosePrintf("...%s...", wakeupText[rand.Intn(len(wakeupText))])
```

brandonSc: Hehe, I love that actually. Especially for when we port this kind of API to the Launch and Delete endpoints

cloud/satellite.go
brandonSc: The failure status was added initially when we started satellites, but it was never used anywhere. I'm torn between marking it as deprecated and leaving it as supported in case there's a use for it eventually (e.g. perhaps we'll implement health checks and then can show this state if that fails)

Nonetheless, I don't think there's any harm in adding that check here as precaution in case the API starts returning that status here one day. I'll add it in.

go.mod
```diff
@@ -11,7 +11,7 @@ require (
 	github.com/docker/distribution v2.8.1+incompatible
 	github.com/docker/go-connections v0.4.0
 	github.com/dustin/go-humanize v1.0.0
-	github.com/earthly/cloud-api v1.0.1-0.20220913124428-8736f7614ec8
```
brandonSc, line 14: Need to update this to released version once the cloud PR is merged

cmd/earthly/buildkit.go
vladaionescu: LoL hilarious :D love it

alexcb: amazing list of statuses :clap:

vladaionescu (approved):

Issue #2202, opened 2022-09-20
Podman - CI tests


r-LaForge:
# Add CI tests for Podman
child of #760 

The goal in this issue will be to enable testing of earthly in this repository via Podman. 
This ensures we have test coverage of Podman and our own working example of running Podman in CI.

## Included
- [ ] ~~Podman runs in CI using most of our tests (except WITH DOCKER - will resolve with #1870).~~
- [x] Ensure the performance of the CI pipeline is comparable to Docker.
- [x] Fixing small issues / creating follow-up issues when problems are discovered.

## Not Included
- [ ] ~Running tests to cover WITH DOCKER usage (#1870).~

**Edit Sept 23rd:**
Due to serious performance issues when running the MacOS Github Runner we have opted to do the majority of testing of Podman on Linux with a nightly build that runs some MacOS tests.

The WITH DOCKER issue specifically discusses rootless containers, which is now out of scope of this initial work (since it is a known limitation with Docker itself already).

For using WITH DOCKER with Podman please either
 - use `sudo` on Linux
 - Start a rootful machine on Mac `podman machine stop && podman machine set --rootful=true && podman machine start`

We now also include testing WITH DOCKER using Podman in CI.

r-LaForge:
Initial test of Podman in CI has shown an extremely slow initial build time for earthly of 55 minutes.
https://github.com/earthly/earthly/actions/runs/3093154502/jobs/5005215312

Thanks to Vlad and Alex, this seems to be a problem with lacking privileges when running the `earthly/earthly` image : https://docs.earthly.dev/docs/guides/using-the-earthly-docker-images/all-in-one#requirements

r-LaForge:
The first draft of Podman tests [have been merged](https://github.com/earthly/earthly/issues?q=is%3Aissue+is%3Aopen+podman+author%3Ar-LaForge) and we will continue to add tests for Podman overtime.

PR #2204, opened 2022-09-20
Podman ci tests - #2202

r-LaForge:
# Support Podman tests in CI
related to #2202

This PR does the following: 
- [x] Switch out usage of setup-actions to use the `main` branch to dogfood out latest changes.
- [x] Make our Github checkout actions use a token to drastically reduce rate limiting. ([See Github Docs for more info](https://docs.github.com/en/developers/apps/building-github-apps/rate-limits-for-github-apps)).
- [x]  Add thorough Podman testing on Linux using our existing workflow logic.
- [x] Add basic, nightly, Podman testing on MacOS.
- [x] Refactor our workflow logic into reusable workflows. This reduces duplicate code and allows us to add more configurations more easily (in the near future nerdctl and perhaps Mac tests using Docker). 

The prefix `ci-` denotes workflows that will run as part of our CI.
The prefix `ci-scheduled-` denotes that will run on a schedule.
The prefix `reusable-` denote workflows that will not run by themselves and are instead callable from other workflows.

Quirks / Limitations:
 - Podman cannot run multiple jobs at the same time due to conflicts on ports. To handle this we run the Podman jobs sequentially by using `needs`.
 - We have conditionals around the BINARY used when running the workflows. Ideally, we'd reduce this by encapsulating the logic in an action.


.github/workflows/satellites.yml
```diff
@@ -1,170 +0,0 @@
-name: Test on Satellites
```
r-LaForge, line 1: All of the logic in this file is now in `ci-docker-ubuntu.yml` and `ci-podman-ubuntu.yml`

.github/workflows/ci.yml
```diff
@@ -1,1107 +0,0 @@
-name: GitHub Actions CI
```
r-LaForge, line 1: Everything in this file is now part of `ci-docker-ubuntu.yml`

.github/workflows/ci-scheduled-podman-mac.yml
```diff
@@ -0,0 +1,93 @@
+name: Podman Mac CI
+
+on:
+  schedule:
+    - cron: '0 6 * * * ' # Every day at 6:00AM UTC (~midnight CST pending DST)
```
r-LaForge, line 5: Due to the slowness of running MacOS Podman in the Guthub Runner  we opt to run it once a day on some simple tests as a sanity check.

.github/workflows/ci-docker-ubuntu.yml
```diff
@@ -0,0 +1,569 @@
...
+          set -euo pipefail
+          EARTHLY_VERSION_FLAG_OVERRIDES="$(tr -d '\n' < .earthly_version_flag_overrides)"
+          echo "EARTHLY_VERSION_FLAG_OVERRIDES=$EARTHLY_VERSION_FLAG_OVERRIDES" >> "$GITHUB_ENV"
+      - name: Enable local registry-based exporter
+        run: ./build/linux/amd64/earthly config global.local_registry_host 'tcp://127.0.0.1:8371'
+      - name: Configure Earthly Conversion Parallelism
+        run: ./build/linux/amd64/earthly config global.conversion_parallelism 5
+      - name: Build linux/arm64 +buildkitd
+        run: |-
+          ./build/linux/amd64/earthly --ci  --platform=linux/arm64 ./buildkitd+buildkitd --BUILDKIT_PROJECT="$BUILDKIT_PROJECT"
```
vladaionescu, line 312: Setting the `BUILDKIT_PROJECT` flag here seems unnecessary?

.github/workflows/ci-docker-ubuntu.yml
```diff
@@ -0,0 +1,569 @@
...
+      - name: Configure Earthly Conversion Parallelism
+        run: ./build/linux/amd64/earthly config global.conversion_parallelism 5
+      - name: Build linux/amd64 +buildkitd
+        run: |-
+          ./build/linux/amd64/earthly --ci  --platform=linux/amd64 ./buildkitd+buildkitd --BUILDKIT_PROJECT="$BUILDKIT_PROJECT"
+      - name: Buildkit logs (runs on failure)
+        run: docker logs earthly-buildkitd
+        if: ${{ failure() }}
+
+  buildkitd-arm64:
```
vladaionescu, line 262: Do I understand correctly that the buildkitd build is now split up into different CI tasks?

.github/workflows/ci-podman-ubuntu.yml
```diff
@@ -0,0 +1,118 @@
...
+    with:
+      BUILT_EARTHLY_PATH: "./build/linux/amd64/earthly"
+      BUILD_EARTHLY_TARGET: "+for-linux"
+      RUNS_ON: "ubuntu-latest"
+      BINARY: "podman"
+      SUDO: "sudo"
+    secrets: inherit
+
+  podman-examples-1:
+    # TODO: Figure out how to run multiple Podman instances in parallel with different ports for buildkitd
```
vladaionescu, line 23: I thought that these different CI tasks ran on different, sandboxed machines. No idea why this is happening.

.github/workflows/reusable-misc-tests.yml
```diff
@@ -0,0 +1,196 @@
...
+      SUDO:
+        type: string
+        required: false
+      RUNS_ON:
+        required: true
+        type: string
+
+
+jobs:
+  misc-tests:
```
vladaionescu, line 24: Is this used anywhere? I think this has been forgotten as I don't see it triggered.

scripts/tests/export.sh
```diff
@@ -167,9 +169,9 @@ EOF
 "$earthly" prune --reset
 "$earthly" +multi6
 
-docker run --rm earthly-export-test-6:test cat /data | grep "hello my world"
-docker run --rm earthly-export-test-6:test cat /data | grep "aarch64"
-if docker inspect earthly-export-test-6:test_linux_arm64 >/dev/null 2>&1 ; then
+"$frontend" run --rm earthly-export-test-6:test cat /data | grep "hello my world"
+"$frontend" run --rm earthly-export-test-6:test cat /data | grep "aarch64"
+if "$frontend" inspect earthly-export-test-6:test_linux_arm64 >/dev/null 2>&1 ; then
```
vladaionescu, line 174: The fact that this export test is now carried over too is fantastic!

tests/save-images/test.sh
```diff
@@ -9,102 +9,103 @@ cd "$(dirname "$0")"
...
-docker rmi -f myimage:buildtest
-docker rmi -f earthly-test-saveimage:test
+"$frontend" rmi -f myimage:buildtest
+"$frontend" rmi -f earthly-test-saveimage:test
 
 "$earthly" ./new-behaviour-remote+myimage-buildtest
-docker inspect myimage:buildtest >/dev/null
-docker inspect earthly-test-saveimage:test >/dev/null
+"$frontend" inspect myimage:buildtest >/dev/null
+"$frontend" inspect earthly-test-saveimage:test >/dev/null
```
vladaionescu, line 109: Nice!

vladaionescu (commented):
This looks good in general! I left a few small comments below...

@alexcb is there any test that we had before that this new version is not running by any chance? (I was able to spot `misc-tests` missing, but didn't see anything else)

r-LaForge:
Reminder: Appears to be an issue with the bootstrap step cache from our actions-setup.
https://github.com/earthly/earthly/actions/runs/3115223038/jobs/5051897565

.github/workflows/ci-docker-ubuntu.yml
r-LaForge: I'm not sure - this was copy + pasted from the previous ci.yml file.
We could try running without it.

.github/workflows/ci-podman-ubuntu.yml
r-LaForge: That's precisely what I thought as well.
However, no matter what I tried they would run into port conflicts if run in parallel. They might not be as "sandboxed" as we think.
My hypothesis is our old tests ran fine because they all used the same Docker environment (reused buildkitd) while the Podman tests do not reuse the same Podman environment (so the container check fails).

.github/workflows/reusable-misc-tests.yml
r-LaForge: Good catch! I've added it now. I definitely missed this one.

scripts/tests/export.sh
r-LaForge: Yep :) 

My plan is for nerdctl to follow the same pattern here. It's really nice since both Podman and Nerdctl ensure they are fully compatible with current Docker CLI commands (to the point where many users literally alias `docker=podman`).

.github/workflows/ci-docker-ubuntu.yml
r-LaForge: I carried this change over from the current [ci.yml file ](https://github.com/earthly/earthly/blob/main/.github/workflows/ci.yml#L758)

I'm completely open to either leaving as-is or combining them.

earthly
```diff
@@ -59,7 +60,7 @@ get_latest_binary() {
     fi
 
     $frontend_bin pull --platform="$bk_platform" docker.io/earthly/buildkitd:prerelease >&2
-    $frontend_bin cp earthly_binary:"$earth_bin_path" "$bindir/earthly-prerelease" >&2
+    $frontend_bin cp earthly_"$earth_bin_path" "$bindir/earthly-prerelease" >&2
```
alexcb, line 63: is this intentional?

I just gave it a try on my machine, and it fails:

```
alex@surfperch:~/gh/earthly/earthly$ git rev-parse HEAD
06a942c51174354a1e31318166d1b9c14a4ab006

alex@surfperch:~/gh/earthly/earthly$ ./earthly +for-linux
checking for latest earthly prerelease binaries
prerelease: Pulling from earthly/earthlybinaries
9621f1afde84: Pull complete 
19c959ea7afa: Pull complete 
Digest: sha256:83bcf827df2995a685452e466c32eaa4a2b9e488c4d1e86fa5d618437676b319
Status: Downloaded newer image for earthly/earthlybinaries:prerelease
docker.io/earthly/earthlybinaries:prerelease
8fa3c939585f36dd739fb28c99fe7c40fde5b18fab9f6dc5dae8a01f0c57d647
prerelease: Pulling from earthly/buildkitd
59bf1c3509f3: Pull complete 
8d0d8be1ef72: Pull complete 
25a03d6a6f10: Pull complete 
23b93116472b: Pull complete 
a91fb26f2997: Pull complete 
05b924ea5781: Pull complete 
c884202d65a3: Pull complete 
0c4a5c06b0b1: Pull complete 
62eb7cb9cf37: Pull complete 
aa6f00fd7a83: Pull complete 
5571a0b4724c: Pull complete 
e4febe4ae775: Pull complete 
1e18899cdcd6: Pull complete 
90a220faa27a: Pull complete 
5b4e8cfbec7c: Pull complete 
097bb3a26302: Pull complete 
0b30b908808d: Pull complete 
d7b739135341: Pull complete 
4d6c569fdbd3: Pull complete 
Digest: sha256:ced85008893f7957cde535ffbc0f1fa94e6c3186b3d00e1994b978e25c3ca71e
Status: Downloaded newer image for earthly/buildkitd:prerelease
docker.io/earthly/buildkitd:prerelease
must specify at least one container source
alex@surfperch:~/gh/earthly/earthly$ echo $?
1
```

r-LaForge: This was not intentional.. Not sure how this change slipped in 🤔 
Great catch!

tests/bootstrap/test-bootstrap.sh
```diff
@@ -8,6 +8,9 @@ cd "$(dirname "$0")"
 earthly=${earthly-"../../build/linux/amd64/earthly"}
 earthly=$(realpath "$earthly")
 
+# docker / podman
+frontend=${frontend-"docker"}
```
alexcb, line 12: It would be helpful if `frontend` would default to docker for cases where it's not specified.

some of these tests scripts (such as this one that references `../../build/...`) are designed to be runnable via GHA and locally (at least for linux users). e.g. I can simply run `./tests/bootstrap/test-bootstrap.sh` on my local machine and this script passes.

At the very least we should test that `$frontend` exists and exit with an error message when it's missing. However, 
 I think having a default is more dev-friendly. (maybe when it's missing we could use `which` to default to docker or podman -- whichever is detected)

earthly
r-LaForge: reverted

Earthfile
```diff
@@ -487,22 +490,47 @@ test:
...
+    BUILD ./tests+ga-qemu \
         --DOCKERHUB_AUTH=$DOCKERHUB_AUTH \
         --DOCKERHUB_USER_SECRET=$DOCKERHUB_USER_SECRET \
         --DOCKERHUB_TOKEN_SECRET=$DOCKERHUB_TOKEN_SECRET \
         --DOCKERHUB_MIRROR=$DOCKERHUB_MIRROR \
         --DOCKERHUB_MIRROR_INSECURE=$DOCKERHUB_MIRROR_INSECURE \
         --GLOBAL_WAIT_END="$GLOBAL_WAIT_END"
 
+test:
+    BUILD +test-no-qemu
```
alexcb, line 517: I think we might need to pass along the `--DOCKERHUB_*` vars here too (similar to how `+test-all` does it)

Earthfile
```diff
@@ -487,22 +490,47 @@ test:
...
+    BUILD +test-qemu
```
alexcb, line 518: (same as above)

Earthfile
r-LaForge: done

Earthfile
r-LaForge: done

scripts/tests/auth/test-self-hosted.sh
```diff
@@ -23,6 +23,9 @@ fi
 earthly=${earthly:=earthly}
 earthly=$(realpath "$earthly")
 echo "running tests with $earthly"
+echo "using frontend $frontend"
```
alexcb, line 26: nit: can we move this after the `frontend="${frontend:-$(which docker || which podman)}"`, so we can see the final value (after `which`)

scripts/tests/export.sh
```diff
@@ -1,9 +1,13 @@
 #!/usr/bin/env bash
 set -xeu
 
+# "$frontend" / podman
```
alexcb, line 4: nit: unneeded comment

alexcb (approved):
looks great! just left a couple minor nits.

alexcb:
looks like there might be a test that's failing; maybe some secrets aren't completely getting passed:

```
WARN: (RUN docker login $DOCKERHUB_MIRROR --username="$USERNAME" --***) unauthorized
```

r-LaForge:
@alexcb I've commented out those tests for this PR and will be looking into them as a separate, future, PR.

alexcb (approved):

alexcb (approved):

Issue #2194, opened 2022-09-19
Running multiple jobs in parallel causes race conditions for buildkit startup


vladaionescu:
Earthly automatically starts up the buildkit daemon if it's not already started. But if two jobs attempt to do this at the same time, the following occurs.

```
docker: Error response from daemon: Conflict. The container name "/earthly-buildkitd" is already in use by container "5bd39c324493f9ca2745b846d8827d70410c8ae9f9ec486a86f61f184bcda740". You have to remove (or rename) that container to be able to reuse that name.
```

We need to make this startup resilient to such race conditions. Given that this is the local machine, perhaps a file-based lock is all we need.

kcampion:
I returned this mistake in your Slack. The goal is to launch 2 jobs in parallel under GitLab CI.

I used this .gitlab-ci.yml
```
image: docker:git

services:
  - docker:dind

before_script:
  - apk update
  - wget https://github.com/earthly/earthly/releases/download/v0.6.22/earthly-linux-amd64 -O /usr/local/bin/earthly
  - chmod +x /usr/local/bin/earthly
  - export FORCE_COLOR=1
  - /usr/local/bin/earthly bootstrap

stages:
  - build

build-api:
  stage: build
  script:
    - earthly --ci +build-api

build-front:
  stage: build
  script:
    - earthly --ci +build-front
```

kcampion:
Here the log from this job
```
Running with gitlab-runner 15.3.0 (bbcb5aba)
  on webapp bHMv_R1Y
Resolving secrets 00:00
Preparing the "docker" executor 00:18
Using Docker executor with image docker:git ...
Pulling docker image registry.gitlab.com/gitlab-org/gitlab-runner/gitlab-runner-helper:x86_64-bbcb5aba ...
Using docker image sha256:eb09578c512493961df34d9fcea26ec9beba73f6888b35dfe36858c36accf8fa for registry.gitlab.com/gitlab-org/gitlab-runner/gitlab-runner-helper:x86_64-bbcb5aba with digest registry.gitlab.com/gitlab-org/gitlab-runner/gitlab-runner-helper@sha256:24641a70251a846357ccb5866ae834e9907eb1d1e52f2dcaa02cf5e88e6481b3 ...
Starting service docker:dind ...
Pulling docker image docker:dind ...
[...]

PR #2207, opened 2022-09-21
Use flock to prevent buildkit startup race conditions

vladaionescu:
* Use flock to prevent buildkit startup race conditions. Fixes https://github.com/earthly/earthly/issues/2194
* This PR also speeds up buildkit start time, after adjusting some timeouts in the startup code

### DIFF
--- a/buildkitd/buildkitd.go
+++ b/buildkitd/buildkitd.go
@@ -749,7 +749,7 @@ func WaitUntilStopped(ctx context.Context, containerName string, opTimeout time.
 	defer cancel()
 	for {
 		select {
-		case <-time.After(1 * time.Second):
+		case <-time.After(200 * time.Millisecond):
 			isRunning, err := isContainerRunning(ctxTimeout, containerName, fe)
 			if err != nil {
 				// The container can no longer be found at all.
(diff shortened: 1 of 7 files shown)


buildkitd/buildkitd.go
```diff
@@ -135,6 +136,19 @@ func ResetCache(ctx context.Context, console conslogging.ConsoleLogger, image, c
...
+		defer cancel()
+		_, err := startLock.TryLockContext(timeoutCtx, 200*time.Millisecond)
+		if err != nil {
+			if errors.Is(err, context.DeadlineExceeded) {
+				return nil, nil, errors.Errorf("timeout waiting for other process to start buildkitd")
+			}
+			return nil, nil, errors.Wrapf(err, "try flock context %s", settings.StartUpLockPath)
+		}
+		defer startLock.Unlock()
+	}
```
vladaionescu, line 151: FWIW, I tried first using a shared lock, which would later be upgraded to an exclusive lock, but it seems that on some systems the shared lock becomes automatically an exclusive lock, which causes strange behaviors.

The logic I ended up using (exclusive locking only) seems to be performant enough in practice (there's no bottleneck that I could notice in my experimentation locally).

alexcb (approved):

PR #2215, opened 2022-09-23
Do not complain if failing to remove dockerd data root dir itself, if dir is empty

vladaionescu:
Help eliminate the extra output when using podman + docker-compose.


alexcb (approved):

r-LaForge (approved):

vladaionescu:
Merging with failing test as the failure of misc-test also appears on `main`.

PR #2219, opened 2022-09-23
Docs updates

jalletto:
Let me know how you feel about the updates to the go-example in the first part. Once we feel good about how things are named, I can go back and update all the other examples in the other sections, since many of them build off this first example.

### DIFF
--- a/docs/basics/part-1-a-simple-earthfile.md
+++ b/docs/basics/part-1-a-simple-earthfile.md
@@ -3,16 +3,16 @@ Below you'll find a simple example of an Earthfile. All the magic of Earthly hap
 ```Dockerfile
 VERSION 0.6
 FROM golang:1.15-alpine3.13
-WORKDIR /go-example
- 
+WORKDIR /go-workdir
+
 build:
     COPY main.go .
-    RUN go build -o build/go-example main.go
-    SAVE ARTIFACT build/go-example /go-example AS LOCAL build/go-example
+    RUN go build -o my_go_build/go-example main.go
+    SAVE ARTIFACT my_go_build/go-example /saved_go_artifact AS LOCAL local_go_build/go-example
 
 docker:
-    COPY +build/go-example .
-    ENTRYPOINT ["/go-example/go-example"]
+    COPY +build/saved_go_artifact .
+    ENTRYPOINT ["/go-workdir/saved_go_artifact"]
     SAVE IMAGE go-example:latest
 ```
 
@@ -36,7 +36,7 @@ We'll slowly build up to the Earthfile we have above. Let's start with these fir
 ```Dockerfile
 VERSION 0.6
 FROM golang:1.15-alpine3.13
-WORKDIR /go-example
+WORKDIR /go-workdir
 ```
 And some simple Hello World code in a `main.go`.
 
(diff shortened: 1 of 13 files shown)


docs/basics/part-1-a-simple-earthfile.md
```diff
@@ -3,16 +3,16 @@ Below you'll find a simple example of an Earthfile. All the magic of Earthly hap
...
-WORKDIR /go-example
- 
+WORKDIR /go-workdir
+
 build:
     COPY main.go .
-    RUN go build -o build/go-example main.go
-    SAVE ARTIFACT build/go-example /go-example AS LOCAL build/go-example
+    RUN go build -o my_go_build/go-example main.go
+    SAVE ARTIFACT my_go_build/go-example /saved_go_artifact AS LOCAL local_go_build/go-example
```
vladaionescu, line 11:
```suggestion
    RUN go build -o output/example main.go
    SAVE ARTIFACT output/example AS LOCAL local-output/go-example
```

docs/basics/part-1-a-simple-earthfile.md
```diff
@@ -3,16 +3,16 @@ Below you'll find a simple example of an Earthfile. All the magic of Earthly hap
...
 
 docker:
-    COPY +build/go-example .
-    ENTRYPOINT ["/go-example/go-example"]
+    COPY +build/saved_go_artifact .
+    ENTRYPOINT ["/go-workdir/saved_go_artifact"]
```
vladaionescu, line 15:
```suggestion
    COPY +build/example .
    ENTRYPOINT ["/go-workdir/example"]
```

docs/basics/part-6-using-docker-with-earthly.md
```diff
@@ -145,6 +145,13 @@ integration-tests:
 ```
 When we use the `--compose` flag, Earthly will start up the services defined in the `docker-compose` file for us. In this case, we built a separate image that copies in our test files and uses the command to run the tests as its `ENTRYPOINT`. We can then load this image into our `WITH DOCKER` command. Note that loading an image will not run it by default, we need to explicitly run the image after we load it.
 
+You'll need to use `--allow-privileged` to run this example.
```
vladaionescu, line 148:
```suggestion
You'll need to use `--allow-privileged` (or `-P` for short) to run this example. 
```

vladaionescu (commented):
Looks good! I provided some suggestions to simplify a little bit while maintaining some kind of consistency, but with different-enough names.

docs/basics/part-6-using-docker-with-earthly.md
```diff
@@ -145,6 +145,13 @@ integration-tests:
 ```
 When we use the `--compose` flag, Earthly will start up the services defined in the `docker-compose` file for us. In this case, we built a separate image that copies in our test files and uses the command to run the tests as its `ENTRYPOINT`. We can then load this image into our `WITH DOCKER` command. Note that loading an image will not run it by default, we need to explicitly run the image after we load it.
 
+You'll need to use `--allow-privileged` (or `-P` for short) to run this example. 
+
+```
```
vladaionescu, line 150:
```suggestion
```bash
```

vladaionescu (approved):
One more very minor comment.

PR #2222, opened 2022-09-28
added an example for a Next.js website deployed to Netlify

thtmnisamnstr:
Added an example for a Next.js website deployed to Netlify

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -29,6 +29,7 @@ Please note that these examples, although similar, are distinct from the ones us
 * [go](./go)
 * [java](./java)
 * [js](./js)
+* [next-js-netlify](./next-js-netlify)
 * [python](./python)
 * [ruby](./ruby)
 * [ruby-on-rails](./ruby-on-rails)
--- a/examples/next-js-netlify/Earthfile
+++ b/examples/next-js-netlify/Earthfile
@@ -8,21 +8,22 @@ deps:
     # Set environment variables required for building and deploying to Netlify
     ENV NETLIFY_AUTH_TOKEN=$auth_token
     ENV NETLIFY_SITE_ID=$site_id
-    # Install the Netlify CLI
+    # Install the Netlify CLI (global)
     RUN npm install netlify-cli -g
-    # Copy files required for building and deploying to Netlify
+    # Copy package.json for installing required packages
+    COPY package.json ./
+    # Install the Netlify CLI (local) and Netlify Next.js plugin and required packages
+    RUN npm install netlify-cli --save-dev
+    RUN npm install @netlify/plugin-nextjs --save-dev
+    RUN npm install
+    # Copy netlify.toml - required for building and deploying to Netlify
     COPY netlify.toml ./
 
 build:
     FROM +deps
-    # Copy files required for building
-    COPY next-env.d.ts package.json styles.module.css tsconfig.json ./
-    # Copy directories required for building
+    # Copy files and directories required for building
+    COPY next-env.d.ts styles.module.css tsconfig.json ./
     COPY --dir pages ./
-    # Install required packages
-    RUN npm install netlify-cli --save-dev
-    RUN npm install @netlify/plugin-nextjs --save-dev
-    RUN npm install
     # Build site
     RUN netlify build --context production
     SAVE ARTIFACT ./node_modules node_modules/ AS LOCAL ./
--- a/examples/next-js-netlify/Earthfile
+++ b/examples/next-js-netlify/Earthfile
@@ -3,11 +3,6 @@ FROM --platform=linux/amd64 node:latest
 WORKDIR /app
 
 deps:
-    ARG auth_token
-    ARG site_id
-    # Set environment variables required for building and deploying to Netlify
-    ENV NETLIFY_AUTH_TOKEN=$auth_token
-    ENV NETLIFY_SITE_ID=$site_id
     # Install the Netlify CLI (global)
     RUN npm install netlify-cli -g
     # Copy package.json for installing required packages
@@ -24,8 +19,8 @@ build:
     # Copy files and directories required for building
     COPY next-env.d.ts styles.module.css tsconfig.json ./
     COPY --dir pages ./
-    # Build site
-    RUN netlify build --context production
+    # Build site using NETLIFY_AUTH_TOKEN and NETLIFY_SITE_ID secrets
+    RUN --secret NETLIFY_AUTH_TOKEN --secret NETLIFY_SITE_ID netlify build --context production
     SAVE ARTIFACT ./node_modules node_modules/ AS LOCAL ./
     SAVE ARTIFACT ./.next .next/ AS LOCAL ./
     SAVE ARTIFACT ./.netlify .netlify/ AS LOCAL ./
@@ -35,4 +30,4 @@ deploy:
     # Copy artifacts required for deploying to Netlify
     COPY +build/node_modules/ +build/.next/ +build/.netlify/ ./
     # Deploy site
-    RUN netlify deploy --prod
+    RUN --push --secret NETLIFY_AUTH_TOKEN --secret NETLIFY_SITE_ID netlify deploy --prod
--- a/examples/next-js-netlify/Earthfile
+++ b/examples/next-js-netlify/Earthfile
@@ -8,9 +8,10 @@ deps:
     # Set environment variables required for building and deploying to Netlify
     ENV NETLIFY_AUTH_TOKEN=$auth_token
     ENV NETLIFY_SITE_ID=$site_id
+    # Install the Netlify CLI
+    RUN npm install netlify-cli -g
     # Copy files required for building and deploying to Netlify
     COPY netlify.toml ./
-    RUN npm install netlify-cli -g
 
 build:
     FROM +deps
--- a/examples/next-js-netlify/Earthfile
+++ b/examples/next-js-netlify/Earthfile
@@ -1,0 +1,36 @@
+VERSION 0.6
+FROM --platform=linux/amd64 node:latest
+WORKDIR /app
+
+deps:
+    ARG auth_token
+    ARG site_id
+    # Set environment variables required for building and deploying to Netlify
+    ENV NETLIFY_AUTH_TOKEN=$auth_token
+    ENV NETLIFY_SITE_ID=$site_id
+    # Copy files required for building and deploying to Netlify
+    COPY netlify.toml ./
+    RUN npm install netlify-cli -g
+
+build:
+    FROM +deps
+    # Copy files required for building
+    COPY next-env.d.ts package.json styles.module.css tsconfig.json ./
+    # Copy directories required for building
+    COPY --dir pages ./
+    # Install required packages
+    RUN npm install netlify-cli --save-dev
+    RUN npm install @netlify/plugin-nextjs --save-dev
+    RUN npm install
+    # Build site
+    RUN netlify build --context production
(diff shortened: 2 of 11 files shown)


examples/README.md
```diff
@@ -29,6 +29,7 @@ Please note that these examples, although similar, are distinct from the ones us
 * [go](./go)
 * [java](./java)
 * [js](./js)
+* [next-js-netlify](./next-js-netlify)
```
r-LaForge, line 32: Make sure you add this to the [base Earthfile](examples/next-js-netlify/.gitignore) too.

examples/next-js-netlify/Earthfile
```diff
@@ -0,0 +1,36 @@
...
+
+deps:
+    ARG auth_token
+    ARG site_id
+    # Set environment variables required for building and deploying to Netlify
+    ENV NETLIFY_AUTH_TOKEN=$auth_token
+    ENV NETLIFY_SITE_ID=$site_id
+    # Copy files required for building and deploying to Netlify
+    COPY netlify.toml ./
+    RUN npm install netlify-cli -g
```
r-LaForge, line 13: I would recommend doing the install first so that a change in the toml doesn't require re-downloading netlify-cli.

examples/README.md
thtmnisamnstr: I'm leaving this off (at least for now). Both the +build and +deploy targets require supplying arguments. So this build would fail unless the user manually included args before executing either target.

examples/next-js-netlify/Earthfile
thtmnisamnstr: I committed this change.

examples/next-js-netlify/Earthfile
```diff
@@ -0,0 +1,37 @@
...
+build:
+    FROM +deps
+    # Copy files required for building
+    COPY next-env.d.ts package.json styles.module.css tsconfig.json ./
+    # Copy directories required for building
+    COPY --dir pages ./
+    # Install required packages
+    RUN npm install netlify-cli --save-dev
+    RUN npm install @netlify/plugin-nextjs --save-dev
+    RUN npm install
```
r-LaForge, line 25: Similar to the above change you made - this means a change to anything in `pages` will break cache and require a full install. 

Is it possible to put all of your `npm install` commands into `deps` and then do the majority of your other commands later? 

The way I like to think about it is "put the commands with content that is most likely to change below other commands that are unlikely to change".

examples/next-js-netlify/README.md
```diff
@@ -0,0 +1,108 @@
...
+1.  `deps` is the base target that `build` and `deploy` use. It copies all of the common files and sets environment variables that both the `build` and `deploy` targets require.
+2.  `build` is a target that copies the local files and directories required for building, installs prerequisites, and builds your website using Netlify's CLI.
+3.  `deploy` is a target that runs the `build` target (if it isn't cached), copies the output from `build` that is required for deploying to Netlify, and deploys your site to Netlify using their CLI.
+  
+**Note:** We specify the platform in the base image to be linux/amd64. This is because the Netlify CLI uses [Deno](https://deno.land/), and Deno doesn't support linux/arm64 at the moment. If you're using an arm64 computer, the `build` target will fail unless the platform is specified as amd64.  
+  
+`Earthfile`
+```
+VERSION 0.6
+FROM --platform=linux/amd64 node:latest
```
r-LaForge, line 51: You can also run `earthly --platform=linux/amd64 +build` or `earthly --platform=linux/amd64 +deploy`

examples/next-js-netlify/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+deps:
+    ARG auth_token
+    ARG site_id
+    # Set environment variables required for building and deploying to Netlify
+    ENV NETLIFY_AUTH_TOKEN=$auth_token
+    ENV NETLIFY_SITE_ID=$site_id
+    # Copy files required for building and deploying to Netlify
+    COPY netlify.toml ./
+    RUN npm install netlify-cli -g
```
r-LaForge, line 62: This is already a little out of sync with the current Earthfile. I would recommend leaving it out or just copying the piece that changed.

r-LaForge (approved):
Had a few small recommendations - nice work! :) 

Some of the tests are failing with `./examples/next-js-netlify/README.md does not end with a newline`.
misc-tests may just need to run.

examples/next-js-netlify/Earthfile
thtmnisamnstr: This is actually the correct order of operations for a Next.js site. All of those installs have to run after files are copied. They install packages that are necessary to the site and defined in the copied package.json file.

examples/next-js-netlify/README.md
thtmnisamnstr: That is true, but, since linux/amd64 is required because of a limitation in one of Netlify CLI's dependencies, I thought it was more foolproof to put it in the Earthfile. If I didn't include it, and someone forgot `--platform=linux/amd64` in the `earthly` command, the build would fail.

examples/next-js-netlify/README.md
thtmnisamnstr: That was my bad. I forgot to update the readme. I'll update.

examples/next-js-netlify/Earthfile
r-LaForge: Oh - does that also include the pages directory? 
I would have thought it would only have been `package.json` and `package-lock.json` that are required for an install. 🤔

examples/next-js-netlify/README.md
r-LaForge: I see - that makes perfect sense.

examples/next-js-netlify/Earthfile
thtmnisamnstr: Actually, you are correct. It just needs `package.json` to install dependencies. I updated the Earthfile and Readme to do all installs on the +deps target.

thtmnisamnstr:
@r-LaForge Any thoughts on why `GitHub Actions CI / misc-tests` and `GitHub Actions CI / +examples1` are failing? `+examples1` passed on previous commits to this PR, and nothing in this PR should impact it anyway.

examples/next-js-netlify/Earthfile
r-LaForge: Awesome 🥳

r-LaForge:
@thtmnisamnstr unfortunately, our tests are a little flakey in this repository. You'll even notice this if you look at the master commits.

I'm not sure why these fail (other than the error messages in them).
examples seems to be 
```
psql: could not connect to server: No such file or directory
```

while misc-tests is `psql: could not connect to server: No such file or directory`.

chances are if you re-run those jobs they will eventually pass.

My Podman PR introduces retries on the troublesome commands.

examples/next-js-netlify/Earthfile
```diff
@@ -0,0 +1,38 @@
+VERSION 0.6
+FROM --platform=linux/amd64 node:latest
+WORKDIR /app
+
+deps:
+    ARG auth_token
+    ARG site_id
+    # Set environment variables required for building and deploying to Netlify
+    ENV NETLIFY_AUTH_TOKEN=$auth_token
```
vladaionescu, line 9: This should be passed in via a secret. See the `RUN --secret` option. I suspect you only need this for deployment - so you would normally put this in the `RUN netlify deploy` step.

Explanation: https://docs.earthly.dev/best-practices#use-secret-not-args-to-pass-secrets-to-the-build.

examples/next-js-netlify/Earthfile
```diff
@@ -0,0 +1,38 @@
...
+    ENV NETLIFY_SITE_ID=$site_id
```
vladaionescu, line 10: Wondering if these are declared too early in the build. If the values change, too much cache would be invalidated. I would declare these where they are needed only (e.g. in `+build` and `+deploy`), and as late as possible - in order to preserve as much cache as we can.

examples/next-js-netlify/Earthfile
```diff
@@ -0,0 +1,38 @@
...
+    # Install the Netlify CLI (global)
+    RUN npm install netlify-cli -g
+    # Copy package.json for installing required packages
+    COPY package.json ./
+    # Install the Netlify CLI (local) and Netlify Next.js plugin and required packages
+    RUN npm install netlify-cli --save-dev
+    RUN npm install @netlify/plugin-nextjs --save-dev
```
vladaionescu, line 17: Wondering if these can be saved in `package.json`, and then installed via `npm install --include dev`.

examples/next-js-netlify/Earthfile
```diff
@@ -0,0 +1,38 @@
...
+    SAVE ARTIFACT ./node_modules node_modules/ AS LOCAL ./
+    SAVE ARTIFACT ./.next .next/ AS LOCAL ./
+    SAVE ARTIFACT ./.netlify .netlify/ AS LOCAL ./
+
+deploy:
+    FROM +deps
+    # Copy artifacts required for deploying to Netlify
+    COPY +build/node_modules/ +build/.next/ +build/.netlify/ ./
+    # Deploy site
+    RUN netlify deploy --prod
```
vladaionescu, line 38: We generally use `--push` commands for things that have effects outside of Earthly.

`RUN --push` commands ensure that the execution takes place every time, even if it is cached. It is also a safety feature - `RUN --push` commands execute only when `earthly` itself is executed with the `--push` flag: `earthly --push +deploy`. A more in-depth explanation here: https://docs.earthly.dev/best-practices#use-run-push-for-deployment-commands.

```suggestion
    RUN --push netlify deploy --prod
```

vladaionescu (commented):
Looks great! I left a few comments...

We should try to add this example to `+examples2` in here, to make sure that it works fine forever (this is ran as part of the CI): https://github.com/earthly/earthly/blob/main/Earthfile#L547

examples/next-js-netlify/Earthfile
thtmnisamnstr: The Netlify CLI requires the auth token and site id for build as well as for deploy (weird, right?). Regardless, I'll update to pass it as a secret.

examples/next-js-netlify/Earthfile
thtmnisamnstr: That is fair. I'll update.

examples/next-js-netlify/Earthfile
thtmnisamnstr: They could be, but they aren't necessary packages for the site itself, only for using the Netlify CLI to build and deploy. If you build outside of the Netlify CLI or let Netlify's CI pick up the changes in your repo, you don't have to have these in `package.json`. So including them in the Earthfile makes more sense to me.

examples/next-js-netlify/Earthfile
thtmnisamnstr: I didn't know that. I though push was more for pushing container images. I'll update.

examples/next-js-netlify/Earthfile
thtmnisamnstr: Do we have any examples that use secrets in `+examples1` or `+examples2`? Both `+build` and `+deploy` will fail without having those secrets passed appropriately. That's why I didn't include it in `+examples2`.

vladaionescu: We don't... I see the dilemma. We'd have to maintain the Netlify account + secrets just for this integration test. Perhaps it's ok not to include in our CI suite in that case. Feels like too much of a maintenance burden.

vladaionescu (approved):
Ready to merge!

Given that this PR is low-risk (and it includes only example code), I believe that we can merge without CI.

Issue #2220, opened 2022-09-27
earthy-buildkit container start timeout value should be configurable


alexcb:
https://github.com/earthly/earthly/commit/422c07a4599a7a6e40f12a600181b8a828058d7e#diff-e63c4b0c699fa73d4d9f7cf9913bb0f7100bc56c1b1419bb7698d0b650894ea6R500 changed 1sec to 200ms, and we had a report from a user having a timeout:

```
Hi! I attempted to upgrade earthly from v0.6.23 to v0.6.24 in self hosted Drone CI. We have more than 10 steps running in parallel, but unfortunately steps start to fail randomly with v0.6.24 within a few seconds after step start. I haven’t looked at the source code changes since v0.6.23, but it looks something to do with the buildkit client. Any help? :)
+ earthly +build --service=horizon-etl
           bootstrap | Bootstrapping successful.
 1. Init 🚀
————————————————————————————————————————————————————————————————————————————————

           buildkitd | Found buildkit daemon as docker container (earthly-buildkitd)
Share your logs with an Earthly account (experimental)! Register for one at https://ci.earthly.dev/.
Error: build new buildkitd client: maybe start buildkitd: maybe restart: timeout
```

vladaionescu:
It might have been a different timeout changed as part of that same PR. The 200ms is the retry interval, I believe.

janishorsts:
I run earthly with `--verbose --debug` and it printed out the following stack trace.

```
$ + earthly --verbose --debug ...
           bootstrap | Bootstrapping successful.
 1. Init 🚀
————————————————————————————————————————————————————————————————————————————————

            frontend | docker-shell frontend initialized.
           buildkitd | Found buildkit daemon as docker container (earthly-buildkitd)
           buildkitd | Comparing running container image ("sha256:c1fe02837a933c70884d8cd73fd44a12c7740d9bddce847ff3f5197d33a1454e") with available image ("sha256:c1fe02837a933c70884d8cd73fd44a12c7740d9bddce847ff3f5197d33a1454e")
           buildkitd | Settings hashes match ("604dbd8e859bbc5c"), no restart required
Share your logs with an Earthly account (experimental)! Register for one at https://ci.earthly.dev.
Error stack trace:
github.com/earthly/earthly/buildkitd.checkConnection
	/earthly/buildkitd/buildkitd.go:604
github.com/earthly/earthly/buildkitd.MaybeRestart
	/earthly/buildkitd/buildkitd.go:213
[...]

PR #2229, opened 2022-09-30
Fix buildkit connection timing out occasionally

vladaionescu:


brandonSc (approved):
Looks good to me. I ran your branch in a loop for a few minutes and didn't see any timeouts. When I tried doing that on the main branch I got a few of them.

brandonSc:
Is this fix related to this issue? https://github.com/earthly/earthly/issues/2220

Mel also reported a similar issue, but the error is `buildkit crashed` , do you think it's related? https://earthlycommunity.slack.com/archives/C01DL2928RM/p1664663129123339

vladaionescu:
> Mel also reported a similar issue, but the error is buildkit crashed , do you think it's related?

Yes, I believe that is should be related.

PR #2238, opened 2022-10-05
Add line numbers for build failures

vladaionescu:
[Sample](https://ci.earthly.dev/logs?logId=0bff2c8a-f26c-4f59-9706-aea64bb4ea9b)


cmd/debugger/main.go
```diff
@@ -304,7 +304,9 @@ func main() {
 		exitCode := 1
 		if exitErr, ok := err.(*exec.ExitError); ok {
 			exitCode = exitErr.ExitCode()
-			conslogger.Warnf("Command %s failed with exit code %d\n", quotedCmd, exitCode)
+			if debuggerSettings.Enabled {
```
alexcb, line 307: nice catch! guess we just got used to this error message always appearing.

outmon/vertexmon.go
```diff
@@ -175,13 +176,54 @@ func (vm *vertexMonitor) printProgress(id string, progress int, verbose bool, sa
 	vm.console.PrintBytes([]byte(strings.Join(builder, "")))
 }
 
+var reErrExitCode = regexp.MustCompile(`^process (".*") did not complete successfully: exit code: ([0-9]+)$`)
```
alexcb, line 179: these regexs feel fragile -- are they covered by any integration tests?

r-LaForge (approved):
Just one question regarding protecting a log line using `Enabled`.

alexcb (commented):
I like the feature.

We should create a test for this.

cmd/debugger/main.go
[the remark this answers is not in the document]
vladaionescu: This piece of code is from the debugger. Every user command being executed is actually wrapped by the debugger. The debugger should only start interacting only if it's enabled. In this case, despite being disabled, it would add some output to failing commands. This kind of output has now been moved to its proper place (see `printError` in another file in this PR)..

outmon/vertexmon.go
vladaionescu: Done

alexcb (approved):

PR #2253, opened 2022-10-07
Use Compute API instead of Pipelines for Satellites

brandonSc:


vladaionescu (approved):

go.mod
```diff
@@ -48,6 +48,6 @@ replace (
 	github.com/earthly/earthly/ast => ./ast
 	github.com/jessevdk/go-flags => github.com/alexcb/go-flags v0.0.0-20210722203016-f11d7ecb5ee5
 
-	github.com/moby/buildkit => github.com/earthly/buildkit v0.0.1-0.20220913002628-9c256c19c539
+	github.com/moby/buildkit => github.com/earthly/buildkit v0.0.1-0.20220922172718-e3950730ae66
```
alexcb, line 51: why is buildkit being updated here?

alexcb (changes requested):
can you remove the buildkit update, unless it's absolutely required.

go.mod
brandonSc: Oh I don't even know where that came from. Glad you noticed.

alexcb (approved):

PR #2257, opened 2022-10-08
Enable console log grouping in context of Github actions

### LINKED ISSUE #2189: Collapsible output in GitHub Actions
This nifty little script can take Earthly output and make it look prettier in GHA. https://gist.github.com/Dkendal/065c7713aea782c232bd0f94985504a7

image (8)

It would be cool if Earthly could do that on its own:
* Detect automatically when it's running in the context of GHA
* In that situation (or if forced via a flag / setting), perform the stage printouts differently to use GHA folding

arajhansa:
Resolves #2189 

As per this PR, Console Logger now reads the `GITHUB_ACTION` env var to determine if logging output is in GHA. If so, it appends the required grouping structures.

### DIFF
--- a/buildkitd/buildkitd.go
+++ b/buildkitd/buildkitd.go
@@ -21,6 +21,7 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 
+	"github.com/earthly/earthly/ast/hint"
 	"github.com/earthly/earthly/conslogging"
 	"github.com/earthly/earthly/util/buildkitutil"
 	"github.com/earthly/earthly/util/cliutil"
@@ -40,7 +41,7 @@ var (
 // NewClient returns a new buildkitd client. If the buildkitd daemon is local, this function
 // might start one up, if not already started.
 func NewClient(ctx context.Context, console conslogging.ConsoleLogger, image, containerName, installationName string, fe containerutil.ContainerFrontend, earthlyVersion string, settings Settings, opts ...client.ClientOpt) (*client.Client, error) {
-	opts, err := addRequiredOpts(settings, installationName, opts...)
+	opts, err := addRequiredOpts(settings, installationName, fe.Config().Setting == containerutil.FrontendPodmanShell, opts...)
 	if err != nil {
 		return nil, errors.Wrap(err, "add required client opts")
 	}
@@ -94,7 +95,7 @@ func ResetCache(ctx context.Context, console conslogging.ConsoleLogger, image, c
 		return errors.New("cannot reset cache of a provided buildkit-host setting")
 	}
 
-	opts, err := addRequiredOpts(settings, installationName, opts...)
+	opts, err := addRequiredOpts(settings, installationName, fe.Config().Setting == containerutil.FrontendPodmanShell, opts...)
 	if err != nil {
 		return errors.Wrap(err, "add required client opts")
 	}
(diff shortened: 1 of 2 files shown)


vladaionescu (commented):
Hey thanks for the contribution! Unfortunately, it doesn't look like it's working:

Screen Shot 2022-10-08 at 10 38 30 AM

https://github.com/earthly/earthly/actions/runs/3209795830/jobs/5249161551#step:17:14

Perhaps it's because the `::group::` part is not the very first thing on the line? You might notice that it's colored - which would mean that the ANSI escape sequences are printed before `::group::` on that line.

Separately, can you also add a line in [CHANGELOG.md](https://raw.githubusercontent.com/earthly/earthly/main/CHANGELOG.md) under Unreleased -> Added?

conslogging/conslogging.go
```diff
@@ -195,6 +196,9 @@ func (cl ConsoleLogger) PrintPhaseHeader(phase string, disabled bool, special st
 	defer cl.mu.Unlock()
 	msg := phase
 	c := cl.color(phaseColor)
+	if cl.shouldGroupPhaseOutput() {
+		msg = fmt.Sprintf("::group::%s", msg)
+	}
```
vladaionescu, line 201: To print uncolored output, IIRC, you could try to use `cl.errW.Write([]byte("::group::")` before the line `c.Fprintf(cl.errW, " %s", msg)`.

arajhansa: I didn't think of this. Might work. I've updated the PR.

vladaionescu: You should do the same for endgroup

arajhansa:
Thanks to the suggestion I think it is working now:

image

https://github.com/earthly/earthly/actions/runs/3213430182/jobs/5254151556#step:17:1

vladaionescu:
I just noticed an unintended side-effect of this.

In this test failure: https://github.com/earthly/earthly/actions/runs/3213430189/jobs/5254151335#step:17:21 (failure is probably an unrelated flake due to a random network error), the output is difficult to browse, because GHA does not automatically scroll to the end when opening the Build step. So hunting for the error becomes very difficult.

There are a couple of places where error output propagates from:
* Place 1: https://github.com/earthly/earthly/blob/2391901a9e29a051ed49152d282f9bacf2549916/builder/builder.go#L498
* Place 2 https://github.com/earthly/earthly/blob/2391901a9e29a051ed49152d282f9bacf2549916/outmon/solvermon.go#L106

What I think we should do is to close the group before the failure output. That way, the failure output is already opened and there is no need to hunt for it. In the first place, it's easy, because all we have to do is move the `PrintPhaseFooter` from just a few lines below to above the `if err != nil`. In the second place, it's more difficult, because it's somewhere deeper in the call stack.

For the second place, I think we should do the following:
* Remove the first param (`errVertex`) from `reprintFailure` as it's already accessible via `sm`.
* Switch up `reprintFailure` into a `maybeReprintFailure` - it only reprints if the `errVertex` is not nil.
* At the end of the call perhaps it also clears `sm.errVertex = nil`, to avoid reprinting the failure too many times.
* Expose this new `maybeReprintFailure` from the `SolverMonitor` as a public function (`MaybeReprintFailure`).
* Expose a new function, also called `MaybeReprintFailure` in [`Solver`](https://github.com/earthly/earthly/blob/main/builder/solver.go), which simply calls the `sm.MeybeReprintFailure`.
* In Builder (the same file as place 1), just after the printing of any phase footer, we call the solver's `MaybeReprintFailure`.

I hope I'm not missing anything in my thinking! Do you want to give this a shot?

arajhansa:
I do agree on the suggestion.

I would like to put in here that the scope of this is more than initial suggestion for this issue. While I'd love to continue work on this, I need more time. With my full time occupied elsewhere, I would kindly like to suggest to close this PR and issue and raise a new issue with required changes. Not to mention, I'd be happy to take that on.

let me know your take on this.

vladaionescu:
Sure thing - I think we can keep this PR open. Either to serve as an inspiration for the remaining work, or to be built on top of, if you end up coming back to this yourself. I'll update the original issue with the new finding.

I would not merge this PR as-is, mainly because finding the source of the build breakage is more trouble with this change than without it.

Issue #2237, opened 2022-10-04
VERSION 0.7


alexcb:
This is a high catch-all ticket for the upcoming VERSION 0.7 release.

The following features will be enabled by default:
 - explicit-global
 - check-duplicate-images
 - earthly-version-arg
 - use-cache-command
 - use-host-command
 - use-copy-link
 - new-platform
 - no-tar-build-output
 - use-no-manifest-list
 - use-chmod
 - shell-out-anywhere
 - use-registry-for-with-docker
 - earthly-locally-arg
- use-project-secrets
- use-pipelines
- wait

The following feature(s) will remain as experimental, due to them (it) being too new:
- try

TODO:
 - [x] Draft release notes
 - [x] Finish up release notes (add information about all the breaking changes, expand on the meaning of each feature flag being flipped)
 - [x] Add docs for PIPELINE, TRIGGER, PROJECT, TRY (+FINALLY)
 - [ ] Update docs to point to 0.7 in all examples #2587 
 - [x] Update docs for commands that are being GA'd in 0.7 to no longer say "beta" or "experimental"
 - [x] Edit existing secrets-related docs to update to the new secrets (+ add migration information)
 - [x] Finish Podman support, including necessary docs
 - [ ] [Later] Update all syntax highlighting packages with the new commands: PIPELINE, TRIGGER, PROJECT, CACHE, WAIT, TRY, FINALLY.
 - [x] Remove `--save-inline-cache` and `--use-inline-cache` from `--ci`. They seem to be problematic in some cases and we don't know what the cause of this is (see e.g. https://github.com/earthly/earthly/issues/1187)
 - [x] Force using the VERSION command at the top of every Earthfile (specifying `VERSION 0.5` is still allowed, but it has to be explicit)
 - [x] #2544
 - [ ] [Later] #2178
 - [x] #1783
 - [x] #2272
 - [x] #2218
 - [x] #2609

Notes:
 - We will do one or more `v0.7-rc<n>` release candidates prior to the full v0.7.0 release

alexcb:
#2269 contains flags for docs

alexcb:
we will aim to release `WAIT` in VERSION 0.7 too; however need to fix #2218 first

r-LaForge:
The change of behaviour that comes with the `--wait-block` flag will fix a segfault encountered with --push and SAVE ARTIFACT.

Closes https://github.com/earthly/earthly/issues/2286

alexcb:
[...]

PR #2271, opened 2022-10-12
version 0.7 RC

alexcb:
New 0.7 version with corresponding default-enabled features, along with
updated tests.


alexcb:
There are a few tests that are left at `VERSION 0.6`, which I will address in follow-up PRs.

examples/secrets/Earthfile
```diff
@@ -1,3 +1,4 @@
+VERSION 0.6
```
nelsam, line 1: Just double-checking that this was intentional.

A comment explaining why this example uses `0.6` would be useful here. I think if I were trying to use earthly and was browsing through the examples, I would be surprised to see old versions pinned with no explanation.

CHANGELOG.md
```diff
@@ -96,6 +125,10 @@ All notable changes to [Earthly](https://github.com/earthly/earthly) will be doc
 
 - Loading Docker images as part of `WITH DOCKER` is now faster through the use of an embedded registry in Buildkit. This functionality was previously hidden (`VERSION --use-registry-for-with-docker`) and was only auto-enabled for Earthly Satellite users. It is now enabled by default for all builds. [#1268](https://github.com/earthly/earthly/issues/1268)
 
+### Changed
+
+- `VERSION` command is now required.
+
```
vladaionescu, line 131: I would move `### Changed` as the first section. This is the stuff that will contain breaking changes too and it feels pretty important.

tests/allow-privileged-import.earth
```diff
@@ -1,8 +1,8 @@
-VERSION 0.6
+VERSION 0.7
 FROM alpine:3.15
 RUN apk add libcap # for capsh
 
-IMPORT --allow-privileged github.com/earthly/test-remote/privileged:main
+IMPORT --allow-privileged github.com/earthly/test-remote/privileged:45359a92acdd7ef27d561d0836099cdacc508e40
```
nelsam: Will we need to update these tags after merge? (this is more of a process question than a review comment)

CHANGELOG.md
```diff
@@ -4,6 +4,35 @@ All notable changes to [Earthly](https://github.com/earthly/earthly) will be doc
 
 ## Unreleased
 
+## v0.7.0-rc1 - 2022-10-12
```
vladaionescu, line 7: Did you mean to use a January date in here?

```suggestion
## v0.7.0-rc1 - 2023-01-13
```

vladaionescu, line 7: The release notes are not final, correct? Can we still edit them after merging to main? We will have more stuff in here to expand on. I would add most of the feature flags in more detail.

tests/remote-cache/Earthfile
```diff
@@ -1,9 +1,28 @@
...
+--no-tar-build-output               \
+--shell-out-anywhere                \
+--use-cache-command                 \
+--use-chmod                         \
+--use-copy-link                     \
+--use-host-command                  \
+--use-no-manifest-list              \
+--use-pipelines                     \
+--use-project-secrets               \
+#--wait-block                        \
```
nelsam, line 19: Was this comment intended to be left in?

nelsam: Oh, I think this is the bug that we're fixing after the RC, right? got it.

alexcb: This whole list comes from https://github.com/earthly/earthly/issues/2237#issuecomment-1340120452 (the features we're enabling for v0.7); I'm inclined to leave it commented out to both illustrate it's been disabled on purpose -- and to make it easier to re-enable when we loop back to fix this after the initial RC.

CHANGELOG.md
alexcb: correct -- even though I'm merging this in -- I do not intend to release it as is.

vladaionescu (approved):

CHANGELOG.md
alexcb: We'll update this when we rewrite the release notes prior to the actual release -- there'll be a few more PRs before then.

tests/remote-cache/Earthfile
nelsam: Yeah, makes sense. I usually use `git revert` for those purposes, so I don't often leave the comments in - but it does serve as a good reminder.

nelsam (approved):

Issue #2247, opened 2022-10-05
Earthly randomly hangs on Init step in `v0.6.24+`


vladaionescu:
`SIGABRT` stack trace: https://gist.github.com/jazzdan/505c9211c30efc533ab687831a169306

The stack trace shows hanging on `docker info`. Though that doesn't make a lot of sense.

The fact that this does not reproduce in `v0.6.23` seems to point to this possible change: https://github.com/earthly/earthly/pull/2207.

EDIT: A workaround has been identified: `brew uninstall earthly && brew install earthly/earthly/earthly`. This only happens on the homebrew core version of Earthly, but not on the version from our own tap.

vladaionescu:
Trying to emulate @jazzdan's usage - I tried running earthly in a loop and killing it randomly. I could never get it to hang. @jazzdan's docker version is the same as mine and we both use M1.

This is the script I used:

```bash
#!/bin/bash

set -eux -o pipefail

earthly=./build/darwin/arm64/earthly

one_test() {
    it="$1"
    echo "Iteration $it"
    "$earthly" --no-output +for-darwin-m1 &
    pid=$!
    sleep $(( ( RANDOM % 10 ) + 0 ))
    kill "$pid" || true
}

for i in {1..100}; do
    one_test "$i" &
done

wait
```

vladaionescu:
This script does cause `docker info` to time out:

```go
package main

import (
	"context"
	"fmt"
	"os"
	"os/exec"
	"strings"
	"time"
)

type commmandContextOutput struct {
	stdout strings.Builder
	stderr strings.Builder
}

func command(ctx context.Context, suffix int) {
	output := &commmandContextOutput{}
	cmd := exec.CommandContext(ctx, "docker", "info")
	cmd.Env = os.Environ() // Ensure all shellouts are using the current environment, picks up DOCKER_/PODMAN_ env vars when they matter
	cmd.Stdout = &output.stdout
	cmd.Stderr = &output.stderr

	err := cmd.Run()
	if err != nil {
		out := output.stdout.String()
		e := output.stderr.String()
		if false {
			fmt.Printf("%s, %s", out, e)
		}
		// fmt.Printf("error during run: %s\n", err)
	} else {
		// fmt.Printf("got info %d\n", suffix)
	}
}

func forever() {
	var i = 0
	for {
		i++
		ctx, _ := context.WithTimeout(context.Background(), time.Minute*5)
		go func() {
			<-ctx.Done()
[...]

PR #2278, opened 2022-10-14
Retry some frontend commands

vladaionescu:
Also, add verbose printing on executed frontend-related commands.

Re #2247

### DIFF
--- a/util/containerutil/shell_shared.go
+++ b/util/containerutil/shell_shared.go
@@ -10,6 +10,7 @@ import (
 	"os/exec"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/containerd/containerd/platforms"
 	"github.com/earthly/earthly/conslogging"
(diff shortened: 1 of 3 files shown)


util/containerutil/shell_shared.go
```diff
@@ -285,24 +286,38 @@ func (cco *commmandContextOutput) string() string {
...
+func (sf *shellFrontend) commandContextOutputWithRetry(ctx context.Context, retries int, timeout time.Duration, args ...string) (*commmandContextOutput, error) {
+	var err error
+	for i := 0; i < retries; i++ {
+		timeoutCtx, cancel := context.WithTimeout(ctx, timeout)
+		defer cancel()
+		output, cmdErr := sf.commandContextOutput(timeoutCtx, args...)
+		if cmdErr == nil {
+			return output, nil
+		}
+		err = cmdErr
```
r-LaForge, line 301: small nit: In this file we already use `multierror.Append(err, cmdErr)`. It might be nice to see all errors... they might be the same, but they might not be.

util/containerutil/shell_shared.go
```diff
@@ -285,24 +286,38 @@ func (cco *commmandContextOutput) string() string {
...
+		if i < retries-1 {
+			sf.Console.VerbosePrintf("Command '%s' failed. Retrying...\n", strings.Join(args, " "))
```
r-LaForge, line 303: Small nit:
This isn't technically the full command we're running. This is missing the binary and the `globalCompatibilityArgs` args we might append.

It would be nice for this log to be accurate, but that might bloat the scope of this PR.

r-LaForge (approved):

util/containerutil/shell_shared.go
alexcb: Always displaying the most recent failure message here would be useful:

```suggestion
			sf.Console.Printf("Command '%s' failed: %s. Retrying...\n", strings.Join(args, " "), cmdErr)
```

If it's hidden behind a verbose, and the command initially fails, some users might now have earthly wait up to 9*10=90 seconds before the 10th occurrence works. The user would have no idea why earthly is running so slowly.

alexcb (approved):
just a small suggestion.

PR #2280, opened 2022-10-15
Update crypto to address "parse error in message type 27" errors

### LINKED ISSUE #2151: Spurious "parse error in message type 27" errors in build output

rrjjvv:
This (re)-fixes #2151 

After upgrading to 0.6.26 I started getting those errors again.  It appears that https://github.com/earthly/earthly/commit/038239929f24ebdadcefc6b11c689a620b8dcda1 is the cause.  The commit message/content doesn't indicate it was intentional.  The downgrade happens to be the exact version that I upgraded from, which makes me think maybe it was a result of something like a weird merge conflict resolution, external automation, etc.

Under the assumption it was not intentional,  this is just the exact same fix from #2152.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -34,7 +34,7 @@ require (
 	github.com/stretchr/testify v1.7.0
 	github.com/tonistiigi/fsutil v0.0.0-20220510150904-0dbf3a8a7d58
 	github.com/urfave/cli/v2 v2.3.0
-	golang.org/x/crypto v0.0.0-20220315160706-3147a52a75dd
+	golang.org/x/crypto v0.0.0-20220826181053-bd7e27e6170d
 	golang.org/x/sync v0.0.0-20210220032951-036812b2e83c
 	golang.org/x/sys v0.0.0-20220919091848-fb04ddd9f9c8 // indirect
 	golang.org/x/term v0.0.0-20220919170432-7a66f970e087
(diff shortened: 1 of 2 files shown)


rrjjvv:
If there was an obvious easy/straightforward test for this (in the sense that "the active version of crypto does/doesn't emit the errors"), I'd have done so.  I would think it would be fairly easy to write a test that simply verifies that the version is at (or newer) the fixed version; if you think it's warranted, I can take a stab at it; let me know what you think.

vladaionescu:
My bad - thanks for catching!

alexcb (approved):

PR #2285, opened 2022-10-18
update buildkit to merge sha b4ccdaf100c3093bf0d4b430a70ec7f541d25a4a

r-LaForge:
Linked to buildkit PR: https://github.com/earthly/buildkit/pull/99
required fsutil update: https://github.com/earthly/fsutil/pull/9

PR #2289, opened 2022-10-19
Add OOM Hook Script

dchw:
Allows you to configure a custom adjusted OOM score for build runs, with an accompanying Buildkit PR.


buildkitd/entrypoint.sh
```diff
@@ -215,12 +215,21 @@ fi
 export TLS_ENABLED
 
 envsubst </etc/buildkitd.toml.template >/etc/buildkitd.toml
+
+# Set up OOM
+OOM_SCORE_ADJ="${BUILDKIT_OOM_SCORE_ADJ:-0}"
+export OOM_SCORE_ADJ
+
+envsubst '${OOM_SCORE_ADJ}' </bin/oom-adjust.sh.template >/bin/oom-adjust.sh
+chmod +x /bin/oom-adjust.sh
```
alexcb, line 224: can we `rm oom-adjust.sh.template` at this point?

buildkitd/entrypoint.sh
```diff
@@ -229,7 +238,10 @@ echo ""
 echo "======== Buildkitd config =========="
 cat /etc/buildkitd.toml
 echo "======== End buildkitd config =========="
-
+echo ""
+echo "======== OOM Adjust config =========="
```
alexcb, line 242: nit: 
```suggestion
echo "======== OOM Adjust script =========="
```

buildkitd/oom-adjust.sh.template
```diff
@@ -0,0 +1,10 @@
+#! /bin/sh
```
alexcb, line 1: can we add `set -e`?

buildkitd/oom-adjust.sh.template
```diff
@@ -0,0 +1,10 @@
...
+
+if [ "${OOM_SCORE_ADJ}" -eq "0" ]; then
+    exit 0
+fi
+
+for PID in $(pidof buildkit-runc)
+do
+    echo ${OOM_SCORE_ADJ} > /proc/$PID/oom_score_adj
+done
```
vladaionescu, line 10: Wondering if this script should be moved to the buildkit repo so that it reduces coupling between earthly and buildkit.

No strong opinion though, if that is not easy to do.

buildkitd/entrypoint.sh
dchw: If we want. We dont rm the other template files either though?

buildkitd/oom-adjust.sh.template
dchw: It doesn't seem like it would reduce coupling; more like move it. Our entrypoint script, at that point, would be templating a file that doesn't exist in this repo; rather than buildkit calling a file that doesn't exist in its repo.

We could take some extra steps and do it "more right", where we have a script wrapper around buildkit that templates this script, and we invoke that wrapper instead of buildkit? But is that wrapping too much?

dchw: Vlad, I think you are right; now that I think deeper. This script template should move over there; and I should include a `COPY` to bring it in in our buildkit Earthfile (to avoid having merges on the Dockerfile).

I'll also add a check for if the variable is set to this script, so regular builds with the hook will still work too.

vladaionescu: Another way to reduce coupling might be to make the hook configurable via an env var, and have that env var + this script (which would be referenced by that env var) be passed in our entrypoint.

dchw: I could make the hooks configurable via buildkit config... is it worth carrying that config tweak burden?

vladaionescu: I think so. It would make buildkit more self-contained.

dchw: I can dig in. Its turning into far more than the 5 lines I wanted :D

buildkitd/entrypoint.sh
dchw: Doesn't matter anymore, hooks (when we want them) are config options, and the new config value obviates this script.

Still willing to `rm` the other template pieces in here if you want.

buildkitd/oom-adjust.sh.template
dchw: Ended up with a full config-based implementation of hooks. Ill squash and force-push this commit with those details.

buildkitd/entrypoint.sh
dchw: So I did some real world testing here. I cannot remove any templates if we want to allow the container to restart; otherwise it fails because the files are missing when it attempts to reconstruct any templated file.

alexcb: that really surprises me -- I assumed a restart would also create a new instance of the filesystem, guess we'll need to keep them around. thanks for taking a look.

buildkitd/Earthfile
```diff
@@ -12,7 +12,7 @@ buildkitd:
             ARG BUILDKIT_BASE_IMAGE=$BUILDKIT_PROJECT+build
         END
     ELSE
-        ARG BUILDKIT_BASE_IMAGE=github.com/earthly/buildkit:9c256c19c5397aee1ba0379590d685c2e87ff17c+build
+        ARG BUILDKIT_BASE_IMAGE=github.com/earthly/buildkit:53d3f36ec9cd31894bd16fa2145b9436031c3b50+build
```
alexcb, line 15: can you coordinate with @r-LaForge regarding #2285

I suspect you're buidkit PR will need to be rebased against the ongoing buildkit-fork-update PR.

dchw: I just did a rebase over there; I thought it was merged already :( I can wait, its no problem

buildkitd/entrypoint.sh
dchw: I do wish restarting would remove the backing volume, but it doesnt :(

buildkitd/entrypoint.sh
```diff
@@ -248,5 +260,16 @@ do
...
         exit "$code"
     fi
+
+    for PID in $(pgrep -P 1)
+    do
+        # Sometimes, child processes can be reparented to the init (this script). One
+        # common instance is when something is OOM killed, for instance. This enumerates
+        # all those PIDs, and kills them to prevent accidential "ghost" loads.
+        if [ "$PID" != "$execpid" ]; then
+            ! ${BUILDKIT_DEBUG} || echo "$(date) | $PID($(cat /proc/$PID/cmdline)) killed" >> /var/log/oom_adj
```
vladaionescu, line 270: Nit: Style.

```suggestion
            ! "${BUILDKIT_DEBUG}" || echo "$(date) | $PID($(cat /proc/$PID/cmdline)) killed" >> /var/log/oom_adj
```

buildkitd/oom-adjust.sh.template
```diff
@@ -0,0 +1,29 @@
...
+    for CHILD_PID in $(pgrep -P $PID)
+    do
+        ! ${BUILDKIT_DEBUG} || echo "$(date) |   $PID has child: $CHILD_PID($(cat /proc/$CHILD_PID/cmdline)) with oom_score_adj: $(/proc/$CHILD_PID/oom_score_adj)" >> /var/log/oom_adj
+    done
+
+    ! ${BUILDKIT_DEBUG} || echo "$(date) |   Adjusting pid $PID oom_score_adj to ${OOM_SCORE_ADJ}" >> /var/log/oom_adj
+    ! ${BUILDKIT_DEBUG} || echo "${OOM_SCORE_ADJ}" > /proc/$PID/oom_score_adj
+done
+
+! ${BUILDKIT_DEBUG} || echo "$(date) | --" >> /var/log/oom_adj
```
vladaionescu, line 29: I *think* that if we quote `"${BUILDKIT_DEBUG}"` (this shouldn't change the functionality), we can add this template to the shellcheck list in our main Earthfile.

vladaionescu (approved):
Just some minor remarks...

buildkitd/oom-adjust.sh.template
dchw: Ill look and see what shellcheck wants and if it affects anything

dchw: Shellcheck is happy with it without quoting these

buildkitd/entrypoint.sh
dchw: I need to tweak this more, yes, partly a copy paste error

buildkitd/oom-adjust.sh.template
vladaionescu: It's usually a good idea to quote for style mainly - in case it ever becomes a variable that contains spaces. I think shellcheck ignores this rule in some operations.

In any case - all looks good from my point of view!

dchw: Quoted

Earthfile
```diff
@@ -78,11 +78,11 @@ lint-scripts-base:
 lint-scripts-misc:
     FROM +lint-scripts-base
     COPY ./earthly ./scripts/install-all-versions.sh ./buildkitd/entrypoint.sh ./earthly-entrypoint.sh \
-        ./buildkitd/dockerd-wrapper.sh ./buildkitd/docker-auto-install.sh \
+        ./buildkitd/dockerd-wrapper.sh ./buildkitd/docker-auto-install.sh ./buildkitd/oom-adjust.sh.template \
         ./release/envcredhelper.sh ./.buildkite/*.sh \
         ./scripts/tests/*.sh \
         ./scripts/*.sh \
-        ./shell_scripts/
+        ./shell_scripts/
```
alexcb, line 85: nit

dchw: dangit

outmon/vertexmon.go
```diff
@@ -190,11 +192,26 @@ func (vm *vertexMonitor) printError() bool {
...
+		// Ignore the Error, default case will print it as a string using the source, so we won't miss any data.
+		exitCode, _ := strconv.ParseUint(m[2], 10, 32)
+
+		switch exitCode {
+		case 137:
+		case math.MaxUint32:
+			errString = fmt.Sprintf(""+
+				"      The%s command\n"+
+				"          %s\n"+
+				"      was terminated because the satellite ran out of memory.",
```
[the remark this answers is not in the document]
dchw: I had it that way originally; the thought was this probably wont be used locally? It could though.

@brandonSc @vladaionescu Y'all had opinions on wording. Im not worried about it; what is the consensus here?

alexcb (approved):
looks good once we figure out the wording of the OOM error message.

outmon/vertexmon.go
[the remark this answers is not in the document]
dchw: `was terminated because of an out of memory error. If you are using a satellite or other remote buildkit, it is the remote system that ran out of memory.`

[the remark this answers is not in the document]
dchw: Updated message, if you are good with it @alexcb let me know and I will merge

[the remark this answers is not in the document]
alexcb: looks good!

dchw:
Merging after CI run passes for pointing at the freshly merged SHA in buildkit.

outmon/vertexmon.go
```diff
@@ -190,11 +192,27 @@ func (vm *vertexMonitor) printError() bool {
...
-			"          %s\n"+
-			"      did not complete successfully. Exit code %s",
-			internalStr, indentOp, m[2])
+
+		// Ignore the Error, default case will print it as a string using the source, so we won't miss any data.
+		exitCode, _ := strconv.ParseUint(m[2], 10, 32)
+
+		switch exitCode {
+		case 137:
+		case math.MaxUint32:
```
vladaionescu, line 201: I just noticed that this should be `case 137, mathUint32:` (otherwise it does nothing on 137).

PR #2297, opened 2022-10-20
Some Improvements to Support Satellite Sleep

brandonSc:
More states are represented during Satellite reserve process, and the `inspect` command will only show full  information when the satellite is awake.

Example awake:

Screen Shot 2022-10-20 at 2 43 07 PM

Example asleep:

Screen Shot 2022-10-20 at 2 43 25 PM


cmd/earthly/satellite_cmds.go
```diff
@@ -296,17 +296,23 @@ func (app *earthlyApp) actionSatelliteInspect(cliCtx *cli.Context) error {
...
 	app.console.Printf("Currently selected: %s", selected)
+	app.console.Printf("")
+
+	if satellite.Status == cloud.SatelliteStatusOperational {
+		err = buildkitd.PrintSatelliteInfo(cliCtx.Context, app.console, Version, app.buildkitdSettings)
+		if err != nil {
+			return errors.Wrap(err, "failed checking buildkit info")
+		}
+	} else {
+		app.console.Printf("More info available when Satellite is awake.")
```
brandonSc, line 314: In a follow up PR, I will add a command for `earthly satellite wake <my-sat>` , but I want to update that endpoint on the backend to use gRPC streams first.

brandonSc: It's also worth mentioning that the state displayed above might now actually say "Asleep" in all cases. 

It could be something else, such as "Stopping", "Starting", or "Destroying" 

however, I think saying "...when Satellite is awake" is still clear enough.

alexcb (approved):

vladaionescu (approved):

PR #2300, opened 2022-10-21
Support for leading Docker `ARG`s in `docker2earthly`

rrjjvv:
Given a trivial (but valid) Dockerfile:
```Dockerfile
ARG img
FROM $img
```
the conversion results in an unusable Earthfile:
```
$ earthly docker2earthly --tag foo --earthfile - 2>/dev/null

# This Earthfile was generated using docker2earthly
# the conversion is done on a best-effort basis
# and might not follow best practices, please
# visit http://docs.earthly.dev for Earthfile guides
subbuild1:
    FROM $img
    SAVE IMAGE foo

build:
    BUILD +subbuild1
```

This change results in a file that's buildable (and has the equivalent docker behavior):
```
$ earthly-dev docker2earthly --tag foo --earthfile - 2>/dev/null
VERSION 0.6

# This Earthfile was generated using docker2earthly
# the conversion is done on a best-effort basis
# and might not follow best practices, please
# visit https://docs.earthly.dev for Earthfile guides
subbuild1:
    ARG img
    FROM $img
    SAVE IMAGE foo

build:
    BUILD +subbuild1
```

### DIFF
--- a/docker2earthly/convert.go
+++ b/docker2earthly/convert.go
@@ -14,6 +14,10 @@ import (
 	"github.com/pkg/errors"
 )
 
+// Ideally this would point to "the current version" rather than being hard-coded, but the single
+// "source of truth" (in ast/validator) isn't currently exported.
+const earthlyCurrentVersion = "0.6"
+
 func getArtifactName(s string) string {
 	split := strings.Split(s, "/")
 	n := len(split)
@@ -41,10 +45,11 @@ func Docker2Earthly(dockerfilePath, earthfilePath, imageTag string) error {
 
 	targets := [][]string{
 		{
+			fmt.Sprintf("VERSION %s\n", earthlyCurrentVersion),
 			"# This Earthfile was generated using docker2earthly",
 			"# the conversion is done on a best-effort basis",
 			"# and might not follow best practices, please",
-			"# visit http://docs.earthly.dev for Earthfile guides",
+			"# visit https://docs.earthly.dev for Earthfile guides",
 		},
 	}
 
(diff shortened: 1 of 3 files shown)


docker2earthly/convert.go
```diff
@@ -64,6 +69,15 @@ func Docker2Earthly(dockerfilePath, earthfilePath, imageTag string) error {
 		targets = append(targets, []string{
 			fmt.Sprintf("FROM %s", stage.BaseName),
 		})
+		// These args are in scope *only* for the very first FROM
+		if i == 0 && len(initialArgs) > 0 {
+			var fromArgs []string
+			for _, arg := range initialArgs {
+				fromArgs = append(fromArgs, fmt.Sprintf("%s", arg.String()))
+			}
+			targets[i+1] = append(fromArgs, targets[i+1]...)
```
alexcb, line 78: this assumption (`i+1`) feels brittle; would it be possible to do something more along the lines of:

```
for i, stage := range stages {
    fromCmd := []string{fmt.Sprintf("FROM %s", stage.BaseName)
    // These args are in scope *only* for the very first FROM
    if i == 0 && len(initialArgs) > 0 {
        var fromArgs []string
            for _, arg := range initialArgs {
                fromArgs = append(fromArgs, fmt.Sprintf("%s", arg.String()))
            }
        fromCmd = append(fromArgs, fromCmd...)
    } 
    targets = append(targets, fromCmd)
}
```

alexcb (commented):
thanks for the PR! Just one suggestion.

docker2earthly/convert.go
rrjjvv: Absolutely (and I agree).  I had reasons for going that way, but I should have followed my instincts on this one!

I'll fix the lint issue at the same time.  I assumed `+test` would always be sufficient to guard against embarrassing failures like that.  I assume not including `+lint` inside `+test` is intentional; would it worth throwing a quick snippet in the contributors file (to advise explicitly running the linter prior to pushing)?

Issue #2301, opened 2022-10-21
User-provided daemon settings (d-in-d) not merged properly with earthly settings


rrjjvv:
This is around undocumented/special-case functionality, so this probably more of a feature request than a bug report.  However, a recent change, combined with the native behavior of `jq`, has resulted in the inability to utilize `WITH DOCKER` with insecure registries.

There's a long history behind this, but in #1780, it was decided that user-defined daemon configs should be respected (within reason).  One of the main use-cases I had for wanting this was the ability to interact with (local) insecure registries (for testing/dev purposes).  At one point I even utilized it in a bug report (https://github.com/earthly/earthly/issues/2072#issuecomment-1205937735).

I tried employing that functionality today, but that specific use-case no longer works.  The cause was d89a55aacb3ef6043a12436e8690a6f53ec29e7d, which added an explicit setting for insecure registries (the internal buildkit registry).  Since the value is an array, the RHS wins (and overwrites).  So in this scenario, providing an insecure registry results in `--load` (and probably `--pull`, but I didn't test) not working (i.e., `Error response from daemon: Get "https://172.30.0.1:8371/v2/": http: server gave HTTP response to HTTPS client`).  And obviously, if I don't provide a value, the buildkit registry works, but I can't interact with any other insecure registries.

What I'm hoping for is either a change in the config merging behavior (for a general solution), or explicit support for insecure registries (even if it remains an undocumented feature).

PR #2302, opened 2022-10-21
Improve merging of user daemon config with earthly-required config

rrjjvv:
Potential fix for #2301 

The addition of two objects (current implementation) will merge them in ways that are mostly desirable and intuitive.  However, arrays are never concatenated; the RHS always wins.  This will now 'merge' arrays as well.

I'm starting this as a draft since I'm not confident that it will be (or should be) accepted as-is.

Potential questions/issues:
1. My PR prioritizes earthly changes over user changes (which *seems* like a good thing, but is it?)
2. This change fixes my immediate problem (buildkit registry works alongside arbitrary insecure registries), but will impact the behavior of any current (e.g., `default-address-pools`) array values, as well as those in the future
3. The algorithm I found makes sense after reading, but I'm non proficient enough to immediately know whether there's risk of overflow or other issues
4. Rather than try to accommodate everyone "in theory", would it make more sense to support individual overrides (even if undocumented)?  (Appending "my" list of insecure registries to "yours" make sense, but maybe some appends don't make sense or could cause harm in the future.)
5. I'm 'pretty sure' that this function doesn't utilize any bleeding-edge `jq` features, but without further investigation, I can't confirm that for a fact.

This change does work (for my desired use-case, as well as not breaking standard usage).  The question is... is that "good enough"?

### DIFF
--- a/buildkitd/dockerd-wrapper.sh
+++ b/buildkitd/dockerd-wrapper.sh
@@ -83,8 +83,24 @@ start_dockerd() {
         echo >/etc/docker/daemon.json '{}'
     fi
 
+    # compliments of https://stackoverflow.com/a/53666584
+    # this will concatenate arrays found in both the LHS and RHS; default jq will overwrite the LHS with the RHS
+    cat <<'EOF' > /tmp/meld.jq
+def meld(a; b):
+  a as $a | b as $b
+  | if ($a|type) == "object" and ($b|type) == "object"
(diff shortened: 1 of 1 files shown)


rrjjvv:
I think I made this mistake before; using "draft" to signify it might warrant discussion, rather than POC/rough draft/etc..  But it was/is "complete".

PR #2309, opened 2022-10-24
Send locally configured Git email address on Satellite reserve

mikejholly:
This PR adds the user's Git email address to Satellite reserve calls to help with the tracking of active users.


util/gitutil/detectgit.go
```diff
@@ -302,6 +302,17 @@ func detectGitAuthor(ctx context.Context, dir string) (string, error) {
 	return strings.SplitN(outStr, "\n", 2)[0], nil
 }
 
+// GlobalEmail returns the user's currently configured (global) email address
+func GlobalEmail(ctx context.Context) (string, error) {
+	cmd := exec.CommandContext(ctx, "git", "config", "--global", "user.email")
```
vladaionescu, line 307: I would use `git config --get user.email` rather than global. If a user happens to use a non-global email for certain projects, it's better to get the email specific email.

vladaionescu (approved):
Minor remark...

util/gitutil/detectgit.go
```diff
@@ -302,6 +302,17 @@ func detectGitAuthor(ctx context.Context, dir string) (string, error) {
 	return strings.SplitN(outStr, "\n", 2)[0], nil
 }
 
+// GlobalEmail returns the user's currently configured (global) email address
+func GlobalEmail(ctx context.Context) (string, error) {
```
r-LaForge, line 306: This is wrong now right? It's not really the global email

mikejholly: Good point. Thanks.

util/gitutil/detectgit.go
```diff
@@ -302,6 +302,17 @@ func detectGitAuthor(ctx context.Context, dir string) (string, error) {
 	return strings.SplitN(outStr, "\n", 2)[0], nil
 }
 
+// ConfigEmail returns the user's currently configured (global) email address
+func ConfigEmail(ctx context.Context) (string, error) {
+	cmd := exec.CommandContext(ctx, "git", "config", "--get", "user.email")
+	cmd.Stderr = nil // force capture of stderr on errors
+	out, err := cmd.Output()
+	if err != nil {
+		return "", errors.Wrap(err, "detect git global email")
```
alexcb, line 311: leftover `global`

PR #2310, opened 2022-10-24
Satellite Wake-up Command

brandonSc:
The satellite wake command will start a satellite that is asleep. This can be useful for using the `inspect` command, since it will not show much info when the satellite is asleep. 

The wake command will also use more "force" in trying to start a satellite than a normal Reserve call will do, so it can be usef to debug or attempt to recover a satellite that has failed or ended up in a bad state for some reason.

Examples below 👇 

Satellite wake command, when satellite is asleep:
Screenshot 2022-10-24 164539

Satellite wake when already awake:
Screenshot 2022-10-24 164607

`inspect` prompts user to run `wake` if the satellite is asleep:
Screenshot 2022-10-24 165652

### DIFF
--- a/cloud/satellite.go
+++ b/cloud/satellite.go
@@ -117,6 +117,7 @@ func (c *client) ReserveSatellite(ctx context.Context, name, orgID, gitAuthor, g
 		})
 		if err != nil {
 			out <- SatelliteStatusUpdate{Err: errors.Wrap(err, "failed opening satellite reserve stream")}
+			return
 		}
 		var update *pb.ReserveSatelliteResponse
 		for {
@@ -126,10 +127,12 @@ func (c *client) ReserveSatellite(ctx context.Context, name, orgID, gitAuthor, g
 			}
 			if err != nil {
 				out <- SatelliteStatusUpdate{Err: errors.Wrap(err, "failed receiving satellite reserve update")}
+				return
 			}
 			status := satelliteStatus(update.Status)
 			if status == SatelliteStatusFailed {
 				out <- SatelliteStatusUpdate{Err: errors.New("satellite is in a failed state")}
+				return
 			}
 			out <- SatelliteStatusUpdate{State: satelliteStatus(update.Status)}
 		}
@@ -149,6 +152,7 @@ func (c *client) WakeSatellite(ctx context.Context, name, orgID string) (out cha
 		})
 		if err != nil {
 			out <- SatelliteStatusUpdate{Err: errors.Wrap(err, "failed opening satellite wake stream")}
+			return
 		}
 		var update *pb.WakeSatelliteResponse
 		for {
@@ -158,10 +162,12 @@ func (c *client) WakeSatellite(ctx context.Context, name, orgID string) (out cha
 			}
 			if err != nil {
 				out <- SatelliteStatusUpdate{Err: errors.Wrap(err, "failed receiving satellite wake update")}
+				return
 			}
 			status := satelliteStatus(update.Status)
 			if status == SatelliteStatusFailed {
 				out <- SatelliteStatusUpdate{Err: errors.New("satellite is in a failed state")}
+				return
 			}
 			out <- SatelliteStatusUpdate{State: satelliteStatus(update.Status)}
 		}
--- a/cloud/satellite.go
+++ b/cloud/satellite.go
@@ -97,62 +97,75 @@ func (c *client) LaunchSatellite(ctx context.Context, name, orgID string, featur
 	return nil
 }
 
-func (c *client) ReserveSatellite(ctx context.Context, name, orgID, gitAuthor string, isCI bool, out chan<- string) error {
-	defer close(out)
-	ctxTimeout, cancel := context.WithTimeout(ctx, 5*time.Minute)
-	defer cancel()
-	stream, err := c.compute.ReserveSatellite(c.withAuth(ctxTimeout), &pb.ReserveSatelliteRequest{
-		OrgId:       orgID,
-		Name:        name,
-		CommitEmail: gitAuthor,
-		IsCi:        isCI,
-	})
-	if err != nil {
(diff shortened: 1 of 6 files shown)


cmd/earthly/satellite_cmds.go
```diff
@@ -387,3 +398,124 @@ func (app *earthlyApp) actionSatelliteUnselect(cliCtx *cli.Context) error {
...
+	}
+
+	orgID, err := app.getSatelliteOrgID(cliCtx.Context, cloudClient)
+	if err != nil {
+		return err
+	}
+
+	out := make(chan string)
+	var wakeErr error
+	go func() { wakeErr = cloudClient.WakeSatellite(cliCtx.Context, app.satelliteName, orgID, out) }()
```
vladaionescu, line 423: Given the async nature of this operation, I would maybe make the function `WakeSatellite` do the goroutine spawning within it, and instead have it return the chan (instead of accepting it as a param).

It makes the code using the `WakeSatellite` function simpler.

It also makes the lifecycle of the channel self-contained in a single location (creation, feeding in data, and closing).

brandonSc: Yeah that seems like a nice way to do it. I'll update the `ReserveSatellite` method to do that too

brandonSc: I've updated this

cloud/satellite.go
```diff
@@ -97,35 +97,76 @@ func (c *client) LaunchSatellite(ctx context.Context, name, orgID string, featur
...
+			out <- SatelliteStatusUpdate{Err: errors.Wrap(err, "failed opening satellite reserve stream")}
+		}
+		var update *pb.ReserveSatelliteResponse
+		for {
+			update, err = stream.Recv()
+			if err == io.EOF {
+				return
+			}
+			if err != nil {
+				out <- SatelliteStatusUpdate{Err: errors.Wrap(err, "failed receiving satellite reserve update")}
```
alexcb, line 128: should there be a `return`, or `continue` here?

cloud/satellite.go
```diff
@@ -97,35 +97,76 @@ func (c *client) LaunchSatellite(ctx context.Context, name, orgID string, featur
...
+			}
+			status := satelliteStatus(update.Status)
+			if status == SatelliteStatusFailed {
+				out <- SatelliteStatusUpdate{Err: errors.New("satellite is in a failed state")}
```
alexcb, line 132: should there be a `return`, or `continue` here?

cloud/satellite.go
```diff
@@ -97,35 +97,76 @@ func (c *client) LaunchSatellite(ctx context.Context, name, orgID string, featur
...
+		ctxTimeout, cancel := context.WithTimeout(ctx, 5*time.Minute)
+		defer cancel()
+		defer close(out)
+		stream, err := c.compute.WakeSatellite(c.withAuth(ctxTimeout), &pb.WakeSatelliteRequest{
+			OrgId: orgID,
+			Name:  name,
+		})
 		if err != nil {
-			return errors.Wrap(err, "failed receiving satellite reserve update")
+			out <- SatelliteStatusUpdate{Err: errors.Wrap(err, "failed opening satellite wake stream")}
```
alexcb, line 151: should there be a `return`, or `continue` here?

cloud/satellite.go
```diff
@@ -97,35 +97,76 @@ func (c *client) LaunchSatellite(ctx context.Context, name, orgID string, featur
...
-		if status == SatelliteStatusFailed {
-			return errors.New("satellite is in failed state")
+		var update *pb.WakeSatelliteResponse
+		for {
+			update, err = stream.Recv()
+			if err == io.EOF {
+				return
+			}
+			if err != nil {
+				out <- SatelliteStatusUpdate{Err: errors.Wrap(err, "failed receiving satellite wake update")}
```
alexcb, line 160: should there be a `return`, or `continue` here?

cloud/satellite.go
```diff
@@ -97,35 +97,76 @@ func (c *client) LaunchSatellite(ctx context.Context, name, orgID string, featur
...
+			}
+			status := satelliteStatus(update.Status)
+			if status == SatelliteStatusFailed {
+				out <- SatelliteStatusUpdate{Err: errors.New("satellite is in a failed state")}
```
alexcb, line 164: should there be a `return`, or `continue` here?

cmd/earthly/satellite_cmds.go
```diff
@@ -387,3 +399,132 @@ func (app *earthlyApp) actionSatelliteUnselect(cliCtx *cli.Context) error {
...
+		switch o.State {
+		case cloud.SatelliteStatusSleep:
+			if !loggedSleep {
+				console.Printf("%s is waking up. Please wait...", satName)
+				loggedSleep = true
+				shouldLogLoading = false
+			}
+		case cloud.SatelliteStatusStopping:
+			if !loggedStop {
+				console.Printf("%s is falling asleep. Please wait...", satName)
```
alexcb, line 462: this message might be confusing to a user -- they want the satellite to wake up, yet the message is saying it's waiting for it to sleep. Maybe it should be pedantic to fully describe what is happening: `%s is falling asleep (once sleep event has completed) it will wake up. Please wait...`

alexcb (commented):
looks good, just a question about some missing returns.

brandonSc:
> looks good, just a question about some missing returns.

Yes, thanks, I dropped them by mistake via refactoring.

cmd/earthly/satellite_cmds.go
brandonSc: I updated this to say `%s is currently falling asleep. Waiting to send wake up signal...`  - I think it gets the right point across while still (somewhat) concise.

vladaionescu (approved):

PR #2320, opened 2022-10-27
rename Load to current utilization

alexcb:
This is to avoid load being misinterpreted as "loading data".

### DIFF
--- a/buildkitd/buildkitd.go
+++ b/buildkitd/buildkitd.go
@@ -22,6 +22,7 @@ import (
 	"google.golang.org/grpc/status"
 
 	"github.com/earthly/earthly/conslogging"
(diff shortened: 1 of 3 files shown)


vladaionescu (commented):
Can you make this consistent across - e.g. https://github.com/earthly/earthly/blob/main/buildkitd/buildkitd.go#L863

alexcb:
How does this look?

I wasn't sure about renaming `Buildkit is currently under significant load` to `Buildkit is currently under significant utilization` though.

vladaionescu:
See also the link I provided above

alexcb:
> See also the link I provided above

Sorry, misread the request. I pulled that string out into a common util since it's referenced across multiple packages now.

vladaionescu (approved):

PR #2323, opened 2022-10-27
Patch frontend tests

r-LaForge:
The frontend tests had a few issues:

- the returned error was being ignored.
- The container was started with an invalid command meaning they weren't up for long.
- We didn't wait for the containers to actually be running before trying to test them.
- We only checked stderr had logs, but now we check both for expected messages.
- We weren't using `t.Cleanup`, so we weren't necessarily waiting for containers to be cleaned up.

### DIFF
--- a/util/containerutil/frontend_test.go
+++ b/util/containerutil/frontend_test.go
@@ -8,14 +8,15 @@ import (
 	"os"
 	"os/exec"
 	"strings"
+	"sync"
 	"testing"
-
-	"github.com/hashicorp/go-multierror"
-	"github.com/pkg/errors"
-	"github.com/stretchr/testify/assert"
+	"time"
 
 	"github.com/earthly/earthly/conslogging"
 	"github.com/earthly/earthly/util/containerutil"
+	"github.com/hashicorp/go-multierror"
+	"github.com/pkg/errors"
+	"github.com/stretchr/testify/assert"
 )
 
 func TestFrontendNew(t *testing.T) {
@@ -121,8 +122,8 @@ func TestFrontendContainerInfo(t *testing.T) {
 
 			testContainers := []string{"test-1", "test-2"}
 			cleanup, err := spawnTestContainers(ctx, tC.binary, testContainers...)
+			t.Cleanup(cleanup)
 			assert.NoError(t, err)
-			defer cleanup()
 
 			fe, err := tC.newFunc(ctx, &containerutil.FrontendConfig{Console: testLogger()})
 			assert.NoError(t, err)
@@ -161,8 +162,8 @@ func TestFrontendContainerRemove(t *testing.T) {
 
 			testContainers := []string{"remove-1", "remove-2"}
 			cleanup, err := spawnTestContainers(ctx, tC.binary, testContainers...)
+			t.Cleanup(cleanup)
 			assert.NoError(t, err)
-			defer cleanup()
 
 			fe, err := tC.newFunc(ctx, &containerutil.FrontendConfig{Console: testLogger()})
 			assert.NoError(t, err)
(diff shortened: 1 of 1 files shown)


util/containerutil/frontend_test.go
```diff
@@ -592,23 +594,96 @@ func isBinaryInstalled(ctx context.Context, binary string) bool {
 	return cmd.Run() == nil
 }
 
+// First, we attempt to stop the containers that may be running.
+// Next, we start the containers
+// After that, we attempt to wait for the containers for up to 20 seconds
+// Then we return
+// Caller is always expected to call cleanup
 func spawnTestContainers(ctx context.Context, feBinary string, names ...string) (func(), error) {
+	removeContainers(ctx, feBinary, names...) // best effort
```
alexcb, line 603:
```suggestion
	_ = removeContainers(ctx, feBinary, names...) // best effort
```

util/containerutil/frontend_test.go
```diff
@@ -592,23 +594,96 @@ func isBinaryInstalled(ctx context.Context, binary string) bool {
...
+	err := startTestContainers(ctx, feBinary, names...)
+	cleanup := func() {
+		removeContainers(ctx, feBinary, names...) // best-effort
```
alexcb, line 606:
```suggestion
		_ = removeContainers(ctx, feBinary, names...) // best-effort
```

util/containerutil/frontend_test.go
```diff
@@ -592,23 +594,96 @@ func isBinaryInstalled(ctx context.Context, binary string) bool {
...
+		go func(name string) {
+			defer wg.Done()
+			attempts := 0
+			for attempts < 100 {
+				attempts++
+				// docker inspect -f {{.State.Running}} CONTAINERNAME`"=="true"
+				cmd := exec.CommandContext(ctx, feBinary, "inspect", "-f", "{{.State.Running}}", name)
+				output, _ := cmd.CombinedOutput()
+				if strings.Contains(string(output), "true") {
+					break
```
alexcb, line 674: might be best to just return here instead,

then if you ever get to the end of the `for` loop, you'll know there's an error.

this would make things more robust since the for loop references `100` attempts, but the error checking below references `1000`.

otherwise, it's probably better to use a maxAttempts const.

r-LaForge: Good catch! I missed that in the refactor.
I'm using a return now and a maxAttempts const.

util/containerutil/frontend_test.go
```diff
@@ -592,23 +594,95 @@ func isBinaryInstalled(ctx context.Context, binary string) bool {
...
+		const maxAttempts = 100
+		wg.Add(1)
+		go func(name string) {
+			defer wg.Done()
+			attempts := 0
+			for attempts < maxAttempts {
+				attempts++
+				// docker inspect -f {{.State.Running}} CONTAINERNAME`"=="true"
+				cmd := exec.CommandContext(ctx, feBinary, "inspect", "-f", "{{.State.Running}}", name)
+				output, _ := cmd.CombinedOutput()
```
alexcb, line 673: nit: can we check the error here too?

here's an edge-case:

```
docker inspect -f {{.State.Running}} true:dat
```

this fails with

```

Error: No such object: true:dat
```

r-LaForge: done

util/containerutil/frontend_test.go
```diff
@@ -592,23 +594,120 @@ func isBinaryInstalled(ctx context.Context, binary string) bool {
...
+		go func(name string) {
+			defer wg.Done()
+			attempts := 0
+			for attempts < maxAttempts {
+				attempts++
+				// docker inspect -f {{.State.Running}} CONTAINERNAME`"=="true"
+				cmd := exec.CommandContext(ctx, feBinary, "inspect", "-f", "{{.State.Running}}", name)
+				output, inspectErr := cmd.CombinedOutput()
+				if inspectErr != nil {
+					err = multierror.Append(err, inspectErr)
```
alexcb, line 696: just noticed a scoping-bug here, `var err error` is shared between all threads in this wait group.

it should be moved to be under:

```
go func(name string) {
    defer wg.Done()
    var err error
```

This bug exists in a couple places in this PR.

alexcb: or maybe it's just missing a `m.Lock()`

r-LaForge: The scope was intentional, but It was missing the `m.Lock` for this appending.
From what I can tell the other places have the correct locking.

alexcb (approved):

PR #2335, opened 2022-11-01
Build ARM and x64 AMIs, move to Packer

dchw:
This makes it so we build an ARM and x64 Earthly AMI on release. I moved from AWS ImageBuilder to this, because the build matrix was a little weird keeping it in Terraform the way it was.

I tested the target called by `release.sh`, but not the full release process.


release/ami/configure.sh
```diff
@@ -0,0 +1,12 @@
...
+
+set -xe
+
+sudo systemctl start docker.service
+sudo systemctl start containerd.service
+
+sudo systemctl enable docker.service
+sudo systemctl enable containerd.service
+
+docker run --rm --privileged aptman/qus -s -- --r -p
```
dchw, line 11: This image has all the same static binaries, its just also a multi-arch one that I can install on arm and x64.

vladaionescu (approved):

release/ami/configure.sh
dchw: It is also well-documented: https://dbhi.github.io/qus/intro.html

alexcb (approved):

PR #2342, opened 2022-11-02
Satellite sleep command

brandonSc:
Added a new command that manually puts a satellite to sleep. When the satellite is already asleep, this is a no-op. 

This command is very useful for us internally for testing, but also could be of use to users who want more fine-grain control over their satellite's up-time. 

Note this will put a satellite to sleep, even if the user has manually disabled the auto-sleep feature. Hence a user could opt out of the auto-sleep feature and completely manage the satellite's state themselves, if they are so inclined. 

Screen Shot 2022-11-02 at 1 32 26 PM

### DIFF
--- a/cmd/earthly/satellite_cmds.go
+++ b/cmd/earthly/satellite_cmds.go
@@ -172,8 +172,11 @@ func (app *earthlyApp) getSatelliteOrgID(ctx context.Context, cloudClient cloud.
 func (app *earthlyApp) actionSatelliteLaunch(cliCtx *cli.Context) error {
 	app.commandName = "satelliteLaunch"
 
-	if cliCtx.NArg() != 1 {
+	if cliCtx.NArg() == 0 {
 		return errors.New("satellite name is required")
+	}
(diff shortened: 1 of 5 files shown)


cmd/earthly/satellite_cmds.go
```diff
@@ -437,6 +446,34 @@ func (app *earthlyApp) actionSatelliteWake(cliCtx *cli.Context) error {
 	return nil
 }
 
+func (app *earthlyApp) actionSatelliteSleep(cliCtx *cli.Context) error {
+	app.commandName = "satelliteSleep"
+
+	if cliCtx.NArg() != 1 {
+		return errors.New("satellite name is required")
```
alexcb, line 453: nit: break this up into two cases: 

```
if cliCtx.NArg() == 0 {
		return errors.New("satellite name is required")
}
if cliCtx.NArg() > 1 {
		return errors.New("only a single satellite name is supported")
}
```

alexcb (approved):

PR #2348, opened 2022-11-03
Handle missing docker or earthly prerelease

jsoref:
I use Rancher Desktop (1.6.1) on an m1 (Monterey 12.6.1).

I don't always have it running, so `docker info` will sometimes exit `1` instead of `0`.

`./earthly` will set `frontend_bin` to `none` when this happens.

On the first run, it will then die trying to run `none ...`.

If one deletes `~/.earthly` without deleting some files from `/tmp`, then future runs of `./earthly` will die:
```sh
+ exec -a earthly /Users/jsoref/.earthly/earthly-prerelease +for-darwin-m1
./earthly: line 156: /Users/jsoref/.earthly/earthly-prerelease: No such file or directory
```

### DIFF
--- a/earthly
+++ b/earthly
@@ -136,7 +136,11 @@ case "$1" in
 
         last_flag_hash=$(cat "$last_flag_hash_path" 2>/dev/null || echo null)
         flagoverride_hash=$(md5sum "$earthly_version_flag_overrides_path" | awk '{print $1}')
-        if [ -z "$COMP_LINE" ]; then
+        if [ ! -x "$bindir/earthly-prerelease" ]; then
+            echo "Could not find an executable '$bindir/earthly-prerelease'."
+            update="true"
+            get_latest_binary_and_update_state
+        elif [ -z "$COMP_LINE" ]; then
             update="false"
             if [ "$last_flag_hash" != "$flagoverride_hash" ]; then
                 echo ".earthly_version_flag_overrides has changed since last run, checking for prerelease binaries" \
--- a/earthly
+++ b/earthly
@@ -103,6 +103,14 @@ do_help() {
     exec -a "$scriptname" "$bindir/earthly-prerelease" --help
 }
 
+get_latest_binary_and_update_state() {
+    get_latest_binary
+    echo "Updated prerelease binary. Version:" >&2
+    "$bindir/earthly-prerelease" --version >&2
+    echo "$now" >"$last_check_state_path"
+    echo "$flagoverride_hash" >"$last_flag_hash_path"
+}
+
 case "$1" in
     reset)
         do_reset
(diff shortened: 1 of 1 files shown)


earthly
```diff
@@ -44,6 +44,10 @@ detect_frontend() {
 frontend_bin=$(detect_frontend)
 
 get_latest_binary() {
+    if [ "$frontend_bin" == "none" ]; then
+        echo "Could not find a frontend, perhaps the daemon isn't running?"
+        false
```
alexcb, line 49: should this be an explicit `exit 1` instead?

Ideally we'll have `set -e` enabled, but relying on that and using `false` to exit seems fragile.

jsoref: `exit 1` would be reasonable, or some other number. I didn't pick `exit 1` because I didn't want to overly prescribe the exit code - that's really a project choice (maybe you prefer 7...).

(In my own code I'd definitely use an explicit exit code for a thing like this)

earthly
```diff
@@ -124,7 +136,11 @@ case "$1" in
 
         last_flag_hash=$(cat "$last_flag_hash_path" 2>/dev/null || echo null)
         flagoverride_hash=$(md5sum "$earthly_version_flag_overrides_path" | awk '{print $1}')
-        if [ -z "$COMP_LINE" ]; then
+        if [ ! -x "$bindir/earthly-prerelease" ]; then
+            echo "Could not find an executable '$bindir/earthly-prerelease'."
```
alexcb, line 140:
```suggestion
            echo "Could not find an executable '$bindir/earthly-prerelease'." >&2
```

alexcb: Could you move this entire check under the 

```
if [ -z "$COMP_LINE" ]; then

...
elif [ "$since" -ge 3600 ]; then
   ...
elif [ ! -x "$bindir/earthly-prerelease" ]; then
   echo "Could not find an executable '$bindir/earthly-prerelease'."
            update="true"
            get_latest_binary_and_update_state

```

block?

I want to avoid any potential updates when resolving auto-completion -- I would rather have earthly fail-fast (and not potentially print anything to stdout)

jsoref: That won't work since the normal case we care about will be `[ -z "$COMP_LINE" ]`.

For autocompletion, we could do something like:
```sh
if [ ! -x "$bindir/earthly-prerelease" ]; then
  if [ -z "$COMP_LINE" ]; then
     echo "Could not find an executable '$bindir/earthly-prerelease'." >&2
              update="true"
              get_latest_binary_and_update_state
    ...
  else
    # handle autocompletion when there's no binary...
  fi
elif [ -z "$COMP_LINE" ]; then

...
elif [ "$since" -ge 3600 ]; then
   ...
```

alexcb: Here's more or less what I was suggesting: #2356

PR #2349, opened 2022-11-03
Switch platform/arch to work on Apple m1

jsoref:
I don't know that the average Mac user knows what `<platform>` would be. But, more importantly, the average new Mac user will not have an `amd64` Mac, so the rest of the command doesn't work.

Alternatively, the text could be `<platform>/<arch>` -- in which case something should really explain both `<platform>` and `<arch>`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -25,7 +25,13 @@ To build Earthly from source for your target system, use
     ./earthly +for-darwin-m1
     ```
 
-This builds the earthly binary in `./build/<platform>/amd64/earthly` and also the buildkitd image.
+This builds the earthly binary in `./build/*/*/earthly`, typically one of:
+
+* `./build/linux/amd64/earthly`
+* `./build/darwin/amd64/earthly`
+* `./build/darwin/arm64/earthly`
+
+It also builds the buildkitd image.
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -25,7 +25,7 @@ To build Earthly from source for your target system, use
     ./earthly +for-darwin-m1
     ```
 
-This builds the earthly binary in `./build/<platform>/amd64/earthly` and also the buildkitd image.
+This builds the earthly binary in `./build/$(uname|tr A-Z a-z)/$(uname -m)/earthly` and also the buildkitd image.
```
vladaionescu, line 28: This isn't entirely accurate, as `uname -m` can vary a lot: `X86_64` vs `amd64`, or `arm64` vs `aarch64`.

Maybe just list out the three common options?

```
* ./build/linux/amd64/earthly
* ./build/darwin/amd64/earthly
* ./build/darwin/arm64/earthly
```

jsoref: I can't think of a remotely good way to write this. I'd almost argue in favor of:

```suggestion
This builds the earthly binary in `./build/*/*/earthly` and also the buildkitd image.
```

I'm assuming that on average you only have a single flavor.

vladaionescu: Hah! This works surprisingly well :-)

I think it makes sense! I would still list those three popular possibilities the first time we say `./build/*/*/earthly`, but then just use `./build/*/*/earthly` in the rest of the doc.

vladaionescu (approved):

PR #2352, opened 2022-11-03
Buildkit graceful shutdown on SIGTERM

brandonSc:
Uses a `trap` to relay the SIGTERM to buildkit in the entrypoint script.

Screen Shot 2022-11-03 at 12 00 20 PM

### DIFF
--- a/buildkitd/entrypoint.sh
+++ b/buildkitd/entrypoint.sh
@@ -253,11 +253,11 @@ echo "Detected container architecture is $(uname -m)"
 execpid=$!
 
 stop_buildkit() {
-  echo "SIGTERM received. Shutting down buildkit..."
+  echo "Shutdown signal received. Stopping buildkit..."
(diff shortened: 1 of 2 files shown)


r-LaForge (approved):

buildkitd/entrypoint.sh
```diff
@@ -252,6 +252,13 @@ echo "Detected container architecture is $(uname -m)"
 "$@" &
 execpid=$!
 
+stop_buildkit() {
+  echo "SIGTERM received. Shutting down buildkit..."
+  kill -SIGTERM "$execpid"
+}
+
+trap stop_buildkit SIGTERM
```
vladaionescu, line 260: Maybe also add `SIGQUIT` and `SIGINT` for completion's sake?

```suggestion
trap stop_buildkit SIGTERM SIGQUIT SIGINT
```

vladaionescu (approved):

PR #2356, opened 2022-11-03
Always Check for earthly prerelease

alexcb:
This fixes an edge case where ~/.earthly/earthly-prerelease is deleted by the user, but the script assumes it still exists.


vladaionescu (approved):

r-LaForge (approved):

jsoref:
What happens if you try to tab complete and the binary is missing?

(The current PR does fix the normal run command case.)

alexcb:
> What happens if you try to tab complete and the binary is missing?
> 
> (The current PR does fix the normal run command case.)

That's an edge case we won't worry about. Normal earthly installations will have autocompletion setup during the bootstrap stage.

PR #2364, opened 2022-11-04
Fix Buildkit Race Test

brandonSc:


brandonSc:
Sorry for extra commits, they're from the previous branch that was merged.

alexcb:
> Sorry for extra commits, they're from the previous branch that was merged.

`git push --force-with-lease` would fix this :)

alexcb (approved):

.github/workflows/reusable-misc-tests.yml
```diff
@@ -74,18 +74,21 @@ jobs:
         run: ${{inputs.SUDO}} ${{inputs.BUILT_EARTHLY_PATH}} config global.local_registry_host 'tcp://127.0.0.1:8371'
       - name: Configure Earthly Conversion Parallelism
         run: ${{inputs.SUDO}} ${{inputs.BUILT_EARTHLY_PATH}} config global.conversion_parallelism 5
+      - name: Stop the Buildkit container
+        run: ${{inputs.SUDO}} ${{inputs.BINARY}} stop earthly-buildkitd
```
alexcb, line 78: I think this should be in the same test as below; otherwise it's not captured in the same retry scope.

alexcb (changes requested):
I take back my approval; I think there's a potential issue between stopping the container and running the test.

.github/workflows/reusable-misc-tests.yml
brandonSc: Sure, I just thought it was getting a bit cluttered down there, but retry is a good point. I also wonder if we just won't need the retry, maybe it's stable now?

alexcb: it's helpful for dealing with network flakes.

vladaionescu (approved):

alexcb (approved):

Issue #1736, opened 2022-03-09
`.env` values (potentially secrets) printed to stderr


alexcb:
```
> Earthfile
echo "secretkey=keep-it-secret" > .env
earthly +this-doesnt-exist
```

will output

```
Error: build target: build main: failed to solve: Earthfile line 0:0 target this-doesnt-exist not found
in		+this-doesnt-exist --secretkey=keep-it-secret
```

SimonAdameit:
Earthly will also output the secrets on the command line when being interrupted (Ctrl-C, SIGINT) during the build.

alexcb:
The .env values are being converted to the `stack string` which gets formatted into an error under https://github.com/earthly/earthly/blob/e5d699740bc65fc97ff325c9c27d132210a02a6e/earthfile2llb/interpretererror.go#L54

alexcb:
An initial work-in-progress has been opened in #1801; however, it sounds like we may need to rework how `.env` is used to be able to list build-args seperate from secrets.

If we're forced to change the syntax, we could implement a credential-helper plugin, perhaps supporting a shell-out like: `MY_PASSWORD=$(cat /etc/password.insecure.example | grep user | awk '{print $1}')`?

or even a generic helper that passes the required token to a program, .e.g. `/bin/secret MY_PASSWORD`.

alexcb:
according to https://github.com/earthly/earthly/issues/179#issuecomment-707234920

> The build args only show up in the history if they are declared as such via ARG. But if an arg is passed on the command-line, but is never used in the build with an ARG declaration, then it will not show up.

makes it sound like they should only be displayed when explicitly used by the target.

In the case of referencing a target that doesn't exist, I think it should be fine to simply say:

```
Earthfile line 0:0 target this-doesnt-exist not found
```

SimonAdameit:
I think the credential helper and separation in `.env` of secrets and variables are a bit complex as solutions to this. How about this approach:

All input (e.g. from `.env`) is effectively treated as secret, until it is explicitly referenced as a build arg at which moment it is promoted to non-secret at exactly the place where it is referenced.

WDYT?

alexcb:
> until it is explicitly referenced as a build arg

This sounds like a good idea.

alexcb:
Some other potential fixes from #2336 

- Update the documentation to no longer recommend putting secrets in .env
- Add support for a different file just for secrets (e.g. .secrets)
- No longer allow loading secrets from .env (could break existing builds)

[...]

PR #2365, opened 2022-11-04
Only display declared ARGs in failures

alexcb:
This fixes an issue where all .env values are displayed during failures, which could potentially print a secret value (since secrets can also read from .env).


alexcb:
This fixes #1736

vladaionescu (approved):
I wonder if these could ever still show up?

* https://github.com/earthly/earthly/blob/main/outmon/vertexmon.go#L243
* https://github.com/earthly/earthly/blob/main/outmon/vertexmon.go#L48

alexcb:
> I wonder if these could ever still show up?
> 
> * https://github.com/earthly/earthly/blob/main/outmon/vertexmon.go#L243
> * https://github.com/earthly/earthly/blob/main/outmon/vertexmon.go#L48

After some testing, it looks like this only shows variables that are referenced via an `ARG`, if there's no `ARG` then they don't get marked as overridden.

Issue #2337, opened 2022-11-01
Updates to loaded images don't invalidate cache for RUN commands in WITH DOCKER


JBergles:
I'm not sure if this is a bug or an issue with my understanding of how caching works in earthly, so please bear with me. 

## Earthfile

```
VERSION 0.6

build:
  FROM alpine:latest
  COPY test.txt /test.txt

example:
  FROM earthly/dind:alpine
  WITH DOCKER --load built-image:latest=+build
    RUN docker run built-image:latest cat /test.txt
  END
```

## Steps: 

1. Run `earthly -P +example` for the included Earthfile
2. Change the contents of `test.txt`
3. Run `earthly -P +example` again

## Expected results: 

The `RUN docker run built-image:latest cat /test.txt` command is executed again

## Observed results:

RUN step is reported as cached. If `--no-cache` is used with the RUN step, I can see that the updated container image is being used.

```
 earthly/dind:alpine | --> Load metadata linux/arm64
       alpine:latest | --> Load metadata linux/arm64
             context | --> local context .
              +build | --> FROM alpine:latest
              +build | [██████████] 100% resolve docker.io/library/alpine:latest@sha256:bc41182d7ef5ffc53a40b044e725193bc10142a1243f395ee852a8d9730fc2ad
             ongoing | context (5 seconds ago)
             context | [          ]   0% transferring .:
             context | sent 1 file stat)
             context | [██████████] 100% transferring .:
              +build | --> COPY test.txt /test.txt
              output | --> exporting outputs
              output | [██████████] 100% exporting layers
              output | [██████████] 100% exporting manifest sha256:4dd11cc29e9f8e60fc2965922107430e5140f5dfe78a1428e689bfeebac41314
              output | [██████████] 100% exporting config sha256:58fb16adcd4da2e90b60d06720473c24aa20d31ec12891988519544af0bea009
              output | [          ]   0% transferring docker.io/library/built-image:latest
Warning: Could not detect digest for image built-image:latest. Please update your buildkit installation.
            +example | --> FROM earthly/dind:alpine
            +example | [██████████] 100% resolve docker.io/earthly/dind:alpine@sha256:9b09804c68a2c68195b71e61a09616657b8f0a5ca01b4e029f710c7ca2396dc6
            +example | *cached* --> WITH DOCKER (install deps)
            +example | *cached* --> WITH DOCKER RUN --privileged docker run built-image:latest cat /test.txt
[...]

Issue #2288, opened 2022-10-18
Incorrect `Please update your buildkit installation` warning


vladaionescu:
Example Earthfile

```Earthfile
VERSION 0.6

postgres:
    FROM postgres:11.8
    SAVE IMAGE earthly/compute-postgres:main

up:
    LOCALLY
    WITH DOCKER --load=+postgres
        RUN echo "There might be a 'could not detect digest' message right above ^"
    END
```

https://ci.earthly.dev/logs?logId=acb8cf1f-b25d-44cf-89ca-aad3e08d8880

mitre-tm:
For reference, the `LOCALLY` is not required:
```earthly
VERSION 0.6

up:
  FROM earthly/dind:alpine
  WITH DOCKER --pull=postgres:13.7
    RUN echo "There might be a 'could not detect digest' message right above ^"
  END
```
Running on my local machine:
```sh
$ earthly -v
earthly version v0.6.26 4a8d66e3b30d8ee758e7d71b40bc027f92e9adf4 linux/amd64; Ubuntu 22.04.1 LTS (Jammy Jellyfish)

$ earthly -P +up
 1. Init 🚀
————————————————————————————————————————————————————————————————————————————————

           buildkitd | Found buildkit daemon as docker container (earthly-buildkitd)

 2. Build 🔧
————————————————————————————————————————————————————————————————————————————————

 earthly/dind:alpine | --> Load metadata linux/amd64
       postgres:13.7 | --> Load metadata linux/amd64
       postgres:13.7 | --> DOCKER PULL postgres:13.7
       postgres:13.7 | [██████████] 100% resolve docker.io/library/postgres:13.7@sha256:03652c675ae177af98ddd50f9f4b4b2cf8ad38d0e116aa68fe670fbc2cf250fc
              output | --> exporting outputs
              output | [██████████] 100% exporting layers
              output | [██████████] 100% exporting manifest sha256:8df44890dd12b59862772032c65555cd2dc3b39fc73b724df0a2cda71a83227b
              output | [██████████] 100% exporting config sha256:b3591b962136e747c72a5d0e788551cb6b8163f8c577d97ce933cc86d22ab973
              output | [          ]   0% transferring docker.io/library/postgres:13.7
Warning: Could not detect digest for image postgres:13.7. Please update your buildkit installation.
                 +up | --> FROM earthly/dind:alpine
[...]

PR #2366, opened 2022-11-05
Fix registry caching issue (#2337, #2288)

rrjjvv:
This improves on b35afd1b; images that are pulled/loaded by abbreviated (but common/accepted) names, and then normalized/canonicalized by earthly/buildkit, are not processed correctly.

I was looking to see if #2288 was within my reach to address, and in the process, discovered it's directly related to #2337.

When using `WITH DOCKER`, if an image specified via `--pull` or `--load` doesn't *exactly* match what earthly/buildkit returns (simplest example being adding `:latest` to an unqualified image)
https://github.com/earthly/earthly/blob/696c9e2e618ba4fe70695d6d22fd4de33991b0a3/builder/image_solver.go#L234-L236
then the warning from #2288 shows up
https://github.com/earthly/earthly/blob/696c9e2e618ba4fe70695d6d22fd4de33991b0a3/builder/image_solver.go#L258-L261

While the warning is an annoyance, it is also impactful.  A side effect of the `.ImageDigest` not being set is that `.FinalImageNameWithDigest` is also unset
https://github.com/earthly/earthly/blob/696c9e2e618ba4fe70695d6d22fd4de33991b0a3/builder/image_solver.go#L249-L251
and they are used
https://github.com/earthly/earthly/blob/696c9e2e618ba4fe70695d6d22fd4de33991b0a3/earthfile2llb/withdockerrunreg.go#L173
to intentionally break the cache
https://github.com/earthly/earthly/blob/696c9e2e618ba4fe70695d6d22fd4de33991b0a3/earthfile2llb/withdockerrunbase.go#L139-L141

Thus, if that metadata isn't processed, the `.FinalImageNameWithDigest` is always the empty string, regardless of any structural changes (which leads to #2337).

### DIFF
--- a/builder/image_solver.go
+++ b/builder/image_solver.go
@@ -231,10 +231,14 @@ func (m *multiImageSolver) SolveImages(ctx context.Context, imageDefs []*states.
 			}
 
 			for k, v := range resp {
-				if !strings.HasPrefix(k, result.FinalImageName+"|") {
+				// discarded prefix is likely a registry/repository from its canonical name, i.e.
+				// 1. user asks for 'busybox'
+				// 2. result.FinalImageName is calculated as 'busybox:latest'
+				// 3. BK response has keys 'docker.io/library/busybox:latest|<type>'
+				_, k2, found := strings.Cut(k, result.FinalImageName+"|")
+				if !found {
 					continue
 				}
-				k2 := strings.TrimPrefix(k, result.FinalImageName+"|")
 				switch k2 {
 				case exptypes.ExporterImageDescriptorKey:
 					vdec, err := base64.StdEncoding.DecodeString(v)


rrjjvv:
Though this change is small, I would have liked to provide a test as well (to prove I fixed something, as opposed to simply knowing I simply didn't break anything).  My most recent attempts haven't been successful, which is the reason I went into so much detail in my summary above.

I tried using `+copy-test-verbose-output` as inspiration (since back-to-back executions is needed), to no avail.  If you folks have a better example, or any other hints in general, I'm all ears.  If this is legitimately a difficult/fragile thing to test and my analysis is sufficient, that works as well.

I'll also add that my implementation was not my first choice (despite being the fewest LOC).  The fact that the original code didn't simply `.Split` on `|` made me think the format/layout of that value might change in the future, so I went with a solution that was relatively conservative.

vladaionescu:
Really nice find, Roberto! I built on top of this idea in this PR here: https://github.com/earthly/earthly/pull/2367. I figured it's better to be a bit more precise with that comparison. Many thanks for the fix!

PR #2372, opened 2022-11-07
lint docs

alexcb:


vladaionescu (approved):

vladaionescu:
I spent some time thinking whether this should fold into an existing GHA file. I couldn't think of anything better.

alexcb:
> I spent some time thinking whether this should fold into an existing GHA file. I couldn't think of anything better.

I too was on the fence about somehow making use of the existing reusable workflow, but it seems distinct enough; it's interesting to note that this job doesn't even build earthly from source.

PR #2373, opened 2022-11-07
Add `SAVE IMAGE` warning to Earthfile reference

dchw:
A user brought up that this behavior was not well documented. I added it to our documentation for `SAVE IMAGE` to make it more apparent.

### DIFF
--- a/docs/earthfile/earthfile.md
+++ b/docs/earthfile/earthfile.md
@@ -677,6 +677,14 @@ SAVE IMAGE my-example-registry.com/another-image:latest
 ```
 {% endhint %}
 
+{% hint style='danger' %}
+##### Important
+
+Since the introduction of the `--referenced-save-only` feature flag, Images are only saved if they are connected to the initial target through a chain of `BUILD` commands. [See the `--referenced-save-only` flag documentation for more details](./features.md#referenced-save-only).
+{% endhint %}
+
+
+
 #### Options
 
 ##### `--push`


docs/earthfile/earthfile.md
```diff
@@ -677,6 +677,14 @@ SAVE IMAGE my-example-registry.com/another-image:latest
 ```
 {% endhint %}
 
+{% hint style='danger' %}
+##### Important
+
+Since the introduction of the `--referenced-save-only` feature flag, images are only saved if they are connected to the initial target through a chain of `BUILD` commands. [See the `--referenced-save-only` flag documentation for more details](./features.md#referenced-save-only).
```
alexcb, line 683: I think it would be helpful to call out this is enabled via `VERSION 0.6`

dchw: That was the point of the link; I was trying to not pull in so many details of how our feature flags work. I gave it a crack in the latest commit

dchw: I added something but I will use your wording I guess

vladaionescu: I think we need to reword this to just say "In VERSION 0.6...". Only advanced users / earthly contributors should know about individual feature flags.

docs/earthfile/earthfile.md
```diff
@@ -677,6 +677,14 @@ SAVE IMAGE my-example-registry.com/another-image:latest
...
+
+This is automatically enabled when setting [`VERSION 0.6`](#version) in your `Earthfile`.
+{% endhint %}
```
vladaionescu, line 685: Note that this applies to `SAVE ARTIFACT ... AS LOCAL ...` too. Should there be a similar warning there?

docs/earthfile/earthfile.md
```diff
@@ -677,6 +677,13 @@ SAVE IMAGE my-example-registry.com/another-image:latest
 ```
 {% endhint %}
 
+{% hint style='danger' %}
+##### Important
+
+As of [`VERSION 0.6`](#version), the `--referenced-save-only` feature flag is enabled by default. Images are only saved if they are connected to the initial target through a chain of `BUILD` commands. [See the `--referenced-save-only` flag documentation for more details](./features.md#referenced-save-only).
```
vladaionescu, line 683: Instead of the feature flag link here, I would link to BUILD and point to "What is output and pushed".

docs/earthfile/earthfile.md
dchw: See latest commit

docs/earthfile/earthfile.md
dchw: Why not both

docs/earthfile/earthfile.md
dchw: Yes, probably

docs/earthfile/earthfile.md
vladaionescu: I'm trying to avoid confusing new users with the term "feature flag" :-)

Just thinking that simpler is better - they don't need to know about our history to get going.

vladaionescu: Friendly ping on this. I know it seems innocuous, but the extra simplicity is worth it.

vladaionescu (approved):

alexcb (approved):

PR #2374, opened 2022-11-07
Bring remote caching and remote runners higher up in the docs

vladaionescu:
Also, Rename shared caching to remote caching, to align with the industry.


docs/remote-runners.md
```diff
@@ -0,0 +1,31 @@
+# Remote runners
+
+Earthly supports running builds on remotely via remote runners. Remote runners allow you to benefit from sharing the cache with other users of that remote runner. This is especially useful in CI environments where you want to share the cache between runs.
```
alexcb, line 3:
```suggestion
Earthly supports running builds remotely via remote runners. Remote runners allow you to benefit from sharing the cache with other users of that remote runner. This is especially useful in CI environments where you want to share the cache between runs.
```

PR #2375, opened 2022-11-07
deltautil: reset pointer value via dereference

mikejholly:


vladaionescu (approved):

vladaionescu (commented):
Mind adjusting the comment of this function? It's safe to say that the original manifest is always mutated now.

mikejholly:
@vladaionescu are these failing tests important?

vladaionescu:
I clicked re-run - let's see if it was simply a flake.

PR #2380, opened 2022-11-08
docs: Satellite sleep behaviour

brandonSc:
Probably don't want to merge until the sleep feature is rolled out completely.


docs/cloud/satellites/managing.md
```diff
@@ -115,6 +115,26 @@ earthly sat launch <satellite-name>
 
 The newly launched satellite will always get the latest version available.
 
+### Managing a satellite's state
+
+To save costs, satellites automatically enter a "sleep" state after 30 min of inactivity. While a satellite is asleep, you are not billed for any compute minutes.
```
vladaionescu, line 120:
```suggestion
To save costs, satellites automatically enter a **sleep** state after 30 min of inactivity. While a satellite is asleep, you are not billed for any compute minutes.
```

docs/cloud/satellites/managing.md
```diff
@@ -115,6 +115,26 @@ earthly sat launch <satellite-name>
...
+
+The satellite will automatically "wake up" when a new build is started while it's in a sleep state. This is visible during the `Init` phase of the Earthly log.
```
vladaionescu, line 122:
```suggestion
The satellite will automatically **wake up** when a new build is started while it's in a sleep state. This is visible during the `Init` phase of the Earthly log.
```

docs/cloud/satellites/managing.md
```diff
@@ -115,6 +115,26 @@ earthly sat launch <satellite-name>
...
+earthly sat sleep <satellite-name>
+```
+
+Similarly, a Satellite can be manually woken up using:
+
+```bash
+earthly sat wake <satellite-name>
+```
+
+Note that the [`inspect`](#Checking status of a satellite) command will show you if a Satellite is currently awake or asleep.
```
vladaionescu, line 136: I don't think this link works - you might need to check the preview to see what the correct anchor should be here.

vladaionescu (commented):
:+1: Some small remarks...

docs/cloud/satellites/managing.md
brandonSc: I think I have the syntax right now, but it doesn't seem like it's creating the new preview link?

vladaionescu: Not sure why - maybe it needs yet another commit?

vladaionescu (approved):

docs/cloud/satellites/managing.md
brandonSc: It looks right now btw, triggered it with some dummy commits

PR #2395, opened 2022-11-10
Improvements related to logstream integration

vladaionescu:
Changes:

* Biggest change: Allow Earthly to run as a different container+volume+.earthly dir name in development. This allows the released Earthly to be logged in as one user and the dev Earthly to be logged in as a different user. This works by changing the name `earthly` in all these places to `earthly-dev`, when not building an official release. So the container `earthly-buildkitd` becomes `earthly-dev-buildkitd`, the cache volume `earthly-cache` becomes `earthly-dev-cache`, and the dir `~/.earthly` becomes `~/.earthly-dev`. The ports also shift. The release ports are the same as before, whereas the ports used in dev are based on official ports + an offset calculated based on a hash of the installation name.
* Removed debugger port as it was no longer used
* Add a flag for enabling uploading logstream: `--logstream-upload`. When set, this replaces the old log uploading feature.
* Allow grpc insecure
* Remove unnecessary logstream batching

### DIFF
--- a/.buildkite/tests.sh
+++ b/.buildkite/tests.sh
@@ -75,10 +75,7 @@ do
     sleep $(( att_num++ ))
 done
 
-"$earthly" config global.conversion_parallelism 5
-
 export EARTHLY_VERSION_FLAG_OVERRIDES="referenced-save-only"
-"$earthly" config global.local_registry_host 'tcp://127.0.0.1:8371'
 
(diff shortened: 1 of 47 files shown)


mikejholly (approved):
LGTM

vladaionescu:
I'll keep this open for @alexcb to take a look too

alexcb (approved):

PR #2397, opened 2022-11-10
exit with error if frontend not detected

alexcb:
If the frontend detection fails, it returns `none`, then the `./earthly` script will attempt to run commands such as `none pull docker.io/earthly/earthlybinaries:prerelease`, which results in a `none command not found` which is difficult to debug.

### DIFF
--- a/earthly
+++ b/earthly
@@ -38,10 +38,14 @@ detect_frontend() {
     return
   fi
 
-  echo "none"
+  echo >&2 "failed to detect docker/podman frontend"
+  exit 1
 }
 
-frontend_bin=$(detect_frontend)
+# some tests run ./earthly when no front-end is present; when EARTHLY_DISABLE_FRONTEND_DETECTION=true, skip detection.
+if [ "$EARTHLY_DISABLE_FRONTEND_DETECTION" != "true" ]; then
+    frontend_bin=$(detect_frontend)
+fi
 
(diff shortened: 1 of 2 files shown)


earthly
```diff
@@ -38,10 +38,13 @@ detect_frontend() {
...
 
-  echo "none"
+  echo >&2 "failed to detect docker/podman frontend"
+  exit 1
 }
 
-frontend_bin=$(detect_frontend)
+if [ "$EARTHLY_DISABLE_AUTO_UPDATE" != "true" ]; then
+    frontend_bin=$(detect_frontend)
+fi
```
vladaionescu, line 47: Is this intended? We should still detect the frontend even if auto-update is disabled.

vladaionescu: Oh nvm - I see the logic now :-)

vladaionescu (approved):

earthly
alexcb: it's needed for our `+earthly-script-no-stdout` test.

alexcb: ah but there's a case where calling `do_reset()` will skip this part if a user has `EARTHLY_DISABLE_AUTO_UPDATE=true`.

I'll introduce a new `EARTHLY_DISABLE_FRONTEND_DETECTION` and add a comment saying it's meant for tests.

PR #2398, opened 2022-11-11
Add EARTHLY_CI arg

### LINKED ISSUE #2261: Expose `--ci` status to targets
A very common pattern in my Earthfile is to check whether or not I'm in a CI environment. This is useful in so many scenarios:

- Should I pull AWS credentials from the command line with `--secrets`, or run `aws login` in Earthly
- Should I automatically fix linter errors (e.g. Prettier), or should I fail the build
- Should I write out files for CI, or echo certain extra data (e.g. for GitHub actions)?

I currently solve this by adding a `dev` argument to most of my targets. It'd be so much nicer if I could use the status of the `--ci` flag. It'd remove a lot of ARG declarations, and I wouldn't have to pass the value down to each target that I'm calling.

I'd love to see these two things:

- Expose the status of `--ci`, e.g. as a `$ci` variable
- Add some easy way of checking. `IF [ "$ci" = "true" ]` is okay, but something like `IF --ci` would be amazing

miles170:
Closes #2261.

This adds an `EARTHLY_CI` builtin arg which is set to `true` when earthly is run with the `--ci` flag.


variables/reserved/names.go
```diff
@@ -14,6 +14,7 @@ const (
 	EarthlyGitTag               = "EARTHLY_GIT_TAG"
 	EarthlyLocally              = "EARTHLY_LOCALLY"
 	EarthlyPush                 = "EARTHLY_PUSH"
+	EarthlyCI                   = "EARTHLY_CI"
```
alexcb, line 17: this will also have to be added to the map in the `init()` function below.

tests/ci-arg.earth
```diff
@@ -0,0 +1,8 @@
+VERSION 0.6
+FROM alpine:3.15
+
+ci-test:
+    ARG EARTHLY_CI
+    RUN test "$EARTHLY_CI" = "true" || test "$EARTHLY_CI" = "false"
+    RUN (if [ "$EARTHLY_CI" = "true" ]; then echo UHVzaCBpdCByZWFsIGdvb2Q=; else echo U2hvb3Agc2hvb3AgYmEtZG9vcA==; fi) | base64 -d | tee chorus
+    SAVE ARTIFACT chorus
```
alexcb, line 8: is this `SAVE ARTIFACT` needed? I don't see `chorus` being used anywhere in the test.

tests/ci-arg.earth
```diff
@@ -0,0 +1,8 @@
+VERSION 0.6
+FROM alpine:3.15
+
+ci-test:
+    ARG EARTHLY_CI
+    RUN test "$EARTHLY_CI" = "true" || test "$EARTHLY_CI" = "false"
+    RUN (if [ "$EARTHLY_CI" = "true" ]; then echo UHVzaCBpdCByZWFsIGdvb2Q=; else echo U2hvb3Agc2hvb3AgYmEtZG9vcA==; fi) | base64 -d | tee chorus
```
alexcb, line 7: Can you pick some different output text rather than copy/pasting the push test.

alexcb (commented):
thanks for the PR, looks good -- just left a couple minor requests.

alexcb (approved):

PR #2400, opened 2022-11-11
Simplify the docker-compose interactive debugger test

### LINKED ISSUE #2263: simplify the docker compose interactive debugger test
The https://github.com/earthly/earthly/tree/main/scripts/tests/interactive-debugger/docker-compose test makes use of `aa8y/postgres-dataset:iso3166` which can take some time to download and start up. As a result, this test frequently times out (and appears to be flaky)

We should use a different smaller image that starts a smaller, faster server.

The test still needs to use docker-compose, and the server should still listen to a port, and use environment variables.

The test should still continue to test that the interactive debugger can connect to the test server, and issue some sort of command the test server should respond to.

idea: perhaps the server could be a tcp echo server, which performs a rot13?

The interactive debugger test can be run via https://github.com/earthly/earthly/blob/main/scripts/tests/interactive-debugger/test-interactive.py

miles170:
Closes #2263.

1. Replace `aa8y/postgres-dataset` with the rot13 echo server.
2. Reduce the interactive debugger test timeout to 180 seconds.

### DIFF
--- a/.github/workflows/reusable-misc-tests.yml
+++ b/.github/workflows/reusable-misc-tests.yml
@@ -93,7 +93,7 @@ jobs:
         with:
           timeout_minutes: 10
           max_attempts: 2
-          command: ./scripts/tests/interactive-debugger/test-interactive.py --earthly ${{inputs.BUILT_EARTHLY_PATH}} --timeout 300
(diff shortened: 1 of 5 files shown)


alexcb (approved):
thanks for the fix! :tada:

PR #2404, opened 2022-11-11
Use request IDs to cloud endpoints, and print them on errors

brandonSc:
Errors are printed with a request ID:

Screenshot 2022-11-11 185823

It works more consistently with gRPC errors, but with HTTP errors, it's only printed on requests that retry (so basically all 500-level errors, which is when this is most important anyway)

On the server log, we can correlate errors to request IDs that a user might send us, for example, by punching it into Kibana:

Screenshot 2022-11-11 190108

### DIFF
--- a/cloud/grpc.go
+++ b/cloud/grpc.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/google/uuid"
 	"github.com/pkg/errors"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
@@ -18,13 +17,13 @@ func (c *client) withAuth(ctx context.Context) context.Context {
 }
 
 func withReqID(ctx context.Context) context.Context {
-	return metadata.AppendToOutgoingContext(ctx, "request-id", uuid.NewString())
+	return metadata.AppendToOutgoingContext(ctx, requestID, newRequestID())
 }
 
 func getReqID(ctx context.Context) string {
 	md, ok := metadata.FromOutgoingContext(ctx)
 	if ok {
-		vals := md.Get("request-id")
+		vals := md.Get(requestID)
 		if len(vals) > 0 {
 			return vals[0]
 		}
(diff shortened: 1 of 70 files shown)


cloud/grpc.go
```diff
@@ -27,18 +44,21 @@ func (c *client) UnaryAuthInterceptor() grpc.UnaryClientInterceptor {
...
+			if ok {
+				err = status.Errorf(s.Code(), fmt.Sprintf("%s {reqID: %s}", s.Err(), getReqID(ctx)))
+				if s.Code() == codes.Unauthenticated {
+					err = c.Authenticate(ctx)
+					if err != nil {
+						return errors.Wrap(err, "failed re-authenticating")
+					}
+					md, ok := metadata.FromOutgoingContext(ctx)
+					if !ok {
+						return errors.New("could not parse metadata")
```
vladaionescu, line 56: Should this really be an error - couldn't we just start a new metadata object if it doesn't exist?

cloud/http.go
```diff
@@ -109,9 +110,10 @@ func (c *client) doCall(ctx context.Context, method, url string, opts ...request
 	body := []byte{}
 	var callErr error
 	duration := time.Millisecond * 100
+	reqID := uuid.NewString()
```
vladaionescu, line 113: Shouldn't the request ID be created by the server and then returned as part of any responses (especially as part of error responses)?

brandonSc: It can be done that way, but I think it makes more sense for the client to generate it.

One reason is idempotency - the client may retry failed requests using the same request-id, and the server can be coded to ensure it doesn't duplicate things it shouldn't by distinguishing between a retry vs an entirely new request. For example, [Stripe APIs work this way](https://stripe.com/docs/api/idempotent_requests), and [Google suggests doing so in gRPC](https://cloud.google.com/apis/design/design_patterns) (although I think their suggestion is in the request message, vs metadata, but close enough).

It's sometimes documented as a standard-ish HTTP header for correlating server logs to client requests, e.g. [heroku](https://http.dev/x-request-id) and [other blogs](https://http.dev/x-request-id) mention it.

Also similar to how distributed tracing works, where a tracked request usually begins at the client in the form of some unique ID, before it's passed through various backend services. P.S. Our backend also passes this request ID to subsequent internal microservices, so we can do some of that tracing via our logs.

cloud/grpc.go
brandonSc: I've updated this to start new metadata as you suggested

cloud/grpc.go
```diff
@@ -16,8 +17,24 @@ func (c *client) withAuth(ctx context.Context) context.Context {
 	return metadata.AppendToOutgoingContext(ctx, "authorization", fmt.Sprintf("Bearer %s", c.authToken))
 }
 
+func withReqID(ctx context.Context) context.Context {
+	return metadata.AppendToOutgoingContext(ctx, "request-id", uuid.NewString())
```
alexcb, line 21: rather than always generate a new one, it would be helpful if we could override this via an environment variable (e.g. `EARTHLY_REQUEST_ID`) -- this could be useful for cases where one programatically runs earthly.

cloud/http.go
alexcb: I'm in favour of generating it client side (see my comment regarding https://github.com/earthly/earthly/pull/2404/files#r1021874460 )

Is there any way we can make use of `withReqID()` here? Having a single code path for generating (and potentially overriding) the request ID would be cleaner.

cloud/grpc.go
vladaionescu: Good thinking

cloud/http.go
vladaionescu: Sounds good then - I was a bit worried about someone being able to spam us. But then again, it's hard to guess other user's UUIDs to create any real issues.

cloud/grpc.go
brandonSc: Yeah I like that idea

brandonSc:
An example on HTTP request to secrets (I forced a 500 error on staging to get a good example)
Screen Shot 2022-11-14 at 1 29 57 PM

cloud/grpc.go
```diff
@@ -16,8 +17,24 @@ func (c *client) withAuth(ctx context.Context) context.Context {
...
+}
+
+func getReqID(ctx context.Context) string {
+	md, ok := metadata.FromOutgoingContext(ctx)
+	if ok {
+		vals := md.Get("request-id")
```
mikejholly, line 27: Should we make `request-id` a const?

cloud/grpc.go
```diff
@@ -27,18 +44,22 @@ func (c *client) UnaryAuthInterceptor() grpc.UnaryClientInterceptor {
 		err := invoker(ctx, method, req, reply, cc, opts...)
 		if err != nil {
 			s, ok := status.FromError(err)
-			if ok && s.Code() == codes.Unauthenticated {
-				err = c.Authenticate(ctx)
-				if err != nil {
-					return errors.Wrap(err, "failed re-authenticating")
+			if ok {
```
mikejholly, line 47: Perhaps `if !ok` with a `return` would limit nesting here.

cloud/grpc.go
```diff
@@ -56,19 +78,23 @@ func (c *client) StreamAuthInterceptor() grpc.StreamClientInterceptor {
...
-			if ok && s.Code() == codes.Unauthenticated {
-				err = c.Authenticate(ctx)
-				if err != nil {
-					return nil, errors.Wrap(err, "failed re-authenticating")
-				}
-				md, ok := metadata.FromOutgoingContext(ctx)
-				if !ok {
-					return nil, errors.New("could not parse metadata")
+			if ok {
+				err = status.Errorf(s.Code(), fmt.Sprintf("%s {reqID: %s}", s.Err(), getReqID(ctx)))
```
mikejholly, line 82: Might be nice to deduplicate this code and the above.

cloud/http.go
```diff
@@ -141,16 +143,16 @@ func (c *client) doCall(ctx context.Context, method, url string, opts ...request
...
-func shouldRetry(status int, body []byte, callErr error, warnFunc func(string, ...interface{})) (bool, error) {
+func shouldRetry(status int, body []byte, callErr error, warnFunc func(string, ...interface{}), reqID string) (bool, error) {
 	if status == http.StatusUnauthorized {
 		return true, nil
 	}
 	if 500 <= status && status <= 599 {
 		msg, err := getMessageFromJSON(bytes.NewReader(body))
 		if err != nil {
-			warnFunc("retrying http request due to unexpected status code %v", status)
+			warnFunc("retrying http request due to unexpected status code %v {reqID: %s}", status, reqID)
```
mikejholly, line 153: Should `warnFunc` just append the request ID to the message string? It seems like we already pass it.

cloud/grpc.go
brandonSc: I've added this, but I've kept it as a hidden flag. I think it might be useful for us debugging at some point, but I realized a lot of our commands make multiple distinct requests, and this would cause them to all have the same request ID. I'm not sure we can guarantee that wouldn't do weird things to the server in the future.

vladaionescu (approved):

cloud/http.go
brandonSc: `warnFunc` is passed in to this func here, but is created much higher up in the earthly code before the requestID is created, so I think it might be simpler to append it here

cloud/grpc.go
brandonSc: Yeah it was getting a bit messy as it grew, I've refactored it quite a bit now if you don't mind taking another look

cloud/http.go
mikejholly: Oh ok, it looks like it's already using printf-style arguments.

alexcb (approved):

Issue #2210, opened 2022-09-22
Adhere to XDG Base Directory Specification


hyperupcall:
This issue is a request to adhere to the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) when managing Earthly files. Currently, files are put inside of [`~/.earthly`](https://github.com/earthly/earthly/blob/2391901a9e29a051ed49152d282f9bacf2549916/util/cliutil/earthlydir.go#L29), but there are [many reasons](https://xdgbasedirectoryspecification.com/) to change that.

I propose to change [`getEarthlyDirAndUser`](https://github.com/earthly/earthly/blob/2391901a9e29a051ed49152d282f9bacf2549916/util/cliutil/earthlydir.go#L29) (or wherever this is best applicable) like so:
1. If `~/.earthly` exists, use it (backwards compatability)
2. If `XDG_DATA_HOME` exists and starts with a `/`, then use `$XDG_DATA_HOME/earthly` (as per xdg base dir spec)
3. Else, use `~/.local/share/earthly` (as per xdg base dir spec)

[Many](https://github.com/python-poetry/poetry/issues/1659) [other](https://github.com/pnpm/pnpm/issues/2574) [tools](https://github.com/basherpm/basher/pull/98) made changes for better adherence, and I would welcome a change for Earthly too :)

alexcb:
Seems like a reasonable proposal; I would consider such a change if a PR were submitted.

hyperupcall:
I've been fiddling around with this, and I'm wondering how should the bootstrap case be handled? In my [current branch](https://github.com/hyperupcall/earthly/compare/main...hyperupcall:earthly:xdg-base-dir-spec-compliance), I change everything to use the XDG-compliant directory (including the `./eartly` and various other shell scripts). I did this because it would seem that the scripts would break if I didn't. It seems that once the new Earthly earthlybinaries:prerelease is built with the new Go code, the scripts would be looking in the wrong place. Since I'm not as familiar with the bootstrap setup, I just wanna be sure - did I get those fixes right?

Also, I changed to using the `XDG_STATE_HOME` instead of `XDG_DATA_HOME` because the directory seemed to contain files that were specific to a particular machine. Specifically, the computer-specific `install_id` file and the architecture-specific `earthly-prerelease` files aren't compatible with the semantics of `XDG_DATA_HOME`, so I used `XDG_STATE_HOME` instead.

alexcb:
[...]

PR #2405, opened 2022-11-13
Allow for bash-completion for non-root user

### LINKED ISSUE #2318: Potential autocomplete UX tweaks
Sorry for the vague/generic title, but I didn't think the individual issues warranted their own issues.  These are all just UX/polish.

#### Errors/warnings
When running earthly bootstrap (with completion) as an unprivileged user, the warnings are presented differently and "look funny" when seen together (the spacing, as well as the bash warning is red in a color-enabled terminal, while the zsh one is plain white):
```
$ earthly bootstrap --with-autocomplete
           bootstrap | Warning: open /usr/share/bash-completion/completions/earthly: permission denied
Warning: unable to enable zsh-completion: none of /usr/local/share/zsh/site-functions, /usr/share/zsh/site-functions does not exist
```
When I try, I can come up with a reason why it could makes sense, but my guess is that it's not intentional.  But I wasn't 100% sure (nor did I have a good guess which one was "more correct"), so I'm just offering feedback rather than a fix.  As well, that zsh error message could use some love ("none of X, Y does not exist").

#### Per-user automated install
In a perfect world, auto-completion would install for unprivileged users as well.  Quick googling and personal experience shows that standardized user-specific locations have been around a while (for bash at least).  If you agree this is a good thing, it might make sense to do this alongside (or after) #2210 

#### Per-user manual installation directions
In the absence of the above, what are your thoughts on providing DIY instructions in the output (if they can't be successfully installed at the system level and/or you detect it's a non-root execution)?  For example:
```
$ earthly bootstrap --with-autocomplete
           bootstrap | Warning: open /usr/share/bash-completion/completions/earthly: permission denied
           bootstrap |   (To install as a user, add "complete -o nospace -C '<bin>' earthly" to your ~/.bashrc or equivalent)
```
Or at the very least, add those manual directions to the [online docs](https://docs.earthly.dev/docs/earthly-command#earthly-bootstrap)?

At present, I have six auto-complete files that were magically installed into my XDG-specified folder and at least 11 manually added in my `.bashrc`.  Earthly is the only one that required going into source code to discover.

rrjjvv:
Closes #2318 

This is the original proposed solution for #2318.  It works, but may not be ideal.

1. A new library (for XDG standards) is introduced; it's arguably overkill for what's needed, but is lightweight and sets the stage for other uses (e.g., #2210).  The required functionality can be trivially re-implemented if the addition is not desired.
2. The inconsistent error messaging is better (I think); errors are printed to the 'console' rather than raw `stderr`
3. While bash has pre-defined standards for user-defined completions, zsh does not (neither established convention nor XDG-based standard), so zsh is untouched in this PR.  This means non-root completion is not present and its error messaging is now inconsistent with bash.

I have an alternate solution (#2406) that solves the problem in a different manner.  They are not mutually exclusive and it's not a given that it will be preferred, so I'm still creating this one.

Addendum
---
There's another limitation (that can be overcome) of this solution.  I followed the bash-completion guidance, and also maintained the spirit of the existing code.  The existing code will not create any parent directories, which I interpret as being an implicit check on whether the completion package is actually installed.  This PR (probably wrongly) maintains that same logic.

On my laptop, this code would not have any effect.  I have a number of completions (from other software) in XDG-defined directories, but the "wrong" ones.    The ones I have are all in the "config" directory (whereas the faq refers to the "data" directory).  That's probably a Ubuntu/Debian extension, which those packages recognized.  Bottom line, the parent directory that is _supposed_ to be the home for user-completions (per the standard)  doesn't even exist on my laptop.

It would be trivial to update this PR to create missing directories, meaning that we may be install completions on a system that doesn't even have completion installed (not a big deal in my opinion).  But there's no guarantee that a distro even _uses_ the standard location.  I think this further points to #2406 being the better overall solution.


cmd/earthly/autocomplete.go
```diff
@@ -82,50 +83,80 @@ func (app *earthlyApp) autoCompleteImp(ctx context.Context) (err error) {
...
+		// https://github.com/scop/bash-completion/blob/master/README.md#faq
+		userPath, ok := os.LookupEnv("BASH_COMPLETION_USER_DIR")
+		if !ok {
+			// This will give a standardized fallback even if XDG isn't active
+			userPath = xdg.DataHome
+		}
+		path = filepath.Join(userPath, "bash-completion/completions/earthly")
+	}
+	ok, err := app.insertBashCompleteEntryAt(path)
+	if err == nil {
```
alexcb, line 114: can you move the `return err` from the end of this function here instead:
```
	if err != nil {
	    return err
	}
	if ok {
		...
	} else {
		...
	}
	return nil
```

It makes the code more robust to future changes where someone might unintentionally reassign to `err` before the final `return err` at the end of the function is hit.

rrjjvv: Sure thing; I'll make that change when I get a few minutes.

alexcb (approved):

PR #2406, opened 2022-11-13
Add `completion` commands

### LINKED ISSUE #2318: Potential autocomplete UX tweaks

rrjjvv:
Closes #2318 

This is an alternative (or possible addition) to #2405 

This will allow for the popular pattern of sourcing the output of `earthly completion <shell>` to enable user-specific completion.  This does put more work on the user, but follows existing patterns (for better or worse) that are likely familiar to users (and removes guesswork/complexity from the codebase).

I didn't know if you'd like the addition of new commands, so I didn't spend a whole lot of time trying to figure out your conventions/preferences around the user-facing aspects, so those may need tweaking.

### DIFF
--- a/cmd/earthly/autocomplete.go
+++ b/cmd/earthly/autocomplete.go
@@ -20,6 +20,8 @@ import (
 	"github.com/pkg/errors"
 )
 
+type completionGenerator func() (string, error)
+
 // to enable autocomplete, enter
 // complete -o nospace -C "/path/to/earthly" earthly
 func (app *earthlyApp) autoComplete(ctx context.Context) {
--- a/cmd/earthly/root_cmds.go
+++ b/cmd/earthly/root_cmds.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"io"
 	"net/url"
 	"os"
 	"path"
@@ -59,6 +60,25 @@ func (app *earthlyApp) rootCmds() []*cli.Command {
 					Name:        "with-autocomplete",
 					Usage:       "Add earthly autocompletions",
 					Destination: &app.bootstrapWithAutocomplete,
+				},
+			},
+		},
+		{
+			Name:        "completion",
+			Usage:       "Output shell completion code for the specified shell (bash, zsh)",
+			Description: "Output shell completion code for the specified shell (bash, zsh)",
+			Subcommands: []*cli.Command{
+				{
+					Name:      "bash",
+					Usage:     "Generate completions for bash",
+					UsageText: "earthly completion bash",
+					Action:    app.actionCompletion(bashCompleteEntry, os.Stdout),
+				},
+				{
+					Name:      "zsh",
+					Usage:     "Generate completions for zsh",
+					UsageText: "earthly completion zsh",
+					Action:    app.actionCompletion(zshCompleteEntry, os.Stdout),
 				},
 			},
 		},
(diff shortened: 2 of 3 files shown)


alexcb:
> I didn't know if you'd like the addition of new commands

I'm hesitant to introduce these new commands since we already have similar (hidden) functionality under the `bootstrap` command.

Here's what we currently have:
```
$ earthly bootstrap --source bash
complete -o nospace -C '/home/alex/.earthly/earthly-prerelease' earthly
```

and

```
$ earthly bootstrap --source zsh
#compdef _earthly earthly

function _earthly {
    autoload -Uz bashcompinit
    bashcompinit
    complete -o nospace -C '/home/alex/.earthly/earthly-prerelease' earthly
}
```

if you run `EARTHLY_SHOW_HIDDEN=1 earthly bootstrap --help`, you'll see a the help text for `--source`:

```
$ earthly bootstrap --help
NAME:
   earthly bootstrap - Bootstraps earthly installation including shell autocompletion and buildkit image download

USAGE:
   earthly bootstrap [command options] [arguments...]

DESCRIPTION:
   Bootstraps earthly installation including shell autocompletion and buildkit image download

OPTIONS:
   --source value       Output source file (for use in homebrew install)       <----------- HERE
   --no-buildkit        Do not bootstrap buildkit (default: false)
   --with-autocomplete  Add earthly autocompletions (default: false)
   --help, -h           show help (default: false)
```

Perhaps the solution here is to unhide them, and improve the help text?

rrjjvv:
Unhiding that (or at least adding it to the docs) would work.  If you also have to provide `--no-buildkit`, it could get a bit unwieldy though  (Granted a user only has to type it once, but it would be an eyesore.)  If the `--source` does (or could) implicitly enable that flag it would be a non-issue though.

But I think one of them (whether it's this PR, exposing what you showed above, or a copy/pasteable snippet in the docs) is needed for the non-privileged zsh user (that can't/won't do a one-time root install for the sake of finding the completion code, or going through the source).

alexcb:
> Unhiding that (or at least adding it to the docs) would work. If you also have to provide `--no-buildkit`, it could get a bit unwieldy though (Granted a user only has to type it once, but it would be an eyesore.) If the `--source` does (or could) implicitly enable that flag it would be a non-issue though.

when `--source` is given, it only prints the completion function/command and then exits:
https://github.com/earthly/earthly/blob/main/cmd/earthly/root_cmds.go#L328

bootstrapping only occurs when no `--source=` flag is given: https://github.com/earthly/earthly/blob/main/cmd/earthly/root_cmds.go#L338

> is needed for the non-privileged zsh user

unhiding it, and mentioning it in our docs seems like it'll cover the choose-your-adventure route.

rrjjvv:
Closing, per discussion above.

PR #2409, opened 2022-11-14
Add size and platform options for Satellites

dchw:
Also include a small fix to the OOM hook script, so we dont't have to enable developer mode to make it work.


vladaionescu (approved):
🚀

release/ami/image.pkr.hcl
```diff
@@ -19,7 +19,7 @@ locals {
 # https://cloud-images.ubuntu.com/locator/ec2/
 
 source "amazon-ebs" "x86_64" {
-  ami_name      = "earthly-${var.earthly_version}-x86_64-${local.timestamp}"
+  ami_name      = "earthly-${var.earthly_version}-amd64-${local.timestamp}"
```
dchw, line 22: Also a small fix to align the AMI names with the architecture component of our platform strings

dchw: I already fixed the existing one by copying the AMI to match this naming pattern.

cmd/earthly/satellite_cmds.go
```diff
@@ -7,6 +7,7 @@ import (
 	"os"
 	"time"
 
+	"github.com/jedib0t/go-pretty/table"
```
alexcb, line 10: is there a reason we can't use `"text/tabwriter"` instead?

we already use it in many of our commands, see  https://github.com/earthly/earthly/blob/main/cmd/earthly/account_cmds.go#L442-L443 for an example.

dchw: It takes a lot more code, but I can rewrite it if you want

dchw: Rewritten

alexcb (approved):

cmd/earthly/satellite_cmds.go
alexcb: thanks! I think it's useful to stay consistent with how we write out these tables.

PR #2413, opened 2022-11-15
fix incorrect session ID passed to llb.SessionID

alexcb:


buildcontext/local.go
```diff
@@ -87,21 +86,24 @@ func (lr *localResolver) resolveLocal(ctx context.Context, gwClient gwclient.Cli
...
+		if _, isTarget := ref.(domain.Target); isTarget {
+			noImplicitIgnore := bf.ftrs != nil && bf.ftrs.NoImplicitIgnore
+			excludes, err := readExcludes(ref.GetLocalPath(), noImplicitIgnore)
+			if err != nil {
+				return nil, err
+			}
+			buildContextFactory = llbfactory.Local(
+				ref.GetLocalPath(),
+				llb.ExcludePatterns(excludes),
+				llb.SessionID(gwClient.BuildOpts().SessionID),
```
vladaionescu, line 100: Apparently, this session ID is a different kind of session ID from the buildkit session ID. I remember Tonis mentioned this over the Docker slack some years ago. Could we just not set this at all?

earthfile2llb/withdockerrun.go
```diff
@@ -364,22 +361,12 @@ func (w *withDockerRunTar) solveImage(ctx context.Context, mts *states.MultiTarg
...
-		// buildkit to use cache when these are the same as before (eg a docker image
-		// that is identical as before).
-		sessionIDKey := fmt.Sprintf("%s-%s", dockerTag, dockerImageID)
-		sha256SessionIDKey := sha256.Sum256([]byte(sessionIDKey))
-		sessionID := hex.EncodeToString(sha256SessionIDKey[:])
 
 		tarContext := pllb.Local(
 			string(solveID),
-			llb.SessionID(sessionID),
+			llb.SessionID(w.c.opt.GwClient.BuildOpts().SessionID),
```
vladaionescu, line 367: I would not change this part of the code. Note that this is only used for tar-based WITH DOCKER loading, which is no longer used anyway.

vladaionescu:
I think we can remove this variable as a result of this PR (and all the places where this is propagated to): https://github.com/earthly/earthly/blob/main/cmd/earthly/main.go#L347

vladaionescu:
Note that VERSION 0.5 and VERSION 0.6 would not make a difference, since the registry-based loading is now enabled retroactively for 0.5 too.

buildcontext/local.go
alexcb: Using the SessionID passed in via the BuildOpts was inspired by how the dockerfile frontend works: https://github.com/moby/buildkit/blob/fad945e9b9ae09dbdd110405e1dd78a2c3cbdc09/frontend/dockerfile/builder/build.go#L186

If it's not set we would use `return ls.snapshotWithAnySession(ctx, g)` under https://github.com/moby/buildkit/blob/b3e8c63a48ad8c015f5631fc1947945b229b3919/source/local/local.go#L92

I'm worried that if we simply use any session -- potentially a session from a completely different client (on a different host). I would need to dig in more to fully understand if such a race is possible.

vladaionescu: I see - yeah, sounds good then. If the Dockerfile frontend does the same then it should be fine. No idea why Buildkit doesn't set that session ID itself but 🤷‍♂️.

alexcb: there's an `InContext` call that jobs get wrapped in, which creates a group containing **only** the current sessionID,

```
func (j *Job) InContext(ctx context.Context, f func(context.Context, session.Group) error) error {
	return f(progress.WithProgress(ctx, j.pw), session.NewGroup(j.SessionID))
}
```

then when `sm.Any` is called, it gets passed that group, which only contains the single session ID; so we're safe to remove this.

> If the Dockerfile frontend does the same

it turns out that the dockerfile frontend is executed by `buildkitd` and not `buildkitctl`, so it appears that the `llb.SessionID` override might only be needed for cases where the server is creating the llb rather than the client.

tl;dr: I'll remove it :)

earthfile2llb/withdockerrun.go
alexcb: I spent some time digging in here, and added to the comment explaining what's going on. When I tried removing the call to `llb.SessionID`, the image was nolonger cached -- so I definitely agree with leaving it as-is.

(thanks for pointing out that `VERSION --no-use-registry-for-with-docker 0.6` was needed to test this)

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ All notable changes to [Earthly](https://github.com/earthly/earthly) will be doc
 - Fixed Earthly on Mac would randomly hang on `1. Init` if Earthly was installed from Homebrew or the Earthly homebrew tap. [#2247](https://github.com/earthly/earthly/issues/2247)
 - Only referenced ARGs from .env are displayed on failures, this prevents secrets contained in .env from being displayed. [#1736](https://github.com/earthly/earthly/issues/1736)
 - Earthly now correctly detects if Podman is running but is under the disguise of the Docker CLI.
+- Slow performance when copying files due to incorrect usage of buildkit Session ID. [#2049](https://github.com/earthly/earthly/issues/2049)
```
vladaionescu, line 16: Make this a little more user-friendly.

```suggestion
- Improved performance when copying files. Fully-cached builds are now dramatically faster as a result [#2049](https://github.com/earthly/earthly/issues/2049)
```

vladaionescu (approved):

PR #2415, opened 2022-11-16
Fix `time.Now().UnixNano()` stripping away the nanos

vladaionescu:
Also, fix panic due to sending on a closed channel in `Logstreamer`.

### DIFF
--- a/logbus/logstreamer/logstreamer.go
+++ b/logbus/logstreamer/logstreamer.go
@@ -127,7 +127,7 @@ func (ls *LogStreamer) Write(delta *logstream.Delta) {
 		// TODO (vladaionescu): We should only log this if verbose is enabled.
 		dt, err := protojson.Marshal(delta)
 		if err != nil {
-			fmt.Fprintf(os.Stderr, "failed to marshal log delta: %v", err)
+			fmt.Fprintf(os.Stderr, "log streamer closed, but failed to marshal log delta: %v", err)
 			return
 		}
 		fmt.Fprintf(os.Stderr, "log streamer closed, dropping delta %v\n", string(dt))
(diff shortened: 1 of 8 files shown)


logbus/logstreamer/logstreamer.go
```diff
@@ -118,14 +119,27 @@ func (ls *LogStreamer) Write(delta *logstream.Delta) {
...
-	ch <- []*logstream.Delta{delta}
+	if ch != nil {
+		ch <- []*logstream.Delta{delta}
+	} else {
+		// TODO (vladaionescu): If these messages show up, we need to rethink
+		//						the closing sequence.
+		// TODO (vladaionescu): We should only log this if verbose is enabled.
+		dt, err := protojson.Marshal(delta)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "failed to marshal log delta: %v", err)
```
r-LaForge, line 130: nit
```suggestion
			fmt.Fprintf(os.Stderr, "log streamer closed, but failed to marshal log delta: %v", err)
```

logbus/logstreamer/logstreamer.go
```diff
@@ -118,14 +119,27 @@ func (ls *LogStreamer) Write(delta *logstream.Delta) {
 	ls.mu.Lock()
 	ch := ls.ch // ls.ch may get replaced on retry
 	ls.mu.Unlock()
```
r-LaForge, line 121: should we also defer the Unlock here?

r-LaForge (approved):

logbus/logstreamer/logstreamer.go
vladaionescu: No.. Channels are already concurrent in Go - they don't need a lock. And we don't want to hold the lock if the channel is congested as it could create a dead-lock.

PR #2416, opened 2022-11-16
OOM score adjustment should tolerate missing pids better

dchw:
It looks like if you have a process that exits before the `poststart` hook of the layer, you may have PIDs that go missing mid-loop. This tries to account for those.


buildkitd/oom-adjust.sh.template
```diff
@@ -8,22 +8,33 @@ fi
 
 for PID in $(pidof buildkit-runc)
 do
-    PID_NAME=$(cat /proc/"$PID"/cmdline)
+    if [ ! -d /proc/"$PID" ]; then
+        ! "${BUILDKIT_DEBUG}" || echo "$(date) | $PID was missing, ignoring" >> /var/log/oom_adj
+        continue
+    fi
+
+    PID_NAME=$(cat /proc/"$PID"/cmdline || echo "unknown")
```
vladaionescu, line 16: Could this be simplified to do the `if` afterward, and instead check for `$PID_NAME == "unknown"`? It would be more atomic this way - there's only one interaction with `/proc/$PID` instead of two.

buildkitd/oom-adjust.sh.template
```diff
@@ -8,22 +8,33 @@ fi
...
     do
-        ! "${BUILDKIT_DEBUG}" || echo "$(date) |   $PID has child: $CHILD_PID($(cat /proc/"$CHILD_PID"/cmdline)) with oom_score_adj: $(cat /proc/"$CHILD_PID"/oom_score_adj)" >> /var/log/oom_adj
+        if [ ! -d /proc/"$CHILD_PID" ]; then
+            ! "${BUILDKIT_DEBUG}" || echo "$(date) |   $CHILD_PID was missing, ignoring" >> /var/log/oom_adj
+            continue
+        fi
+
+        CHILD_PID_NAME=$(cat /proc/"$CHILD_PID"/cmdline || echo "unknown")
+        CHILD_OON_ADJ=$(cat /proc/"$CHILD_PID"/oom_score_adj || echo "unknown")
+        ! "${BUILDKIT_DEBUG}" || echo "$(date) |   $PID has child: $CHILD_PID($CHILD_PID_NAME) with oom_score_adj: $CHILD_OON_ADJ)" >> /var/log/oom_adj
```
vladaionescu, line 33: Typo

```suggestion
        CHILD_OOM_ADJ=$(cat /proc/"$CHILD_PID"/oom_score_adj || echo "unknown")
        ! "${BUILDKIT_DEBUG}" || echo "$(date) |   $PID has child: $CHILD_PID($CHILD_PID_NAME) with oom_score_adj: $CHILD_OOM_ADJ)" >> /var/log/oom_adj
```

vladaionescu (approved):
Just a couple of nits

buildkitd/oom-adjust.sh.template
dchw: Eh I went with it

vladaionescu (approved):

Issue #2340, opened 2022-11-01
Some shell-out expressions don't work well


vladaionescu:
The following expression does not work with `VERSION --shell-out-anywhere`:

```Earthfile
ARG foo = "$(echo '()')"
```

which is surprising. The following versions do work, however:

```Earthfile
ARG foo = "$(echo \\(\\))"
ARG foo = $(echo \\(\\))
```

Incidentally, without `--shell-out-anywhere`, all three versions work.

PR #2419, opened 2022-11-16
fix singlequote parsing for --shell-out-anywhere

alexcb:
lexer was incorrectly removing first character when parsing single quotes within a shellout. fixes https://github.com/earthly/earthly/issues/2340

### DIFF
--- a/util/shell/lex.go
+++ b/util/shell/lex.go
@@ -355,9 +355,7 @@ func (sw *shellWord) processDollar() (string, error) {
 	switch peek {
 	case '(':
 		sw.scanner.Next()
-		//return sw.processDollarShellOut()
-		shout, err := sw.processDollarShellOut()
-		return shout, err
+		return sw.processDollarShellOut()
 	case '{':
 		sw.scanner.Next()
 		return sw.processDollarCurlyBracket()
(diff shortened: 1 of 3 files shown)


util/shell/lex.go
```diff
@@ -374,27 +372,35 @@ func (sw *shellWord) processDollar() (string, error) {
...
 	escaped := false
 	for {
-		ch := sw.scanner.Next()
+		ch := sw.scanner.Peek()
 		if ch == scanner.EOF {
 			return "", errors.New("syntax error: missing ')'")
 		}
 		if escaped {
 			escaped = false
+			ch = sw.scanner.Next()
```
alexcb, line 392: not sure if it's more readable as `ch = sw.scanner.Next()` or `_ = sw.scanner.Next()`.

as `ch` is set above in `sw.scanner.Peek()`.

vladaionescu: Slight preference for `_ =`, but not a strong opinion.

vladaionescu (approved):

util/shell/lex.go
alexcb: I should have looked earlier, but the upstream code already uses the `ch = sw.scanner.Next()` pattern, so i'll leave it as is.

PR #2423, opened 2022-11-17
Add all the non-buildkit output to logstream too

vladaionescu:
Known issue: There are some extra new lines in the output. Will be fixed in a future PR.

### DIFF
--- a/cmd/earthly/main.go
+++ b/cmd/earthly/main.go
@@ -414,7 +414,6 @@ func (app *earthlyApp) before(cliCtx *cli.Context) error {
 		app.logstream = true
 	}
 	if app.logstream {
-		// TODO (vladaionescu): Set category correctly via the console logger.
 		app.console = app.console.WithPrefixWriter(app.logbus.Run().Generic())
 	}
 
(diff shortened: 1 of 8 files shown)


cmd/earthly/main.go
```diff
@@ -400,15 +400,22 @@ func (app *earthlyApp) before(cliCtx *cli.Context) error {
...
+		disableOngoingUpdates, app.logstreamDebugFile, app.buildID)
 	if err != nil {
 		return errors.Wrap(err, "logbus setup")
 	}
+	if app.logstreamUpload {
+		app.logstream = true
+	}
+	if app.logstream {
+		app.console = app.console.WithPrefixWriter(app.logbus.Run().Generic())
+	}
```
r-LaForge, line 418: nit:
```suggestion
	if app.logstreamUpload {
		app.logstream = true
		app.console = app.console.WithPrefixWriter(app.logbus.Run().Generic())
	}
```

r-LaForge (approved):

cmd/earthly/main.go
vladaionescu: This is incorrect because `app.logstream` can be enabled separately too (not just via `app.logstreamUpload`).

PR #2434, opened 2022-11-22
Make the Earthly Docker image work without privileged and without docker in some situations

vladaionescu:
Also, add tests for it.


brandonSc (approved):
LGTM

PR #2437, opened 2022-11-22
homebrew-release: don't change ldflags

alexcb:
The homebrew earthly formula creates it's own ldflags string, replacing it via the release.sh will lead to conflicting changes.

### DIFF
--- a/release/Earthfile
+++ b/release/Earthfile
@@ -205,7 +205,6 @@ release-homebrew:
         -e 's^\burl ".*"^url "'"$NEW_URL"'"^' \
         -e 's^\bsha256 ".*"$^sha256 "'$(cat /params/downloadsha256)'"^' \
         -e 's^\btags = ".*"^tags = "'"$(cat /params/tags)"'"^' \
-        -e 's^\bldflags = ".*"^ldflags = "'"$(cat /params/ldflags)"'"^' \
         ./Formula/earthly.rb
     RUN echo "Diff:" && git diff
     RUN version=${RELEASE_TAG#v} ;\


alexcb:
without this change, we've been auto-generating diffs such as:
```
@@ -24,7 +24,7 @@ class Earthly < Formula
     # the ldflags string will break the upstream release process.
     earthly_gitsha = "7e4f1df4c124db1644d51d312b19313217cbe478"
 
-    ldflags = "-X main.DefaultBuildkitdImage=earthly/buildkitd:v#{version} -X main.Version=v#{version} " \
+    ldflags = "-X main.DefaultBuildkitdImage=docker.io/earthly/buildkitd:v#{version} -X main.Version=v#{version} -X main.GitSha=60f1b31dae87864e2e9f0bf60d81a9a492e5344e " \
               "-X main.GitSha=#{earthly_gitsha}"
     tags = "dfrunmount dfrunsecurity dfsecrets dfssh dfrunnetwork dfheredoc forceposix"
     system "go", "build",
```

note that we have a `"-X main.GitSha=#{earthly_gitsha}"` which follows the auto-generated `-X main.GitSha=60f1b31dae87864e2e9f0bf60d81a9a492e5344e "` entry.

vladaionescu (approved):
Wasn't the generated version better?

alexcb:
> Wasn't the generated version better?

Not really -- our version of https://github.com/earthly/homebrew-earthly/blob/main/Formula/earthly.rb differs from https://github.com/Homebrew/homebrew-core/blob/master/Formula/earthly.rb

and this auto-generated code makes it impossible to move towards unifying the two.

The homebrew-core version gets updated via a `brew bump-cask-pr`, I think we should aim to slowly replace this custom homebrew-earthly release code to make use of the underlying `bump-cask-pr` code.

PR #2438, opened 2022-11-22
update homebrew-release

alexcb:
The earthly/homebrew-earthly's Formula/earthly.rb has been updated to match the homebrew-core version (which makes use of git rather than downloading a source tar).


alexcb:
This is a follow-up to https://github.com/earthly/earthly/pull/2437#issuecomment-1324249408

our `Formula/earthly.rb` has been changed to mimic the version from homebrew-core. It's wasn't obvious how to pull out the `brew cask-pr` code that performs the bump (without using the full PR, since our process requires a simple push to a `release-*` branch).

vladaionescu (approved):

PR #2444, opened 2022-11-23
fix wait-end SAVE IMAGE/ARTIFACT race

alexcb:
This fixes a race condition where the same target is referenced from
both a FROM or COPY, then a BUILD.

When the converter is run via a FROM or COPY, it produces wait-items
which do not have a doSave (aka localExport) flag set; when it's
re-referenced from a BUILD it must scan through all wait items and set
the flag to true.

### DIFF
--- a/earthfile2llb/state_wait_item.go
+++ b/earthfile2llb/state_wait_item.go
@@ -1,0 +1,26 @@
+package earthfile2llb
+
+import (
+	"github.com/earthly/earthly/util/llbutil/pllb"
+	"github.com/earthly/earthly/util/waitutil"
+)
+
+type stateWaitItem struct {
+	c     *Converter
+	state *pllb.State
+}
+
+// SetDoPush has no effect, but exists to satisfy interface
+func (swi *stateWaitItem) SetDoPush() {
+}
+
+// SetDoSave has no effect, but exists to satisfy interface
+func (swi *stateWaitItem) SetDoSave() {
+}
+
+func newStateWaitItem(state *pllb.State, c *Converter) waitutil.WaitItem {
+	return &stateWaitItem{
+		c:     c,
+		state: state,
+	}
+}
--- a/util/waitutil/wait_block.go
+++ b/util/waitutil/wait_block.go
@@ -1,0 +1,11 @@
+package waitutil
+
+import "context"
+
+// WaitBlock stores items within a WAIT / END block
+type WaitBlock interface {
+	Wait(ctx context.Context, push, save bool) error
+	AddItem(item WaitItem)
+	SetDoSaves()
+	SetDoPushes()
+}
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -220,8 +220,12 @@ func (b *Builder) convertAndBuild(ctx context.Context, target domain.Target, opt
 			}
 		}
 		if opt.GlobalWaitBlockFtr {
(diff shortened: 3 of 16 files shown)


util/waitutil/wait_block.go
```diff
@@ -0,0 +1,11 @@
+package waitutil
+
+import "context"
+
+// WaitBlock needs work
```
alexcb, line 5: :facepalm: gotta update this comment

alexcb:
```suggestion
// WaitBlock stores items within a WAIT / END block
```

earthfile2llb/state_wait_item.go
```diff
@@ -0,0 +1,26 @@
...
+// SetDoSave has no effect, but exists to satisfy interface
+func (swi *stateWaitItem) SetDoSave() {
+}
+
+func newStateWaitItem(state *pllb.State, c *Converter) waitutil.WaitItem {
+	return &stateWaitItem{
+		c:     c,
+		state: state,
+	}
+}
```
vladaionescu, line 26: I would move the WaitItems to their own sub-package, to keep things a little more organized.

earthfile2llb/state_wait_item.go
```diff
@@ -0,0 +1,26 @@
+package earthfile2llb
+
+import (
+	"github.com/earthly/earthly/util/llbutil/pllb"
+	"github.com/earthly/earthly/util/waitutil"
+)
+
+type stateWaitItem struct {
```
vladaionescu, line 8: What is a `stateWaitItem`?

EDIT: Oh I see - it's a state we wait on to finish forced execution within a WAIT block, right?

vladaionescu (approved):
Looks good as far as I can tell!

Just one minor comment.

earthfile2llb/state_wait_item.go
alexcb: yeah it's quite tied to the converter, it was moved out of `wait_block.go` which was getting rather large,

`wait_block.go` contains some code that uses the stateWaitItem's converter reference, such as:

```
return item.c.forceExecution(ctx, *item.state, item.c.platr)
```

(where `c` is the `Converter`)

the higher-level methods of the wait items are defined as an interface under `util/waitutil`, but we need to keep all converter-references out of that to prevent cyclical dependencies.

earthfile2llb/state_wait_item.go
alexcb: can we move all the `wait_block.go` code into `earthfile2llb/waitblock`? it accesses a lot of converter internals.

or are you thinking simply these saveImageWaitItem/saveArtifactWaitItem/saveStateItem structs?

vladaionescu: I was thinking just these wait items. If it's even possible anyway.

alexcb:
```
            +earthly | 	imports github.com/earthly/earthly/earthfile2llb/waititem
            +earthly | 	imports github.com/earthly/earthly/earthfile2llb: import cycle not allowed
```

unfortunately it's not going to work, since `waititem.SaveImageWaitItem` need to reference `earthfile2llb.Converter`, and the converter code needs to access `waititem.NewSaveImageWaitItem`.

vladaionescu: Yeah, ok makes sense.

PR #2445, opened 2022-11-24
Logstream metadata improvements

vladaionescu:
* Set the org and project from the PROJECT command
* Set the main target ID
* Add runner name
* Fix canceled status

### DIFF
--- a/cmd/earthly/build_cmd.go
+++ b/cmd/earthly/build_cmd.go
@@ -187,14 +187,15 @@ func (app *earthlyApp) actionBuildImp(cliCtx *cli.Context, flagArgs, nonFlagArgs
 	}
 
 	// Default upload logs, unless explicitly configured
+	doLogstreamUpload := false
+	var logstreamURL string
 	if !app.cfg.Global.DisableLogSharing {
 		if cloudClient.IsLoggedIn(cliCtx.Context) {
 			if app.logstreamUpload {
-				app.logbusSetup.StartLogStreamer(cliCtx.Context, cloudClient)
-				urlStr := fmt.Sprintf("https://ci.earthly.dev/todourl/%s", app.logbusSetup.InitialManifest.GetBuildId())
-				app.console.Printf("Streaming logs to %s\n", urlStr)
+				doLogstreamUpload = true
+				logstreamURL = fmt.Sprintf("https://ci.earthly.dev/todourl/%s", app.logbusSetup.InitialManifest.GetBuildId())
 				defer func() {
-					app.console.Printf("View logs at %s\n", urlStr)
+					app.console.Printf("View logs at %s\n", logstreamURL)
 				}()
 			} else {
 				// If you are logged in, then add the bundle builder code, and configure cleanup and post-build messages.
@@ -475,6 +476,7 @@ func (app *earthlyApp) actionBuildImp(cliCtx *cli.Context, flagArgs, nonFlagArgs
 		EnableGatewayClientLogging: app.debug,
 		BuiltinArgs:                builtinArgs,
 		LocalArtifactWhiteList:     localArtifactWhiteList,
+		MainStsIDFuture:            make(chan string, 1),
 
 		// feature-flip the removal of builder.go code
 		// once VERSION 0.7 is released AND support for 0.6 is dropped,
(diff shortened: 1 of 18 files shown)


cmd/earthly/main.go
```diff
@@ -617,17 +617,20 @@ func (app *earthlyApp) run(ctx context.Context, args []string) int {
 			}
 		}
 
-		if strings.Contains(err.Error(), "security.insecure is not allowed") {
+		switch {
+		case strings.Contains(err.Error(), "security.insecure is not allowed"):
 			app.logbus.Run().SetFatalError(time.Now(), "", "", logstream.FailureType_FAILURE_TYPE_NEEDS_PRIVILEGED, err.Error())
 			app.console.Warnf("Error: earthly --allow-privileged (earthly -P) flag is required\n")
-		} else if strings.Contains(err.Error(), "failed to fetch remote") {
+			return 9
```
r-LaForge, line 624: Out of curiosity, why is this one `9` and most of the rest `1`?
Should we return a unique number for each case?

earthfile2llb/earthfile2llb.go
```diff
@@ -220,6 +240,14 @@ func Earthfile2LLB(ctx context.Context, target domain.Target, opt ConvertOpt, in
 	if err != nil {
 		return nil, err
 	}
+	if opt.MainTargetDetailsFuture != nil {
+		opt.MainTargetDetailsFuture <- TargetDetails{
+			ID:                 sts.ID,
+			EarthlyOrgName:     bc.EarthlyOrgName,
+			EarthlyProjectName: bc.EarthlyProjectName,
+		}
+		opt.MainTargetDetailsFuture = nil
```
r-LaForge, line 249: Should we close the channel here?
```suggestion
        close(opt.MainTargetDetailsFuture)
		opt.MainTargetDetailsFuture = nil
```

logbus/formatter/formatter.go
```diff
@@ -374,12 +374,14 @@ func (f *Formatter) printBuildFailure() {
...
+	if failure.GetCommandId() != "" {
+		c.Printf("Repeating the failure error...\n")
+		f.printHeader(failure.GetTargetId(), failure.GetCommandId(), tm, cm, true)
+		if len(failure.GetOutput()) > 0 {
+			c.PrintBytes(failure.GetOutput())
+		} else {
+			c.Printf("[no output]\n")
+		}
 	}
 	if failure.GetErrorMessage() != "" {
```
r-LaForge, line 386: nit: We know this is the case now thanks to the check on `364`

r-LaForge:
@vladaionescu Looks like you may have to update the protos in this branch

vladaionescu:
@r-LaForge I'm waiting on proto review too

cmd/earthly/main.go
mikejholly: Are these exit codes? I agree that it would be nice to use named consts or add comments. Minor thing.

earthfile2llb/earthfile2llb.go
vladaionescu: Nope - not needed.

cmd/earthly/main.go
vladaionescu: They are - but it's pretty self-explanatory what they indicate given the warnings printed for them. Nothing really depends on these specific exit codes currently. I figured I would give special number numbers to some specific, more important, situations as a nice-to-have.

cmd/earthly/build_cmd.go
```diff
@@ -489,6 +502,23 @@ func (app *earthlyApp) actionBuildImp(cliCtx *cli.Context, flagArgs, nonFlagArgs
 		buildOpts.OnlyArtifact = &artifact
 		buildOpts.OnlyArtifactDestPath = destPath
 	}
+	// Kick off logstream upload only when we have the main target ID reported back from
```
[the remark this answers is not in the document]
vladaionescu: Currently, some fields cannot be part of deltas and can only be sent via the first reset message. The main target ID and the org + project names are such fields.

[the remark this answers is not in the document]
r-LaForge: The target + org makes sense, but the main target ID is on the `FieldsDelta` proto and applied here: https://github.com/earthly/earthly/blob/033124d837c3afbd371441e941930cdf233db836/util/deltautil/apply.go#L51

r-LaForge (approved):
LGTM - just had a question about the main target id

cmd/earthly/build_cmd.go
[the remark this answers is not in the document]
vladaionescu: Ah yeah - true. In any case - I wouldn't change the implementation - this is still the easiest way to extract the main target ID. Perhaps I should update the comment, though.

PR #2448, opened 2022-11-24
Fix logstream target statuses and output prefix names

vladaionescu:


logbus/run.go
```diff
@@ -11,21 +11,23 @@ import (
...
 	run := &Run{
-		b:        b,
-		targets:  make(map[string]*Target),
-		commands: make(map[string]*Command),
-		generic:  nil, // set below
+		b:          b,
+		targets:    make(map[string]*Target),
+		commands:   make(map[string]*Command),
+		generic:    nil, // set below
+		mainTarget: true,
```
r-LaForge, line 30: nit:
For clarity, we should rename this to `hasMainTarget` as that seems to be the purpose.
(If we did change it, we'd have to negate the boolean).

r-LaForge (approved):
LGTM - pending proto updates

Issue #2447, opened 2022-11-24
conflicting `SAVE IMAGE` commands produce inconsistent results


alexcb:
```Earthfile
VERSION 0.6

a:
  FROM alpine
  RUN echo a > /data
  SAVE IMAGE testimg:latest

b:
  FROM alpine
  RUN echo b > /data
  SAVE IMAGE testimg:latest

all:
  BUILD +a
  BUILD +b
```

Now run it:

```
$ earthly-v0.6.30 +all
 1. Init 🚀
————————————————————————————————————————————————————————————————————————————————

           buildkitd | Found buildkit daemon as docker container (earthly-buildkitd)

 2. Build 🔧
————————————————————————————————————————————————————————————————————————————————

              alpine | --> Load metadata linux/amd64
                  +b | --> FROM alpine
                  +b | [██████████] 100% resolve docker.io/library/alpine@sha256:8914eb54f968791faf6a8638949e480fef81e697984fba772b3976835194c6d4
                  +a | *cached* --> RUN echo a > /data
                  +b | *cached* --> RUN echo b > /data
              output | --> exporting outputs
              output | [██████████] 100% exporting layers
              output | [██████████] 100% exporting manifest sha256:acb3a1060a78285c2b7519dda7e7df43194e97b38a122650b2d028bcdd5ed948
              output | [██████████] 100% exporting config sha256:e773588bc732c488a89644973830841a0633d3c4278fa662736967a512b0a846
              output | [██████████] 100% transferring docker.io/library/testimg:latest
              output | 
 3. Push ⏫ (disabled)
————————————————————————————————————————————————————————————————————————————————

To enable pushing use earthly --push

 4. Local Output 🎁
————————————————————————————————————————————————————————————————————————————————

Image +a output as testimg:latest
Image +b output as testimg:latest


========================== 🌍 Earthly Build  ✅ SUCCESS ==========================

Shareable link: https://ci.earthly.dev/logs?logId=1f999b54-14da-4ff2-8dbe-cc5088271fd0
```

sometimes the image will contain data from `a`:

```
alex@mah:~/test/multi-save-image$ docker run --rm testimg:latest /bin/sh -c 'cat /data'
a
```

and other times it will contain `b`:

```
$ docker run --rm testimg:latest /bin/sh -c 'cat /data'
b
```

however in all cases the `Local Output` portion is always sorted alphabetically.

alexcb:
[...]

PR #2451, opened 2022-11-25
unused args should not be propagated

alexcb:
Propitiating all args, even unused args will cause the same target to be converted multiple times, which would lead to duplicate save image commands.

### DIFF
--- a/earthfile2llb/converter.go
+++ b/earthfile2llb/converter.go
@@ -114,7 +114,6 @@ func NewConverter(ctx context.Context, target domain.Target, bc *buildcontext.Da
 		Final:   sts,
 		Visited: opt.Visited,
 	}
-	sts.AddOverridingVarsAsBuildArgInputs(opt.OverridingVars)
 	newCollOpt := variables.NewCollectionOpt{
 		Console:          opt.Console,
 		Target:           target,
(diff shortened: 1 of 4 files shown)


alexcb:
fixes #2447

alexcb:
Without this change, when `Earthfile2LLB` is called, the args `myarg=notused` and `myarg=unused` cause the underlying

```
same, err := CompareTargetInputs(target, platr, allowPrivileged, overridingVars, sts.TargetInput())
```

to return `same=false` when evaluating `BUILD +a`, even though `myarg` isn't defined under the `a` target.

This then bubbles up to `found` being returned as `false`, which causes a second `SAVE IMAGE` to get evaluated as a result.

alexcb:
Any thoughts if there's some additional tests that we should write to ensure this doesn't break other things?

tests/wait-block/save-image-indirect-unused-args/Earthfile
```diff
@@ -0,0 +1,21 @@
...
+indirect:
+  FROM alpine:3.15
+  ARG myarg
+  BUILD +a
+
+test:
+  WAIT
+    BUILD +indirect --myarg=notused
+    BUILD +indirect --myarg=unused
+  END
```
vladaionescu, line 21: What is the expected outcome of this test. From the looks of it, it should result in an error?

earthfile2llb/converter.go
```diff
@@ -114,7 +114,6 @@ func NewConverter(ctx context.Context, target domain.Target, bc *buildcontext.Da
 		Final:   sts,
 		Visited: opt.Visited,
 	}
-	sts.AddOverridingVarsAsBuildArgInputs(opt.OverridingVars)
```
vladaionescu, line 117: This doesn't look right..?

tests/wait-block/save-image-indirect-unused-args/Earthfile
alexcb: Without this fix, we get the error:

```
Error: build target: build main: failed to solve: SAVE IMAGE earthly/testimg:f8874c9c-50bd-4634-aa5f-c53f26b69b87 was already declared (none had --platform values)
```

earthfile2llb/converter.go
alexcb: They should be added in via `buildTarget`'s `propagateBuildArgs`, but perhaps there's other places we would need to add them in.

```
if !reserved.IsBuiltIn(bai.Name) {
				c.mts.Final.AddBuildArgInput(bai)
}
```

PR #2453, opened 2022-11-25
Allow files to be greater than 64kB

### LINKED ISSUE #2452: Artifact saving in FINALLY is buggy
I want to save a test report even on failure, so I'm trying to use `--try` from #587. If the artifact is not tiny, I see it getting cropped somehow! Check it out:

```
VERSION --try 0.6
FROM mambaorg/micromamba:jammy

works:
  RUN head -c 500k < /dev/zero > hello
  RUN ls -l hello
  TRY
    RUN true
  FINALLY
    SAVE ARTIFACT hello AS LOCAL hello
  END

buggy:
  RUN head -c 500k < /dev/zero > hello
  RUN ls -l hello
  TRY
    RUN false
  FINALLY
    SAVE ARTIFACT hello AS LOCAL hello
  END
```

Output:

```
$ rm -f hello; earthly --no-cache +works; ls -l hello
              +works | --> RUN head -c 500k < /dev/zero > hello
              +works | --> RUN ls -l hello
              +works | -rw-r--r-- 1 mambauser mambauser 512000 Nov 25 20:49 hello
              +works | --> RUN true
              +works | --> SAVE ARTIFACT hello +works/hello AS LOCAL hello
-rw-r--r-- 1 darabos darabos 512000 Apr 16  2020 hello
```

```
$ rm -f hello; earthly --no-cache +buggy; ls -l hello
              +buggy | --> RUN head -c 500k < /dev/zero > hello
              +buggy | --> RUN ls -l hello
              +buggy | -rw-r--r-- 1 mambauser mambauser 512000 Nov 25 20:45 hello
              +buggy | --> RUN false
              +buggy | ERROR Earthfile line 17:4
              +buggy |       The command
              +buggy |           RUN false
              +buggy |       did not complete successfully. Exit code 1
-rw-r--r-- 1 darabos darabos 53248 Nov 25 21:45 hello
```

It still manages to create `hello` on failure, but it's just 53,248 bytes! Where's the rest of my file?

For sizes up to 65,535 bytes it works. For sizes larger than that I get (SIZE % 65536) bytes.

Looks like an all-zero file is kind of a special case. With `head -c 500k < /dev/random` the file doesn't get created if it's longer than 65,535 bytes. And this is what I see in my real use case with an 8 MB zip file.

darabos:
Fixes #2452.

At first I tried using Read/WriteDataPacket in a loop, but this is simpler.

I have no idea about tests, sorry. But it works on my test case when I try it manually.

### DIFF
--- a/cmd/debugger/main.go
+++ b/cmd/debugger/main.go
@@ -117,7 +117,7 @@ func sendFile(ctx context.Context, sockAddr, src, dst string) error {
 		return err
 	}
 
-	err = common.WriteDataPacket(conn, len(b), b)
+	err = common.WriteFileTransfer(conn, b)
 	if err != nil {
 		return err
 	}
--- a/cmd/earthly/build_cmd.go
+++ b/cmd/earthly/build_cmd.go
@@ -549,7 +549,7 @@ func getTryCatchSaveFileHandler(localArtifactWhiteList *gatewaycrafter.LocalArti
 		}
 
 		// data
-		_, data, err := debuggercommon.ReadDataPacket(conn)
+		data, err := debuggercommon.ReadFileTransfer(conn)
 		if err != nil {
 			return err
 		}
--- a/debugger/common/common.go
+++ b/debugger/common/common.go
@@ -4,6 +4,9 @@ import (
 	"bytes"
 	"encoding/binary"
 	"io"
+	"math"
+
+	"github.com/pkg/errors"
 )
 
 //******************************************************************************************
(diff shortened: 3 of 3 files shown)


alexcb:
Thanks for the contribution, I gave it a test and it works great locally; however, we will need to make this change backwards compatible with older versions of (remote) buildkit instances (e.g. satellites).

However this will require a brand new versioning scheme to be implemented; 

https://github.com/earthly/earthly/blob/4d6b0fe99c2dc02418e4c9409f8c6695a302428a/debugger/terminal/terminal_other.go#L44

will have to first read an int (outside of the loop), and if it's a `1` (`StartShellSession`), then we'll jump to legacy mode, otherwise we'll enter into some sort of capabilities-connection-handshake, which could be used to determine if we support large files.

I'm happy to take this PR over and work on the backwards logic if it's too big of an ask.

darabos:
> I'm happy to take this PR over and work on the backwards logic if it's too big of an ask.

Oh, yes, please! I don't know enough about this. Thank you!

alexcb:
> > I'm happy to take this PR over and work on the backwards logic if it's too big of an ask.
> 
> Oh, yes, please! I don't know enough about this. Thank you!

It turns out that we already had a protocol versioning scheme implemented at the send file layer; for version 2 we'll send files over in 64k chunks: #2458

PR #2458, opened 2022-11-28
fix: support saving large files on failure

### LINKED ISSUE #2452: Artifact saving in FINALLY is buggy

alexcb:
Fixes #2452

### DIFF
--- a/cmd/debugger/main.go
+++ b/cmd/debugger/main.go
@@ -1,9 +1,12 @@
 package main
 
 import (
+	"bufio"
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
+	"math"
 	"net"
 	"os"
 	"os/exec"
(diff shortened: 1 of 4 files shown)


cmd/debugger/main.go
```diff
@@ -21,6 +23,8 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+const uint16Max = 65535
```
alexcb, line 26: does this exist in an official go package? I couldn't find it anywhere.

vladaionescu (approved):

PR #2459, opened 2022-11-29
Perform semver compatibility check between Earthly and Buildkit

vladaionescu:


util/semverutil/semverutil_test.go
```diff
@@ -0,0 +1,47 @@
...
+		{"1.2.3-alpha.1", args{"1.2.3-alpha.1"}, Version{1, 2, 3, "-alpha.1"}, false},
+		{"v1.2.3-alpha.1+001", args{"v1.2.3-alpha.1+001"}, Version{1, 2, 3, "-alpha.1+001"}, false},
+		{"1.2.3-alpha.1+001", args{"1.2.3-alpha.1+001"}, Version{1, 2, 3, "-alpha.1+001"}, false},
+		{"v1.2.3+001", args{"v1.2.3+001"}, Version{1, 2, 3, "+001"}, false},
+		{"1.2.3+001", args{"1.2.3+001"}, Version{1, 2, 3, "+001"}, false},
+		{"not-a-version", args{"not-a-version"}, Version{}, true},
+		{"v.1.2", args{"v.1.2"}, Version{}, true},
+		{"v1.2", args{"v1.2"}, Version{}, true},
+		{"1.2", args{"1.2"}, Version{}, true},
+		{"1", args{"1"}, Version{}, true},
```
vladaionescu, line 33: Fun fact: GH Copilot generated most of these.

r-LaForge (approved):

util/semverutil/semverutil.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+// String returns the string representation of the version.
+func (v Version) String() string {
+	return fmt.Sprintf("v%d.%d.%d%s", v.Major, v.Minor, v.Patch, v.Tail)
+}
+
+// IsCompatible returns true if the two versions are compatible.
+func IsCompatible(a, b Version) bool {
+	return a.Major == b.Major && a.Minor == b.Minor
```
alexcb, line 40: based on the name `IsCompatible`, is your idea that we would eventually replace the logic in this function to list known pairs that are compatible?

buildkit already provides a capabilities feature under https://github.com/earthly/buildkit/blob/earthly-main/solver/pb/caps.go

We've extended this in the past to force the client to require the `CapExecMountSock` capability: e.g. https://github.com/earthly/buildkit/commit/df14b0779eabc92481915e8a87d88065253832ba

I worry that having two mechanisms for this will be confusing.

vladaionescu: True... Not sure what the right solution is.

I would like to show a warning of some sort if Earthly and Buildkit have drastically different versions. I think that today we don't have a good way to track every buildkit change through capabilities (e.g. changes to the debugger, or changes to the WITH DOCKER wrapper script), so I think that this is not a good use-case for capabilities.

I figured that using major + minor comparison would give us a rough idea about whether Earthly and Buildkit would be close in terms of version number (e.g. 1.2.3 is close to 1.2.5, but not to 1.3.3). I wasn't thinking of extending this with specific version pairs - just to do semver comparison for this warning message only, and not for detecting support for specific pieces of functionality (where I think capabilities are the way to go).

alexcb: If we don't want to expand this to suggest stable pairings, then this seems manageable.

I suspect we've abused our patch number where we really should have increased our minor version.

Ultimately if this will  just issue a warning if, for instance, someone runs 0.6 against 0.5, that seems fine.

alexcb (approved):

PR #2460, opened 2022-11-29
lint for consistent import orders

alexcb:
This enables the gci linter which enforces imports are grouped consistently; it is currently set to be:

    standard

    github.com/earthly/earthly

    github.com/earthly/cloud-api

    3rd party imports

The imports can be automatically re-written with

    gci write --skip-generated --custom-order -s 'standard,prefix(github.com/earthly/earthly),prefix(github.com/earthly/cloud-api),default' .

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,2 +1,15 @@
 run:
   timeout: 10m
+
+linters:
+  enable:
+  - gci
+
+linters-settings:
+  gci:
+    custom-order: true
+    sections:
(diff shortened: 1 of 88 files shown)


vladaionescu:
IDEs have auto-formatters like these set up

Screenshot 2022-11-29 at 12 50 43 PM

It might be hard for everyone to maintain a specific import order that is not automatically supported by one of these more standard formatters.

PR #2462, opened 2022-11-30
Add builtin support for git author timestamps

### LINKED ISSUE #2432: Have a git commit date in addition to the git author date
Earthly provides a `EARTHLY_GIT_COMMIT_TIMESTAMP` to make builds that depends on commit which is quite nice

However, to each commit, there are actually two timestamp:
- a author date (what `EARTHLY_GIT_COMMIT_TIMESTAMP` is actually pointing to): when someone first creates a commit with `git commit`
- a commit date: the last edit of the commit (either through `git rebase` or `git commit --amend`)

In our workflow that relies on rebasing branches, this would be useful to have access to both timestamps

miles170:
Closes #2432.

Add `EARTHLY_GIT_COMMIT_AUTHOR_TIMESTAMP`, which points to the git author timestamp.


alexcb (changes requested):
this is awesome, thanks for the contribution!

Can you create a new feature flag for this? that way users will have to explicitly enable it via `VERSION --git-commit-author-timestamp 0.6`

that way if an Earthfile depends on this feature, and a user is using an old version of earthly, the user will get an error such as "unknown feature-flag git-commit-author-timestamp" rather than have the ARG set to `""`.

Check out this commit (in particular the addition to `features/features.go`) to see how we've done this for previous newly introduced ARGs:

https://github.com/earthly/earthly/commit/e20c10483#diff-a8b6135d50534471326ea7bcd20e0f5eae25353f7788338060f718128a6a0b34R30

alexcb (approved):
thanks for the changes!

PR #2465, opened 2022-12-01
Fix source location missing in logstream manifest

vladaionescu:


logbus/run.go
```diff
@@ -85,16 +85,18 @@ func (run *Run) NewCommand(commandID string, command string, targetID string, ca
...
-				IsCached:       cached,
-				IsLocal:        local,
-				SourceLocation: sourceLocationToProto(repoURL, repoHash, fileRelToRepo, sourceLocation),
+				Name:              command,
+				TargetId:          targetID,
+				Category:          category,
+				Platform:          platform,
+				IsCached:          cached,
+				IsLocal:           local,
+				HasSourceLocation: (sl != nil),
```
vladaionescu, line 98: This line is the key change.

r-LaForge (approved):

PR #2466, opened 2022-12-01
Update the logstream logs URL

vladaionescu:
* Also, fix a race detector issue that randomly popped up.

### DIFF
--- a/cmd/earthly/build_cmd.go
+++ b/cmd/earthly/build_cmd.go
@@ -188,7 +188,11 @@ func (app *earthlyApp) actionBuildImp(cliCtx *cli.Context, flagArgs, nonFlagArgs
 		if cloudClient.IsLoggedIn(cliCtx.Context) {
 			if app.logstreamUpload {
 				doLogstreamUpload = true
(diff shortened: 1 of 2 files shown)


cmd/earthly/build_cmd.go
```diff
@@ -188,7 +188,11 @@ func (app *earthlyApp) actionBuildImp(cliCtx *cli.Context, flagArgs, nonFlagArgs
 		if cloudClient.IsLoggedIn(cliCtx.Context) {
 			if app.logstreamUpload {
 				doLogstreamUpload = true
-				logstreamURL = fmt.Sprintf("https://ci.earthly.dev/todourl/%s", app.logbusSetup.InitialManifest.GetBuildId())
+				if strings.Contains(app.cloudGRPCAddr, "staging") {
```
vladaionescu, line 191: This is a hack... but I think it's good enough for now. Thoughts?

r-LaForge (approved):

cmd/earthly/build_cmd.go
r-LaForge: This works for now.
The server may return this url as well at some point.

Issue #2469, opened 2022-12-02
ENV variables in secret targetpath 


danielschlegel:
I try to mount a secret as a file inside the $HOME directory. Unfortunately this does not work. 

```
  RUN echo $HOME # outputs /root
  RUN --mount=type=secret,target=/root/.ssh/id_rsa_akeneo,id=+secrets/AKENEO_SSH_KEY \ 
      cat /root/.ssh/id_rsa_akeneo  # works
  RUN --mount=type=secret,target=$HOME/.ssh/id_rsa_akeneo,id=+secrets/AKENEO_SSH_KEY \ 
      cat $HOME/.ssh/id_rsa_akeneo # does not work
```

I assume that earthly is not replacing the env variable $HOME in the target.

alexcb:
I would have expected this to work with the `VERSION --shell-out-anywhere 0.6` feature enabled, but even then this doesn't seem to work. Thanks for raising the issue.

alexcb:
It's interesting to note that the `HOME` env does **not** exist in a standard image such as alpine; as we can see by running

```
$ docker image inspect alpine | jq '.[].Config.Env'
```

which only shows that `PATH` is set in the alpine image:

```
[
  "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
]
```

Under the hood, when earthly is run with the `--shell-out-anywhere` feature, if it detects a var such as `$KEY`, it simply looks up the equivalence of `Config.Env`.

However if the shell-out-anywhere lexer encounters `$(....)`, it will shell out to the container, and execute a command within `/bin/sh`; and it appears that `/bin/sh` sets additional envs upon execution (e.g. `HOME`, `SHLVL`, `TERM`, `PWD`).

so as a work-around, you can force the run mount path to be expanded using a shellout by running `$(echo $HOME)`.

Here's a example:

```
VERSION --shell-out-anywhere 0.6
test:
  FROM alpine
  RUN --mount=type=secret,target="$(echo $HOME)/.ssh/id_rsa",id=+secrets/mykey,mode=0600 \
    ls -la $HOME/.ssh/id_rsa && cat $HOME/.ssh/id_rsa
```

Then running `earthly --no-cache --secret mykey=helloworld +test` produces:

```
              alpine | --> Load metadata alpine linux/amd64
               +test | --> FROM alpine
[...]

PR #2475, opened 2022-12-05
automatic shell-out for $HOME expansion

alexcb:
The HOME env is set by the shell and not in the docker image; if a user references $HOME we will automatically shellout to determine the value.

### DIFF
--- a/variables/collection.go
+++ b/variables/collection.go
@@ -15,6 +15,10 @@ import (
 	specs "github.com/opencontainers/image-spec/specs-go/v1"
 )
 
+var ShellOutEnvs = map[string]struct{}{
+	"HOME": struct{}{},
+}
+
 type stackFrame struct {
 	frameName string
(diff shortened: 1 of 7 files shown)


alexcb:
fixes #2469

variables/collection.go
```diff
@@ -15,6 +15,14 @@ import (
 	specs "github.com/opencontainers/image-spec/specs-go/v1"
 )
 
+var ShellOutEnvs map[string]struct{}
+
+func init() {
+	ShellOutEnvs = map[string]struct{}{
+		"HOME": struct{}{},
+	}
+}
```
vladaionescu, line 24: I believe that you can initialize the map without the use of `init`. See e.g. https://stackoverflow.com/a/41078296/1353666

vladaionescu (approved):
One minor comment...

PR #2499, opened 2022-12-12
Add a web command

r-LaForge:
This command will attempt to open the web UI with query parameters set.
If the provider is set, the UI will start the flow for that provider automatically.
If the user is logged in via Earthly CLI the GitHub login should be merged with the existing user.

TODO:
 - [x] Manual testing
 - [x] Add some tests.
 - [x] Documentation in our `next` branch -> added an internal ticket to come back to if we don't make changes to this command.


cmd/earthly/common.go
```diff
@@ -26,6 +26,14 @@ func (app *earthlyApp) newCloudClient() (cloud.Client, error) {
 	return cloudClient, nil
 }
 
+// returns protocol:hostname
```
vladaionescu, line 29:
```suggestion
// getCIHost returns protocol://hostname
```

r-LaForge:
I'm not sure what is wrong with the Podman tests and other tests. They appear unrelated to the changes here.
I'm taking a look though.

cmd/earthly/root_cmds.go
```diff
@@ -311,6 +311,30 @@ Set up a whole custom git repository for a server called example.com, using a si
...
+					Required:    false,
+					Destination: &app.loginProvider,
+				},
+				&cli.StringFlag{
+					Name:        "final",
+					EnvVars:     []string{"EARTHLY_LOGIN_FINAL"},
+					Usage:       "The final destination after accessing the web ui",
+					Required:    false,
+					Destination: &app.loginFinal,
+				},
```
vladaionescu, line 334: I'm not sure I understand the function of this flag. What are other possible values for it?

(Forgive my ignorance if this was already covered in previous conversations).

cmd/earthly/root_cmds.go
```diff
@@ -311,6 +311,30 @@ Set up a whole custom git repository for a server called example.com, using a si
 				},
 			},
 		},
+		{
+			Name:        "web-ui",
```
vladaionescu, line 315: Shouldn't this have been under `account`? `earthly account web-ui`?

cmd/earthly/root_cmds.go
```diff
@@ -311,6 +311,30 @@ Set up a whole custom git repository for a server called example.com, using a si
...
+			Description: "Access the web UI",
+			Usage:       "Access the web UI via your default browser and print the url",
+			UsageText:   "earthly web-ui (--provider=github|--final=ci)",
+			Hidden:      true,
+			Flags: []cli.Flag{
+				&cli.StringFlag{
+					Name:        "provider",
```
vladaionescu, line 322: Is this flag even needed? I imagine that the user would be able to just select in the UI itself the form of login, right?

r-LaForge: If this flag is provided, and the provider is valid, we attempt to log the user in automatically - without the need to click a button.
If / when we add more providers this will also work with those.

As-is, I can run
```
earthly web-ui --provider=github
```
to open the web ui and attempt a login via GitHub. 

If / when we add more providers we can simply use:
```
earthly web-ui --provider=bitbucket
```
without zero changes to the CLI.

cmd/earthly/root_cmds.go
r-LaForge: There was a minor remark in the internal Slack thread where simply `earthly web-ui` was proposed.
I opted for that, but open to change it.

cmd/earthly/root_cmds.go
r-LaForge: Right now the only valid values are `information` and `auth`.
Think of it as where a user should be redirected to after login.

vladaionescu: I wonder if this should be a URL instead? Maybe with some validation server-side such that the URL can only be within Earthly. But I'm thinking that a URL gives us the most flexibility.

cmd/earthly/root_cmds.go
vladaionescu: I'm kinda inclined towards the more structured option, under `account`, but it's not a strong opinion.

Separately, thoughts on using just `web`? Like `earthly web` or `earthly account web`? Feels cleaner.

cmd/earthly/root_cmds.go
vladaionescu: Ok sounds good

vladaionescu: Ok sounds good

cmd/earthly/root_cmds.go
r-LaForge: Funny enough, I was having an internal debate when I was writing this and just `earthly web` crossed my mind.
I'm open to it.

cmd/earthly/root_cmds.go
r-LaForge: While I like that idea (it's fairly common in login scenarios as `nextUrl`) I think that decision is outside of the scope of this PR because it already exists in our cloud backend.
If we want to add support for a url instead, I'm completely open to discussing that internally.

For this PR, I think we should either:
- Include it as-is because it's already supported.
- Decide to omit it and wait to discuss our `nextUrl` implementation.

cmd/earthly/root_cmds.go
r-LaForge: Votes are in - `web` it is! :)

cmd/earthly/root_cmds.go
vladaionescu: I think omitting it makes the most sense, if it only has one possible value and also we're planning on replacing it anyway.

r-LaForge: It has been removed.

scripts/tests/web.sh
```diff
@@ -0,0 +1,66 @@
...
+
+earthly=${earthly:=earthly}
+earthly=$(realpath "$earthly")
+echo "running tests with $earthly"
+
+## Test that we display web link correctly without being logged in and with no params
+NO_COLOR=0 "$earthly" web 2>&1 | grep -qE "https://ci(-beta)?.earthly.dev/login"
+RESULT=$?
+if [ $RESULT -ne 0 ]; then
+  echo "failed without login and no arguments"
```
alexcb, line 12: how is this possible, there's a `set -e` above?

scripts/tests/web.sh
```diff
@@ -0,0 +1,66 @@
...
+fi
+
+# Test with argument passed for provider
+NO_COLOR=0 "$earthly" web --provider github 2>output-file
+output=$(cat output-file)
+rm output-file
+
+RESULT=$?
+if [ $RESULT -ne 0 ]; then
+  echo "failed without login provider argument passed"
```
alexcb, line 23: in addition to there being a `set -e`, this appears to be testing if the `rm` failed.

scripts/tests/web.sh
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env bash
```
alexcb, line 1: general comment on this test: if we run this test within earthly (i.e. testing earthly in earthly), then we should do that instead. Typically the outside of earthly tests are only required when the embedded earthly makes use of `WITH DOCKER`.

alexcb (changes requested):
seems reasonable, but is it possible to move the test inside earthly?

cmd/earthly/web_cmds.go
```diff
@@ -0,0 +1,50 @@
...
+	if token != "" {
+		query.Set("token", token)
+	}
+
+	urlToOpen.RawQuery = query.Encode()
+	urlString := urlToOpen.String()
+
+	err = browser.OpenURL(urlString)
+	if err != nil {
+		app.console.Printf("failed to open UI in browser")
```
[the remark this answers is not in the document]
r-LaForge: Yes, we can include the error here

r-LaForge:
@alexcb I've moved the test under `tests//web/Earthfile` I think this should address your comments

tests/web/Earthfile
```diff
@@ -0,0 +1,21 @@
+VERSION 0.6
+
+FROM ../..+earthly-integration-test-base --DOCKERHUB_AUTH=false
```
alexcb, line 3: can you follow the same pattern we use elsewhere:

https://github.com/earthly/earthly/blob/main/tests/config/Earthfile#L3-L13

i.e. passing along the dockerhub mirror credentials.

without this we will run into rate limiting issues.

r-LaForge: done

cmd/earthly/web_cmds.go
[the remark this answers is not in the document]
r-LaForge: included the error message.

tests/web/Earthfile
```diff
@@ -0,0 +1,21 @@
...
+WORKDIR /test
+
+test:
+    RUN earthly web 2>&1 | grep -qE "https://ci(-beta)?.earthly.dev/login"
+    RUN earthly web --provider github 2>output-file \
+        && output=$(cat output-file) \
+        && rm output-file \
+        && if [[ "$output" != *"provider=github"* ]]; then echo "failed to find provider=github in output" && exit 1; fi
+
+    RUN --secret EARTHLY_TOKEN=+secrets/earthly-technologies/earthly/fake-user/token earthly web 2>&1 | grep -qE "https://ci(-beta)?.earthly.dev/login\?token=(.*)+"
```
alexcb, line 26: I think there's a bug in the regex, `(.*)` could match an empty string, then `(.*)+` exepects one or more empty strings.

should it be this instead?
```suggestion
    RUN --secret EARTHLY_TOKEN=+secrets/earthly-technologies/earthly/fake-user/token earthly web 2>&1 | grep -qE "https://ci(-beta)?.earthly.dev/login\?token=.+"
```

alexcb (approved):

r-LaForge:
Podman tests are being looked at separately.

vladaionescu (approved):

r-LaForge:
There appears to be a `git-metadata` test failing that I think is unrelated to the changes in this PR: 
image

PR #2509, opened 2022-12-15
Support for Satellite Auto-Updates

brandonSc:
* Supports new flags in `satellite launch` flags to control features related to auto-update
* Adds a new `satellite update` command
* Can show an 'Updating' state in case of starting a build during an update
* Ensure satellite name is shown in satellite cmds (requested by a user https://github.com/earthly/earthly/issues/2578)


cmd/earthly/satellite_cmds.go
```diff
@@ -50,6 +50,23 @@ func (app *earthlyApp) satelliteCmds() []*cli.Command {
 					Hidden:      true,
 					Destination: &app.satelliteFeatureFlags,
 				},
+				&cli.StringFlag{
+					Name:    "maintenance-window",
```
brandonSc, line 54: Wondering if we should make this flag `--maintenance-window-start` to leave room for `--maintenance-window-end` if we decide to support that later. It may be a breaking change to switch that later.

alexcb: > It's a bit verbose of a flag name though

tab completion makes it manageable.

I think it's fine as-is, we can always change it later if we absolutely want to.

alexcb: although it does seem a bit odd since `window` implies it should be a range.

alternatively we could expand the accepted value from being `HH:MM` to also, one day, accept `HH:MM-HH:MM`

alexcb (approved):

dchw (approved):

alexcb (approved):

cmd/earthly/satellite_cmds.go
brandonSc: Yeah, I like this option the best. We thought we might expand that string to support more details eventually anyway (like certain days of the week, for example)

PR #2511, opened 2022-12-15
Don't pass --platform to docker/podman

alexcb:
The `--platform` flag is only ever passed to docker/podman when both the
user and server platform values are equal. When they mismatch, the
`--platform` flag is ommitted, and the server's native platform is used.

Since the `--platform` value is only ever passed when the user and
server platforms are equal, it shouldn't matter; however in practice
there is a podman bug, which causes a pull to occur whenver the
`--platform` flag is specified: https://github.com/containers/podman/issues/15711

This bug will cause podman to always pull the earthly/buildkitd image from
docker hub, which will either 1) overwrite the local image if the image exists in docker hub,
and ultimately will cause our tests to run against an incorrect image
version, or 2) result in the following 404-error if the tag does not
exist:

    exit status 125: Trying to pull docker.io/earthly/buildkitd:dev-HEAD...
    Error: initializing source docker://earthly/buildkitd:dev-HEAD: reading manifest dev-HEAD in docker.io/earthly/buildkitd: manifest unknown: manifest unknown: exit status 125

This commit additionally updates the podman tests to use capsh rather
than grep for the `RUN --privileged` tests, as the Effective capabilities (CapEff)
bits are not always the same between docker and podman

### DIFF
--- a/.github/workflows/reusable-test.yml
+++ b/.github/workflows/reusable-test.yml
@@ -61,7 +61,7 @@ jobs:
         if: inputs.BINARY == 'docker'
       - name: remove Docker
         run: ${{inputs.SUDO}} apt-get purge docker-engine docker docker.io docker-ce docker-ce-cli ; ${{inputs.SUDO}} rm -rf /usr/bin/docker
-        if: inputs.binary == 'podman'
+        if: inputs.binary != 'docker'
       - name: Install Podman (with apt-get)
         run: ${{inputs.SUDO}} apt-get update && ${{inputs.SUDO}} apt-get install -y podman
         if: inputs.binary == 'podman'
--- a/ast/tests/privileged.ast.json
+++ b/ast/tests/privileged.ast.json
@@ -7,6 +7,16 @@
         ],
         "name": "FROM"
       }
+    },
+    {
+      "command": {
+        "args": [
+          "apk",
+          "add",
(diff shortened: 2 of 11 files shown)


alexcb:
I tested this on the m1:

under `/tmp/uname/Earthfile`:

```
FROM alpine
RUN uname -a
```

Then I ran:

```
./earthly +for-darwin-m1 && ./build/darwin/arm64/earthly +for-darwin-m1
export DOCKER_HOST=ssh://alex@<ip>:<magic-port>
./build/darwin/arm64/earthly --verbose /tmp/uname+base
```

which produced

```
/tmp/uname+base | Linux buildkitsandbox 5.19.0-26-generic #27-Ubuntu SMP PREEMPT_DYNAMIC Wed Nov 23 20:44:15 UTC 2022 x86_64 Linux
```

alexcb:
and when running on the `m1` without remote docker host, I get linux/arm64:

```
Linux buildkitsandbox 5.15.49-linuxkit #1 SMP PREEMPT Tue Sep 13 07:51:32 UTC 2022 aarch64 Linux
```

nelsam (approved):
Everything seems reasonable to my (untrained) eyes. I particularly like the capsh changes to scrap the magic hex value.

PR #2512, opened 2022-12-16
add export target

idelvall:
Needed to reference the docs from an external Earthfile

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -632,7 +632,3 @@ npm-update-all:
         RUN cd $nodepath && npm update
         SAVE ARTIFACT --if-exists $nodepath/package-lock.json AS LOCAL $nodepath/package-lock.json
     END
-
-export:
-  COPY docs docs
-  SAVE ARTIFACT docs docs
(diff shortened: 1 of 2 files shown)


Earthfile
```diff
@@ -632,3 +632,7 @@ npm-update-all:
         RUN cd $nodepath && npm update
         SAVE ARTIFACT --if-exists $nodepath/package-lock.json AS LOCAL $nodepath/package-lock.json
     END
+
+export:
```
vladaionescu, line 636: Should this go under the `docs` subdir instead?

idelvall: > another potential idea is to just use `GIT CLONE github.com/earthly/earthly/docs .` in the external Earthfile.

Thanks Alex. I didn't realize that option. In this particular case the artifact reference is preferred for that external Earthfile since it is more versatile and can be parameterized to point to other artifacts not necessarily on GIT

idelvall: > Should this go under the `docs` subdir instead?

Changing it

idelvall:
Addressed feedback

alexcb (changes requested):
If we're just simply copying files, I'm much more of favour of using `GIT CLONE github.com/earthly/earthly/docs .` in the project that's copying them in (and simply close this PR). This would reduce coupling.

vladaionescu:
Either way is fine with me. There is anyway some implied coupling because the project that imports this anyway makes assumptions about the location and structure of the `docs` folder. 🤷

alexcb:
@idelvall can we close this, or is this still needed?

Issue #1954, opened 2022-06-22
LOCALLY commands don't seem to run in parallel


mrjgreen:
Use case: 

- I want to run a single command to start my application locally. 
- The application consists of a backend API and a front end UI.
- I want the commands to run locally to benefit from hot reloading.

I have defined earthfiles in the root and the two project folders that look like something like this:

```
// ./Earthfile
start:
    BUILD ./api+start
    BUILD ./ui+start
```
```
// ./api/Earthfile
start:
    LOCALLY
    RUN yarn start
```
```
// ./ui/Earthfile
start:
    LOCALLY
    RUN yarn start
```

#### Expected behaviour
I would expect the two independent build commands to run concurrently, starting both applications and remaining running until the command is cancelled in the terminal.

#### Actual behaviour
The commands run sequentially. Because the commands are blocking, the second command never runs.

#### Simple test case

```
VERSION 0.6

FROM busybox:latest

start:
    BUILD +test
    BUILD +lint

test:
    RUN sleep 5 && echo "Done Test"

lint:
    RUN sleep 5 && echo "Done Lint"


start-locally:
    BUILD +test-locally
    BUILD +lint-locally

test-locally:
    LOCALLY
    RUN sleep 5 && echo "Done Test"

lint-locally:
    LOCALLY
    RUN sleep 5 && echo "Done Lint"
```

`+start` Runs commands in parallel taking about 8 seconds

```
$ time earthly --no-cache +start

...

      busybox:latest | --> Load metadata linux/amd64
               +base | --> FROM busybox:latest
               +base | [██████████] 100% resolve docker.io/library/busybox:latest@...
               +test | --> RUN sleep 5 && echo "Done Test"
               +lint | --> RUN sleep 5 && echo "Done Lint"
             ongoing | +test (4 seconds ago), +lint (4 seconds ago)
               +lint | Done Lint
               +test | Done Test
              output | --> exporting outputs

========================== 🌍 Earthly Build  ✅ SUCCESS ==========================

real	0m8.834s
user	0m1.673s
sys	0m0.908s
```

`+start-locally` Runs commands sequential taking about 15 seconds

```
$ time earthly --no-cache +start-locally

      busybox:latest | --> Load metadata linux/amd64
       +lint-locally *local* | --> RUN sleep 5 && echo "Done Lint"
       +test-locally *local* | --> RUN sleep 5 && echo "Done Test"
[...]

PR #2515, opened 2022-12-19
Add support for opt-in parallelism for LOCALLY commands

Alphasite:
This is intended to resolve: [!1954](https://github.com/earthly/earthly/issues/1954)

The default behaviour for LOCALLY is to execute each step in serial, as its not possible to infer from context if the task is parallel safe, this commit adds a flag which allows you to mark your LOCALLY block as safe for parallel execution and speeds up the relevant tasks.

### DIFF
--- a/earthfile2llb/flags.go
+++ b/earthfile2llb/flags.go
@@ -125,3 +125,7 @@ type pipelineOpts struct {
 type cacheOpts struct {
 	Sharing string `long:"sharing" description:"The cache sharing mode: locked (default), shared, private"`
 }
+
+type locallyOpts struct {
+	Parallel bool `long:"allow-parallel" description:"Allow this locally command to run in parallel with other commands"`
+}
--- a/earthfile2llb/interpreter.go
+++ b/earthfile2llb/interpreter.go
@@ -37,6 +37,7 @@ type Interpreter struct {
 	isWith          bool
 	pushOnlyAllowed bool
 	local           bool
+	denyParallel    bool
 	allowPrivileged bool
 
 	withDocker    *WithDockerOpt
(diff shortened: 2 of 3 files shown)


earthfile2llb/interpreter.go
```diff
@@ -37,6 +37,7 @@ type Interpreter struct {
 	isWith          bool
 	pushOnlyAllowed bool
 	local           bool
+	denyParallel    bool
```
Alphasite, line 40: The default value of a bool is false, so this should default to allowing parallelism, unless explicitly prohibited, as it is for `LOCALLY` blocks (without `--allow-parallelism`.

earthfile2llb/interpreter.go
```diff
@@ -124,7 +125,7 @@ func (i *Interpreter) handleBlock(ctx context.Context, b spec.Block) error {
 }
 
 func (i *Interpreter) handleBlockParallel(ctx context.Context, b spec.Block, startIndex int) error {
-	if i.local {
+	if i.denyParallel {
```
vladaionescu, line 128: I wouldn't change this part, since we don't want to parallelize the commands within a single LOCALLY target. We want to parallelize multiple unrelated LOCALLY targets.

earthfile2llb/flags.go
```diff
@@ -125,3 +125,7 @@ type pipelineOpts struct {
 type cacheOpts struct {
 	Sharing string `long:"sharing" description:"The cache sharing mode: locked (default), shared, private"`
 }
+
+type locallyOpts struct {
+	Parallel bool `long:"allow-parallel" description:"Allow this locally command to run in parallel with other commands"`
```
vladaionescu, line 130: This should be called `AllowParallel` to better match the setting name.

tests/local/Earthfile
```diff
@@ -32,6 +32,30 @@ test-local:
         fi
 
 
+##################################################
+# tests for running locally with --allow-parallel
+
+test-parallel:
```
vladaionescu, line 38: This isn't really testing parallelization - perhaps add a comment that it's merely a smoke test, and not a test verifying that the targets indeed execute in parallel.

vladaionescu (commented):
The key line that restricts parallelism in LOCALLY commands is this one: https://github.com/earthly/earthly/blob/main/earthfile2llb/converter.go#L2011. Earthly forces execution after every RUN command when under LOCALLY. That line, plus a few other spots for COPY and SAVE ARTIFACT. You might want to play with that.

There might be other things too that require updates. @alexcb might know of some Buildkit-side changes needed too. It might be pretty tricky to pull this off - I'm not sure.

alexcb:
this PR appears to have stalled, please feel free to reopen.

PR #2516, opened 2022-12-20
Allow toggling TLS and setting CA from CLI/environment

rrjjvv:
This also adds aliases to other existing TLS-related flags to more closely match existing conventions around 'translating' between a CLI flag, environment variable, and config entry.

When I started this, it was 'obviously' a bug, but was only 95% certain towards the end.  It was easier to continue rather than abort and move to a question/request.

Originally I was trying to configure a local earthly client to execute against a remote buildkit setup using an `.env` file.  Needless to say, it was not possible.  No matter what I did, the only result I could achieve was a timeout.  After many hours, I tried specifying the equivalent CLI flags, and upon getting an error for `--tlsca` (but not `--tlskey` or `--tlscert`), the actual issue starting became clearer.

If this (perceived) limitation was actually intentional, let me know and I'll revise (or scrap) as necessary.

If you guys object to the aliases, I can remove them (but would in turn propose adding docs to call out the discrepancy).

There are currently no tests for this; I was only 95% sure this might be accepted, and there is no prior art in terms of testing this mechanism.  I saw there's a github action that tests the "local-remote", but that's the closest I could find.  If you're good with my change but feel tests are needed, I can see what I can do.

### DIFF
--- a/cmd/earthly/buildkit.go
+++ b/cmd/earthly/buildkit.go
@@ -72,7 +72,7 @@ func (app *earthlyApp) initFrontend(cliCtx *cli.Context) error {
 	app.buildkitdSettings.BuildkitAddress = app.buildkitHost
 	app.buildkitdSettings.LocalRegistryAddress = app.localRegistryHost
 	app.buildkitdSettings.UseTCP = bkURL.Scheme == "tcp"
-	app.buildkitdSettings.UseTLS = app.cfg.Global.TLSEnabled
+	app.buildkitdSettings.UseTLS = app.tlsEnabled || app.cfg.Global.TLSEnabled
 	app.buildkitdSettings.MaxParallelism = app.cfg.Global.BuildkitMaxParallelism
 	app.buildkitdSettings.CacheSizeMb = app.cfg.Global.BuildkitCacheSizeMb
 	app.buildkitdSettings.CacheSizePct = app.cfg.Global.BuildkitCacheSizePct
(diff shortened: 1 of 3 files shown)


alexcb:
Maybe we need to revisit our documentation in parallel? I'm surprised to see a new flag for `tlsca` -- does this mean there was no way to specify a self-signed ca previously?

I see there's https://docs.earthly.dev/ci-integration/remote-buildkit which documents how to configure the remote buildkit server, but there's no corresponding how-to section in the docs for show how to configure earthly (the cli) to securely connect to the buildkit server.

rrjjvv:
> Maybe we need to revisit our documentation in parallel? I'm surprised to see a new flag for `tlsca` -- does this mean there was no way to specify a self-signed ca previously?

There was (is), but can *only* be specified via config file.  It's not possible to specify it via CLI or environment variable.

> I see there's https://docs.earthly.dev/ci-integration/remote-buildkit which documents how to configure the remote buildkit server, but there's no corresponding how-to section in the docs for show how to configure earthly (the cli) to securely connect to the buildkit server.

Yeah.  There's nothing incorrect with the current docs.  They just don't go into much detail.  And I think that's 'fine', given that it's outside of the happy-path (not to mention you have a vested interest in having users use satellites, where they don't have to mess with this stuff).  But these particular group of settings don't match the conventions used elsewhere in Earthly.  Out of the various ways to improve this area, getting the conventions lined up had the biggest bang-for-the-buck (in my opinion).

So just to be clear, it is entirely possible to get the remote setup (with TLS) working with self-signed (or any certs).  So this isn't a blocker.  The PR is what I _wish_ was in place (as I had originally wanted to be environment-driven).  This is just around "configuration parity".

alexcb:
I rebased this code and got it merged in.

rrjjvv:
Nice, thanks!  There's a couple times I was ready to reach out to get a pulse, but wanted to give you some breathing room after your v7 milestone.  Much appreciated!

PR #2524, opened 2022-12-22
[logstream]: add target if it is seen on command

r-LaForge:
If we haven't yet obtained a TargetManifest, but we have obtained a command WITH a targetID the logbus will fail.

### DIFF
--- a/util/deltautil/apply.go
+++ b/util/deltautil/apply.go
@@ -52,14 +52,7 @@ func WithDeltaManifest(m *pb.RunManifest, d *pb.Delta) (*pb.RunManifest, error) 
 			ret.MainTargetId = f.GetMainTargetId()
 		}
 		for targetID, t2 := range f.GetTargets() {
-			if ret.Targets == nil {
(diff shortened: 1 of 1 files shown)


vladaionescu (commented):
This shouldn't be necessary theoretically. Perhaps we can look at what issue you were seeing and try to fix that the right way?

r-LaForge:
This PR really should have added more context 😅 

@vladaionescu We were seeing a "target not found" error: https://github.com/earthly/earthly/blob/f7db424d808838d3c47d91aca858b32ec7ab22e3/logbus/formatter/formatter.go#L208

It appeared as though we'd call `WithDeltaManifest` to apply the manifest changes and then call `handleDeltaManifest`. 
The hypothesis was a command might be added to the manifest with a parent targetId before the target itself is added.
Using this change seemed to fix the problem.

We can dig more into "why" the command may be saved first, which would also resolve the issue.
It didn't seem too bad to just create the target if it didn't exist.

PR #2530, opened 2022-12-29
explicitly check docker login vars are set

alexcb:
This is to help with sanity checking docker login is passed variables correctly.

For example, I had some tests that were failing on:

    WARN Earthfile line 369:12: The command 'RUN docker login $DOCKERHUB_MIRROR --username="$USERNAME" --***' failed: unauthorized

which were due to upstream server issues, but I wanted to eliminate the possibility that secrets were not correctly passed.

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -345,6 +345,8 @@ earthly-integration-test-base:
         IF [ "$DOCKERHUB_AUTH" = "true" ]
             RUN --secret USERNAME=$DOCKERHUB_USER_SECRET \
                 --secret TOKEN=$DOCKERHUB_TOKEN_SECRET \
+                (test -n "$USERNAME" || (echo "ERROR: USERNAME not set"; exit 1)) && \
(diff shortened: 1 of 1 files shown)


nelsam (approved):
The error message when they're not set is not very obvious - would it make sense to print them out using orange or yellow colors?

It still works as is, though - LGTM.  The message just gets a little lost in the output.

alexcb:
> The error message when they're not set is not very obvious - would it make sense to print them out using orange or yellow colors?

I added an `ERROR: ` prefix, which should make it easier to find in case someone searches for `error`.

earthly will display the full command in red in case of a failure. I think individual colours is overkill.

nelsam:
Yeah, that's fair.  In the failure output, `TOKEN is not set` comes out as default colors right above the `ERROR Earthfile line 346:12` trace, so it's not _completely_ hidden. adding the `ERROR:` prefix should be enough.

PR #2542, opened 2023-01-04
Decouple the `ast` submodule from the `util` package

nelsam:
The `ast` submodule was a little broken because it was importing `github.com/earthly/earthly/util/shell` but did not have `github.com/earthly/earthly/util/shell` in its `go.mod`.  Running `go mod tidy` fixed that; but the whole point of having the `ast` package as a separate submodule was to allow other packages to import `ast` without downloading the whole `earthly` project.

The quick and dirty solution is to reverse the imports so that the `util/shell` package is importing `ast`, not the other way around.

This PR also fixes our `+unit-test` target to run tests against our current go submodules, since `./...` ignores submodules.

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -188,10 +188,15 @@ unit-test:
     FROM +code
     COPY podman-setup.sh .
     ARG testname # when specified, only run specific unit-test, otherwise run all.
+
+    # pkgname determines the package name (or names) that will be tested. The go
+    # submodules must be specified explicitly or they will not be run, as
+    # "./..." does not match submodules.
+    ARG pkgname = ./... github.com/earthly/earthly/ast/... github.com/earthly/earthly/util/deltautil/...
     WITH DOCKER
         RUN ./podman-setup.sh && \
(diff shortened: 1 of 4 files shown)


alexcb:
interesting! looks like I broke it with 74bb85960

I wonder if we could add an Earthfile test to prevent this in the future, it could be as simple as:

    test "$(grep -PRI '^\t*"github.com/earthly/earthly/' ast | grep -v github.com/earthly/earthly/ast | wc -l)" = 0

alexcb (approved):

nelsam:
Yeah, although I think it might be cleaner to use `go list` or something. `go list -f '{{.Imports}}' ./...` will get us a list of all packages imported, recursively. And since we can use go template syntax there, we could probably write a reasonable template that only outputs offending imports.

Alternatively, `go mod tidy` followed by checking the `go.mod` file might also work.  I think I prefer the `go list` approach though.

alexcb:
> I think I prefer the `go list` approach though.

That sounds much cleaner!

PR #2545, opened 2023-01-04
Earthfile: add a target to check for core dependencies in submodules

nelsam:
This adds a (very slow) target to the Earthfile that checks for any dependencies in our submodules that cycle back to the core earthly project, failing if it finds any.

Discussion points:

- [ ] `.Deps` vs `.Imports`
    - `.Deps` is definitely more comprehensive, but maybe _too_ much so? it recurses through _everything_.
    - `.Imports` gets the imports only for the package being checked, which is probably good enough in most cases. But we don't need this check to be run constantly. Does the speed of the check matter that much?
- [x] `FOR` is certainly slow, we probably should shell out when we loop through deps.
- [x] Where should this go in `.github/workflows`?
- [x] Right now we're hardcoding `github.com/earthly/earthly/ast` and `github.com/earthly/earthly/util/deltautil` as our submodules. Is it worth the effort to do a `find . -name go.mod` to find submodules?

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -199,6 +199,20 @@ unit-test:
             go test -timeout 20m $testarg $pkgname
     END
 
+submodule-decouple-check:
+    FROM +code
+    RUN for submodule in github.com/earthly/earthly/ast github.com/earthly/earthly/util/deltautil; \
+    do \
+        for dep in $(go list -f '{{range .Deps}}{{.}} {{end}}' $submodule/...); \
+        do \
+            if [ "$(go list -f '{{if .Module}}{{.Module}}{{end}}' $dep)" == "github.com/earthly/earthly" ]; \
+            then \
+               echo "FAIL: submodule $submodule imports $dep, which is in the core 'github.com/earthly/earthly' module"; \
+               exit 1; \
+            fi; \
+        done; \
+    done
+
 changelog:
     FROM scratch
     COPY CHANGELOG.md .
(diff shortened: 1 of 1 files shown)


alexcb:
nice check -- here are some thoughts:

> .Deps vs .Imports

I'm not sure to be honest -- how much a time difference is there? We should evaluate if speed is an issue once we convert it to a single `RUN` command.

> FOR is certainly slow, we probably should shell out when we loop through deps.

If we're not performing any earthly commands withing this loop (e.g. `BUILD +target`), then we can write it as a single `RUN` command using `sh`, see https://github.com/earthly/earthly/blob/main/Earthfile#L138-L146 for an example.

> Where should this go in .github/workflows?

If we squint, this could almost be a lint type test -- Maybe we can add a `BUILD +submodule-decouple-check` to the existing `+lint-all` target.

> Right now we're hardcoding github.com/earthly/earthly/ast and github.com/earthly/earthly/util/deltautil as our submodules. Is it worth the effort to do a find . -name go.mod to find submodules?

I think the current hardcoded lists are sufficient.

nelsam:
> > .Deps vs .Imports
> 
> I'm not sure to be honest -- how much a time difference is there? We should evaluate if speed is an issue once we convert it to a single `RUN` command.

Turns out to be a significant _percentage_ difference, but the literal value isn't super high.  `.Imports` is very fast, under a second. `.Deps` is between 5 and 10 seconds on my system - a noticeable increase, but not awful IMO.

alexcb (approved):

Issue #2548, opened 2023-01-05
`WITH DOCKER` does not get `CACHE` mounts


rrjjvv:
Example:
```Earthfile
VERSION --use-cache-command 0.6

with:
    FROM earthly/dind:alpine
    CACHE /some/dir
    RUN echo "yo" > /some/dir/greet
    WITH DOCKER
        RUN --no-cache ls -al /some/dir
    END
```
```
$ earthly --version
earthly version v0.6.30 743687991d227f2f7e2e6ba11a90f443b4f99c67 linux/amd64; Ubuntu 22.04.1 LTS (Jammy Jellyfish)

$ earthly -P +with
<snip>
 earthly/dind:alpine | --> Load metadata linux/amd64
               +with | --> FROM earthly/dind:alpine
               +with | [██████████] 100% resolve docker.io/earthly/dind:alpine@sha256:9b09804c68a2c68195b71e61a09616657b8f0a5ca01b4e029f710c7ca2396dc6
               +with | --> RUN echo "yo" > /some/dir/greet
               +with | --> WITH DOCKER (install deps)
               +with | /usr/bin/jq
               +with | /usr/local/bin/dockerd
               +with | --> WITH DOCKER RUN --privileged --no-cache ls -al /some/dir
               +with | Starting dockerd with data root /var/earthly/dind/234e6b660326d22f9e997893fe6b3d24513d7865067df5b47fef9840337608e8/tmp.igGKDh
               +with | total 8
               +with | drwxr-xr-x    2 root     root          4096 Jan  5 18:13 .
               +with | drwxr-xr-x    3 root     root          4096 Jan  5 18:13 ..
<snip>
```
My original problem was actually the reverse; making changes to the mount within `WITH DOCKER` were not persisted (thus not visible/usable in future invocations), but this easier to illustrate.

I can't think of a reason why this shouldn't work, so is probably a logical extension of #2331.  I'm guessing it's as simple as taking https://github.com/earthly/earthly/blob/0946ea9c2e1349fe773473efc2444867b7c3d78b/earthfile2llb/converter.go#L594-L596 and adding it prior to any/all applicable docker variants, e.g. https://github.com/earthly/earthly/blob/0946ea9c2e1349fe773473efc2444867b7c3d78b/earthfile2llb/withdockerrun.go#L240 and https://github.com/earthly/earthly/blob/0946ea9c2e1349fe773473efc2444867b7c3d78b/earthfile2llb/withdockerrunreg.go#L230

[...]

PR #2549, opened 2023-01-05
Fix CACHE when used with WITH DOCKER

alexcb:
This fixes https://github.com/earthly/earthly/issues/2548


earthfile2llb/converter.go
```diff
@@ -1361,6 +1361,11 @@ func (c *Converter) WithDockerRun(ctx context.Context, args []string, opt WithDo
 
 	enableParallel := allowParallel && c.opt.ParallelConversion && c.ftrs.ParallelLoad
 
+	opt.extraRunOpts = nil
```
vladaionescu, line 1364: Nit: Is this line really necessary?

`opt.extraRunOpts = nil`

Feels like it could create more surprises than it addresses if the logic changes somehow in the future.

vladaionescu (approved):

earthfile2llb/converter.go
alexcb: good question -- I thought it was shared in the interrupter and would be re-used for multiple `WITH DOCKER`, but it turns out it's created each time and can be safely removed.

vladaionescu (approved):

PR #2550, opened 2023-01-06
dont attempt secrets lookup when org or project not set

alexcb:
This fixes a bug where using `VERSION --use-project-secrets 0.6` causes secrets to be lookedup in the cloud with an empty org or project, which returns an unauthorized error rather than the expected `unable to lookup secret SECRET3: not found` error message (which causes the `+secrets-test` test to fail).


alexcb:
I'm torn between expanding our integration tests to catch this -- it's a growing issue where introducing feature flags cause a combinatorics-expansion of tests required to maintain full coverage.

When #2237 is released, we'll be updating this to be `VERSION 0.7`, which will flip it on (this was discovered while working on our initial RC in #2271)

vladaionescu (approved):

PR #2559, opened 2023-01-10
Add a few 0.7 commands to the docs

vladaionescu:


docs/earthfile/earthfile.md
```diff
@@ -1547,6 +1598,54 @@ VERSION --use-host-command 0.6
...
+
+#### Synopsis
+
+* `PIPELINE [--push]`
+
+#### Description
+
+The `PIPELINE` command is used to declare that the current target is an Earthly CI pipeline. The `PIPELINE` command must be the first command in the target.
+
+A pipeline is a target that is executed by Earthly CI in when a certain trigger is activated. Triggers can be declared via the `TRIGGER` command. Pipeline targets allow only the commands `TRIGGER`, `ARG` and `BUILD`. Other commands may be used indirectly in other targets that can be then referenced by `BUILD`.
```
nelsam, line 1616:
```suggestion
A pipeline is a target that is executed by Earthly CI when a certain trigger is activated. Triggers can be declared via the `TRIGGER` command. Pipeline targets allow only the commands `TRIGGER`, `ARG` and `BUILD`. Other commands may be used indirectly in other targets that can be then referenced by `BUILD`.
```

vladaionescu: ✔️ Fixed

nelsam (approved):

PR #2561, opened 2023-01-10
use dockerhub mirror for earthly-image.sh test

alexcb:


vladaionescu (approved):
Thanks for fixing this!

PR #2562, opened 2023-01-10
buildkit http and insecure options are mutually exclusive

alexcb:
If both http=true and insecure=true are set, buildkit will create two registry entries, one on port 80, and the other on port 443 and will attempt to connect to either server.

### DIFF
--- a/docs/guides/registries/self-signed.md
+++ b/docs/guides/registries/self-signed.md
@@ -41,7 +41,6 @@ To configure Earthly to use an insecure registry, use the following [Earthly con
 global:
   buildkit_additional_config: |
     [registry."<registry-hostname>"]
-      http = true
       insecure = true
 ```
 
@@ -55,6 +54,15 @@ build:
     ENTRYPOINT cat motd
     SAVE IMAGE --push --insecure <registry-hostname>/hello-earthly:with-love
 ```
+
+{% hint style='danger' %}
+##### Note
+
+The `http` and `insecure` settings are mutually exclusive. Setting `insecure=true` should only be used when the registry is https and is configured with an insecure certificate.
+Setting `http=true` is only for the case where a standard http-based registry is used (i.e. no SSL encryption). If both are set buildkit will attempt to connect to the registry using either http (port 80), or https (port 443).
+
+{% endhint %}
+
 
 ## Other BuildKit options
 
--- a/tests/remote-cache/Earthfile
+++ b/tests/remote-cache/Earthfile
@@ -36,15 +36,14 @@ ARG --global EARTHLY_BUILD_ARGS="REGISTRY"
 ARG --global REGISTRY_CONFIG="
 [registry.\"$REGISTRY\"]
   http = true
-  insecure = true
 "
 
 COPY test.earth ./Earthfile
 
 all:
     BUILD +test1
-    #BUILD +test2
-    #BUILD +test3
+    BUILD +test2
+    BUILD +test3
 
 test1:
     RUN echo "content" >./input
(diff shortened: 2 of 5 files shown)


tests/remote-cache/Earthfile
```diff
@@ -36,15 +36,14 @@ ARG --global EARTHLY_BUILD_ARGS="REGISTRY"
...
-  insecure = true
 "
 
 COPY test.earth ./Earthfile
 
 all:
     BUILD +test1
-    #BUILD +test2
-    #BUILD +test3
+    BUILD +test2
```
alexcb, line 45: these were accidentally commented out in the initial VERSION 0.7 PR; I want to re-enable them to ensure the `insecure = true` removal doesn't break them.

docs/guides/registries/self-signed.md
```diff
@@ -56,6 +55,15 @@ build:
     SAVE IMAGE --push --insecure <registry-hostname>/hello-earthly:with-love
 ```
 
+{% hint style='danger' %}
+##### Note
+
+The `http` and `insecure` settings are mutually exclusive. Setting `insecure=ture` should only be used when the registry is https, and you deliberately want to accept invalid or unverifiable certificates.
```
nelsam, line 61:
```suggestion
The `http` and `insecure` settings are mutually exclusive. Setting `insecure=true` should only be used when the registry is https, and you deliberately want to accept invalid or unverifiable certificates.
```

nelsam: I also think that "...the registry is https, and you deliberately..." is a bit ... ambiguous maybe? saying `...the registry is https _and_ you deliberately..." seems to read more clearly to me - the comma is perhaps a misleading separator in this context.

nelsam (approved):
Minor doc grammar aside, lgtm.

docs/guides/registries/self-signed.md
alexcb: good point -- I simplified it (and also fixed the typo).

docs/guides/registries/self-signed.md
```diff
@@ -56,6 +55,15 @@ build:
     SAVE IMAGE --push --insecure <registry-hostname>/hello-earthly:with-love
 ```
 
+{% hint style='danger' %}
+##### Note
+
+The `http` and `insecure` settings are mutually exclusive. Setting `insecure=true` should only be used when the registry is https and is configured with an insecure certificates.
```
nelsam, line 61:
```suggestion
The `http` and `insecure` settings are mutually exclusive. Setting `insecure=true` should only be used when the registry is https and is configured with insecure certificates.
```

rrjjvv:
@alexcb I know it's considered bad practice to comment on PRs as a non-reviewer (let alone PRs that are closed), but thought it slightly more appropriate than opening a new issue for something so hot-off-the-press, or putting it in the user-facing Slack channel.  Feel free to slap me if this post is actually the worst evil...

I just happened to sync my fork minutes after this change, and this change cause my execution of the tests almost immediately.  For the sake of creating my recent buildkit PR, I added `http = true` back.

The tests have always worked for me with both `http = true` and `insecure = true`.  Having `insecure = true` (only) makes everything fail immediately, as it's trying to hit my caching registry via HTTPS.  Having only `http = true` also works.  So it would _appear_ that `http = true` implies `insecure = true`, but definitely not the other way around.

Despite the fact I was unable to run the tests, this PR makes total sense, given that the environment variable controlling this is named `EARTHLY_MIRROR_INSECURE`.  The 'obvious' remedy would be making a parallel `EARTHLY_MIRROR_HTTP`.  But I'm painfully aware that that would entail adding copy/paste boilerplate to nearly every single test file.

If I'm the only contributor using a non-HTTPS local pull-through cache, I have no problem simply making that ad-hoc patch whenever I need.

Another option would be to wire up my cache with some self-signed certs.  I have no problem doing that, if you want that to be the "paved road".  But in that case, I'd think you'd want to update the self-signed registry guide with instructions on how to do that...?

Let me know your thoughts.

(Have you guys every considered creating an #earthly-contrib channel, for your non-core contributors to pose dev-specific questions/issues to you folks?)

PR #2565, opened 2023-01-10
git-metadata-test: set author via env variable

alexcb:
git will make use of GIT_AUTHOR_NAME and GIT_AUTHOR_EMAIL while detecting the author identity (see ident.c under git source).

We will attempt to explicitly set the env variables, in case that prevents the following error we periodically see in GHA:

    *** Please tell me who you are.

    Run

      git config --global user.email "you@example.com"
      git config --global user.name "Your Name"

    to set your account's default identity.
    Omit --global to set the identity only in this repository.

    fatal: unable to auto-detect email address (got 'root@buildkitsandbox.(none)')


alexcb:
this did not fix the problem

PR #2573, opened 2023-01-13
introduce a new flag for plain HTTP docker mirrors

alexcb:
434b98a8ad42d85883a6644bd22611678eadb101 removed the ability for a user to run tests against a plain http-based mirror. This introduces a new flag to allow configuring buildkit to use plain http.

Simply adding back the http option (and including the insecure option), would result in the mirror being hit on both port 80 and por 443 (or if a non-standard port was used, that port would be hit with both http and https requests, which can lead to lots of noisy error logs).


Earthfile
```diff
@@ -374,10 +375,18 @@ earthly-integration-test-base:
   mirrors = [\"$DOCKERHUB_MIRROR\"]"
 
         IF [ "$DOCKERHUB_MIRROR_INSECURE" = "true" ]
+            IF [ "$DOCKERHUB_MIRROR_HTTP" = "true" ]
+                RUN --no-cache echo "DOCKERHUB_MIRROR_INSECURE and DOCKERHUB_MIRROR_HTTP are mutually exclusive" && exit 1
```
alexcb, line 379: technically not true -- a user might have a server that runs a mirror that accepts either. Maybe we should allow it?

rrjjvv: My own two cents: I think it should be allowed.  At the very least, I've been running with both true on my host (not just the early test setup) since day <span>#</span>1 with no ill effect.  I'd trust the user to know what they're doing and understand the ramifications (even if they actually don't 😄).

If upstream decides to disallow that combo, then I do think your enforcement is a good thing, simply because the the buildkit error would likely be hidden away.

vladaionescu (approved):

PR #2579, opened 2023-01-13
earthly doc: parse and output documentation comments on targets

nelsam:
This is a first pass on an `earthly doc` subcommand. For this first pass:

1. Doc comments are only output for targets. More docs about `ARG` and `SAVE TARGET` will  come later.
2. Comments _must_ be on the line immediately before the target.
3. Comments _must_ reference  the target name as the first word in the comment block.


Earthfile
```diff
@@ -192,13 +209,24 @@ unit-test:
...
-    ARG pkgname = ./... github.com/earthly/earthly/ast/... github.com/earthly/earthly/util/deltautil/...
+    ARG pkgname = ./...
+
+    ARG setuppodman = true
     WITH DOCKER
-        RUN ./podman-setup.sh && \
-            if [ -n "$testname" ]; then testarg="-run $testname"; fi && \
-            go test -timeout 20m $testarg $pkgname
+        IF [ "$setuppodman" = "true" ]
+            RUN ./podman-setup.sh && \
```
alexcb, line 217: nit: to avoid duplicating the `go test ...` logic, maybe we could do something like:

```
RUN \
    if [ "setuppodman" = "true" ]; then ./podman-setup.sh; fi && \
    if [ -n "$testname" ]; then testarg="-run $testname"; fi && \
    go test -timeout 20m $testarg $pkgname
```

ast/ast_test.go
```diff
@@ -0,0 +1,247 @@
...
+}
+
+// handleMockFile helps us perform slightly more black-box testing by handling a
+// mockNamedReader as if it were a file-like io.ReadSeeker. This way, we don't
+// need to know in the test how many times the file is seeked back to zero and
+// re-read.
+//
+// This cannot handle non-zero seeks and will fail if it receives a non-zero
+// seek call.
+func handleMockFile(t *testing.T, r *mockNamedReader, body []byte) {
```
alexcb, line 212: It took me a bit to understand what

```
		pers.ConsistentlyReturn(t, tt.reader.NameOutput, "Earthfile")
		handleMockFile(tt.t, tt.reader, []byte(`
```
was doing (especially the `ConsistentlyReturn` function as well as where `NameOutput` was coming from -- at first I thought it might have been testing a function called `NameOutput()` consistently returned the value `Earthfile` -- After looking over the `hel` source it makes more sense that it's a mock-helper)

The `handleMockFile` name seems more focused on the implementation  (handling lower-level mocked syscalls) rather than the task that it's doing -- mocking out an Earthfile at a higher-level.

What if we created a new helper function to combine both mock-setup tasks:

```
func mockEarthfile(tt testCtx, earthfileBody []byte) {
    pers.ConsistentlyReturn(tt.t, tt.reader.NameOutput, "Earthfile")
    handleMockFile(tt.t, tt.reader, earthfileBody)
}
```

It could make it more quicker to understand for first time readers (and it'll reduce a line from all the tests).

ast/ast_test.go
```diff
@@ -0,0 +1,247 @@
...
+
+		tt.expect(f.Version.Args).To(haveLen(1))
+		tt.expect(f.Version.Args[0]).To(equal("0.6"))
+
+		tt.expect(f.Targets).To(haveLen(1))
+		tgt := f.Targets[0]
+		tt.expect(tgt.Name).To(equal("foo"))
+		tt.expect(tgt.Recipe).To(haveLen(1))
+		rcp := tgt.Recipe[0]
+		tt.expect(rcp.Command).To(not(beNil()))
```
alexcb, line 98: ultra-nit:

This completely comes down to preferemce, but I can't say I'm a fan of this (rspec style?) testing library.

`To(not(beNil()))` doesn't feel very go-like (which could be a compliment depending on who you ask :laughing: )

We already use the `testify` library in the code base, which wouldn't require any nesting or chaining, e.g. `assert.NotNil(t, rcp.Command)`.

Is it worth standardising on a single testing framework? 

If an existing framework doesn't support what we need (or makes it cumbersome), then we should obviously re-evaluate it.

At the same time I don't want to get distracted on a tabs-vs-spaces discussion :smile:

alexcb (commented):
This is looking great!

Adding an integration test to `tests/Earthfile` would be good to validate the `earthly doc` command is correctly plugged into the parsing code -- it doesn't need to be complicated as your unit test coverage looks great!

Earthfile
nelsam: Sounds good.  I was wondering, though, should this be using `-race`? I know it runs some integration tests, and `-race` is not really recommended with integration tests due to the goroutine limit...

ast/ast_test.go
nelsam: I was one of the very early contributors to testify, and I have a lot of problems with it. Some of my problems with it are simply problems I have with testing in go in general, but many are easily solved. One particularly easy one: in testify, it's very difficult to know which function parameter is the `expected` and which is the `actual`. I also recall some situations where the order of the parameters had to be reversed for some reason (probably something about variadic functions).

I started using rspec-style assertions in 2015 with [`ginkgo`](https://onsi.github.io/ginkgo/) and [`gomega`](https://onsi.github.io/gomega/). I was not a fan at first, but they've certainly grown on me. While I still think that `gomega` (and especially `ginkgo`) are over-complicated, I really appreciate that I never have a question about which parameter is the `actual` and which is the `expected`. And while `onpar/expect` isn't particularly heavily used, `gomega` _is_ - so I don't think the style is completely foreign in the go world any more.

I haven't found an assertion style that I _really_ like, but while I've grown more and more frustrated with testify-style assertion functions over time, I have definitely grown _less_ frustrated with rspec/gomega/expect style expectation matchers. `goconvey`-style `So(actual, ShouldEqual, expected)` doesn't look awful to me either, but I am dubious about `goconvey` because last I checked, it imports `gls` (goroutine-local-storage). I don't like having that in my import graph.

I still _really_ want to find a good assertion style that can make use of type parameters to avoid the "untyped nil" problem. I [tried to do it in `onpar`](https://github.com/poy/onpar/pull/35) but it made the built-in matchers feel awful.

---

For this PR, I'd be willing to go through and replace all the `tt.expect(...)` with `if ... { tt.t.Fatalf(...) }` instead.  I would _prefer_ to avoid using testify, personally, but if you _really_ like it I'll work through my dislike of that library.

Alternatively, we could wrap up `tt.expect(...).To(...)` in different verbiage. Maybe `tt.assert(...).Does(...)`?

alexcb: > One particularly easy one: in testify, it's very difficult to know which function parameter is the expected and which is the actual

I have definitely mixed them up before.

>  also recall some situations where the order of the parameters had to be reversed for some reason 

interesting point! I see `WithinRange` is an example where actual comes first (and then there's a range of expected vals).

vladaionescu: Some good arguments for both. Let's talk about this briefly in our Core sync-up to get to a resolution.

tests/Earthfile
```diff
@@ -1128,6 +1129,23 @@ help:
...
+    It also has a separator between paragraphs."
+    DO +RUN_EARTHLY --earthfile=target-docs.earth --extra_args="doc" --target="" --grep_flags="-v" --output_contains="undocumented-target"
+    DO +RUN_EARTHLY --earthfile=target-docs.earth --extra_args="doc" --target="" --grep_flags="-v" --output_contains="incorrectly-documented-target"
+    DO +RUN_EARTHLY --earthfile=target-docs.earth --extra_args="doc" --target="+documented-target" --output_contains="+documented-target
+  documented-target is a target with documentation
+  that spans multiple lines.
+
+  It also has a separator between paragraphs."
+    DO +RUN_EARTHLY --earthfile=target-docs.earth --extra_args="doc" --target="+undocumented-target" --output_contains="No doc comment found"
+    DO +RUN_EARTHLY --earthfile=target-docs.earth --extra_args="doc" --target="+incorrectly-documented-target" --output_contains="No doc comment found"
```
alexcb, line 1147: should this cause earthly to exit with a non-zero exit code? there's a `--should_fail=true` flag for `DO +RUN_Earthly` that can be used.

ast/tests/chown.ast.json
```diff
@@ -68,10 +68,7 @@
               "test",
               "testuser",
               "==",
-              "$(stat",
-              "-c",
-              "%U",
-              "./a.txt)"
+              "$(stat -c %U ./a.txt)"
```
alexcb, line 71: woohoo!

tests/Earthfile
nelsam: Good point.  I noticed that arg, but just didn't think about whether or not `doc` should have a non-zero exit code for this case - but it probably should.

alexcb (approved):

tests/Earthfile
nelsam: FWIW, I made `earthly doc` actually error out if a specific target is requested but no documentation exists for that target. I think it seems reasonable. Plus as an experimental sub-command, we can always soften up those errors later.

alexcb (approved):

ast/tests/documentation.ast.json
```diff
@@ -0,0 +1,33 @@
...
+        }
+      ]
+    }
+  ],
+  "version": {
+    "args": [
+      "0.7"
+    ]
+  }
+}
```
vladaionescu, line 33: Yay!

ast/tests/no-cache-local-artifact.ast.json
```diff
@@ -16,20 +16,7 @@
...
-              "'a-zA-Z0-9'",
-              "|",
-              "fold",
-              "-w",
-              "32",
-              "|",
-              "head",
-              "-n",
-              "1)",
+              "$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 32 | head -n 1)",
```
vladaionescu, line 19: This looks so much better :D

cmd/earthly/root_cmds.go
```diff
@@ -113,6 +115,12 @@ func (app *earthlyApp) rootCmds() []*cli.Command {
 			Usage:       "Earthly organization administration *experimental*",
 			Subcommands: app.orgCmds(),
 		},
+		{
+			Name:      "doc",
+			Usage:     "Document targets from an Earthfile *experimental*",
```
vladaionescu, line 120: Shall we launch it straight as "beta", since there is no risk of messing up existing builds?

Perhaps we also move `ls` into "beta".

vladaionescu (commented):
Looks good for the most part - the only thing this is missing, I think, is a CHANGELOG entry. Perhaps this is part of the next 0.7 RC?

cmd/earthly/root_cmds.go
nelsam: Sounds good. Does this mean that we drop `*experimental*`, or that we replace it with `*beta*`? I don't see any current subcommands marked beta.

vladaionescu: Good question - it seems that historically we haven't marked beta features in a specific way. I guess mostly because in beta, the API is anyway finalized. Dropping makes sense to me.

vladaionescu (approved):

PR #2585, opened 2023-01-16
move staging builds to separate docker image

alexcb:
push staging builds to a separate earthly/earthly-staging and earthly/buildkitd-staging docker images, to make it easier for users to differentiate between version tags when viewing docker hub.


vladaionescu (approved):
Good idea

PR #2590, opened 2023-01-18
fix release.sh not passing prerelease arg

alexcb:
The prerelease value was not correctly passed to the +release-github target, which caused the RC to show up as latest.

This additionally re-orders the earthly version detection logic to handle the case where the earthly command override is missing.

### DIFF
--- a/release/release.sh
+++ b/release/release.sh
@@ -34,8 +34,19 @@ set -ex
 #    env -i HOME="$HOME" PATH="$PATH" SSH_AUTH_SOCK="$SSH_AUTH_SOCK" RELEASE_TAG=v0.6.0 ./release.sh
 #
 
+SCRIPT_DIR="$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+cd "$SCRIPT_DIR"
+
+test -n "$HOME" || (echo "ERROR: HOME is not set"; exit 1);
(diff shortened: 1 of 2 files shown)


alexcb:
It's difficult to fully test this PR, so let's leave it as a draft, and when we do a release for rc2, we'll rebase this, then run it, and then merge it once we're more confident.

release/release.sh
```diff
@@ -34,8 +34,19 @@ set -ex
 #    env -i HOME="$HOME" PATH="$PATH" SSH_AUTH_SOCK="$SSH_AUTH_SOCK" RELEASE_TAG=v0.6.0 ./release.sh
 #
 
+SCRIPT_DIR="$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
```
alexcb, line 37: these were re-ordered from below.

alexcb:
This script was used for v0.7.0-rc2 (then we added in the extra check for pre-release being set to true when rc is in the tag name).

nelsam (approved):
we did some of this while pairing, so i only took a cursory glance now, but it like like the same script we paired on 😁 lgtm

alexcb:
> we did some of this while pairing, so i only took a cursory glance now, but it like like the same script we paired on grin lgtm

yeah, all I did was remove the various `exit 0`s I had sprinkled in as we were testing stuff.

PR #2597, opened 2023-01-20
Allow registry login with cloud-stored secrets

alexcb:
Allows a user to set registry credentials in the earthly-cloud secrets system rather than `~/.docker/config.json`.

If a user is currently logged in to earthly-cloud, then earthly will use credentials for the registry by searching:

1) /user/std/registry/<host>/username and
   /user/std/registry/<host>/password,

2) /org/project/std/registry/<host>/username and
   /org/project/std/registry/<host>/password,

and finally 3) the `~/.docker/config.json` file.

Note that all projects from both the directly and indirectly referenced Earthfiles will be searched for authentication credentials. The directly-referenced Earthfile will take prescedence, however there is no guarentee the indirectly referenced Earthfiles will be consistently ordered. This limitation means that it is not possible to associate a particular `SAVE IMAGE --push ...` with a paticular org/project credential storage location.

### DIFF
--- a/scripts/tests/cloud-docker-credentials-integration.sh
+++ b/scripts/tests/cloud-docker-credentials-integration.sh
@@ -20,7 +20,7 @@ test -n "$EARTHLY_TOKEN"
 # And our earthly-user that GHA uses has been configured to have access:
 #   earthly projects --org earthly-technologies --project core-test-cloud-docker-credentials-test members add other-service+earthly@earthly.dev read+secrets
 
-# sanity check that secrets is working
+# check that secrets is working, and we are running with the correct user
 test "$("$earthly" secret --org earthly-technologies --project core-test-cloud-docker-credentials-test get std/registry/registry-1.docker.io/username)" = "verygoodusername"
 
 echo ==== test that the private verygoodimage can be fetched using the credentials from secrets ====
--- a/scripts/tests/cloud-docker-credentials-integration.sh
+++ b/scripts/tests/cloud-docker-credentials-integration.sh
@@ -1,0 +1,53 @@
+#!/usr/bin/env bash
+set -eu
+
+earthly=${earthly:=earthly}
+earthly=$(realpath "$earthly")
+echo "running tests with $earthly"
+
+# prevent the self-update of earthly from running (this ensures no bogus data is printed to stdout,
+# which would mess with the secrets data being fetched)
+date +%s > /tmp/last-earthly-prerelease-check
+
+# ensure earthly login works (and print out who gets logged in)
+test -n "$EARTHLY_TOKEN"
+"$earthly" account login
+
+# A username / password has been stored in the cloud to a docker hub user (that is not part of earthly) via:
+#   earthly secret --org earthly-technologies --project core-test-cloud-docker-credentials-test set std/registry/registry-1.docker.io/username verygoodusername
+#   earthly secret --org earthly-technologies --project core-test-cloud-docker-credentials-test set std/registry/registry-1.docker.io/password verygoodpassword # just kidding
+#
+# And our earthly-user that GHA uses has been configured to have access:
+#   earthly projects --org earthly-technologies --project core-test-cloud-docker-credentials-test members add other-service+earthly@earthly.dev read+secrets
+
+# sanity check that secrets is working
+test "$("$earthly" secret --org earthly-technologies --project core-test-cloud-docker-credentials-test get std/registry/registry-1.docker.io/username)" = "verygoodusername"
+
+echo ==== test that the private verygoodimage can be fetched using the credentials from secrets ====
+rm -rf /tmp/earthly-cloud-docker-credentials-test-1
+
+mkdir /tmp/earthly-cloud-docker-credentials-test-1
+cd /tmp/earthly-cloud-docker-credentials-test-1
+cat >> Earthfile <<EOF
+VERSION 0.7
+PROJECT earthly-technologies/core-test-cloud-docker-credentials-test
+test1:
+    FROM verygoodusername/verygoodimage:verygoodtag
+    RUN base64 -d /my-test-data | grep verygooddata
+EOF
+
+# first make sure docker can't access the verygoodimage
+test "$(docker images -q verygoodusername/verygoodimage | wc -l)" = "0"
+if docker pull verygoodusername/verygoodimage:verygoodtag 2> docker-pull.log; then
+  cat docker-pull.log
+  echo "error: this test requires that docker does not have access to pull the verygoodimage"
+  exit 1
+fi
+if ! grep 'requested access to the resource is denied' docker-pull.log >/dev/null; then
+  cat docker-pull.log
+  echo expected denied failed, but got somthing else
+  exit 1
+fi
+
+# then test that earthly can access the verygoodimage (by using the cloud-hosted registry credentials)
+"$earthly" --no-cache +test1
--- a/util/llbutil/authprovider/cloudstoredauthprovider/authprovider.go
+++ b/util/llbutil/authprovider/cloudstoredauthprovider/authprovider.go
@@ -1,0 +1,371 @@
+// Package cloudstoredauthprovider was forked from buildkit/session/auth/authprovider in order to allow using
+// registry credentials that are stored on a server rather than local filesystem.
+// This package is distributed under the original file's license, The Apache License, which is defined under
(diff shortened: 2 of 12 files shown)


alexcb:
depends on #2610

util/llbutil/authprovider/cloudstoredauthprovider/authprovider.go
```diff
@@ -0,0 +1,371 @@
...
+
+const defaultExpiration = 60
+
+var ErrNoCredentialsFound = fmt.Errorf("no credentials found")
+
+type ProjectBasedAuthProvider interface {
+	AddProject(org, project string)
+}
+
+func NewCloudAuthProvider(cfg *configfile.ConfigFile, cloudClient cloud.Client, console conslogging.ConsoleLogger) session.Attachable {
```
nelsam, line 48: When calling this function, we'll be calling `cloudstoredauthprovider.NewCloudAuthProvider`, which stutters. I think this function can just be named `New`, or if we rename the package to `cloudauth`, maybe `NewProvider` (so it reads as `cloudauth.NewProvider`).

util/llbutil/authprovider/cloudstoredauthprovider/authprovider.go
```diff
@@ -0,0 +1,371 @@
...
+	ap.mu.Lock()
+	defer ap.mu.Unlock()
+
+	// check the cache
+	ac, ok := ap.authConfigCache[host]
+	if ok {
+		return ac, nil
+	}
+
+	// check unser user's secrets
```
nelsam, line 288: `unser`? Is that a typo?

util/llbutil/authprovider/cloudstoredauthprovider/authprovider.go
```diff
@@ -0,0 +1,371 @@
...
+	ac, err := ap.getAuthConfigForProject(ctx, "", "", host)
+	if err == nil {
+		ap.authConfigCache[host] = ac
+		return ac, nil
+	} else if err != ErrNoCredentialsFound {
```
nelsam, line 293: `else` is unnecessary here - the `if` ends in a `return`
```suggestion
	}
	if err != ErrNoCredentialsFound {
```

util/llbutil/authprovider/cloudstoredauthprovider/authprovider.go
```diff
@@ -0,0 +1,371 @@
...
+		ap.console.Warnf("failed to lookup username/password for %s under user secrets: %s", host, err.Error())
+	}
+
+	// fall back to project's secrets (starting with the root-level Earthfile's org/project)
+	for _, op := range ap.orgProjects {
+		ac, err := ap.getAuthConfigForProject(ctx, op.org, op.project, host)
+		if err == nil {
+			ap.authConfigCache[host] = ac
+			return ac, nil
+		} else if err != ErrNoCredentialsFound {
```
nelsam, line 303:
```suggestion
		}
		if err != ErrNoCredentialsFound {
```

util/llbutil/authprovider/cloudstoredauthprovider/tokenseed.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	l := flock.New(filepath.Join(ts.dir, ".token_seed.lock"))
+	if err := l.Lock(); err != nil {
+		if !errors.Is(err, syscall.EROFS) && !errors.Is(err, os.ErrPermission) {
+			return nil, err
+		}
+	} else {
+		defer func() { _ = l.Unlock() }()
+	}
```
nelsam, line 44: I'm a little confused - if `l.Lock()` errors but the error is _not_ `syscall.EROFS` or `os.ErrPermission`, then we continue executing but never call `l.Unlock()`? Is that intentional? If so, isn't that a little risky?

nelsam (commented):
I have one thing I'm looking for a little clarification on; otherwise my comments are stylistic.

vladaionescu (approved):
Looks good!

* Will the new CLI sub-command go in a follow-up PR?
* Changelog entry? Or perhaps this goes together with the point above.

scripts/tests/cloud-docker-credentials-integration.sh
```diff
@@ -0,0 +1,53 @@
...
+"$earthly" account login
+
+# A username / password has been stored in the cloud to a docker hub user (that is not part of earthly) via:
+#   earthly secret --org earthly-technologies --project core-test-cloud-docker-credentials-test set std/registry/registry-1.docker.io/username verygoodusername
+#   earthly secret --org earthly-technologies --project core-test-cloud-docker-credentials-test set std/registry/registry-1.docker.io/password verygoodpassword # just kidding
+#
+# And our earthly-user that GHA uses has been configured to have access:
+#   earthly projects --org earthly-technologies --project core-test-cloud-docker-credentials-test members add other-service+earthly@earthly.dev read+secrets
+
+# sanity check that secrets is working
```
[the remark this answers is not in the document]
alexcb: https://en.wikipedia.org/wiki/Sanity_check is a valid term -- it's just a quick test to ensure things are working before proceeding (and also helps as provide a base reference if one is debugging a problem backwards). Either or seems fine to me.

util/llbutil/authprovider/cloudstoredauthprovider/authprovider.go
alexcb: hah yeah that's not a word :laughing:

scripts/tests/cloud-docker-credentials-integration.sh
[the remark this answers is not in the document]
nelsam: Yeah, it's just a bit archaic and ableist.  From the link you provided:
> Some software projects, such as [Android](https://en.wikipedia.org/wiki/Android_(operating_system))[[1]](https://en.wikipedia.org/wiki/Sanity_check#cite_note-1), [MediaWiki](https://en.wikipedia.org/wiki/MediaWiki)[[2]](https://en.wikipedia.org/wiki/Sanity_check#cite_note-2) and [Twitter](https://en.wikipedia.org/wiki/Twitter)[[3]](https://en.wikipedia.org/wiki/Sanity_check#cite_note-3), discourage use of the phrase sanity check in favour of just test or soundness check, as part of a wider attempt to avoid [ableist](https://en.wikipedia.org/wiki/Ableist) language and increase [inclusivity](https://en.wikipedia.org/wiki/Inclusivity).
Not a huge deal, though.

util/llbutil/authprovider/cloudstoredauthprovider/tokenseed.go
alexcb: the original buildkit code simply did a `defer l.Unlock()`; however we have a linter that wants an explicit `_ = ...` to show errors are intentionally dropped, and this is the pattern we've been using to show that.

scripts/tests/cloud-docker-credentials-integration.sh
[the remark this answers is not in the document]
alexcb: ah interesting point! I hadn't considered that.

util/llbutil/authprovider/cloudstoredauthprovider/tokenseed.go
nelsam: Gah, the context here isn't showing as many lines as I'd like. What I was trying to comment on is that the `if err != nil` block does not fully stop execution on line `41`, so we can enter the `if` block due to a non-`nil` error, but still end up exiting the `if` block without entering the `else` block.

That means that we will continue execution without access to the lock (and we will never call `Unlock`). Is that both intentional and safe?

alexcb: honestly I have no idea; it comes from upstream: https://github.com/moby/buildkit/blob/master/session/auth/authprovider/tokenseed.go#L38-L44

alexcb:
> Looks good!
> 
> * Will the new CLI sub-command go in a follow-up PR?
> * Changelog entry? Or perhaps this goes together with the point above.

I was thinking these would be done in a separate PR -- I wanted to implement this functionality in order to unblock the CI side of things.

util/llbutil/authprovider/cloudstoredauthprovider/tokenseed.go
nelsam: Ah, gotcha! At least it's an upstream thing :laughing:

nelsam (approved):

PR #2600, opened 2023-01-20
perform docker login in unit-test target

alexcb:
The unit-test target (which isn't actually a unit-test), depends on access to dockerhub. To avoid being rate-limited, a docker login is now performed if the dockerhub (or mirror) credentials are passed in via secrets.


not-a-unit-test.sh
```diff
@@ -0,0 +1,69 @@
+#!/bin/sh
+set -e # dont use -x, as it will leak credentials
```
nelsam, line 2: Does this test run on PRs? Is the output public?

Or in other words: what happens if a malicious user submits a PR that adds `-x`?

nelsam (approved):
It all looks good to me, so long as this doesn't run and have public output on PRs.

not-a-unit-test.sh
alexcb: https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks protects us from such a case -- we would hopefuly see the `-x` was added in and then reject the PR (without running it).

alexcb: we have this reminder in a few of our test scripts, such as https://github.com/earthly/earthly/blob/9aa64fa1bbb5d39f2d0dfcd01a0f51e857bc9d1b/scripts/tests/auth/test-hello-world-with-non-authorized-key.sh

PR #2601, opened 2023-01-21
Fix description for `--use-chmod` feature

rrjjvv:
This is something I had noticed a long time ago, but minor enough that it wasn't worth a dedicated PR, and then forgot by the time I had a real PR where I could shoehorn it in.

When @vladaionescu pointed me to https://github.com/earthly/earthly/blob/next/docs/earthfile/features.md and I saw the bad description, I immediately recognized it and became 99% sure that doc was auto-generated from code.  So this is more about fixing the docs, and assuming this is first step.  And since this is truly "code", I assume it should be merged to 'main' (and you'll do something after to get it and/or the generated content to 'next').


vladaionescu:
We should fix `next` separately - what's in `next` is probably just a copy-paste error. CC @alexcb

PR #2605, opened 2023-01-21
Updated changelog for 0.7 RC2

vladaionescu:


CHANGELOG.md
```diff
@@ -4,6 +4,129 @@ All notable changes to [Earthly](https://github.com/earthly/earthly) will be doc
 
 ## Unreleased
 
+## v0.7.0-rc2 - 2023-01-18
```
nelsam, line 7: Is this date (`2023-01-18`) a placeholder or a copy/paste error?

alexcb: fortunately the release process requires `+lint-changelog` to pass, which will error if the date isn't today's date.

CHANGELOG.md
```diff
@@ -4,6 +4,129 @@ All notable changes to [Earthly](https://github.com/earthly/earthly) will be doc
...
+
+```Earthfile
+test-and-push:
+  FROM +some-image
+  WAIT
+    BUILD +test
+  END
+  SAVE IMAGE --push my-org/my-image:latest
+```
+
```
alexcb, line 64: It might be helpful to additionally include a CD-type use-case here.

e.g.

The `WAIT` clause can also be used to first push an image, then run a command that deploys that image on a remote server, for example

```
my-image:
  ...
  SAVE IMAGE --push my-org/my-image:latest

push-and-deploy:
  FROM +some-image
  WAIT
    BUILD +my-image
  END
  RUN --push ./deploy my-org/my-image:latest
```

alexcb (commented):
looks good, but think a continuous-deployment example could be helpful.

CHANGELOG.md
vladaionescu: Good idea - added!

PR #2609, opened 2023-01-23
rename output phases to summary

alexcb:
WAIT/END makes it possible to perform a push or local output during the build phase, since these operations may have occurred prior to the push and local output phases, these phases will be renamed to be a summary to make it more clear they don't imply the order.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -41,13 +41,13 @@ import (
 
 const (
 	// PhaseInit is the phase text for the init phase.
-	PhaseInit = "1. Init 🚀"
+	PhaseInit = "Init 🚀"
 	// PhaseBuild is the phase text for the build phase.
-	PhaseBuild = "2. Build 🔧"
+	PhaseBuild = "Build 🔧"
 	// PhasePush is the phase text for the push phase.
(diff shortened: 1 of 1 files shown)


builder/builder.go
```diff
@@ -45,9 +45,9 @@ const (
 	// PhaseBuild is the phase text for the build phase.
 	PhaseBuild = "2. Build 🔧"
 	// PhasePush is the phase text for the push phase.
-	PhasePush = "3. Push ⏫"
+	PhasePush = "3. Push Summary ⏫"
 	// PhaseOutput is the phase text for the output phase.
-	PhaseOutput = "4. Local Output 🎁"
+	PhaseOutput = "4. Local Output Summary 🎁"
```
vladaionescu, line 50: Should we additionally remove the numbers? I introduced the numbers to indicate an order of operations - but that's no longer applicable since pushes and local outputs can take place any time now.

alexcb: that's a good point -- they no longer imply an order.

alexcb: her's an example with them all removed:

image

vladaionescu (approved):

PR #2616, opened 2023-01-26
commands for cloud-stored registry auth

alexcb:
Adds `earthly registry login/list/logout` commands which are used to manage cloud-stored registry credentials.

### DIFF
--- a/cmd/earthly/registry_cmds.go
+++ b/cmd/earthly/registry_cmds.go
@@ -25,14 +25,12 @@ func (app *earthlyApp) registryCmds() []*cli.Command {
 			Flags: []cli.Flag{
 				&cli.StringFlag{
 					Name:        "org",
-					EnvVars:     []string{"EARTHLY_ORG"},
 					Usage:       "The organization to which the project belongs.",
 					Required:    false,
 					Destination: &app.orgName,
 				},
 				&cli.StringFlag{
 					Name:        "project",
-					EnvVars:     []string{"EARTHLY_PROJECT"},
 					Usage:       "The organization project in which to store registry credentials, if empty credentials will be stored under the user's secret storage.",
 					Required:    false,
 					Destination: &app.projectName,
@@ -69,14 +67,12 @@ func (app *earthlyApp) registryCmds() []*cli.Command {
 			Flags: []cli.Flag{
 				&cli.StringFlag{
 					Name:        "org",
-					EnvVars:     []string{"EARTHLY_ORG"},
 					Usage:       "The organization to which the project belongs.",
 					Required:    false,
 					Destination: &app.orgName,
 				},
 				&cli.StringFlag{
 					Name:        "project",
-					EnvVars:     []string{"EARTHLY_PROJECT"},
 					Usage:       "The organization project in which to store registry credentials, if empty credentials will be stored under the user's secret storage.",
 					Required:    false,
 					Destination: &app.projectName,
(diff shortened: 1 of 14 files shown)


alexcb:
Note that `earthly registry login <host> --username myuser --password mypassword` is not allowed due to `github.com/urfave/cli` requiring that the flags come before the args. This unfortunately prevents us from mimicing the `docker login <host> --username ...` command.

It looks like such a feature was requested in https://github.com/urfave/cli/issues/976 ; however the issue was closed without a fix.

cmd/earthly/registry_cmds.go
```diff
@@ -0,0 +1,273 @@
...
+					Name:        "username",
+					EnvVars:     []string{"EARTHLY_REGISTRY_USERNAME"},
+					Usage:       "The username to use when logging into the registry.",
+					Required:    true,
+					Destination: &app.registryUsername,
+				},
+				&cli.StringFlag{
+					Name:        "password",
+					EnvVars:     []string{"EARTHLY_REGISTRY_PASSWORD"},
+					Usage:       "The password to use when logging into the registry.",
```
nelsam, line 50: Is it worth having a warning here about shell history? I believe that still catches younger developers off-guard. Maybe something like, `CAUTION: if you use this, your password will usually end up in your shell history file in plain text.`

nelsam (approved):
lgtm

cmd/earthly/registry_cmds.go
alexcb: I opened #2624 to tackle this, and secrets too.

cmd/earthly/root_cmds.go
```diff
@@ -287,6 +287,13 @@ Set up a whole custom git repository for a server called example.com, using a si
 			},
 			Subcommands: app.secretCmds(),
 		},
+		{
+			Name:        "registry",
+			Aliases:     []string{"registries"},
+			Description: "Manage registry access *experimental*",
+			Usage:       "Manage registry access *experimental*",
```
vladaionescu, line 294: I *think* we have to make this *beta* from the get-go, given that it's critical for CI.

vladaionescu (commented):
Changelog entry?

cmd/earthly/root_cmds.go
```diff
@@ -290,8 +290,8 @@ Set up a whole custom git repository for a server called example.com, using a si
 		{
 			Name:        "registry",
 			Aliases:     []string{"registries"},
-			Description: "Manage registry access *experimental*",
-			Usage:       "Manage registry access *experimental*",
+			Description: "Manage registry access *beta*",
+			Usage:       "Manage registry access *beta*",
```
nelsam, line 294: Ref. the `earthly doc` PR: it sounds like we historically [don't include `beta` tags in the CLI usage](https://github.com/earthly/earthly/pull/2579#discussion_r1083100766). Doesn't bother me to have this here, but it looks like it's inconsistent with other CLI commands that are in beta.

alexcb: ah ha, thanks for pointing that out.

vladaionescu (approved):

vladaionescu:
One more thought - not sure if worth doing or not. Some people have EARTHLY_ORG and EARTHLY_PROJECT set up in their `.bashrc`. I wonder if it would be better to use a special flag for `login` to point to the user secrets, instead of relying on the presence of `--org` and `--project`.

alexcb:
> Some people have EARTHLY_ORG and EARTHLY_PROJECT set up in their .bashrc. I wonder if it would be better to use a special flag for login to point to the user secrets, instead of relying on the presence of --org and --project.

Maybe we should disable setting these via environment variables for the specific case of performing a login -- it's not a command people will frequently have to run.

cmd/earthly/root_cmds.go
vladaionescu: Sorry for the confusion previously, @nelsam. I think we do want to add `beta` tags in the end.

nelsam: Yeah, gotcha. I think that I saw a few instances in git history where the CHANGELOG mentioned sub-commands in beta, but the `usage` text didn't mention anything - IIRC that's what I was going on when I asked. But I don't think it's a bad idea to include those tags. I probably just found a few instances where they weren't applied consistently.

PR #2622, opened 2023-01-26
fix requirement of --org and --project flags

alexcb:
secrets v2 introduced the concept of a `--org` and `--project` flags. This change deprecated the concept of global secrets at the org level; however, this deprecation did not affect the /user/... secret namespace.

The /user namespace can still be accessed via `earthly secrets ls /user/...`; however, the `--org` and `--project` flags were incorrectly required all the time.

This commit changes the `--org` and `--project` flags to only be required when the path is not prefixed by `/user/`.


alexcb:
old behaviour:
```
earthly secrets ls /user
NAME:
   earthly secret - Manage cloud secrets *experimental*

USAGE:
   earthly secret command [command options] [arguments...]

DESCRIPTION:
   Manage cloud secrets *experimental*

COMMANDS:
   set                      Stores a secret in the secrets store
   get                      Retrieve a secret from the secrets store
   ls                       List secrets in the secrets store
   rm                       Removes a secret from the secrets store
   migrate                  Migrate existing secrets into the new project-based structure
   permission, permissions  Manage user-level secret permissions.
   help, h                  Shows a list of commands or help for one command

OPTIONS:
   --org value      The organization to which the project belongs. [$EARTHLY_ORG]
   --project value  The organization project in which to store secrets. [$EARTHLY_PROJECT]
   --help, -h       show help (default: false)
   
Error: Required flags "org, project" not set
```

the old work-around:

```
earthly secrets --org ignore --project this ls /user
/user/earthly_integration_tests/my_test_file
/user/std/registry/asdf/password
```

example of new behaviour:
```
./build/linux/amd64/earthly secrets ls /user
/user/earthly_integration_tests/my_test_file
/user/std/registry/asdf/password
```

an error is still displayed for non-user secrets:

```
./build/linux/amd64/earthly secrets ls /earthly-technologies/
Error: the --org flag is required
```

cmd/earthly/secret_cmds.go
```diff
@@ -104,7 +104,7 @@ func (app *earthlyApp) secretCmds() []*cli.Command {
 	}
 }
 
-func (app *earthlyApp) actionSecretsListV2(cliCtx *cli.Context) error {
+func (app *earthlyApp) actionSecretsListV2(cliCtx *cli.Context) (err error) {
```
mikejholly, line 107: Just out of curiosity. why name the returned error here? It seems like it's referenced by name in each of the return instances.

mikejholly (approved):
LGTM.

cmd/earthly/secret_cmds.go
alexcb: I was on the fence about changing `path, err = app.fullSecretPath(path)` to `path, err := app.fullSecretPath(path)`; but thinking back on that, i'll revisit :)

alexcb: yeah.... not sure what I was thinking :)

vladaionescu (approved):

Issue #2625, opened 2023-01-27
unable to set feature-flags to `false`


alexcb:
consider an Earthfile where a feature flag is set to false:
```Earthfile
VERSION --referenced-save-only=false 0.6
foo:
  FROM alpine
  RUN echo foo
```

This does not work as expected;

```
diff --git a/earthfile2llb/converter.go b/earthfile2llb/converter.go
index 4762c75b..4ea407da 100644
--- a/earthfile2llb/converter.go
+++ b/earthfile2llb/converter.go
@@ -1670,6 +1670,7 @@ func (c *Converter) prepBuildTarget(ctx context.Context, fullTargetName string,
                opt.waitBlock = nil
        }
 
+       fmt.Printf("here with %v\n", c.opt.Features.ReferencedSaveOnly)
        if c.opt.Features.ReferencedSaveOnly {
                // DoSaves should only be potentially turned-off when the ReferencedSaveOnly feature is flipped
                opt.DoSaves = (cmdT == buildCmd && c.opt.DoSaves && !c.opt.OnlyFinalTargetImages)
```

this outputs `here with true`

alexcb:
We have decided to disallow setting feature flags to false; if one really wants to disable a feature, then they must use the previous version and only select the features they want.

For example, if a user wants to disable `--referenced-save-only`, then they will need to instead use version 0.5:

```
VERSION \
  --use-copy-include-patterns \
  --for-in \
  --require-force-for-unsafe-saves \
  --no-implicit-ignore \
  0.5
```

Note that the above `VERSION` definition was taken from the [CHANGELOG](https://github.com/earthly/earthly/blob/main/CHANGELOG.md#v060---2021-11-24) entry for the initial earthly version 0.6.0 release. (and then the `--referenced-save-only \` line was deleted)

PR #2627, opened 2023-01-27
fix: raise error when setting feature flag values

alexcb:
Boolean overrides for feature flags should not be possible; this change
will now raise an error instead of silently ignoring them.

While it is technically possible to implement disabling feature flags,
we chose to raise an error instead to encourage users to fully adopt
a new VERSION.

This is because feature flags are for experimental features, once they
are promoted they become part of the forward-direction of earthly;
allowing users to adopt the latest version while individually flipping a
feature off may lead to buggy experiences.

If a user really wants to disable a feature, then they will have to
remain at a previous VERSION, and instead select the (experimental) features
they wish to enable (with caution).

### DIFF
--- a/features/features.go
+++ b/features/features.go
@@ -171,7 +171,7 @@ func GetFeatures(version *spec.Version) (*Features, bool, error) {
 		return nil, false, errUnexpectedArgs
 	}
 
-	parsedArgs, err := flagutil.ParseArgsWithValueModifierAndOptions("VERSION", &ftrs, version.Args, instrumentVersion, goflags.PassDoubleDash|goflags.PassAfterNonOption|goflags.AllowBoolValues)
+	parsedArgs, err := flagutil.ParseArgsWithValueModifierAndOptions("VERSION", &ftrs, version.Args, instrumentVersion, goflags.PassDoubleDash|goflags.PassAfterNonOption)
 	if err != nil {
 		return nil, false, err
 	}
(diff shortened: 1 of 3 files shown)


alexcb:
potential fix for #2625

alexcb:
@vladaionescu what are your thoughts on the high-level direction of this fix?

nelsam:
My two cents: it sure would be nice to eventually be able to delete flags and their related branching logic. If we reject `--feature-flag=false`, we can delete feature flags and their related branching logic if/when we remove the `VERSION` that they were tied to. If we allow explicitly disabling feature flags in a newer version to fall back to logic from old versions, then it's more difficult to eventually deprecate/delete those flags (and the branching logic related to those flags).

I also think it's pretty uncommon to allow this sort of stuff. e.g. `python2` has `from __future__ import ...`, but I don't think `python3` has anything (official, at least) like `from __past__ import ...`.

alexcb:
>  If we reject --feature-flag=false, we can delete feature flags and their related branching logic if/when we remove the VERSION that they were tied to.

I thought we had some docs that stated our lifecycle of old VERSION support; but can't find it.

I think we could still maintain the goal of deleting feature flags when we mark a VERSION as end-of-life.

> from __past__ import ...

this is a great point! :laughing:

vladaionescu:
I think @nelsam makes some good points. If something is very problematic we can still tell people to downgrade completely to 0.6, but generally, we should keep pushing everyone to use the latest as much as possible.

vladaionescu (approved):

nelsam (approved):

PR #2629, opened 2023-01-28
Add beta flags to CI commands

r-LaForge:

PR #2632, opened 2023-01-30
cmd/earthly: add beta tags to 'ls' and 'doc' subcommands

nelsam:


alexcb (approved):
maybe this should be moved into #2629 as a single commit? but no strong opinion.

r-LaForge:
We talked in Slack and decided to merge this change into https://github.com/earthly/earthly/pull/2629

Issue #2178, opened 2022-09-12
WAIT/END doesn't export to remote cache


alexcb:
consider an `Earthfile` like:

```Earthfile
VERSION --wait-block 0.6

deps:
   FROM ubuntu:22.04
   RUN install test dependencies…
   SAVE IMAGE --cache-hint

ci:
    BUILD +deps
    FROM +deps
    COPY . ./
    RUN --no-cache pytest --junitxml=report.xml ; echo $? > exit-code
    WAIT
        SAVE ARTIFACT report.xml AS LOCAL report.xml
    END
    RUN exit $(cat exit-code)
```

The `SAVE IMAGE --cache-hint` doesn't get triggered by the `END`, even though we wait for `SAVE ARTIFACT` which depends on the previous llb states being executed.

based on a post by @glehmann in https://github.com/earthly/earthly/issues/988#issuecomment-1243495712_

alexcb:
I suspect we will need to extend

```
type ExportRequest struct {
	Refs     map[string]Reference
	Metadata map[string][]byte
}
```

to support referencing

```
type SolveOpt struct {

	CacheExports          []CacheOptionsEntry
	CacheImports          []CacheOptionsEntry
```

alexcb:
debugging notes:

I've been running:

```
./test.sh --build-arg DOCKERHUB_AUTH=true --build-arg DOCKERHUB_USER_SECRET=+secrets/earthly-technologies/dockerhub-mirror/user --build-arg DOCKERHUB_TOKEN_SECRET=+secrets/earthly-technologies/dockerhub-mirror/pass --build-arg DOCKERHUB_MIRROR=registry-1.docker.io.mirror.corp.earthly.dev
```

under `tests/remote-cache`.

when I run `curl localhost:5000/v2/test1/manifests/latest` on `VERSION 0.6`, I get:

```
...
   "history": [
      {
[...]

PR #2635, opened 2023-01-31
work-around for implicit WAIT/END inline cache bug

alexcb:
If a user is using VERSION 0.7 without any explicit WAIT block, we will use the build function to perform the export due to a bug that exists in the earthly-specific Export buildkit grpc call, which skips the inline cache.


alexcb:
note that I did some extra investigation around why it's not working under https://github.com/earthly/earthly/issues/2178#issuecomment-1411121506

and making it work in the buildkit's export function is way too hacky; so we'll need to do this work-around until we find time to re-evaluate an appropriate fix.

vladaionescu (approved):

PR #2639, opened 2023-02-01
Shorten GitHub prefix during the `prettyPrefix` function

r-LaForge:
# Before
image

# After

image


conslogging/conslogging.go
```diff
@@ -467,11 +468,27 @@ func (cl ConsoleLogger) color(c *color.Color) *color.Color {
...
+)
+
 func prettyPrefix(prefixPadding int, prefix string) string {
 	if prefixPadding == NoPadding {
 		return prefix
 	}
 
+	// If the prefix contains a GitHub hash - shorten it
+	gitHashPart := githubRegExp.FindString(prefix)
+	resultingHashLength := 10 // 1 for ':', 1 for '+', 8 for hash
```
vladaionescu, line 484: I would make the hash length 7, to match what GitHub does too.

```suggestion
	resultingHashLength := 9 // 1 for ':', 1 for '+', 7 for hash
```

vladaionescu (approved):

conslogging/conslogging.go
r-LaForge: done

conslogging/conslogging.go
```diff
@@ -467,11 +468,27 @@ func (cl ConsoleLogger) color(c *color.Color) *color.Color {
 	return noColor
 }
 
+var (
+	// githubRegExp Matches :2dd88e53f2e59e96ec1f9215f24a3981e5565edf+ in a target.
+	// 	Prefix containing hash may resemble: g/e/hello-world:2dd88e53f2e59e96ec1f9215f24a3981e5565edf+base
+	githubRegExp = regexp.MustCompile(`:[a-zA-Z0-9]*\\+`)
```
alexcb, line 474: this feels too loose; it could match `g/e/hello-world:feedfacecafe`, which isn't guaranteed to be a git sha.

instead we should match on exactly 40 chars.

```suggestion
	githubRegExp = regexp.MustCompile(`:[a-zA-Z0-9]{40}+`)
```

alexcb (changes requested):
can you add a unit test for `prettyPrefix` to make sure we don't truncate branch names.

conslogging/conslogging.go
mikejholly, line 474: Will this match any string after `:`? Branch name, etc? Do SHAs ever contain upper-case chars?

Maybe we should add a length range like `{8,}`?

conslogging/conslogging.go
```diff
@@ -467,11 +468,27 @@ func (cl ConsoleLogger) color(c *color.Color) *color.Color {
...
 		return prefix
 	}
 
+	// If the prefix contains a GitHub hash - shorten it
+	gitHashPart := githubRegExp.FindString(prefix)
+	resultingHashLength := 9 // 1 for ':', 1 for '+', 7 for hash
+	if len(gitHashPart) > resultingHashLength {
+		hash := gitHashPart[1:]
+		hash = hash[:len(gitHashPart)-1]
+		hash = hash[:7]
```
mikejholly, line 488: Are we truncating the string twice here (L487)? Can that be done in one op?

conslogging/conslogging.go
r-LaForge: I've done just that - and also found pre-existing bugs handling branch names with special characters.
Great push for tests ended with an extra bug fix :)

r-LaForge: The code now properly handles branch name vs sha

conslogging/conslogging.go
r-LaForge: done

conslogging/conslogging.go
```diff
@@ -467,11 +468,29 @@ func (cl ConsoleLogger) color(c *color.Color) *color.Color {
...
+)
+
 func prettyPrefix(prefixPadding int, prefix string) string {
-	if prefixPadding == NoPadding {
+	if prefixPadding <= NoPadding {
 		return prefix
 	}
 
+	// tempReplacement used to temporarily remove branch name / sha from string
+	const tempReplacement = "{{}}{{{}{}}}"
```
alexcb, line 486: this feels fragile, I think it would be better to check if `githubRegExp` matches the input, and then if so do a replacement.

if that regex doesn't match, then no replacements should be required.

conslogging/conslogging.go
alexcb: thanks for adding the tests! nice find with the extra bug.

I think it would be useful to add a test that has a branch name that matches `a-f`, such as the original example I gave `feedfacecafe`.

I also just noticed that the regex should be `a-f` and not `a-z`, since a git sha is encoded as a hexidecimal.

conslogging/conslogging.go
```diff
@@ -467,11 +468,29 @@ func (cl ConsoleLogger) color(c *color.Color) *color.Color {
 	return noColor
 }
 
+var (
+	// githubRegExp Matches :2dd88e53f2e59e96ec1f9215f24a3981e5565edf+ in a prefix.
+	// 	Prefix containing hash may resemble: g/e/hello-world:2dd88e53f2e59e96ec1f9215f24a3981e5565edf+base
+	//	Prefix must be exactly 40 characters
+	githubRegExp = regexp.MustCompile(`:[a-z0-9]{40}\+`)
```
alexcb, line 475:
```suggestion
	githubRegExp = regexp.MustCompile(`:[a-f0-9]{40}\+`)
```

conslogging/conslogging.go
```diff
@@ -467,11 +468,29 @@ func (cl ConsoleLogger) color(c *color.Color) *color.Color {
...
 		return prefix
 	}
 
+	// tempReplacement used to temporarily remove branch name / sha from string
+	const tempReplacement = "{{}}{{{}{}}}"
+	branchOrSHA := branchOrSHARegExp.FindString(prefix)
+	prefix = strings.Replace(prefix, branchOrSHA, tempReplacement, 1)
+	sha := githubRegExp.FindString(branchOrSHA)
+	if sha != "" {
+		branchOrSHA = sha[:7] + "+"
```
alexcb, line 491: can we stick to the standard 8 chars (which is what `git rev-parse --short HEAD` gives us)
```suggestion
		branchOrSHA = sha[:8ball: ] + "+"
```

r-LaForge: Sure - [Vlad had commented before](https://github.com/earthly/earthly/pull/2639#discussion_r1093845118) to bring it down to 7.

alexcb: lol, too many cooks :)

`EARTHLY_GIT_SHORT_HASH` provides 8, which comes from the default git implementation.

but if 7 is picked specifically to match github, then ignore my original comment.

conslogging/conslogging.go
r-LaForge: This particular replacement change is meant to tackle when the branch name has special characters; specifically, a `/`. I just use it in both scenarios.

The previous version of this function would cut off a branch name like `ryan/implement-feature` due to the `/`.
I'll admit I took a shortcut here - doing this rather than refactoring the entire function.

It looks odd - hacky even - though I'm not sure it's fragile. Could you elaborate?

conslogging/conslogging_test.go
```diff
@@ -0,0 +1,85 @@
...
+			prefix:        "+run",
+			expected:      "+run",
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			assert.Equal(t, tc.expected, prettyPrefix(tc.prefixPadding, tc.prefix))
+		})
+	}
```
nelsam, line 70: Not a huge deal in this case, but I find that always running tests in parallel (a) helps uncover bugs and (b) keeps feedback cycles very quick. These tests don't strictly need it, but as far as I can tell they shouldn't be harmed by it.
```suggestion
	for _, tc := range testCases {
		tc := tc
		t.Run(tc.name, func(t *testing.T) {
			t.Parallel()
			assert.Equal(t, tc.expected, prettyPrefix(tc.prefixPadding, tc.prefix))
		})
	}
```

conslogging/conslogging_test.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+func padding(n int) string {
+	s := ""
+	for i := 0; i < n; i++ {
+		s += " "
+	}
+	return s
```
nelsam, line 78:
```suggestion
	return strings.Repeat(" ", n)
```

conslogging/conslogging.go
r-LaForge: done

conslogging/conslogging.go
r-LaForge: Refactored to remove the tempReplacement hack

conslogging/conslogging.go
r-LaForge: I've added that test now as well

Earthfile
```diff
@@ -230,6 +230,7 @@ unit-test:
     END
     BUILD ./ast+unit-test
     BUILD ./util/deltautil+unit-test
+    RUN go test ./conslogging
```
nelsam, line 233: This should be handled by the `ARG pkgname = ./...` (used by `./not-a-unit-test.sh`). The only reason that `ast` and `util/deltautil` are run separately is that they are separate go modules (i.e. they have their own `go.mod` and `go.sum`), so `./...` skips them (among a few other problems).

nelsam (changes requested):
A few minor comments, nothing critical.

Earthfile
r-LaForge: Oh - I didn't see `pkgname` used and the whole `not-a-unit-test.sh` didn't make it obvious. 

Could we pass it explicitly to make it more obvious?

nelsam: Yeah, I think that would be a good change.

conslogging/conslogging_test.go
r-LaForge: done

Earthfile
nelsam: It might also be worth commenting above `BUILD ./ast+unit-test` something like
```
# The following are separate go modules and need to be tested separately.
```

r-LaForge: done and added comment

conslogging/conslogging_test.go
r-LaForge: now they're in parallel.

conslogging/conslogging_test.go
```diff
@@ -0,0 +1,80 @@
...
+		{
+			name:          "negative padding results in no change",
+			prefixPadding: -10,
+			prefix:        "+run",
+			expected:      "+run",
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
```
nelsam, line 74: One extra catch here: `tc` is reused in each loop, which means each test run will use the same memory for that variable. In serial, you usually don't notice (unless you're storing values on `tc` during the test), but it's technically test pollution.

With `t.Parallel()`, the test pollution is much more obvious - only one test case will actually run (and it will run many times), because by the time the first goroutine kicks off, the `tc` variable has already been set to the very last test case.

The fix is to add `tc := tc` between lines 73 and 74 - this allocates new memory for each test so that they're not sharing memory. I should have made that more clear in my original review comments.

nelsam (approved):

conslogging/conslogging_test.go
r-LaForge: ah 🤦 I changed this too fast without considering that.
Thanks for pointing it out.

nelsam: No problem, I caught it because I've made that mistake more times than I can count :sweat_smile: 

I should add a helper to `onpar` to fix that, since `onpar`'s whole thing is forcing tests to run in parallel...

alexcb (approved):

Issue #2640, opened 2023-02-02
COPY --keep-own not working


Itxaka:
earthfile version: `earthly-linux-amd64 version v0.6.30 743687991d227f2f7e2e6ba11a90f443b4f99c67 linux/amd64; Fedora Linux 37 (Workstation Edition)`

example earthfile:

```
VERSION 0.6

test1:
  FROM alpine
  WORKDIR /build
  RUN mkdir testperms
  RUN adduser -D testuser
  RUN chown testuser testperms
  RUN ls -ltra 
  SAVE ARTIFACT /build build

test2:
  FROM alpine
  WORKDIR /build
  COPY --keep-own +test1/build .
  RUN ls -ltra 
```

output from `earthly-linux-amd64 +test2` :

```
 1. Init 🚀
————————————————————————————————————————————————————————————————————————————————

           buildkitd | Found buildkit daemon as docker container (earthly-buildkitd)

 2. Build 🔧
————————————————————————————————————————————————————————————————————————————————

              alpine | --> Load metadata linux/amd64
              +test1 | --> FROM alpine
              +test1 | [██████████] 100% resolve docker.io/library/alpine@sha256:f271e74b17ced29b915d351685fd4644785c6d1559dd1f2d4189a5e851ef753a
              +test1 | [██████████] 100% sha256:8921db27df2831fa6eaa85321205a2470c669b855f3ec95d5a3c2b46de0442c9
              +test1 | [██████████] 100% extracting sha256:8921db27df2831fa6eaa85321205a2470c669b855f3ec95d5a3c2b46de0442c9
              +test1 | --> WORKDIR /build
              +test1 | --> RUN mkdir testperms
              +test1 | --> RUN adduser -D testuser
              +test1 | --> RUN chown testuser testperms
              +test1 | --> RUN ls -ltra
              +test1 | total 24
              +test1 | drwxr-xr-x    1 testuser root          4096 Feb  2 14:14 testperms
              +test1 | drwxr-xr-x    1 root     root          4096 Feb  2 14:14 ..
              +test1 | drwxr-xr-x    1 root     root          4096 Feb  2 14:14 .
              +test1 | --> SAVE ARTIFACT /build +test1/build
              +test2 | --> COPY +test1/build ./
              +test2 | --> RUN ls -ltra
              +test2 | total 20
              +test2 | drwxr-xr-x    2 root     root          4096 Apr 16  2020 testperms
[...]

PR #2643, opened 2023-02-02
doc: --keep-own must also exist in SAVE ARTIFACT

alexcb:

### DIFF
--- a/docs/earthfile/earthfile.md
+++ b/docs/earthfile/earthfile.md
@@ -406,6 +406,10 @@ Instructs Earthly to not overwrite the file creation timestamps with a constant.
 
 Instructs Earthly to keep file ownership information. This applies only to the *artifact form* and has no effect otherwise.
 
+{% hint style='info' %}
+Note that one must include the flag in the corresponding `SAVE ARTIFACT --keep-own ...` command.
+{% endhint %}
+
 ##### `--if-exists`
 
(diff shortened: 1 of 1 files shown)


alexcb:
relates to #2640

although I'm not sure if the `SAVE ARTIFACT --keep-own` flag should even exist -- maybe we should have it always apply, because without the corresponding `COPY --keep-own`, it'll end up back as the root owner.

docs/earthfile/earthfile.md
```diff
@@ -406,6 +406,10 @@ Instructs Earthly to not overwrite the file creation timestamps with a constant.
 
 Instructs Earthly to keep file ownership information. This applies only to the *artifact form* and has no effect otherwise.
 
+{% hint style='info' %}
+Note that one must include the flag in the corresponding `SAVE ARTIFACT --keep-own ...` command.
```
vladaionescu, line 410: Nit: We generally use "you" as a tone in other pages.

Also, added a clarification that this remark applies to certain COPY's only.


```suggestion
Note that you must include the flag in the corresponding `SAVE ARTIFACT --keep-own ...` command, if using *artifact form*.
```

vladaionescu (approved):

vladaionescu:
> although I'm not sure if the SAVE ARTIFACT --keep-own flag should even exist -- maybe we should have it always apply, because without the corresponding COPY --keep-own, it'll end up back as the root owner.

Good point - it might require some investigation / testing to make sure other things are not impacted though.

PR #2644, opened 2023-02-02
always apply SAVE ARTIFACT --keep-own flag

### LINKED ISSUE #2640: COPY --keep-own not working

alexcb:
In order to copy a file from one target to another with the same ownership information requires the `--keep-own` flag be applied to both `SAVE ARTIFACT` and `COPY`.

The final `COPY`'s `--keep-own` flag should be sufficient for determining if the original file's permissions are to be copied over or not.

This fixes #2640

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,21 +76,22 @@ Declaring `VERSION 0.7` is equivalent to
 
 ```
 VERSION \
-  --explicit-global \
   --check-duplicate-images \
+  --earthly-git-author-args \
+  --earthly-locally-arg \
   --earthly-version-arg \
-  --use-cache-command \
-  --use-host-command \
-  --use-copy-link \
+  --explicit-global \
   --new-platform \
   --no-tar-build-output \
-  --use-no-manifest-list \
-  --use-chmod \
+  --save-artifact-keep-own \
   --shell-out-anywhere \
-  --earthly-locally-arg \
-  --use-project-secrets \
+  --use-cache-command \
+  --use-chmod \
+  --use-copy-link \
+  --use-host-command \
+  --use-no-manifest-list \
   --use-pipelines \
-  --earthly-git-author-args \
+  --use-project-secrets \
   --wait-block \
   0.6
 ```
@@ -110,6 +111,7 @@ For more information on the individual Earthfile feature flags see the [Earthfil
 - `earthly ls` has been promoted from *experimental* to *beta* status.
 - Setting a `VERSION` feature flag boolean to false (or any other value) will now raise an error; previously it was syntactically valid but had no effect.
 - `SAVE ARTIFACT <path> AS LOCAL ...` when used under a `TRY` / `FINALLY` can fail to be fully transferred to the host when the `TRY` command fails (resulting in an partially transferred file); an underflow can still occur, and is now detected and will not export the partial file. [2452](https://github.com/earthly/earthly/issues/2452)
+- The `--keep-own` flag for `SAVE ARTIFACT` is now applied by default; note that `COPY --keep-own` must still be used in order to keep ownership
 
 ### Added
 
--- a/earthfile2llb/interpreter.go
+++ b/earthfile2llb/interpreter.go
@@ -1048,6 +1048,13 @@ func (i *Interpreter) handleSaveArtifact(ctx context.Context, cmd spec.Command) 
 		return nil
 	}
 
+	if i.converter.ftrs.SaveArtifactKeepOwn {
+		if opts.KeepOwn {
+			fmt.Fprintf(os.Stderr, "Deprecation: SAVE ARTIFACT --keep-own is now applied by default, setting it no longer has any effect\n")
+		}
+		opts.KeepOwn = true
+	}
+
 	err = i.converter.SaveArtifact(ctx, saveFrom, expandedSaveTo, expandedSaveAsLocalTo, opts.KeepTs, opts.KeepOwn, opts.IfExists, opts.SymlinkNoFollow, opts.Force, i.pushOnlyAllowed)
 	if err != nil {
 		return i.wrapError(err, cmd.SourceLocation, "apply SAVE ARTIFACT")
--- a/features/features.go
+++ b/features/features.go
@@ -40,6 +40,7 @@ type Features struct {
 	GitCommitAuthorTimestamp bool `long:"git-commit-author-timestamp" description:"include EARTHLY_GIT_COMMIT_AUTHOR_TIMESTAMP arg"`
 	NewPlatform              bool `long:"new-platform" description:"enable new platform behavior"`
 	NoTarBuildOutput         bool `long:"no-tar-build-output" description:"do not print output when creating a tarball to load into WITH DOCKER"`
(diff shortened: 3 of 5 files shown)


CHANGELOG.md
```diff
@@ -76,21 +76,22 @@ Declaring `VERSION 0.7` is equivalent to
...
   --earthly-version-arg \
-  --use-cache-command \
-  --use-host-command \
-  --use-copy-link \
+  --explicit-global \
   --new-platform \
   --no-tar-build-output \
-  --use-no-manifest-list \
-  --use-chmod \
+  --save-artifact-keep-own \
```
alexcb, line 86: here's the new flag; the rest were simply ordered alphabetically.

Itxaka:
nice!

Also maybe it could fit in this patch, when you pass the --keep-own flag to a COPY it doesn show up in the log while other flags do. Maybe we need to add it to https://github.com/earthly/earthly/blob/main/earthfile2llb/converter.go#L508 and https://github.com/earthly/earthly/blob/main/earthfile2llb/converter.go#L549 ?

vladaionescu:
I'm a bit nervous about the implications of this flag regarding caching. Maybe my worries are not well-founded though as I can't think of a specific case where this is a problem. Just wanted to ask ourselves if we're not missing anything.

nelsam (approved):
It all looks reasonable, code-wise. As long as Vlad's concerns are addressed, I have no objections.

alexcb:
> I'm a bit nervous about the implications of this flag regarding caching. Maybe my worries are not well-founded though as I can't think of a specific case where this is a problem. Just wanted to ask ourselves if we're not missing anything.

If we think of a `SAVE ARTIFACT` as copying `(filename, filedata, user id, group id)` currently by default we will copy `("some-filename", <binary data here>, 0, 0)` where as when the `--keep-own` flag is used, instead we will copy `("some-filename", <binary data here>, 1000, 1000)` (or some other userid/groupid if the image has many users).

New users will start at `1000`, and be incremented sequentially.

If this data was being copied from different hosts then it would produce very different results, since it would be affected by the number of users on the host. (additionally it would be very odd to pull in `/etc/passwd` from the host, where as in the test in this PR shows two targets that both derive from the same target containing `/etc/passwd`)

I can't think of a case where introducing these deterministic user IDs will lead to caching issues. If we were also pulling across the timestamp value, then that would be problematic.

alexcb:
Here's another quick test of earthly executing the test from this PR, but directly (without any earthly-in-earthly);
on the second run, all steps are cached:

```
y$ ./build/linux/amd64/earthly ~/test/keepown+test
 Init 🚀

 Build 🔧

         alpine:3.15 | --> Load metadata alpine:3.15 linux/amd64
/h/a/t/keepown+withtestuser | --> FROM alpine:3.15
/h/a/t/keepown+withtestuser | [----------] 100% resolve docker.io/library/alpine:3.15@sha256:cf34c62ee8eb3fe8aa24c1fab45d7e9d12768d945c3f5a6fd6a63d901e898479
/h/a/t/keepown+withtestuser | *cached* --> RUN adduser -D testuser
/h/a/t/keepown+withtestuser | *cached* --> WORKDIR /test
/h/a/t/keepown+producer | *cached* --> RUN touch testperms
/h/a/t/keepown+producer | *cached* --> RUN chown testuser:testuser testperms
/h/a/t/keepown+test-unknown-user | *cached* --> WORKDIR /test
/h/a/t/keepown+test-unknown-user | *cached* --> COPY /home/alex/test/keepown+producer/testperms ./
/h/a/t/keepown+test-unknown-user | *cached* --> RUN test "$(stat -c '%u' testperms)" = "1000"
/h/a/t/keepown+test-unknown-user | *cached* --> RUN test "$(stat -c '%U' testperms)" = "UNKNOWN"
/h/a/t/keepown+test-unknown-user | *cached* --> RUN test "$(stat -c '%g' testperms)" = "1000"
/h/a/t/keepown+test-unknown-user | *cached* --> RUN test "$(stat -c '%G' testperms)" = "UNKNOWN"
/h/a/t/keepown+test-known-user | *cached* --> WORKDIR /test
/h/a/t/keepown+test-known-user | *cached* --> COPY /home/alex/test/keepown+producer/testperms ./
/h/a/t/keepown+test-known-user | *cached* --> RUN test "$(stat -c '%u' testperms)" = "1000"
/h/a/t/keepown+test-known-user | *cached* --> RUN test "$(stat -c '%U' testperms)" = "testuser"
/h/a/t/keepown+test-known-user | *cached* --> RUN test "$(stat -c '%g' testperms)" = "1000"
/h/a/t/keepown+test-known-user | *cached* --> RUN test "$(stat -c '%G' testperms)" = "testuser"
```

PR #2646, opened 2023-02-03
split .env to .arg and .secret files

alexcb:
splits up `.env` into `.arg` and `.secret` files, to make it more clear which are build args vs secrets. `.env` will only remain for setting environment variables (for changing `EARTHLY_*` config options).


cmd/earthly/flags.go
```diff
@@ -141,11 +141,25 @@ func (app *earthlyApp) rootFlags() []cli.Flag {
...
 		},
+		&cli.StringFlag{
+			Name:        "arg-path",
+			EnvVars:     []string{"EARTHLY_ARG_PATH"},
+			Usage:       "Use values from this file as earthly buildargs",
+			Value:       defaultArgFile,
+			Destination: &app.argFile,
+		},
+		&cli.StringFlag{
+			Name:        "secret-path",
```
alexcb, line 157: unfortunately `--secret-file` is already taken:

```
-secret-file value                    A secret override, specified as <key>=<path> [$EARTHLY_SECRET_FILES]
```

I'm not sure what a good name for this is:
 - `dot-secret-file`
 - `secret-file-path`
 - `secret-path`
 - `override-secret-file`

Ideally we should keep all the `.env`, `.arg`, and `.secret` override options following the same naming scheme.

cmd/earthly/flags.go
```diff
@@ -141,11 +141,25 @@ func (app *earthlyApp) rootFlags() []cli.Flag {
 		},
 		&cli.StringFlag{
 			Name:        envFileFlag,
-			EnvVars:     []string{"EARTHLY_ENV_FILE"},
-			Usage:       "Use values from this file as earthly environment variables, buildargs, or secrets",
+			EnvVars:     []string{"EARTHLY_ENV_PATH"},
```
alexcb, line 144: not sure if we should keep the old value for backwards compatibility? Since this PR is backwards incompatable, we can rename this too. See my comment below regarding following the same naming scheme.

cmd/earthly/flags.go
nelsam: My two copper go to `secret-file-path`.  `dot-secret-file` feels weird if they're not using a dot-file, e.g. `--dot-secret-file secrets`. `secret-path` worries me from a forward compatibility sentiment - if we add the ability to load secrets from a directory (instead of a file) or if we integrate something like vault, `secret-path` may all of a sudden be confusing. `override-secret-file` just feels like too many syllables.

My preference is pretty slight, though. I think the other options aren't bad, these are just a few reasons I like them slightly less.

cmd/earthly/build_cmd.go
```diff
@@ -296,8 +297,30 @@ func (app *earthlyApp) actionBuildImp(cliCtx *cli.Context, flagArgs, nonFlagArgs
...
+	for k := range dotEnvMap {
+		if _, found := validEnvNames[k]; !found {
+			app.console.Warnf("unexpected env \"%s\": as of v0.7.0, --build-arg values must be defined in .arg (and --secret values in .secret)", k)
+		}
+	}
+
+	argMap, err := godotenv.Read(app.argFile)
+	if err != nil {
+		// ignore ErrNotExist when using default .env file
+		if app.argFile != defaultArgFile || !errors.Is(err, os.ErrNotExist) {
```
nelsam, line 310: The `cli.StringFlag` type has an `IsSet()` method. I'm not familiar enough with `urfave` to know for sure, but would that allow us to check `if app.argFile.IsSet()` instead of `if app.argFile != defaultArgFile`? It would be nice to still error if a user explicitly passes the default.

cmd/earthly/build_cmd.go
```diff
@@ -296,8 +297,30 @@ func (app *earthlyApp) actionBuildImp(cliCtx *cli.Context, flagArgs, nonFlagArgs
...
+			return errors.Wrapf(err, "read %s", app.argFile)
+		}
+	}
+
+	secretsFileMap, err := godotenv.Read(app.secretFile)
+	if err != nil {
+		// ignore ErrNotExist when using default .env file
+		if app.secretFile != defaultSecretFile || !errors.Is(err, os.ErrNotExist) {
```
nelsam, line 318: Same comment as `argFile` about differentiating between the user explicitly passing the default file and us falling back to the default file.

nelsam (approved):
Minor comments about defaults vs explicitly assigned defaults, but otherwise lgtm

cmd/earthly/build_cmd.go
alexcb: good point -- I thought there was a reason for this, but I think it's only for the initial loading of `.env` which must occur before `urfave` is initialised; however, there's no reason I can think of for this code not to use `IsSet()`. I've made the switch.

cmd/earthly/flags.go
alexcb: I switched to `secret-file-path`, as it has two votes now. thanks for the feedback.

vladaionescu (commented):
Add an entry in the changelog?

alexcb:
> Add an entry in the changelog?

:facepalm: thanks for the reminder.

I copy/pasted the previous rc entry in one commit, then the second commit is the new addition.

PR #2653, opened 2023-02-07
Medium is Default Size for Satellites

brandonSc:
Newly launched satellites will now be in the medium size instead of large (which is the server default).

### DIFF
--- a/cloud/satellite.go
+++ b/cloud/satellite.go
@@ -43,6 +43,11 @@ const (
 	SatelliteSizeXLarge = "xlarge"
 )
 
+const (
+	SatellitePlatformAMD64 = "linux/amd64"
+	SatellitePlatformARM64 = "linux/arm64"
+)
+
 const DefaultSatelliteSize = SatelliteSizeMedium
 
 // SatelliteInstance contains details about a remote Buildkit instance.
@@ -319,4 +324,13 @@ var validSizes = map[string]bool{
 
 func ValidSatelliteSize(size string) bool {
 	return validSizes[size]
+}
+
+var validPlatforms = map[string]bool{
+	SatellitePlatformAMD64: true,
+	SatellitePlatformARM64: true,
+}
+
+func ValidSatellitePlatform(size string) bool {
+	return validPlatforms[size]
 }
--- a/cloud/satellite.go
+++ b/cloud/satellite.go
@@ -35,6 +35,15 @@ const (
 	SatelliteStatusUnknown = "Unknown"
 )
 
+const (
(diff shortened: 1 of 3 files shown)


cloud/satellite.go
```diff
@@ -298,3 +308,16 @@ func satelliteStatus(status pb.SatelliteStatus) string {
 		return SatelliteStatusUnknown
 	}
 }
+
+var validSizes = map[string]bool{
+	"":                  true,
```
brandonSc, line 313: I left empty-string as supported here because in theory the server accepts it and defaults to large in that case. The outer command code defaults it to medium now though 🤷

nelsam (approved):
Looks fine to me. Left a minor comment about `IsSet` on the size flag, but that seems like an odd edge case.

cmd/earthly/satellite_cmds.go
```diff
@@ -310,6 +310,13 @@ func (app *earthlyApp) actionSatelliteLaunch(cliCtx *cli.Context) error {
 		return err
 	}
 
+	if size == "" {
```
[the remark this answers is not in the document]
brandonSc: I didn't think about that, nor did I even know this method existed 😄 . I suppose that would be an case worth checking for. I wonder how many other places in the code could use a check like that.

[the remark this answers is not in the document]
brandonSc: Added this.

cmd/earthly/satellite_cmds.go
```diff
@@ -310,6 +312,14 @@ func (app *earthlyApp) actionSatelliteLaunch(cliCtx *cli.Context) error {
 		return err
 	}
 
+	if cliCtx.IsSet(satelliteSizeFlag) {
+		if !cloud.ValidSatelliteSize(size) {
+			return errors.Errorf("not a valid size: %s", size)
+		}
+	} else {
+		size = cloud.DefaultSatelliteSize
```
alexcb, line 320: Could this be defined via the default `Value` `StringFlag` field instead?

brandonSc: Yes, for sure. Wasn't aware of that feature.

brandonSc: Added this.

vladaionescu (commented):
Can you add this to the changelog too?

alexcb (approved):

brandonSc:
I double-checked in Grammarly for any more typos, and it looked good. Thanks to everyone for helping me refine this!

PR #2654, opened 2023-02-07
Docs for Satellite Updates

brandonSc:
Documents new features for auto-updates and manual updates, as well explaining the revision system in general for users.

Technically these features still require `--feature-flag revisions`, however, we want to make that standard with the release of 0.7 earthly.

### DIFF
--- a/docs/cloud/satellites/managing.md
+++ b/docs/cloud/satellites/managing.md
@@ -98,16 +98,21 @@ Currently selected: No
 
 ### Clearing cache
 
-The easiest way to clear cache on a satellite is to restart it using the `update` command with `--drop-cache` flag.
+There are two ways to clear the cache on satellite.
+
+#### Recreating the Underlying Satellite Instance (often faster)
+
+Running the `update`  command with `--drop-cache` will relaunch the instance with an empty cache volume.
+Note that this operation can take a while, and the satellite may also receive any available updates during the process.
 
 ```bash
 earthly satellite update --drop-cache my-satellte
 ```
+#### Using the prune command (slower)
 
-Note that this operation can take a while, and the satellite may also receive any available updates during the process.
-
-The `earthly prune` command can also be used. In many cases, however, it takes longer than using `satellite update`.
-The prune command also requires the satellite to be selected before running.
+The `earthly prune` command also works on satellites.
+It usually takes longer than running `satellite update`, however, it does not trigger a relaunch.
+The prune command requires the satellite to be selected before running.
 
 ```bash
 earthly prune -a
@@ -115,14 +120,13 @@ earthly prune -a
 
 ### Updating a satellite
 
-By default satellites will automatically update as new versions become available. 
-Alternativly, satellite can also be pinned to specific earthly versions, however they will still receive patches and security fixes.
+Satellites receive version updates by default, unless they are pinned to a specific version (by using the `--version` launch flag).
+Pinned versions will still receive minor patches, such as security updates.
 
 #### Auto-Update Maintenance Windows
 
 Maintenance windows are set between 2AM and 4AM in the timezone where the satellite is launched by default.
-If you prefer a different maintenance window, you can specify it in the launch command using the `--maintenance-window` flag.
-Using this flag, you can pass the start time in 24-hr format. The satellite will update within 2 hours of specified start time.
+The start time of the 2 hour window can be explicitly set by passing a 24-hr formatted time to the `--maintenance-window` flag of the launch command.
 
 Here's an example showing a satellite launched with a custom maintenance window of 4AM to 6AM:
 
@@ -131,7 +135,7 @@ earthly satellite launch --maintenance-window 04:00 my-satellite
 ```
 
 Note that updates will only happen during the maintenance window while the satellite is asleep. If the satellite remains in use 
-for the entire duration of the maintenance window, then the update will be aborted until the next day.
+for the entire duration of the maintenance window, then the update will be re-attempted the next day.
 
 #### Version Pinning
 
--- a/docs/cloud/satellites/managing.md
+++ b/docs/cloud/satellites/managing.md
@@ -117,7 +117,6 @@ earthly prune -a
 
 By default satellites will automatically update as new versions become available. 
 Alternativly, satellite can also be pinned to specific earthly versions, however they will still receive patches and security fixes.
-Satellite updates will only run during their specific maintenance window.
 
 #### Auto-Update Maintenance Windows
 
--- a/docs/cloud/satellites/managing.md
+++ b/docs/cloud/satellites/managing.md
@@ -137,7 +137,7 @@ for the entire duration of the maintenance window, then the update will be abort
 #### Version Pinning
 
 If you want to prevent your satellite from automatically upgrading to a new earthly version, you can pin your version using the `--version` flag.
-Note that satellites on pinned versions may still receive auto-updates during a maintenance window, however, these updaets will be important stability or security patches that will not change the version of earthly.
+Note that satellites on pinned versions may still receive auto-updates during a maintenance window; however, these updaets will be important stability or security patches that will not change the version of earthly.
 
 ```bash
 earthly satellite launch --version v0.6.29 my-satellite
--- a/docs/cloud/satellites/managing.md
+++ b/docs/cloud/satellites/managing.md
@@ -122,7 +122,7 @@ Satellite updates will only run during their specific maintenance window.
 #### Auto-Update Maintenance Windows
 
 Maintenance windows are set between 2AM and 4AM in the timezone where the satellite is launched by default.
-If you prefer a different maintenance window, however, you can specify it in the launch command using the `--maintenance-window` flag.
+If you prefer a different maintenance window, you can specify it in the launch command using the `--maintenance-window` flag.
 Using this flag, you can pass the start time in 24-hr format. The satellite will update within 2 hours of specified start time.
 
 Here's an example showing a satellite launched with a custom maintenance window of 4AM to 6AM:
@@ -137,7 +137,7 @@ for the entire duration of the maintenance window, then the update will be abort
 #### Version Pinning
 
 If you want to prevent your satellite from automatically upgrading to a new earthly version, you can pin your version using the `--version` flag.
-Note that satellites on pinned versions may still receive auto-updates during a maintenance window; however, these updaets will be important stability or security patches that will not change the version of earthly.
+Note that satellites on pinned versions may still receive auto-updates during a maintenance window; however, these updates will be important stability or security patches that will not change the version of earthly.
 
 ```bash
 earthly satellite launch --version v0.6.29 my-satellite
@@ -155,7 +155,7 @@ The following example updates a satellite to the latest version, respecting any 
 earthly satellite update my-satellite
 ```
 
-The following command updates the satellite and clear its cache during the process. If no update is available, this command will still clear the cache.
+The following command updates the satellite and clears its cache during the process. If no updates are available, this command will still clear the cache.
 
 ```bash
 earthly satellite update --drop-cache my-satellite
--- a/docs/cloud/satellites/managing.md
+++ b/docs/cloud/satellites/managing.md
@@ -98,22 +98,87 @@ Currently selected: No
 
 ### Clearing cache
 
-To clear the cache of a satellite, run the following while a satellite is selected:
+The easiest way to clear cache on a satellite is to restart it using the `update` command with `--drop-cache` flag.
+
+```bash
+earthly satellite update --drop-cache my-satellte
+```
+
+Note this operation will take around 6-minutes, and the satellite may also receive any available updates during the process.
+
+The `earthly prune` command can also be used. In many cases, however, it may take longer than using `satellite update`. 
+The prune command requires the satellite to be selected before running.
 
 ```bash
 earthly prune -a
 ```
 
-### Upgrading a satellite
+### Updating a satellite
 
-Currently, satellites do not have an auto-update mechanism built in. In order to get a newer version of a satellite, you need to manually remove and re-launch the satellite. Note that this operation resets the cache.
+By default satellites will automatically update as new versions become available. 
+Alternativly, satellite can also be pinned to specific earthly versions, however they will still receive patches and security fixes.
+Satellite updates will only run during their specific maintenance window.
+
+#### Auto-Update Maintenance Windows
+
+Maintenance windows are set between 2AM and 4AM in the timezone where the satellite is launched by default.
+If you prefer a different maintenance window, however, you can specify it in the launch command using the `--maintenance-window` flag.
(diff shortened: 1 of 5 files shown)


docs/cloud/satellites/managing.md
```diff
@@ -98,22 +98,87 @@ Currently selected: No
...
+earthly satellite launch --version v0.6.29 my-satellite
+```
+
+#### Manually Updating a Satellite
+
+Satellites can also be manually updated using the `update` command. The update command can be used to not only trigger a version upgrade, but also to change other parameters of the satellite, such as feature-flags or it's cache. 
+Satellites must be in a sleep state before an update can be started. You can use the `earthly satellite sleep` command to do this manually.
+Below are some examples of how you can use the `update` command.
+
+The following example updates a satellite to the latest version, respecting any pinned versions:
```
brandonSc, line 152: > respecting any pinned versions

@dchw I expected this to be the case, but after testing it I see it updates to the latest version when you have it pinned. Is that expected?

docs/cloud/satellites/managing.md
```diff
@@ -98,22 +98,87 @@ Currently selected: No
...
+Here's an example showing a satellite launched with a custom maintenance window of 4AM to 6AM:
 
 ```bash
-earthly sat rm <satellite-name>
-earthly sat launch <satellite-name>
+earthly satellite launch --maintenance-window 04:00 my-satellite
 ```
 
-The newly launched satellite will always get the latest version available.
+Note that updates will only happen during the maintenance window while the satellite is asleep. If the satellite remains in-use
```
nelsam, line 134:
```suggestion
Note that updates will only happen during the maintenance window while the satellite is asleep. If the satellite remains in use 
```

docs/cloud/satellites/managing.md
```diff
@@ -98,22 +98,87 @@ Currently selected: No
...
 ```
 
-The newly launched satellite will always get the latest version available.
+Note that updates will only happen during the maintenance window while the satellite is asleep. If the satellite remains in-use 
+for the entire duration of the maintenance window, then the update will be aborted until the next day.
+
+#### Version Pinning
+
+If you want to prevent your satellite from automatically upgrading to a new earthly version, you can pin your version using the `--version` flag.
+Note that satellites on pinned versions may still receive auto-updates during a maintenance window, however, these updaets will be important stability or security patches that will not change the version of earthly.
```
nelsam, line 140: This feels nitty, but I think a semicolon reads a little better here.
```suggestion
Note that satellites on pinned versions may still receive auto-updates during a maintenance window; however, these updaets will be important stability or security patches that will not change the version of earthly.
```

docs/cloud/satellites/managing.md
```diff
@@ -98,22 +98,87 @@ Currently selected: No
...
+
+```bash
+earthly satellite launch --version v0.6.29 my-satellite
+```
+
+#### Manually Updating a Satellite
+
+Satellites can also be manually updated using the `update` command. The update command can be used to not only trigger a version upgrade, but also to change other parameters of the satellite, such as feature-flags or it's cache.
```
nelsam, line 148: `it's` is a contraction meaning `it is`; `its` is possessive. Possessive pronouns never (to my knowledge) use apostrophes (whose, its, their, my, her, etc).
```suggestion
Satellites can also be manually updated using the `update` command. The update command can be used to not only trigger a version upgrade, but also to change other parameters of the satellite, such as feature-flags or its cache. 
```

docs/cloud/satellites/managing.md
```diff
@@ -98,22 +98,87 @@ Currently selected: No
...
+Satellites must be in a sleep state before an update can be started. You can use the `earthly satellite sleep` command to do this manually.
+Below are some examples of how you can use the `update` command.
+
+The following example updates a satellite to the latest version, respecting any pinned versions:
+
+```bash
+earthly satellite update my-satellite
+```
+
+The following command updates the satellite and clear it's cache during the process. If no update is available, this command will still clear the cache.
```
nelsam, line 158:
```suggestion
The following command updates the satellite and clear its cache during the process. If no update is available, this command will still clear the cache.
```

docs/cloud/satellites/managing.md
```diff
@@ -98,22 +98,87 @@ Currently selected: No
...
+
+Feature flags can be set during an update as well. Note feature-flags enable preview features which are considered highly expirimental.
+```bash
+earthly satellite update --feature-flag cache-pct=30 my-satellite
+```
+
+#### Satellite Revision System
+
+In more detail, satellite versions are controlled via Earthly's internal revisioning system. 
+A satellite revision includes an earthly version, plus a revision increment, where each earthly version may contain a number or ordered revisions.
```
nelsam, line 177: Another nit, but the comma here feels like it puts too much separation between "earthly version" and "revision increment", and makes it ambiguous whether or not "where each earthly version..." applies to the whole preceding fragment or just to "plus a revision increment".
```suggestion
A satellite revision includes an earthly version plus a revision increment, where each earthly version may contain a number or ordered revisions.
```

brandonSc:
Thanks for the proof-read @nelsam! I've just added your suggestions

nelsam (approved):

nelsam:
> Thanks for the proof-read @nelsam! I've just added your suggestions

No problem! :smile:

docs/cloud/satellites/managing.md
```diff
@@ -98,22 +98,87 @@ Currently selected: No
 
 ### Clearing cache
 
-To clear the cache of a satellite, run the following while a satellite is selected:
+The easiest way to clear cache on a satellite is to restart it using the `update` command with `--drop-cache` flag.
```
alexcb, line 101: I would rephrase this to mention there's two different ways (rather than easier vs harder)

maybe something along the lines of:

```
There are two ways to clear the satellite cache:

#### recreating the underlying satellite instance (often faster)

    ```bash
    earthly satellite update --drop-cache my-satellite
    ```

Note that this operation can take a while, and the satellite may also receive any available updates during the process.

#### Using the prune command (slower)

The `earthly prune` command can also be used. In many cases, however, it takes longer than using `satellite update`.
The prune command also requires the satellite to be selected before running.

    ```bash
    earthly prune -a
    ```

```

docs/cloud/satellites/managing.md
```diff
@@ -98,22 +98,87 @@ Currently selected: No
...
 
 ```bash
 earthly prune -a
 ```
 
-### Upgrading a satellite
+### Updating a satellite
 
-Currently, satellites do not have an auto-update mechanism built in. In order to get a newer version of a satellite, you need to manually remove and re-launch the satellite. Note that this operation resets the cache.
+By default satellites will automatically update as new versions become available.
```
alexcb, line 118: I would combine these two sentences:

Satellites receive version updates by default, unless they are pinned to a specific version (by using the --some-flag-here). Pinned versions will still receive minor patches, such as security updates.

docs/cloud/satellites/managing.md
```diff
@@ -98,22 +98,87 @@ Currently selected: No
...
 earthly prune -a
 ```
 
-### Upgrading a satellite
+### Updating a satellite
 
-Currently, satellites do not have an auto-update mechanism built in. In order to get a newer version of a satellite, you need to manually remove and re-launch the satellite. Note that this operation resets the cache.
+By default satellites will automatically update as new versions become available. 
+Alternativly, satellite can also be pinned to specific earthly versions, however they will still receive patches and security fixes.
+Satellite updates will only run during their specific maintenance window.
```
alexcb, line 120: This seems repetitive -- the next section header implies updates occur during maintenance windows.
```suggestion
```

brandonSc: Yeah, good point

docs/cloud/satellites/managing.md
```diff
@@ -98,22 +98,87 @@ Currently selected: No
...
+
+#### Auto-Update Maintenance Windows
+
+Maintenance windows are set between 2AM and 4AM in the timezone where the satellite is launched by default.
+If you prefer a different maintenance window, you can specify it in the launch command using the `--maintenance-window` flag.
```
alexcb, line 124: Maybe reword this to be a bit more concise:
```
Maintenance windows are set between 2AM and 4AM in the timezone where the satellite is launched by default.
The start time of the 2 hour window can be explicitly set by passing a 24-hr formatted time to the `--maintenance-window` flag of the launch command.
```

docs/cloud/satellites/managing.md
brandonSc: I like it, thanks!

docs/cloud/satellites/managing.md
```diff
@@ -98,22 +98,87 @@ Currently selected: No
...
 
 ```bash
-earthly sat rm <satellite-name>
-earthly sat launch <satellite-name>
+earthly satellite launch --maintenance-window 04:00 my-satellite
 ```
 
-The newly launched satellite will always get the latest version available.
+Note that updates will only happen during the maintenance window while the satellite is asleep. If the satellite remains in use 
+for the entire duration of the maintenance window, then the update will be aborted until the next day.
```
alexcb, line 134:
```suggestion
for the entire duration of the maintenance window, then the update will be re-attempted the next day.
```

docs/cloud/satellites/managing.md
```diff
@@ -98,22 +98,87 @@ Currently selected: No
...
+
+#### Version Pinning
+
+If you want to prevent your satellite from automatically upgrading to a new earthly version, you can pin your version using the `--version` flag.
+Note that satellites on pinned versions may still receive auto-updates during a maintenance window; however, these updates will be important stability or security patches that will not change the version of earthly.
```
alexcb, line 139:
```suggestion
Note that satellites on pinned versions may still receive auto-updates during a maintenance window; however, these updates will be limited to stability or security patches, rather than version updates.
```

docs/cloud/satellites/managing.md
```diff
@@ -98,22 +98,87 @@ Currently selected: No
...
+Updates can also specify a new pinned earthly version:
+```bash
+earthly satellite update --version v0.6.29 my-satellite
+```
+
+Feature flags can be set during an update as well. Note feature-flags enable preview features which are considered highly expirimental.
+```bash
+earthly satellite update --feature-flag cache-pct=30 my-satellite
+```
+
```
alexcb, line 172: is it possible to clear out feature flags? e.g.

```
earthly satellite update --feature-flag ""
```

alexcb: similarly, if one does an update the pinned version (and omits the `--feature-flag` flag), will the existing feature flags persist?

alexcb (commented):
looks great, just left a few (more) minor suggestions.

docs/cloud/satellites/managing.md
brandonSc: I think the ability to clear feature flags entirely might be missing. We may want to add it. 

When no feature-flags are passed-in, the existing flags are retained. When a new set of feature-flags are passed in, it will replace the entire set of existing flags. This behaviour should probably be better documented.

alexcb: maybe we should have a 

```
{% hint style='info' %}
##### Note
It's not currently possible to completely clear out the flags; you will have to destroy and recreate the satellite. 
{% endhint %}
```

in this case?

vladaionescu:
Should this be merged into `next` too?

docs/cloud/satellites/managing.md
brandonSc: Yes makes sense, and we can improve support for that later. We don't really encourage using the feature flags anyway.

vladaionescu:
In that case, it needs to be merged to next only. The next branch had the 0.7 documentation

docs/cloud/satellites/managing.md
```diff
@@ -98,22 +98,100 @@ Currently selected: No
 
 ### Clearing cache
 
-To clear the cache of a satellite, run the following while a satellite is selected:
+There are two ways to clear the cache on satellite.
+
+#### Recreating the Underlying Satellite Instance (often faster)
+
+Running the `update`  command with `--drop-cache` will relaunch the instance with an empty cache volume.
```
alexcb, line 105:
```suggestion
Running the `update` command with `--drop-cache` will relaunch the instance with an empty cache volume.
```

brandonSc: Thanks!

docs/cloud/satellites/managing.md
```diff
@@ -98,22 +98,100 @@ Currently selected: No
...
+Note that this operation can take a while, and the satellite may also receive any available updates during the process.
+
+```bash
+earthly satellite update --drop-cache my-satellte
+```
+#### Using the prune command (slower)
+
+The `earthly prune` command also works on satellites.
+It usually takes longer than running `satellite update`, however, it does not trigger a relaunch.
```
alexcb, line 114:
```suggestion
It usually takes longer than running `satellite update`; however, it does not trigger a relaunch.
```

docs/cloud/satellites/managing.md
brandonSc: We determined this is a bug on the backend that we will fix shortly. It should respect version pinning as I've explained here in the doc.

docs/cloud/satellites/managing.md
```diff
@@ -98,22 +98,100 @@ Currently selected: No
...
+```bash
+earthly satellite update --version v0.6.29 my-satellite
 ```
 
-The newly launched satellite will always get the latest version available.
+Feature flags can be set during an update as well.
+When any feature flags are passed in, the entire set of existing feature flags are replaced with the new set.
+Passing no feature flags will retain the existing flags.
+
+Note feature-flags are typically used to preview unreleased features. They are considered highly expirimental.
```
alexcb, line 176:
```suggestion
Note feature-flags are typically used to preview unreleased features. They are considered highly experimental.
```

docs/cloud/satellites/managing.md
```diff
@@ -98,22 +98,100 @@ Currently selected: No
...
-To clear the cache of a satellite, run the following while a satellite is selected:
+There are two ways to clear the cache on satellite.
+
+#### Recreating the Underlying Satellite Instance (often faster)
+
+Running the `update` command with `--drop-cache` will relaunch the instance with an empty cache volume.
+Note that this operation can take a while, and the satellite may also receive any available updates during the process.
+
+```bash
+earthly satellite update --drop-cache my-satellte
```
alexcb, line 109:
```suggestion
earthly satellite update --drop-cache my-satellite
```

alexcb (approved):

docs/cloud/satellites/managing.md
brandonSc: Wow I should've just pasted this into Grammarly or something.

brandonSc:
I double-checked in Grammarly for any more typos, and it looked good. Thanks to everyone for helping me refine this!

PR #2658, opened 2023-02-07
use gRPC to send analytics; add org and project

idodod:
org and project from the (primary) Earthfile will take precedence over the values provided from flags.
This means project will only be provided in either of the following cases:
 - A build command was invoked and org/project was set in the Earthfile
 - A `secret` command was invoked
 - A `project` command was invoked

@alexcb, @nelsam, @vladaionescu 
Please let me know if you think I'm handling the org and project correctly 

This PR references a branch of cloud-api so I'm going to leave it as a draft for now.


cmd/earthly/build_cmd.go
```diff
@@ -32,17 +45,11 @@ import (
...
-	"github.com/moby/buildkit/session/socketforward/socketprovider"
-	"github.com/moby/buildkit/session/sshforward/sshprovider"
-	"github.com/moby/buildkit/util/entitlements"
-	"github.com/pkg/errors"
-	"github.com/urfave/cli/v2"
+)
+
+var (
+	orgMu     sync.Mutex
+	projectMu sync.Mutex
```
alexcb, line 52: we try to avoid globals as much as possible; I would mimic https://github.com/earthly/earthly/blob/main/analytics/counters.go and create `analytics/project_tracker.go`

```
type ProjectTracker struct {
        org string
        project string
	mutex  sync.Mutex
}

var projectTracker ProjectTracker

func (pt *ProjectTracker) AddProject(org, project string) {
     pt.Lock()
     defer pt.Unlock()
     if pt.org == "" {
        pt.org = org
        pt.project = project
    }

...ad similar method for getting
}
```

cmd/earthly/build_cmd.go
```diff
@@ -243,7 +250,7 @@ func (app *earthlyApp) actionBuildImp(cliCtx *cli.Context, flagArgs, nonFlagArgs
 		runnerName = fmt.Sprintf("local:%s", hostname)
 	} else {
 		if app.satelliteName != "" {
-			runnerName = fmt.Sprintf("sat:%s/%s", app.orgName, app.satelliteName)
+			runnerName = fmt.Sprintf("sat:%s/%s", app.cliFlags.orgName, app.satelliteName)
```
alexcb, line 253: I'm curious about the reason for explicitly referencing the embedded struct here?

idodod: It's because I added the same field in https://github.com/earthly/earthly/pull/2658/files#diff-c5311aaeb839beacd8c0108569b5e9330f54233d2b80b16093f6e492782ca4c5R181 and it caused a collision

alexcb: ah ha! That wasn't immediately obvious to me. Moving those fields to the new `analytics.ProjectTracker` struct will prevent this collision.

cmd/earthly/build_cmd.go
```diff
@@ -558,6 +567,44 @@ func (app *earthlyApp) actionBuildImp(cliCtx *cli.Context, flagArgs, nonFlagArgs
 	return nil
 }
 
+func (app *earthlyApp) getOrgForAnalytics() string {
```
nelsam, line 570: From [`Effective Go`](https://go.dev/doc/effective_go#Getters):

> it's neither idiomatic nor necessary to put Get into the getter's name.

I also think `analyticsOrg` reads better than `orgForAnalytics`.
```suggestion
func (app *earthlyApp) analyticsOrg() string {
```

cmd/earthly/build_cmd.go
```diff
@@ -558,6 +567,44 @@ func (app *earthlyApp) actionBuildImp(cliCtx *cli.Context, flagArgs, nonFlagArgs
...
+	if app.analyticsMetadata.orgName != "" {
+		return app.analyticsMetadata.orgName
+	} else {
+		orgMu.Lock()
+		defer orgMu.Unlock()
+		if app.analyticsMetadata.orgName != "" {
+			return app.analyticsMetadata.orgName
+		}
+	}
```
nelsam, line 579: There's a return in the `if`, so the `else` is unnecessary. Also from [Effective Go](https://go.dev/doc/effective_go#if):

> when an if statement doesn't flow into the next statement—that is, the body ends in break, continue, goto, or return—the unnecessary else is omitted.

```suggestion
	}
	orgMu.Lock()
	defer orgMu.Unlock()
	if app.analyticsMetadata.orgName != "" {
		return app.analyticsMetadata.orgName
	}
```

cmd/earthly/build_cmd.go
```diff
@@ -558,6 +567,44 @@ func (app *earthlyApp) actionBuildImp(cliCtx *cli.Context, flagArgs, nonFlagArgs
...
+func (app *earthlyApp) getProjectForAnalytics() string {
+	if app.analyticsMetadata.projectName != "" {
+		return app.analyticsMetadata.projectName
+	} else {
+		projectMu.Lock()
+		defer projectMu.Unlock()
+		if app.analyticsMetadata.projectName != "" {
+			return app.analyticsMetadata.projectName
+		}
+	}
```
nelsam, line 592:
```suggestion
	}
	projectMu.Lock()
	defer projectMu.Unlock()
	if app.analyticsMetadata.projectName != "" {
		return app.analyticsMetadata.projectName
	}
```

cmd/earthly/build_cmd.go
```diff
@@ -558,6 +567,44 @@ func (app *earthlyApp) actionBuildImp(cliCtx *cli.Context, flagArgs, nonFlagArgs
...
+		orgMu.Lock()
+		defer orgMu.Unlock()
+		if app.analyticsMetadata.orgName != "" {
+			return app.analyticsMetadata.orgName
+		}
+	}
+	return app.cliFlags.orgName
+}
+
+func (app *earthlyApp) getProjectForAnalytics() string {
```
nelsam, line 583:
```suggestion
func (app *earthlyApp) analyticsProject() string {
```

cmd/earthly/build_cmd.go
```diff
@@ -558,6 +567,44 @@ func (app *earthlyApp) actionBuildImp(cliCtx *cli.Context, flagArgs, nonFlagArgs
...
+		projectMu.Lock()
+		defer projectMu.Unlock()
+		if app.analyticsMetadata.projectName != "" {
+			return app.analyticsMetadata.projectName
+		}
+	}
+	return app.cliFlags.projectName
+}
+
+func (app *earthlyApp) setOrgForAnalytics(orgName string) {
```
nelsam, line 596:
```suggestion
func (app *earthlyApp) setAnalyticsOrg(orgName string) {
```

cmd/earthly/build_cmd.go
```diff
@@ -558,6 +567,44 @@ func (app *earthlyApp) actionBuildImp(cliCtx *cli.Context, flagArgs, nonFlagArgs
...
+	orgMu.Lock()
+	app.analyticsMetadata.orgName = orgName
+	orgMu.Unlock()
+}
+
+func (app *earthlyApp) setProjectForAnalytics(projectName string) {
```
nelsam, line 602:
```suggestion
func (app *earthlyApp) setAnalyticsProject(projectName string) {
```

nelsam (commented):
In addition to the in-line comments: it feels like some (but not all) of these files used a different import organization tool than `goimports` or `gopls`. I can't say I'm a big fan of splitting the import clause up like this, mostly because `goimports` and `gopls` will usually automatically tack imports on at the end and then all the fancy organization is just lost.

What are you using for import organization? Maybe I'm just on an old version of `gopls`.

cmd/earthly/build_cmd.go
vladaionescu: I like this suggestion

idodod:
> In addition to the in-line comments: it feels like some (but not all) of these files used a different import organization tool than `goimports` or `gopls`. I can't say I'm a big fan of splitting the import clause up like this, mostly because `goimports` and `gopls` will usually automatically tack imports on at the end and then all the fancy organization is just lost.
> 
> What are you using for import organization? Maybe I'm just on an old version of `gopls`.

I set the imports rules in Goland, according to what I noticed was the convention in earthly/cloud (You can find the rules [here](https://github.com/earthly/cloud/pull/869)).

I'll revert those changes in this PR though

cmd/earthly/main.go
```diff
@@ -70,6 +70,7 @@ type earthlyApp struct {
 	commandName string
 	cliFlags
 	analyticsMetadata
+	projectTracker *analytics.ProjectTracker
```
idodod, line 73: @alexcb not sure if this is the place you had in mind for this object.
Alternatively, I can  define it as a global var in the analytics package, but it would then need to be exported, right?

alexcb: I think a global in analytics is better to be honest -- because that's the pattern we already have for analytics. And it prevents you from having to pass them everywhere.

Currently `Count` is a global function which just forwards it on to the unexported global `counts` pointer.

```
// Count increases the global count of (subsystem, key) which then gets reported when CollectAnalytics is called.
func Count(subsystem, key string) {
        counts.Count(subsystem, key)
}
```

maybe you can do something similar:

```
func AddProjectFromCommandLine(org, project string) {
    projectTracker.AddProjectFromCommandLine(org, prject)
}
```

normally for regular features, these global (call 'em singletons?) might be frowned upon, but analytics are designed to just collect data from the lifespan of the cli running, and fires them off just at the end.

nelsam: Just for some extra context on singletons being frowned upon: usually the biggest downside of the singleton pattern is that any tests against singletons are going to experience test pollution. The tests usually _must_ be run in serial (it's hard to assert on your test's effects when other tests may or may not be running) and often have to be run in a predictable order. And that often means that `go test -run TestSomething` will fail when `go test` passes, because `TestSomething` relies on the effects of a previous test.

But yeah, analytics are a case where it's mostly expected :shrug:

I think a shorter name works just as well for that function, though. e.g. `AddCLIProject(org, project string)` seems to provide roughly the same amount of information in ~half the characters.

nelsam (approved):

analytics/project_tracker_test.go
```diff
@@ -0,0 +1,182 @@
+package analytics
+
+import (
+	"testing"
+
+	. "github.com/stretchr/testify/assert"
+)
+
+type mockMutex struct {
```
alexcb, line 9: this seems overkill for testing, but it's already been written so it's fine.

cloud/analytics.go
```diff
@@ -2,53 +2,15 @@ package cloud
...
-	payload, err := json.Marshal(data)
-	if err != nil {
-		return errors.Wrap(err, "failed to marshal data")
-	}
-	opts := []requestOpt{
-		withBody(payload),
-		withHeader("Content-Type", "application/json; charset=utf-8"),
-	}
-	if c.IsLoggedIn(ctx) {
-		opts = append(opts, withAuth())
```
alexcb, line 44: Have you verified that we still track if a user is logged in or not?

analytics/project_tracker.go
```diff
@@ -0,0 +1,44 @@
...
+	pt.cliProject = project
+}
+
+func (pt *ProjectTracker) ProjectDetails() (string, string) {
+	if pt.earthfileOrg != "" && pt.earthfileProject != "" {
+		return pt.earthfileOrg, pt.earthfileProject
+	}
+	org := pt.cliOrg
+	project := pt.cliProject
+	pt.mutex.Lock()
```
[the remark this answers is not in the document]
idodod: What I had in mine here is that I can avoid lock/unlock if the two values are already populated,
and then if either one of them is not, the lock is acquired and we check again if the values are populated/empty.

I don't have a problem to move it to the top and avoid checking the values twice if you find it unnecessary.

cloud/analytics.go
idodod: yes, but I plan to do another round of testing when this PR seems close to done.

alexcb: perfect! just wanted to double check.

analytics/project_tracker.go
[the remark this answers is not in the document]
alexcb: > What I had in mine here is that I can avoid lock/unlock if the two values are already populated,

But you need a memory barrier to occur, otherwise you might have one thread whos CPU has updated the ram memory; however another thread on a completely different cpu (with a completely different L1/L2 cache) might return a cached copy of the memory without being forced to check the value stored in ram.

one could omit the lock in this case if there were a call to https://man7.org/linux/man-pages/man2/membarrier.2.html ; however, moving the Lock() call higher up would achieve this while being following standard go idioms.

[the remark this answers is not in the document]
idodod: discussed offline, moved lock to the top of the function

idodod:
This PR is ready for another review, but I will need to update earthly/cloud-api ref once it is merged

alexcb (approved):

cloud/analytics.go
idodod: so, good call!
of course I was planning on testing, and glad I caught it, but when I originally tested I used my own client and I overlooked the fact that earthly uses a client interceptor to authenticate, so I had to address that.

idodod:
@alexcb , @nelsam 
I made some changes to correctly send the auth header (if authenticated) + skip the reAuth process in the client interceptor for `SendAnalytics`.

If you don't mind taking another look at the PR. Thanks!

alexcb (commented):
I see our grpc connection uses

```
grpc_retry.WithBackoff(grpc_retry.BackoffExponential(100 * time.Millisecond))
```

Will this also apply to our analytics?

If yes, this needs to change; previously we had a very tight timeout for analytics -- if we're having any sort of outage (or delay), that shouldn't cause earthly to hang (unless they are using satellites, or cloud secrets).

alexcb (approved):

analytics/project_tracker.go
```diff
@@ -0,0 +1,43 @@
...
+	earthfileOrg     string
+	earthfileProject string
+	cliOrg           string
+	cliProject       string
+	mutex            sync.Locker
+}
+
+var projectTracker = ProjectTracker{
+	mutex: &sync.Mutex{},
+}
```
nelsam, line 15: I think it's perfectly fine to use a `sync.Locker`, but just FYI, if you used `sync.Mutex` directly instead, you wouldn't have to allocate the mutex here - you could just declare `var projectTracker ProjectTracker`. The zero value of a `sync.Mutex` is a fully working mutex lock for this express purpose.

nelsam (approved):

analytics/project_tracker.go
idodod: That's good to know, Thanks!
I used the interface so I can mock it in the tests, even though Alex already commented that was an overkill.

PR #2668, opened 2023-02-10
support for ecr-login

alexcb:
This implements support for using ecr-login with AWS_ACCESS_KEY_ID / AWS_SECRET_ACCESS_KEY credentials.

This extends the notion of storing credentials under `std/registry/<host>`, by first looking for a `cred_helper` file.

if the `cred_helper` is empty (or missing), we will look for username / password files (as we already do); however, if the value is `ecr-login`, we will then read from `std/registry/<host>/AWS_ACCESS_KEY_ID` and `std/registry/<host>/AWS_SECRET_ACCESS_KEY` secrets, and pass those to the ecr-login credential library. Note that it is not required to have `docker-credential-ecr-login` installed anywhere on your PATH.


alexcb:
Additionally, we will now display:

    Warning: PROJECT does/not-exist was defined; however it doesn't exist (or you don't have access)

rather than

    failed to lookup username/password for registry-1.docker.io.mirror.corp.earthly.dev under PROJECT does/not-exist: failed to lookup project secret does/not-exist/std/registry/registry-1.docker.io.mirror.corp.earthly.dev/username: failed to list secrets: {"message":"resource not found"}

alexcb:
A follow up PR will be submitted for the `earthy registry login ...` command.

tests/registry-command/test-ecr-project.sh
```diff
@@ -0,0 +1,61 @@
...
+
+echo "here we go in test-ecr-user.sh"
+test -n "$earthly_config" # set by earthly-entrypoint.sh
+which earthly
+
+# clear out secrets from previous test
+clearusersecrets
+
+# test dockerhub credentials do not exist
+earthly registry list | grep -v 404851345508.dkr.ecr.us-west-2.amazonaws.com
```
idodod, line 19: perhaps we could move the host into a variable?

util/llbutil/authprovider/cloudauth/authprovider.go
```diff
@@ -274,6 +283,50 @@ func (ap *authProvider) getAuthConfigForProject(ctx context.Context, org, projec
...
+	secretAccessKey string
+}
+
+// Retrieve implements the CredentialsProvider interface, and only allows for auth based on accessKeyID and secretAccessKey
+func (hc awsAccessKeyCredentials) Retrieve(context.Context) (aws.Credentials, error) {
+	return aws.Credentials{
+		AccessKeyID:     hc.accessKeyID,
+		SecretAccessKey: hc.secretAccessKey,
+	}, nil
+}
```
idodod, line 297: would it make sense to move these to `ecr_login.go`?

util/llbutil/authprovider/cloudauth/authprovider.go
```diff
@@ -274,6 +283,50 @@ func (ap *authProvider) getAuthConfigForProject(ctx context.Context, org, projec
...
+
+func (ap *authProvider) getAuthConfigForProject(ctx context.Context, org, project, host string) (*authConfig, error) {
+	pathPrefix := fmt.Sprintf("std/registry/%s/", host)
```
idodod, line 300: nit: might be nice to move `std/registry` to a const and/or even abstract the paths generation for host,  org/project, `cred_helper`
into their own functions

tests/registry-command/test-ecr-project.sh
```diff
@@ -0,0 +1,61 @@
...
+set -ex
+
+# WARNING -- RACE-CONDITION: this test is not thread-safe (since it makes use of a shared user's secrets)
+# the lock.sh and unlock.sh scripts must first be run
+
+clearusersecrets() {
+    earthly secrets ls /user/std/ | xargs -r -n 1 earthly secrets rm
+}
+
+echo "here we go in test-ecr-user.sh"
```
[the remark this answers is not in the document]
nelsam: I would prefer either `echo "here we go in $0"` or `echo "here we go"`. If we leave the file name in the string, then we are prone to misleading information in the output (e.g. if someone changes the file name or copies/pastes this section of the script), which is worse than no output.

tests/registry-command/test-ecr-user.sh
```diff
@@ -0,0 +1,56 @@
...
+set -ex
+
+# WARNING -- RACE-CONDITION: this test is not thread-safe (since it makes use of a shared user's secrets)
+# the lock.sh and unlock.sh scripts must first be run
+
+clearusersecrets() {
+    earthly secrets ls /user/std/ | xargs -r -n 1 earthly secrets rm
+}
+
+echo "here we go in test-ecr-user.sh"
```
nelsam, line 11: Same comment here about either using `$0` or just not including the name of the file.

util/llbutil/authprovider/cloudauth/authprovider.go
```diff
@@ -287,27 +340,38 @@ func (ap *authProvider) getAuthConfig(ctx context.Context, host string) (*authCo
...
 	ac, err := ap.getAuthConfigForProject(ctx, "", "", host)
-	if err == nil {
+	if err != nil {
+		if !errors.Is(err, ErrNoCredentialsFound) {
+			return nil, fmt.Errorf("failed to lookup credentials for %s: %w", host, err)
+		}
+	} else {
 		ap.authConfigCache[host] = ac
 		return ac, nil
 	}
```
nelsam, line 350: I think it's a little unclear that the `if` block flows back into the main logic, here - the `return` at the end of the `else` and in the nested `if !errors.Is` branch kind of hides that the `if err != nil` case actually merges back to the parent scope. This makes it clearer (to me, at least) how the logic flows into the main logic:

```suggestion
	if err == nil {
		ap.authConfigCache[host] = ac
		return ac, nil
	}
	if !errors.Is(err, ErrNoCredentialsFound) {
		return nil, fmt.Errorf("failed to lookup credentials for %s: %w", host, err)
	}
```

util/llbutil/authprovider/cloudauth/authprovider.go
```diff
@@ -287,27 +340,38 @@ func (ap *authProvider) getAuthConfig(ctx context.Context, host string) (*authCo
...
 		ac, err := ap.getAuthConfigForProject(ctx, op.org, op.project, host)
-		if err == nil {
+		if err != nil {
+			if !errors.Is(err, ErrNoCredentialsFound) {
+				return nil, fmt.Errorf("failed to lookup credentials for %s: %w", host, err)
+			}
+		} else {
 			ap.authConfigCache[host] = ac
 			return ac, nil
 		}
```
nelsam, line 371: Similar comment here wrt it being unclear how the `if` branch flows back to the main logic:
```suggestion
		if err == nil {
			ap.authConfigCache[host] = ac
			return ac, nil
		}
		if !errors.Is(err, ErrNoCredentialsFound) {
			return nil, fmt.Errorf("failed to lookup credentials for %s: %w", host, err)
		}
```

util/llbutil/authprovider/cloudauth/ecr_login.go
```diff
@@ -0,0 +1,70 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to extract aws ecr registry details from %s: %w", host, err)
+	}
+
+	if registry.FIPS {
+		return nil, fmt.Errorf("FIPS registry %s not supported", host)
+	}
+
+	accessKeyIDPath := pathPrefix + "AWS_ACCESS_KEY_ID"
+	secretAccessKeyPath := pathPrefix + "AWS_SECRET_ACCESS_KEY"
```
nelsam, line 24: The term `pathPrefix` makes me think we should be using `path.Join` here, to ensure that we get a `/` separator.

util/llbutil/authprovider/cloudauth/ecr_login.go
```diff
@@ -0,0 +1,70 @@
...
+
+	ap.console.VerbosePrintf("looking up %sAWS_ACCESS_KEY_ID", fullPathPrefix)
+	accessKeyID, err := ap.getProjectOrUserSecret(ctx, org, project, accessKeyIDPath)
+	if err != nil {
+		return nil, err
+	}
+	accessKeyID = strings.TrimSpace(accessKeyID)
+	if accessKeyID == "" {
+		return nil, fmt.Errorf("%sAWS_ACCESS_KEY_ID is missing (or empty), but %scred_helper was set to ecr-login", fullPathPrefix, fullPathPrefix)
```
nelsam, line 33: `%[1]s` can be used to reuse the first argument to a formatting function. But we may also want to use `path.Join` here.

nelsam (approved):
I don't really see anything that should block merging. A few minor stylistic comments.

tests/registry-command/test-ecr-user.sh
alexcb: I was supposed to remove this before submitting this PR; thanks for catching this :)

util/llbutil/authprovider/cloudauth/ecr_login.go
alexcb: > %[1]s

neat! I'll keep that in mind for next time.

I actually rewrote a lot of this path joining code to make it easier to follow

util/llbutil/authprovider/cloudauth/ecr_login.go
alexcb: we need `/` even if running under windows; but the majority of this code has been changed in the latest commit.

alexcb: edit.... looks like path.Join does that; whereas filepath.Join might be a `\\`.

nelsam: `path` always uses `/` (even on windows). `filepath` is the one that changes the separator character depending on OS. Think of it like: `path` is intended for URL paths, `filepath` is intended for local filesystem paths. In this case, `path` should be correct in all scenarios.

idodod (approved):

PR #2670, opened 2023-02-10
import testify functions explicitly

alexcb:
Rather than do a dot import, i.e.

    . "github.com/stretchr/testify/assert"

testify assertion functions will be defined in a corresponding `imports_test.go` file, to make it more obvious where these functions are coming from. This makes it easier when grepping.

Furthermore, an upcoming linter change started complaining about these imports.

### DIFF
--- a/analytics/analytics_test.go
+++ b/analytics/analytics_test.go
@@ -5,8 +5,6 @@ package analytics
 
 import (
 	"testing"
-
-	. "github.com/stretchr/testify/assert"
 )
 
 // TestGetRepoHash tests the git repo hashing never changes
--- a/analytics/imports_test.go
+++ b/analytics/imports_test.go
@@ -1,0 +1,14 @@
+package analytics
+
+import (
+	"github.com/stretchr/testify/assert"
+)
+
+var (
+	NoError = assert.NoError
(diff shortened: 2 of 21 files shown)


alexcb:
> So I prefer equal, noError, etc. over the exported variants.

Unfortunately this won't work for functions like `True`, because `true` is already reserved.

> Also, this may be worthy of an entry in CONTRIBUTING.md or something.

#2668 causes some go.mod updates which updates the linter, which will now display errors if a dot import is used; I suspect that will be obvious enough if someone were to contribute a PR that made use of a dot import.

nelsam:
> > So I prefer equal, noError, etc. over the exported variants.
> 
> Unfortunately this won't work for functions like `True`, because `true` is already reserved.
Good point. I mean, we could always alias e.g. `isTrue = require.True`, but I see how that could get annoying over time.

> > Also, this may be worthy of an entry in CONTRIBUTING.md or something.
> 
> #2668 causes some go.mod updates which updates the linter, which will now display errors if a dot import is used; I suspect that will be obvious enough if someone were to contribute a PR that made use of a dot import.
Yeah, fair.

nelsam (approved):

PR #2672, opened 2023-02-10
earthly registry setup for ecr-login support

alexcb:
This implements a new login command to support ecr-login:

    export AWS_ACCESS_KEY_ID=...
    export AWS_SECRET_ACCESS_KEY=...
    earthly registry login --cred-helper ecr-login <aws_account_id>.dkr.ecr.<region>.amazonaws.com

and gcloud:

    export GCP_KEY=...
    earthly registry login --cred-helper gcloud <gcp-registry-host>


alexcb:
corresponding docs in #2680

vladaionescu (commented):
I only have one important remark regarding org and project flags, in the documentation PR, which may be relevant for this PR: https://github.com/earthly/earthly/pull/2680

cmd/earthly/registry_cmds.go
```diff
@@ -55,50 +63,43 @@ func (app *earthlyApp) registryCmds() []*cli.Command {
...
 					Required:    false,
-					Destination: &app.orgName,
+					Destination: &app.awsSecretAccessKey,
 				},
 				&cli.StringFlag{
-					Name:        "project",
-					Usage:       "The organization project in which to store registry credentials, if empty credentials will be stored under the user's secret storage.",
+					Name:        "gcp-key",
+					EnvVars:     []string{"GCP_KEY"},
+					Usage:       "GCP key to use for artifact or container registry.",
```
idodod, line 83: perhaps it would be good to clarify the key is a service account json or maybe even rename it.
This link might offer some good options - https://cloud.google.com/iam/docs/creating-managing-service-account-keys#creating

cmd/earthly/registry_cmds.go
```diff
@@ -152,6 +153,53 @@ func (app *earthlyApp) actionRegistryLogin(cliCtx *cli.Context) error {
 		return fmt.Errorf("hosts is malformed")
 	}
 
+	switch app.registryCredHelper {
+	case "", "none":
+		return app.actionRegistrySetupUsernamePassword(cliCtx, path, cloudClient, host)
+	case "ecr-login":
+		return app.actionRegistrySetupECRLogin(cliCtx, path, cloudClient, host)
+	case "gcloud":
```
idodod, line 161: might be nice to switch these to an enum and reuse them in the `actionRegistrySetupX` functions, and the other switch cases scenarios

cmd/earthly/registry_cmds.go
```diff
@@ -152,6 +153,53 @@ func (app *earthlyApp) actionRegistryLogin(cliCtx *cli.Context) error {
...
+}
+
+func (app *earthlyApp) actionRegistrySetupECRLogin(cliCtx *cli.Context, path string, cloudClient *cloud.Client, host string) error {
+	if app.awsAccessKeyID == "" {
+		return fmt.Errorf("--aws-access-key-id is missing (or empty)")
+	}
+	if app.awsSecretAccessKey == "" {
+		return fmt.Errorf("--aws-secret-access-key is missing (or empty)")
+	}
+	err := cloudClient.SetSecret(cliCtx.Context, path+host+"/cred_helper", []byte("ecr-login"))
```
idodod, line 175:
```suggestion
	err := cloudClient.SetSecret(cliCtx.Context, path.Join(path, host, "cred_helper"), []byte("ecr-login"))
```

tests/registry-command/test-gcp.sh
```diff
@@ -1,14 +1,14 @@
...
-function finish {
+finish() {
   status="$?"
   ./unlock.sh
   if [ "$status" = "0" ]; then
-    echo "test-gcp passed"
+    echo "test-ecr passed"
   else
-    echo "test-gcp failed with $status"
+    echo "test-ecr failed with $status"
```
idodod, line 11: copied by mistake?

idodod (approved):
LGTM
I think there is only the test-gcp.sh change that is actually a "blocker"

cmd/earthly/registry_cmds.go
```diff
@@ -55,50 +63,43 @@ func (app *earthlyApp) registryCmds() []*cli.Command {
...
+					Usage:       "AWS secret access key to use for elastic-container-registry.",
 					Required:    false,
-					Destination: &app.orgName,
+					Destination: &app.awsSecretAccessKey,
 				},
 				&cli.StringFlag{
-					Name:        "project",
-					Usage:       "The organization project in which to store registry credentials, if empty credentials will be stored under the user's secret storage.",
+					Name:        "gcp-key",
+					EnvVars:     []string{"GCP_KEY"},
```
idodod, line 82: Something that came to mind after reviewing the documentation - 
I wonder if we should prefix the env var name with `EARTLY_CRED_HELPER` in case a `GCP_KEY` might be used in a different context in the future? This also regards the AWS env vars

tests/registry-command/test-gcp.sh
alexcb: good catch, I've changed these to `$0`.

cmd/earthly/registry_cmds.go
alexcb: @vladaionescu what are your thoughts on us reading from the standard `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` env variables?

alexcb: so for the case of gcp, as per https://github.com/GoogleCloudPlatform/docker-credential-gcr#gcr-credentials
we can use the standard `GOOGLE_APPLICATION_CREDENTIALS`?

vladaionescu: Sounds good to me!

PR #2674, opened 2023-02-13
Support for gcp-login

idodod:
This implements support for using gcp-login with a gcloud service account json config.
Based on steps described [here](https://cloud.google.com/artifact-registry/docs/docker/authentication#token).

Similar to ecr-login, this extends the notion of storing credentials under std/registry/<host>, by first looking for a cred_helper file, and in this PR, adding another option - `gcp-login`, which can be used for either  Google Artifact Registry or Google Container Registry.

If the value is `gcp-login`, we will read from std/registry/\<host\>/GCP_KEY secret, and pass those to the gcp-login credential library. Note that it is not required to have `gcloud` or `docker-credential-gcr` installed anywhere on your PATH.

~Note: waiting for https://github.com/earthly/earthly/pull/2668 to be merged first.~ - Done


util/llbutil/authprovider/cloudauth/gcp_login.go
```diff
@@ -0,0 +1,47 @@
...
+		return nil, err
+	}
+	keyJSON = strings.TrimSpace(keyJSON)
+	if keyJSON == "" {
+		return nil, fmt.Errorf("%sGCP_KEY is missing (or empty), but %scred_helper was set to gcp-login", fullPathPrefix, fullPathPrefix)
+	}
+
+	ap.console.VerbosePrintf("creating new a new JWT config for %s", host)
+
+	jwtCFG, err := google.JWTConfigFromJSON([]byte(keyJSON), OauthScope)
```
alexcb, line 31: missing err handling?

tests/registry-command/test-gcp.sh
```diff
@@ -0,0 +1,28 @@
...
+trap finish EXIT
+
+# Test Google artifact registry
+export ARTIFACT_SERVER="us-west1-docker.pkg.dev"
+export ARTIFACT_FULL_ADDRESS="$ARTIFACT_SERVER/ci-cd-302220"
+export IMAGE="integration-test/test"
+./test-gcp-user.sh
+./test-gcp-project.sh
+
+# Test Google container registry
```
alexcb, line 23: awesome to see it follows the same test-flow :tada:

util/llbutil/authprovider/cloudauth/gcp_login.go
idodod: good catch, looks like I removed it by mistake.
That makes me wonder though - should I add integration tests to for the err path (secret is missing, json in the secret is invalid, etc)?

util/llbutil/authprovider/cloudauth/ecr_login.go
```diff
@@ -10,6 +10,10 @@ import (
 	"github.com/docker/cli/cli/config/types"
 )
 
+const (
+	ECRCredHelper = "ecr-login"
```
idodod, line 14: Using a constant but I can undo this change (made it to match with `GCPCredHelper`)

tests/registry-command/test-ecr.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/usr/bin/env sh
```
alexcb: For some reason I thought this trap required bash, but if that's not the case, then there's no need for `env` either. (I only used `env` since `bash` as a matter of habit since it doesn't always get installed in the same location)
```suggestion
#!/bin/sh
```

idodod: sorry, that was just a mistake after I rebased, I reverted this change for now.

util/llbutil/authprovider/cloudauth/ecr_login.go
alexcb: nice change!

alexcb (approved):

nelsam (approved):
It all seems reasonable to me. I wanted to wait until Alex's PR was merged to avoid reviewing his changes amidst yours, but yeah, lgtm.

PR #2675, opened 2023-02-13
new cloud.GetUserOrProjectSecret for v2 secrets

alexcb:
Creates a new function to fetch v2 user or project secrets

### DIFF
--- a/cloud/secret_new.go
+++ b/cloud/secret_new.go
@@ -8,7 +8,6 @@ import (
 	"time"
 
 	secretsapi "github.com/earthly/cloud-api/secrets"
-	"github.com/moby/buildkit/session/secrets"
 	"github.com/pkg/errors"
 )
 
@@ -66,6 +65,7 @@ func (c *client) ListSecrets(ctx context.Context, path string) ([]*Secret, error
 	return secrets, nil
 }
 
+// GetProjectSecret gets a secret from a project secret store
 func (c *client) GetProjectSecret(ctx context.Context, org, project, secretName string) (*Secret, error) {
 	if org == "" {
 		return nil, fmt.Errorf("GetProjectSecret called with empty org")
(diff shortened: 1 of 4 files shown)


cloud/secret_new.go
```diff
@@ -86,6 +85,17 @@ func (c *client) GetUserSecret(ctx context.Context, secretName string) (*Secret,
 	return c.getSecretV2(ctx, fmt.Sprintf("/user/%s", secretName))
 }
 
+func (c *client) GetUserOrProjectSecret(ctx context.Context, path string) (*Secret, error) {
```
idodod, line 88: Mind adding a comment?
Would `GetSecret` be a fitting name too?

cmd/earthly/secret_cmds.go
```diff
@@ -168,16 +169,15 @@ func (app *earthlyApp) actionSecretsGetV2(cliCtx *cli.Context) error {
 		return err
 	}
 
-	secrets, err := cloudClient.ListSecrets(cliCtx.Context, path)
+	secret, err := cloudClient.GetUserOrProjectSecret(cliCtx.Context, path)
 	if err != nil {
+		if errors.Is(err, cloud.ErrNotFound) {
+			return errors.New("no secret found for that path")
```
idodod, line 175: should we mention the path that was used in the error?

idodod (approved):
LGTM
made some minor comments.

cloud/secret_new.go
alexcb: > Would GetSecret be a fitting name too?

I want to make it clear that this is for the new project-based secrets and not the original secrets implementation.

Maybe when we fully remove support for the old secrets, we could then simplify these names.

cmd/earthly/secret_cmds.go
alexcb: seems a bit overkill, malformed paths should generally be caught higher up, e.g. parts will append a `/` prefix when none are given by the user.

nelsam (approved):
lgtm

cloud/secret_new.go
```diff
@@ -66,6 +65,7 @@ func (c *client) ListSecrets(ctx context.Context, path string) ([]*Secret, error
 	return secrets, nil
 }
 
+// GetProjectSecret gets a secret from a project secret store
```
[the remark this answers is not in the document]
alexcb: Good idea; however, I need to stay focused on the ecr-login work at the moment. Maybe we can do some client cleanup down the road after v0.7 is out.

PR #2676, opened 2023-02-13
logstreamer.LogStreamer: avoid deadlock and data race scenarios

nelsam:
I _finally_ got a consistent failure scenario for the deadlock - but I had to write a `WriteAsync` method on the `LogStreamer` for it. Technically, we could unexport the `WriteAsync` method, but that would require the `logstreamer_chaos_test.go` file to be in the `logstreamer` package (instead of the `logstreamer_test` package), and that always makes me feel a little weird. Providing a test case that users of the package cannot exercise (because they don't have access to the same methods) feels like a hack.

Anyway, this both tests for and solves the following scenarios:

## Deadlock

1. When a `LogStreamer` is heavily congested, calls to `Write` will gradually start to back up.
2. When those calls to `Write` have been backing up for long enough, there may be `> bufferSize` goroutines blocked waiting on space in the buffer for deltas.
3. When the `LogStreamer` receives a transient error from the cloud client, it will re-allocate the deltas channel and attempt to decongest it in a separate goroutine.
4. If the scheduler schedules the decongestion goroutine often enough, the `default:` case will be hit and the decongestion goroutine will exit.
5. Some calls to `Write` may still be blocked on the old channel.

## Data Race

1. When a `LogStreamer` is constructing its buffer, `Write` may still be called, queuing up deltas to add to the channel once it is done allocating.
2. The `LogStreamer` releases its lock after constructing its buffer, but before queuing up the initial manifest on that buffer.
3. The scheduler schedules one  of the `Write` goroutines.
4. The delta from the `Write` goroutine is added to the buffer before the initial manifest is.


nelsam:
I do think that it's worth exploring non-channel goroutine-safe queues for the underlying data structure here, but that doesn't seem critical for fixing these issues. I'd like to see that the issues @r-LaForge mentioned are solved before it's changed _too_ drastically.

cloud/logstream.go
```diff
@@ -40,30 +43,28 @@ func (c *client) StreamLogs(ctx context.Context, buildID string, deltasCh chan [
...
+			if errors.Is(err, io.EOF) {
+				msg := &logstream.StreamLogRequest{
 					BuildId: buildID,
-					Deltas:  deltas,
-				})
-				if err != nil {
-					return errors.Wrap(err, "failed to send log delta")
+					Eof:     true,
+				}
+				if err := streamClient.Send(msg); err != nil {
```
vladaionescu, line 52: Nit: I generally prefer not putting things that have side effects in `if` expressions. It helps with readability. There are pros and cons regarding the scope of `err` created here. Not a very strong opinion.

cloud/logstream.go
```diff
@@ -2,32 +2,35 @@ package cloud
...
+
+func (c *client) StreamLogs(ctx context.Context, buildID string, deltas Deltas) error {
 	streamClient, err := c.logstream.StreamLogs(c.withAuth(ctx), grpc_retry.Disable())
 	if err != nil {
 		return errors.Wrap(err, "failed to create log stream client")
 	}
 	eg, ctx := errgroup.WithContext(ctx)
-	var mu sync.Mutex
-	finished := false
+	var finished int32
```
vladaionescu, line 25: Is the change to atomics necessary as part of the bug fix?

It feels more readable with just a mutex.

logbus/logstreamer/logstreamer.go
```diff
@@ -17,29 +20,149 @@ import (
...
+
+func (d *deltasIter) close() {
+	d.mu.Lock()
+	defer d.mu.Unlock()
+	decongest(d.ch)
+	close(d.ch)
+	d.ch = nil
+}
+
+func (d *deltasIter) Next(ctx context.Context) ([]*logstream.Delta, error) {
```
vladaionescu, line 113: I like this API

logbus/logstreamer/logstreamer.go
```diff
@@ -17,29 +20,149 @@ import (
 	"google.golang.org/protobuf/encoding/protojson"
 )
 
+const (
+	DefaultChSize = 10240
```
vladaionescu, line 24: Nit: Make private?

vladaionescu (approved):

cloud/logstream.go
nelsam: Yeah, my counter-point is specifically about the `err` variable and go's dumb typed nils. sticking to `if err := ...; err != nil` helps avoid the typed vs untyped nil problems. But I don't have strong opinions - as long as I can still use `:=` instead of reassignment.

cloud/logstream.go
nelsam: Interesting, I find the atomics more readable. I mean, it's mildly annoying that the variable isn't a boolean in this section; but later in the file, a single `atomic.Store...` to my eyes reads cleaner than `mu.Lock(); finished = true; mu.Unlock()`. Mostly because it's a fairly uncomplicated part of the logic, and taking up three lines for it looks weird to me.

I'll make it an `atomic.Bool` instead and see if that's a reasonable compromise?

logbus/logstreamer/logstreamer.go
nelsam: I usually export defaults like this so that anyone who reads the `godoc` can see what the default should be. I could alternatively put it in a comment, but then we're in the `comments are lies waiting to happen` territory - it feels safer to just export the value. It's a constant, so there's no risk of anything external modifying it.

cloud/logstream.go
nelsam: Just my luck, `atomic.Bool` is new as of go 1.19 :sweat_smile: 

Hey @alexcb how do you feel about bumping our go version? AFAIK the `ast` submodule is already at 1.19.

alexcb: seems fine to me, especially if we already have a submodule using it.

cloud/logstream.go
```diff
@@ -2,32 +2,35 @@ package cloud
...
 		for {
 			resp, err := streamClient.Recv()
 			if err != nil {
 				return errors.Wrap(err, "failed to read log stream response")
 			}
 			if resp.GetEofAck() {
-				mu.Lock()
-				defer mu.Unlock()
-				if !finished {
+				if finished.Load() {
```
r-LaForge, line 33: Shouldn't this be `!finished.Load()`?
The previous code was checking if that finished was false.
I'd expect finished to be set to true here after we sent an EOF and the server responded with an ack.

cloud/logstream.go
nelsam: @vladaionescu this is now an `atomic.Bool`. Let me know what you think. I don't particularly mind going back to a mutex, it just is perhaps against my nature to use a whole mutex lock for a single line :sweat_smile: especially when the logic around the `finished` variable is mostly uninteresting.

cloud/logstream.go
nelsam: :nod: adding test, fixing; thanks!

logbus/logstreamer/helheim_test.go
```diff
@@ -0,0 +1,144 @@
+// Code generated by git.sr.ht/~nelsam/hel/v4. DO NOT EDIT.
```
r-LaForge, line 1: There are just mocks in here...Should we add `mock` to the test name?

nelsam: We can. So far, we've just [used the default name](https://github.com/earthly/earthly/blob/main/ast/helheim_test.go), but `hel` does have a flag to override that.

`helheim_mocks_test.go`?

logbus/logstreamer/logstreamer.go
```diff
@@ -17,29 +20,149 @@ import (
...
+
+func (d *deltasIter) Next(ctx context.Context) ([]*logstream.Delta, error) {
+	select {
+	case <-ctx.Done():
+		return nil, errors.Wrap(ctx.Err(), "logstreamer: context closed while waiting on next delta")
+	case delta, ok := <-d.deltas():
+		if !ok {
+			return nil, errors.Wrap(io.EOF, "logstreamer: channel closed")
+		}
+		return delta, nil
```
r-LaForge, line 121: **very** tiny nit - plural `deltas` (feel free to ignore and mark as resolved).
```suggestion
	case deltas, ok := <-d.deltas():
		if !ok {
			return nil, errors.Wrap(io.EOF, "logstreamer: channel closed")
		}
		return deltas, nil
```

logbus/logstreamer/logstreamer.go
```diff
@@ -17,29 +20,149 @@ import (
...
+const (
+	DefaultChSize = 10240
+	drainTimeout  = 60 * time.Second
+)
+
+type CloudClient interface {
+	StreamLogs(ctx context.Context, buildID string, deltas cloud.Deltas) error
+}
+
+func decongest(ch <-chan []*logstream.Delta) {
```
r-LaForge, line 32: nit:
Since `decongest` and `deltasIter` are internal implementation details - does it make sense to move this code below the logstreamer?

logbus/logstreamer/logstreamer.go
```diff
@@ -124,40 +218,66 @@ func (ls *LogStreamer) tryStream(ctx context.Context) (bool, error) {
...
-// Write writes the given delta to the log streamer.
+// WriteAsync queues delta up for writing and returns immediately. The returned
+// channel will be closed when delta has been passed off to the buffer.
+//
+// This is different than Write in two important ways:
+//
+// 1. It will never block.
+//
+// 2. If the buffer is full, it cannot guarantee order the order that delta will
+// be added to the buffer in the order in which WriteAsync was called.
```
r-LaForge, line 229: nit: This comment reads a little oddly
```suggestion
// 2. If the buffer is full, it cannot guarantee the order that delta will
// be added to the buffer matches the order in which WriteAsync was called.
```

r-LaForge (approved):

logbus/logstreamer/helheim_test.go
r-LaForge: Yeah, I like `helheim_mocks_test.go`.

I don't mind helheim being in the name - I like the word `mock(s)` in the name for the same reason. 
While playful, and I like the nod to Norse mythology, it may not be as obvious as `mock(s)` to one unfamiliar with it.

logbus/logstreamer/logstreamer.go
nelsam: That's fair. I like to organize my code such that a person could feasibly read from top to bottom - I find it a little jarring to jump back and forth between definitions too often. So I put these at the top so that a reader would have already understood what they do by the time they get to the place where they are used (in the LogStreamer).

However, that's just my own personal style, and I'm not particularly married to it. Keeping exported code above unexported code is also a valid style, and I'd rather be consistent with the project than stick to my own personal choices.

logbus/logstreamer/logstreamer.go
```diff
@@ -124,40 +218,66 @@ func (ls *LogStreamer) tryStream(ctx context.Context) (bool, error) {
...
 
-// Write writes the given delta to the log streamer.
+// WriteAsync queues delta up for writing and returns immediately. The returned
+// channel will be closed when delta has been passed off to the buffer.
+//
+// This is different than Write in two important ways:
+//
+// 1. It will never block.
+//
+// 2. If the buffer is full, it cannot guarantee the order that delta will
```
alexcb, line 228: do we have a use case for WriteAsync?

The out of order writes here seem to go against what we'll be using this for.

I do see that we have a regular `Write`, which doesn't have this limitation, but i'm worried someone might use `WriteAsync` one day without reading this comment.

alexcb (approved):

cloud/logstream.go
vladaionescu: Sounds good to me!

logbus/logstreamer/logstreamer.go
vladaionescu: Sure.. then make the other one public too? We do add comments for everything exported.

logbus/logstreamer/logstreamer.go
nelsam: Yeah, I thought about that. The main thing is that I could not test it without `WriteAsync`, and I hate putting tests in the main (non-`_test`) package. I was wondering what others' thoughts were about that.

Quoting the text above (in the description of this PR):
> Technically, we could unexport the WriteAsync method, but that would require the logstreamer_chaos_test.go file to be in the logstreamer package (instead of the logstreamer_test package), and that always makes me feel a little weird. Providing a test case that users of the package cannot exercise (because they don't have access to the same methods) feels like a hack.

logbus/logstreamer/logstreamer.go
nelsam: Exported, commented. I also changed the name a bit to avoid referencing "channel" - since one of the main reasons for changing it to use an iterator interface is so that we can change the underlying channel to a different type if we want.

PR #2677, opened 2023-02-14
cloud.Client should just be a struct

alexcb:
As per https://github.com/golang/go/wiki/CodeReviewComments#interfaces an interface should be defined wherever it is used, and not in the cloud package.

If an interface is needed by a specific caller of the client, then it's best to create a much smaller interface tied to the use of the client, which will make it more clear which methods must bet implemented for that specific portion of the code.

### DIFF
--- a/analytics/analytics.go
+++ b/analytics/analytics.go
@@ -190,7 +190,7 @@ type Meta struct {
 }
 
 // CollectAnalytics sends analytics to api.earthly.dev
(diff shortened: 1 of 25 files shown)


idodod (approved):

nelsam (approved):
Some of the areas that `*cloud.Client` are used are good candidates for a local interface type. But it's reasonable to change those as we go. We don't need to fix everything in one commit.

PR #2680, opened 2023-02-14
document earthly registry setup commands

alexcb:


docs/earthly-command/earthly-command.md
```diff
@@ -646,6 +646,148 @@ List secrets the current account has access to.
 
 Removes a secret from the secrets store.
 
+## earthly registry
+
+Contains sub-commands for managing registry access in cloud-based secrets.
+
+#### earthly registry setup
```
vladaionescu, line 653: Did you mean to use 3 levels here?

```suggestion
### earthly registry setup
```

docs/earthly-command/earthly-command.md
```diff
@@ -646,6 +646,148 @@ List secrets the current account has access to.
...
+
+* ```
+  earthly registry setup [--org <org> --project <project>] [--cred-helper <none|ecr-login|gcloud>] ...
+  ```
+
+###### username/password based registry (`--cred-helper=none`)
+
+* ```
+  earthly registry setup --username <username> --password <password> [<host>]
+  earthly registry setup --org <org> --project <project> --username <username> --password <password> [<host>]
```
vladaionescu, line 665: Many of our other commands have org and project flags one level up, so that they can be shared between the various sub-commands.

So for example

```
earthly registry --org <org> --project <project> setup --username <username> --password <password> [<host>]
```

Perhaps we should do the same, for consistency.

alexcb: good idea.

vladaionescu: I'll change other docs / FAQ docs on my end too.

docs/earthly-command/earthly-command.md
alexcb: good catch, I started adding two levels per subheading :facepalm:

docs/earthly-command/earthly-command.md
```diff
@@ -646,6 +646,130 @@ List secrets the current account has access to.
...
+
+##### `--aws-secret-access-key <secret>`
+
+The AWS secret access key to use when requesting a registry token, only applicable when `--cred-helper=ecr-login`.
+
+Also available as an env var setting: `AWS_SECRET_ACCESS_KEY=<secret>`.
+
+##### `--gcp-key <key>`
+
+The GCP key to use when requesting a registry token, only applicable when `--cred-helper=gcloud`.
```
idodod, line 746: similarly to my comment in the implementation PR, perhaps it would be good to expand on what this key is, I could be wrong but I'm concerned it might not be as clear as the AWS access key and secret.

idodod, line 746: Should we elaborate on the method of authentication, since Google offers using the key directly for auth or alternatively to generate short lived token?

alexcb: I'm not sure;  if the key was used directly, would a credential helper be required? To me it implies some sort of token is being fetched.

How would you word this to make it clear (while remaining concise)?

alexcb: I did rename it to reflect the changed arg, maybe that's enough?

vladaionescu (approved):

docs/earthly-command/earthly-command.md
idodod: Yeah, you make a very good point. The other authentication method I was referring to can be set up by going through the username/password flow, so I think this is good, thanks!

idodod (approved):

PR #2682, opened 2023-02-15
cloud: make *cloud.Client unit testable

nelsam:
In order to actually unit test a *cloud.Client, we needed some way to pass in a mock connection. To facilitate that, there is a new constructor (NewClientOpts) which uses functional options to control the client, including optionally passing in a (mockable) grpc client in instead of passing in an address and relying on the constructor to dial the address.

The nature of functional options means that if/when we need to unit test HTTP requests, we can also add a mock HTTP client option without making much of an impact on the rest of the cloud.Client.  The same goes for any other tests we might need to add. Then the unit tests themselves can simply pass the options that they care about, while relying on defaults for everything else.

Some particular things I'd like feedback about:
- There's a note in here about the somewhat cyclic dependency between `cloud.Client` and `grpc.Conn` - this feels like a dependency born out of the `cloud.Client` doing a little too much. I didn't want to spend too much time on that rabbit hole though. I made a very quick attempt at moving the grpc middleware methods off of the `cloud.Client`, but they depend on the auth methods, so I decided that I've done enough damage for now.
    - What are everyone's thoughts about splitting this up? I think we could move a lot of these options off to `Authenticator` and `Interceptor` types and just pass those in to the `cloud.Client`.
- I made every parameter a functional option. Usually, I would ensure that parameters that are required for basic functionality are required - but in the case of the `cloud.Client`, you can feasibly only give it an http address or only give it a grpc address (or client) and it will be able to work. Typically you would want both, but the client actually works just fine with only one of the two set.
    - Is this okay? do we want to force callers to always provide both an HTTP and a gRPC connection?


idodod (approved):
I think you did a really good job without going too deep into the rabbit hole (I bet resisting it was the biggest challenge)

I agree that it would be nice to break the client into multiple components.

One alternative though (and perhaps you considered and decided against it), is to pass 
the different gRPC clients (pipelines, compute, logstream) to the Client struct instead of having to provide the client connection.  I think that for the purpose of testing logstream it would allow us making less changes in the Client, until we decide how to split it into multiple components. 

Regarding mocking, my suggestion would be to use to a more familiar library (unless there is some critical functionality that is not provided elsewhere), because I believe it would be easier to maintain in the future, especially by a member of the community. We use https://github.com/vektra/mockery in other projects, and it supports generating in the same directory and/or package as the interface (which I noticed was one of your motivations for https://git.sr.ht/~nelsam/hel)

nelsam:
> I think you did a really good job without going too deep into the rabbit hole (I bet resisting it was the biggest challenge)
> 
> I agree that it would be nice to break the client into multiple components.
> 
> One alternative though (and perhaps you considered and decided against it), is to pass the different gRPC clients (pipelines, compute, logstream) to the Client struct instead of having to provide the client connection. I think that for the purpose of testing logstream it would allow us making less changes in the Client, until we decide how to split it into multiple components.

Yeah, I thought about it. I think that's probably a good change to make _some day_ (in part because it makes mocking cleaner, but also because these services may not always exist on the same address), but I was trying to tightly constrain my changes to this package for the time being. The cyclic dependency between `*cloud.Client` and `*grpc.Conn` in the constructor makes that difficult.

> Regarding mocking, my suggestion would be to use to a more familiar library (unless there is some critical functionality that is not provided elsewhere), because I believe it would be easier to maintain in the future, especially by a member of the community. We use https://github.com/vektra/mockery in other projects, and it supports generating in the same directory and/or package as the interface (which I noticed was one of your motivations for https://git.sr.ht/~nelsam/hel)

Well, one of my motivations with hel was to _enforce_ interfaces being defined where they were used (instead of where they're implemented), not just to _allow_ it :smirk: 

A motivation that the readme currently neglects to mention: My team (loggregator) was hand-writing mocks using channels, the same way that `hel` generates them, mostly due to data races we encountered with other generated mocks. Being a log streaming service, a lot of our production code had to run concurrently with our tests, and if you're not super careful, most mocks generated by other tools make it easy to trigger data races. Channels solved that for our team.

`hel` is just a tool I built to generate mocks the way we were already hand-writing them. I've been using it to generate mocks for about 7 years now in all of my projects (both personal and professional). It's not _super_ heavily used, but the mocks it generates are simple enough that I don't think that's a big deal.

brandonSc:
Looks great, thanks for cleaning this up! I was hoping we'd start writing some unit tests for the cloud client soon 😄  

FYI we've been hoping to convert all of the HTTP endpoints to gRPC soon. I may have contributed to it being in the messy state it is, as I thought it would be "temporary". We haven't converted them as fast as I thought we would, but I think we'll get there soon, especially now that we have the basic CI working.

brandonSc (approved):

alexcb (approved):
very nice!

PR #2693, opened 2023-02-20
Logstream orchestrator + wait for logs to be done before existing CLI

r-LaForge:
- Separate retry / orchestration logic into an `Orchestrator`
- Remove the retry logic from `Logstreamer`
- Remove reset / drain semantics and rely on just closing the previous delaIter / Logstreamer.
- deltaIter is now the the object being written to (not Logstreamer)


util/deltautil/apply.go
```diff
@@ -145,3 +132,27 @@ func WithDeltaManifest(m *pb.RunManifest, d *pb.Delta) (*pb.RunManifest, error)
 	}
 	return ret, nil
 }
+
+func ensureTargetExists(r *pb.RunManifest, targetID string) *pb.TargetManifest {
```
r-LaForge, line 136: I ran into these issues while testing. 
I'm not sure why these happen yet, but adding this fixed it.
```
	* formatter: 6 errors occurred:
	* failed to handle delta manifest: target dd01c9fd-f9aa-4156-af62-31d7f8b56c55 not found in manifest
	* failed to handle delta manifest: target dd01c9fd-f9aa-4156-af62-31d7f8b56c55 not found in manifest
	* failed to handle delta manifest: target dd01c9fd-f9aa-4156-af62-31d7f8b56c55 not found in manifest
	* failed to handle delta manifest: target dd01c9fd-f9aa-4156-af62-31d7f8b56c55 not found in manifest
	* failed to handle delta manifest: target dd01c9fd-f9aa-4156-af62-31d7f8b56c55 not found in manifest
	* failed to handle delta manifest: target dd01c9fd-f9aa-4156-af62-31d7f8b56c55 not found in manifest
```
This case isn't hit every time.  I think we'll want to find out why this case is hit, but I also think t his is an ok patch until we figure out why.

logbus/logstreamer/logstreamer.go
```diff
@@ -41,88 +30,40 @@ type LogBus interface {
...
+// This also returns a bool indicating whether the error that occurred should be considered retryable
+func (ls *LogStreamer) Stream(ctx context.Context) (bool, error) {
 	ctxTry, cancelTry := context.WithCancel(ctx)
 	defer cancelTry()
 	ls.mu.Lock()
 	if ls.cancelled {
-		// TODO (vladaionescu): It would be nice if on cancellation we could
-		// 						still go through the entire retry loop.
-		//						This would require that we close ls.ch on each
-		//						retry somehow safely.
```
r-LaForge, line 117: I'm pretty sure the new wait introduced from the iterator `Done` handles this.
We also seem to stream double the formatted logs to the iterator than we need.

logbus/logstreamer/logstreamer_chaos_test.go
```diff
@@ -1,231 +0,0 @@
-//go:build chaos
```
r-LaForge, line 1: With the introduction of the orchestrator, the cases these tests were exercising have changed / no longer exist.
It's possible the changes introduced different edge-cases and I'd like to sync with @nelsam

cmd/earthly/flags.go
```diff
@@ -188,6 +188,13 @@ func (app *earthlyApp) rootFlags() []cli.Flag {
 			Destination: &app.logstreamDebugManifestFile,
 			Hidden:      true, // Internal.
 		},
+		&cli.StringFlag{
+			Name:        "logstream-address",
```
r-LaForge, line 192: I've been using this flag to setup an integration test between the CLI and cloud.
It's been instrumental in diagnosing and fixing issues.

cmd/earthly/build_cmd.go
```diff
@@ -538,10 +540,29 @@ func (app *earthlyApp) actionBuildImp(cliCtx *cli.Context, flagArgs, nonFlagArgs
 	// Kick off logstream upload only when we've passed the necessary information to logbusSetup.
 	// This information is passed back right at the beginning of the build within earthfile2llb.
 	go func() {
+		beforeSelect := time.Now()
 		select {
 		case <-cliCtx.Context.Done():
+			if app.verbose {
+				now := time.Now()
+				app.console.Warnf(
```
vladaionescu, line 548: Should this be on for verbose output only?

r-LaForge: I think so

vladaionescu: Oh wait.. I now see the `app.verbose`. We have a `console.Verbosef` BTW, perhaps that's more appropriate. This works too though.

r-LaForge: using that now

vladaionescu:
I'll let others do a more complete review.

logbus/logstreamer/logstreamer_orchestrator.go
```diff
@@ -0,0 +1,170 @@
...
+	d.Write(delta)
+}
+
+// Done returns a channel that is closed once the Logstreamer has finished
+// communicating wit the server.
+// Callers should listen to the Done channel before exiting
+func (l *LogstreamOrchestrator) Done() chan struct{} {
+	l.doneMU.Lock()
+	defer l.doneMU.Unlock()
+	return l.doneCH
```
nelsam, line 157: This channel gets closed _on every retry_. Is that what we want?

Also, we will (I think) return a `nil` channel if the `start()` goroutine hasn't been scheduled yet, meaning that it will block forever.

I get the feeling that we should be setting `doneCH` in the constructor and closing it after all retries. That would also allow us to remove the `doneMU`.

logbus/logstreamer/logstreamer_orchestrator.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+	go func() {
+		for i := 0; i < l.retries; i++ {
+			l.start()
+			l.CloseLastLogstreamer()
+
+			l.mu.Lock()
+			l.deltas = newDeltasIter(l.deltaBuffer, l.initialManifest, l.verbose)
+			l.streamer = New(l.c, l.buildID, l.deltas)
+			l.mu.Unlock()
```
nelsam, line 80: It always makes me nervous when we can't `defer` an `Unlock()`. Can we move this to a separate method, e.g. `l.reset()` so that we can `defer l.mu.Unlock()` in that function?

logbus/logstreamer/logstreamer_orchestrator.go
```diff
@@ -0,0 +1,170 @@
...
+
+			go l.bus.AddSubscriber(l.deltas)
```
nelsam, line 82: It's important to note that doing this in a separate goroutine means that the subscriber _might not be added_ before it is removed. Since `RemoveSubscriber` doesn't check for that failure case (i.e. if the subscriber didn't exist, it happily returns as if the subscriber was removed), this can leave us with leaked subscribers.

We probably need some kind of communication for this - either a channel or a mutex lock:

```go
l.subCh = make(chan struct{})
go func() {
    defer close(l.subCh)
    l.bus.AddSubscriber(l.deltas)
}()
```

And in `l.CloseLastLogstreamer()`:
```go
if l.deltas != nil {
    <-l.subCh
    l.bus.RemoveSubscriber(l.deltas)
    l.deltas.close()
}
```
NOTE: this would need to be protected by `l.mu` as well, since it is reallocating the channel on each loop. Maybe the "separate method" I mentioned should be named `l.subscribe()` instead of `l.reset()` and take care of the `subCh` too.

nelsam (dismissed):
I see some stylistic things, an opportunity to use `sync.Once`, and some potential data races. Overall it seems reasonable, but I'd like to see some of this addressed (even if it's just "Oh, that's handled over here -->").

cloud/client.go
```diff
@@ -89,20 +103,30 @@ func NewClient(httpAddr, grpcAddr string, useInsecure bool, agentSockPath, authC
...
 	c.analytics = analytics.NewAnalyticsClient(conn)
+	if c.logstreamAddressOverride == "" {
+		c.logstream = logstream.NewLogStreamClient(conn)
+	} else {
+		logstreamConn, err := grpc.DialContext(ctx, c.logstreamAddressOverride, dialOpts...)
+		if err != nil {
+			return nil, errors.Wrap(err, "failed dialing logstream grpc")
+		}
+		c.logstream = logstream.NewLogStreamClient(logstreamConn)
+	}
```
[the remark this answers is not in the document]
r-LaForge: Good point - I agree that's more readable. I'll make the change

cmd/earthly/build_cmd.go
```diff
@@ -538,10 +540,25 @@ func (app *earthlyApp) actionBuildImp(cliCtx *cli.Context, flagArgs, nonFlagArgs
...
 	// This information is passed back right at the beginning of the build within earthfile2llb.
 	go func() {
+		beforeSelect := time.Now()
 		select {
 		case <-cliCtx.Context.Done():
+			now := time.Now()
+			app.console.VerbosePrintf(
+				"========== CONTEXT DONE BEFORE LOGSTREAMER STARTED AT %s (%d ms later) ==========",
+				now.Format(time.RFC3339Nano),
+				now.Sub(beforeSelect).Milliseconds(),
```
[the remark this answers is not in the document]
r-LaForge: Oh! Good point. I'll switch to that

logbus/logstreamer/logstreamer_orchestrator.go
```diff
@@ -0,0 +1,170 @@
...
+import (
+	"context"
+	"sync"
+	"sync/atomic"
+
+	"github.com/earthly/cloud-api/logstream"
+	"github.com/hashicorp/go-multierror"
+)
+
+type LogstreamOrchestrator struct {
```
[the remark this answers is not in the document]
r-LaForge: done

logbus/logstreamer/logstreamer_orchestrator.go
r-LaForge: I've handled this in a `subscribe` function now.
Thanks for the feedback + good catch

logbus/logstreamer/logstreamer_orchestrator.go
r-LaForge: `doneCH` is instantiated during struct instantiation in `NewOrchestrator`

r-LaForge: I think you're right about only needing to close the channel once... I'm going to test that and remove it if it's unnecessary

r-LaForge: Works great with just a single close.

r-LaForge:
Thanks for the in-depth feedback @nelsam! :)

nelsam (approved):

PR #2694, opened 2023-02-21
Example of exec form for a common bashism

goretkin:


alexcb:
Thanks for the contribution; I'm hesitant to merge this, due to keep the documentation of unsupported commands concise. 

The RUN command already has:

```
RUN [[<flags>...], "<executable>", "<arg1>", "<arg2>", ...] (exec form)
```

documented. It also supports a shell form, e.g. `RUN source foo.sh && printenv`.

Can you elaborate on what you were trying to solve when you came across the `SHELL` documentation? I'm curious to see if there's a different part of our docs that are lacking.

goretkin:
I figured an example would help users who were migrating from a Dockerfile.
When I read it, I did not know what "exec form" was, and an example would
have immediately clarified.

I certainly understand wanting to keep this section concise, in any case.

alexcb:
thanks for the response; I'm going to lean towards keeping it as-is.

goretkin:
That is fine. I came across another part of the documentation that is in a similar spirit, I believe: https://docs.earthly.dev/docs/earthfile#options-2

> --from
Although this option is present in classical Dockerfile syntax, it is not supported by Earthfiles. You may instead use a combination of SAVE ARTIFACT and COPY artifact form commands to achieve similar effects. For example, the following Dockerfile [...] (examples of how to migrate from Dockerfile to Earthfile)

PR #2699, opened 2023-02-22
Docs: Weekend only mode for satellite updates

brandonSc:
Apologies for accidentally pushing this to `next` then reverting it.


nelsam:
We've changed how docs branches work - I think it's only been mentioned in slack so far, though (maybe we should change that). I'll hit you up there with a link to the conversation.

brandonSc:
Yes, I missed the memo on that. Closing the PR and re-opening on to main.

PR #2708, opened 2023-02-23
ast: handle unindented comments in recipe bodies

### LINKED ISSUE #2697: 0.7.0 Regression? Unindented comments in recipe bodies cause parser errors
The following fails to execute after upgrade from 0.6.30 to 0.7.0:

```
1: my-target:
2:     RUN some_command
3:     # Comment regarding something
4:     SAVE ARTIFACT /stuff
```
`earthly` will fail with an error like:

```
Error: build target: build main: failed to solve: <...> Earthfile line 4:0 '    ': invalid syntax

Hints:
  - I got lost looking for '<EOF>'
```

That is, it will fail on the first line after an inline comment in a target.

This might very well be "as expected", but I couldn't find a mention regarding it in the docs.
Besides, it *did* work in 0.6.30 ;-) 

Regards, Jan-Helge

nelsam:
This solves an issue where unindented comments in recipe bodies would trigger a DEDENT token, causing parser errors when the comments really should just be ignored.

This also incidentally allows documentation to be on a line immediately following a comment if that comment is in an indented block (e.g. if the previous target ended with a comment).

Resolves #2697

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ All notable changes to [Earthly](https://github.com/earthly/earthly) will be doc
 
 - The `unexpected env` warning can now be silenced by creating a `.arg` or `.secret` file. [#2696](https://github.com/earthly/earthly/issues/2696).
 
+### Fixed
+
+- Unindented comments in the middle of recipe blocks no longer cause parser errors. [#2697](https://github.com/earthly/earthly/issues/2697)
+
 ## v0.7.0 - 2023-02-21
 
 The documentation for this version is available at the [Earthly 0.7 documentation page](https://docs.earthly.dev/v/earthly-0.7/).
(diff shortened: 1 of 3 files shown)


alexcb (approved):

CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ All notable changes to [Earthly](https://github.com/earthly/earthly) will be doc
 
 ## Unreleased
 
+### Fixed:
+- Unindented comments in the middle of recipe blocks no longer cause parser errors [#2697](https://github.com/earthly/earthly/issues/2697)
```
alexcb, line 8:
```suggestion
- Unindented comments in the middle of recipe blocks no longer cause parser errors. [#2697](https://github.com/earthly/earthly/issues/2697)
```

alexcb (approved):

alexcb (approved):

PR #2718, opened 2023-02-24
display --ssh in RUN command

alexcb:
running `RUN --push --ssh ...` didn't correctly display the --ssh flag in the output logs:

Before:

    --> RUN --push stat $SSH_AUTH_SOCK

After:

    --> RUN --push --ssh stat $SSH_AUTH_SOCK


nelsam (approved):
lgtm. in the future, it might be nice to make a `cmd` type with a `String()` method - feels easier to read than counting `%s` sequences. but that's not critical in this pr.

alexcb:
> lgtm. in the future, it might be nice to make a `cmd` type with a `String()` method - feels easier to read than counting `%s` sequences. but that's not critical in this pr.

being able to join them up rather than `%s%s...%s` would be a nice addition.

A stretch goal would be maintaining the original Earthfile ordering.

PR #2719, opened 2023-02-24
Allow toggling TLS and setting CA from CLI/environment

alexcb:
This also adds aliases to other existing TLS-related flags to more closely match existing conventions around 'translating' between a CLI flag, environment variable, and config entry.


alexcb:
These changes were pulled in from #2516, and had a minor merge conflict that was resolved.

PR #2731, opened 2023-02-28
docs: update remote caching to reflect --ci not implying inline cache flags

nelsam:


docs/remote-caching.md
```diff
@@ -39,18 +39,18 @@ The key benefit of this approach is that you get the upload for free if you anyw
...
 #### How to use inline caching
 
-In order to enable inline caching, simply add `--ci` in your invocation of `earthly` in your CI, or `--use-inline-cache` on individual developer's machines. If the `--push` command is also specified, the use of the cache will be read-write.
+In order to enable inline caching, add `--ci --use-inline-cache --save-inline-cache` in your invocation of `earthly` in your CI, or `--use-inline-cache` on individual developer's machines. If the `--push` command is also specified, the use of the cache will be read-write.
 
 In CI, read-only inline cache (typically in PR builds):
 
 ```bash
-earthly --ci +some-target
+earthly --ci --use-inline-cache --save-inline-cache +some-target
```
vladaionescu, line 47: Save is not necessary here, since there is no --push.

docs/remote-caching.md
```diff
@@ -39,18 +39,18 @@ The key benefit of this approach is that you get the upload for free if you anyw
 
 #### How to use inline caching
 
-In order to enable inline caching, simply add `--ci` in your invocation of `earthly` in your CI, or `--use-inline-cache` on individual developer's machines. If the `--push` command is also specified, the use of the cache will be read-write.
+In order to enable inline caching, add `--ci --use-inline-cache --save-inline-cache` in your invocation of `earthly` in your CI, or `--use-inline-cache` on individual developer's machines. If the `--push` command is also specified, the use of the cache will be read-write.
```
vladaionescu, line 42: Perhaps this can be unified a bit, to no longer make the distinction between CI and non-CI?

docs/remote-caching.md
alexcb, line 47: should this be:

```suggestion
earthly --ci --use-inline-cache +some-target
```

since this is for "read-only inline cache (typically in PR builds)"

docs/remote-caching.md
nelsam: Yeah, makes sense. All of the inline caching stuff is distinct from `--ci` now, so maybe we don't need to reference the `--ci` flag at all.

alexcb (commented):
I wonder if the `--ci` flag should be given in these examples? or is it a distraction from the `--use/save-inline-cache` flags?

vladaionescu (approved):

alexcb (approved):

PR #2756, opened 2023-03-07
ARG: add an --explicit-env feature

nelsam:
When --arg-explicit-env is set in the feature flags, commands will only be cache-busted when 'ARG --env' is used or the ARG in question is explicitly used by the command.


nelsam:
This ended up being too confusing. We couldn't figure out a reasonable way to enable it by default.

In the long term, we may revisit this as a `--no-env` or `--no-run` flag.

Issue #2742, opened 2023-03-02
make cause of error more clear


alexcb:
the cause of an error can be unclear for brand new users, for instance consider the output:

```
============================= ❌ FAILURE [Build 🔧] ==============================

Repeating the output of the command that caused the failure
/home/alex/test/alpine-glic+test *failed* | --> RUN apk --no-cache add binutils curl && curl -sL https://alpine-pkgs.sgerrand.com/sgerrand.rsa.pub -o /etc/apk/keys/sgerrand.rsa.pub && curl -sLO https://github.com/sgerrand/alpine-pkg-glibc/releases/download/${GLIBC_VER}/glibc-${GLIBC_VER}.apk && curl -sLO https://github.com/sgerrand/alpine-pkg-glibc/releases/download/${GLIBC_VER}/glibc-bin-${GLIBC_VER}.apk && apk add --no-cache glibc-${GLIBC_VER}.apk glibc-bin-${GLIBC_VER}.apk
/home/alex/test/alpine-glic+test *failed* | fetch https://dl-cdn.alpinelinux.org/alpine/v3.17/main/x86_64/APKINDEX.tar.gz
/home/alex/test/alpine-glic+test *failed* | fetch https://dl-cdn.alpinelinux.org/alpine/v3.17/community/x86_64/APKINDEX.tar.gz
/home/alex/test/alpine-glic+test *failed* | (1/8) Installing libgcc (12.2.1_git20220924-r4)
/home/alex/test/alpine-glic+test *failed* | (2/8) Installing libstdc++ (12.2.1_git20220924-r4)
/home/alex/test/alpine-glic+test *failed* | (3/8) Installing binutils (2.39-r2)
/home/alex/test/alpine-glic+test *failed* | (4/8) Installing ca-certificates (20220614-r4)
/home/alex/test/alpine-glic+test *failed* | (5/8) Installing brotli-libs (1.0.9-r9)
/home/alex/test/alpine-glic+test *failed* | (6/8) Installing nghttp2-libs (1.51.0-r0)
/home/alex/test/alpine-glic+test *failed* | (7/8) Installing libcurl (7.88.1-r0)
/home/alex/test/alpine-glic+test *failed* | (8/8) Installing curl (7.88.1-r0)
/home/alex/test/alpine-glic+test *failed* | Executing busybox-1.35.0-r29.trigger
/home/alex/test/alpine-glic+test *failed* | Executing ca-certificates-20220614-r4.trigger
[...]

PR #2757, opened 2023-03-07
make RUN ... errors easier to read

alexcb:
This detects and silences failures of the form

    Error: build target: build main: failed to solve: unlazy force execution: process "/bin/sh -c GLIBC_VER=2.31-r0 PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin /usr/bin/earth_debugger /bin/sh -c '... did not complete successfully: exit code: 1

As they are displayed in a clearer format higher up in the output.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to [Earthly](https://github.com/earthly/earthly) will be doc
 
 ## Unreleased
 
+### Changed
+
+- Simplified error message when a RUN command fails with an exit code. [#2742](https://github.com/earthly/earthly/issues/2742)
+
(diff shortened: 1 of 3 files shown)


alexcb:
#2742

vladaionescu:
Note that errors are already being processed in multiple places. This PR seems to add yet another point in the stack where errors are filtered in an arbitrary way. I would perhaps tweak some of the existing spots instead?

* https://github.com/earthly/earthly/blob/main/logbus/solvermon/vertexmon.go#L41
* https://github.com/earthly/earthly/blob/main/cmd/earthly/main.go#L649

vladaionescu (approved):

PR #2760, opened 2023-03-07
add test for silent analytics failure

alexcb:
the "failed to send analytics" message should only ever appear under verbose logging.

### DIFF
--- a/tests/Earthfile
+++ b/tests/Earthfile
@@ -227,6 +227,7 @@ ga-no-qemu:
     BUILD +pipelines
     BUILD +doc
     BUILD +no-network
+    BUILD +test-works-without-earthly-server
 
     # Forcing the implicit global wait/end block, causes some tests, which rely
     # on the ability to have two different targets issue the same SAVE IMAGE tag name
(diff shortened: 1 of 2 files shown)


tests/Earthfile
```diff
@@ -1269,6 +1270,25 @@ pipelines:
...
+      echo \"blocking \$host\"
+      for ip in \$(getent hosts \$host | awk '{ print \$1 }'); do
+        iptables -A OUTPUT -d \$ip -j DROP
+      done
+    done
+    date +%s > start-time
+    " > /tmp/precommand && chmod +x /tmp/precommand
+    DO +RUN_EARTHLY --verbose=false --pre_command=/tmp/precommand --post_command=" && date +%s > end-time" --earthfile=true.earth --target=+true
+    RUN if grep "failed sending analytics" earthly.output; then echo "analytics should have failed silently, but didnt" && exit 1; fi
+    RUN export duration=$(echo "$(cat end-time) - $(cat start-time)" | bc) && test "$duration" -lt 45 || (echo "it took $duration seconds to run +true, which is too long" && exit 1)
```
alexcb, line 1290: I really don't like this part -- it looks like we have a 30 second timeout that occurs if our server doesn't respond.

nelsam (approved):

tests/Earthfile
alexcb: CC @idodod

Issue #2762, opened 2023-03-08
referencing target with `FROM` and `BUILD` does not result in a `SAVE IMAGE --push`


alexcb:
```Earthfile
VERSION 0.7
common-base:
    FROM alpine
    SAVE IMAGE --push fakeuser/common:base

use-common-base:
    FROM +common-base
    RUN echo "this won't trigger the SAVE IMAGE in common-base"

test:
    BUILD +use-common-base
    FROM alpine
    IF sleep 3
      BUILD +common-base
    END
```

Running `earthly-v0.7.1 --push +test` should cause `fakeuser/common:base` to be pushed (resulting in a 403 error); however doesn't:

```
$ earthly-v0.7.1 --push +test
 Init 🚀
————————————————————————————————————————————————————————————————————————————————

           buildkitd | Found buildkit daemon as docker container (earthly-buildkitd)

 Build 🔧
————————————————————————————————————————————————————————————————————————————————

              alpine | --> Load metadata alpine linux/amd64
        +common-base | --> FROM alpine
        +common-base | [          ]   0% resolve docker.io/library/alpine@sha256:69665d02cb32192e52e07644d76bc6f25abeb5410edc1c7a81a10ba3f0efb90a              [----------] 100% resolve docker.io/library/alpine@sha256:69665d02cb32192e52e07644d76bc6f25abeb5410edc1c7a81a10ba3f0efb90a
    +use-common-base | *cached* --> RUN echo "this won't trigger the SAVE IMAGE in common-base"
               +test | *cached* --> IF sleep 3
              output | --> exporting outputs
              output | [----------] 100% exporting layers
              output | [----------] 100% exporting manifest sha256:1aa22a31d07845e1fee90248125c3ca5631c4633cbc7049bf48fa69caaeec5df
              output | [----------] 100% exporting config sha256:2367e5f196d5f4c056ad4b388390e2c81acfa22fa4e07047aca4cb6df4070c01
              output | [----------] 100% transferring docker.io/fake/common:base
              output | --> exporting outputs

 Push Summary ⏫
————————————————————————————————————————————————————————————————————————————————

Did not push image fake/common:base                 <---------- BAD, expected this to be pushed

 Local Output Summary 🎁
————————————————————————————————————————————————————————————————————————————————

Image +common-base output as fake/common:base


========================== 🌍 Earthly Build  ✅ SUCCESS ==========================
```

alexcb:
The use case comes from #2218 but only appears to be dropping the push flag.

PR #2768, opened 2023-03-09
propagate push to wait block items

alexcb:
Fixes a race bug where the --push option was not correctly propagated to
targets that were first indirectly referenced by a `FROM` or `COPY`, but
then later referenced by an explicit `BUILD`.

This was missed in eee3db502797bfb74e91886db59e685434e15ccc
and fixes https://github.com/earthly/earthly/issues/2762.

### DIFF
--- a/earthfile2llb/earthfile2llb.go
+++ b/earthfile2llb/earthfile2llb.go
@@ -247,12 +247,19 @@ func Earthfile2LLB(ctx context.Context, target domain.Target, opt ConvertOpt, in
 		opt.MainTargetDetailsFuture = nil
 	}
 	if found {
+		// The found target may have initially been created by a FROM or a COPY;
+		// however, if it is referenced a second time by a BUILD, it may contain items that
+		// require a save (export to the local host) or a push
 		if opt.DoSaves {
-			// Set the do saves flag, in case it was not set before.
 			sts.SetDoSaves()
+		}
+		if opt.DoPushes {
+			sts.SetDoPushes()
+		}
+		if opt.DoSaves || opt.DoPushes {
 			err := sts.Wait(ctx)
 			if err != nil {
-				return nil, errors.Wrapf(err, "resolve build context for target %s", target.String())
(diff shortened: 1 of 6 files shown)


earthfile2llb/earthfile2llb.go
```diff
@@ -250,9 +250,16 @@ func Earthfile2LLB(ctx context.Context, target domain.Target, opt ConvertOpt, in
 		if opt.DoSaves {
 			// Set the do saves flag, in case it was not set before.
 			sts.SetDoSaves()
+		}
+		if opt.DoPushes {
+			// Set the do saves flag, in case it was not set before.
```
nelsam, line 255:
```suggestion
			// Set the do pushes flag, in case it was not set before.
```

nelsam (approved):
lgtm, other than the minor copy/paste error :)

earthfile2llb/earthfile2llb.go
alexcb: I reworded this comment entirely, mind taking a second look?

earthfile2llb/earthfile2llb.go
```diff
@@ -247,12 +247,19 @@ func Earthfile2LLB(ctx context.Context, target domain.Target, opt ConvertOpt, in
 		opt.MainTargetDetailsFuture = nil
 	}
 	if found {
+		// The found target may have been initially referenced by a FROM or a COPY
+		// if it is referenced a second time by a BUILD, it may contain items that
+		// require a save (export to the local host) or a push
```
nelsam, line 252: Is this two sentences or one? `The found target may have been initially referenced by a FROM or a COPY if it is referenced a second time by a BUILD` reads as if it should be two sentences, I think, but I'm not 100% sure.

alexcb: `; however, ` to the rescue :)

nelsam (approved):

PR #2772, opened 2023-03-10
Project command tweaks

brandonSc:
* Be more clear about what is being deleted in a project
* Ask for confirmation before deleting the project (unless `-f` flag is provided).
* Can use `-p` flag as alias to `--project`

Screen Shot 2023-03-10 at 2 24 22 PM

### DIFF
--- a/cmd/earthly/project_cmds.go
+++ b/cmd/earthly/project_cmds.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strings"
 	"text/tabwriter"
 
 	"github.com/pkg/errors"
@@ -25,8 +26,8 @@ func (app *earthlyApp) projectCmds() []*cli.Command {
 		},
 		{
 			Name:        "rm",
-			Usage:       "Remove an existing project from the organization *beta*",
-			Description: "Remove an existing project from the organization *beta*",
+			Usage:       "Remove an existing project and all of its associated pipelines and secrets *beta*",
+			Description: "Remove an existing project and all of its associated pipelines and secrets *beta*",
 			UsageText:   "earthly project [--org <organization-name>] rm <project-name>",
 			Action:      app.actionProjectRemove,
 		},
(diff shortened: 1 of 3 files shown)


mikejholly (approved):
Good stuff. I think it makes sense as it's a pretty destructive action. Wonder if we should include a `--force` option as well.

brandonSc:
> Good stuff. I think it makes sense as it's a pretty destructive action. Wonder if we should include a `--force` option as well.

I think that's a good call, in case someone wants to script this.

brandonSc:
> > Good stuff. I think it makes sense as it's a pretty destructive action. Wonder if we should include a `--force` option as well.
> 
> I think that's a good call, in case someone wants to script this.

With `--force` option:

Screen Shot 2023-03-10 at 2 59 51 PM

nelsam (approved):

cmd/earthly/project_cmds.go
```diff
@@ -124,6 +125,17 @@ func (app *earthlyApp) actionProjectRemove(cliCtx *cli.Context) error {
...
+	answer, err := promptInput(cliCtx.Context,
+		"WARNING: you are about to permanently delete this project and all of its associated pipelines, build history and secrets.\n"+
+			"Would you like to continue?\n"+
+			"Type 'y' or 'yes': ")
+	answer = strings.TrimSpace(strings.ToLower(answer))
+
+	if answer != "y" && answer != "yes" {
+		app.console.Printf("Operation aborted.")
+		return nil
+	}
```
vladaionescu, line 137: We generally try not to make anything interactive, unless we have no choice. Reason being that it's harder to work with it in automation.

I think it's ok not to have any prompt here, the same way `rm -rf ~` doesn't present a prompt either.

vladaionescu: `-f` could be a decent compromise if we want to keep the prompt.

vladaionescu: If we keep it, can you make the format of the question consistent with other prompts? e.g. I think we had things like `[yN]` in others.

vladaionescu: I was saying "go ahead and always delete the project". But prompt + option to use -f to skip the prompt is also ok.

Issue #2771, opened 2023-03-09
CircleCI example does not work


andrew-jones:
https://docs.earthly.dev/ci-integration/vendor-specific-guides/circle-integration

Is using an old image deprecated by CircleCI.

PR #2773, opened 2023-03-10
Update machine image used by circle ci in test and doc

idodod:
Addresses #2771


nelsam (approved):
lgtm. don't forget that docs changes also need to go on the `docs-0.7` (and maybe `docs-0.6`) branches.

vladaionescu (approved):

PR #2775, opened 2023-03-11
Misc. doc updates (0.7 specific)

rrjjvv:
My original intent was only to update the `.env`/`.arg`/`.secret` changes, as that was very noticeable.  Then I started noticing other things on the way.

There are still a number of `0.6.x` references that I left untouched for various reasons.  The subset I changed were mostly around the 'language' (e.g., `VERSION 0.7`), which are very user-facing and 'static'.  The rest were largely around advanced deployment, esp. download links and docker tags, which are probably less visible and are moving targets.


vladaionescu:
Wow! Thank you for fixing this, Roberto!

Alex and Sam will review next week.

alexcb (approved):

alexcb:
This is some amazing attention to detail; thank you very much for discovering AND fixing this.

PR #2803, opened 2023-03-20
ARG scoping changes and the SET command

nelsam:
This implements two pieces of #1575:

# ARG scoping

Now ARG will error if the same variable is declared twice in the same scope. Non-global ARGs may be declared that override global ARGs, but global ARGs will never override non-global ARGs.

Default values will also respect this scoping, so a global ARG's default cannot override a non-global's default (and vice versa).

# SET

A new command has been implemented to reassign existing ARGs - SET. SET works on all ARGs that  were declared in the same scope as the SET command. This means that SET does not work on globals outside of the base target.

All of this means that the following works as you would expect it to:

```Earthfile
VERSION --arg-scope-and-set 0.7
FROM alpine

ARG --global foo = bar
RUN echo $foo # echoes 'bar'

ARG foo = baz
RUN echo $foo # echoes 'baz'

foo:
    ARG foo = bacon
    SET foo = $foo and eggs
    RUN echo $foo # echoes 'bacon and eggs'

bar:
    RUN echo $foo # echoes 'bar'
```


variables/scope.go
```diff
@@ -32,34 +32,32 @@ func (s *Scope) Clone() *Scope {
...
-
-// GetActive returns an active variable by name.
-func (s *Scope) GetActive(name string) (string, bool) {
-	variable, found := s.variables[name]
-	active := false
-	if found {
-		active = s.activeVariables[name]
+// Get gets a variable by name.
+func (s *Scope) Get(name string, opts ...ScopeOpt) (string, bool) {
+	opt := applyOpts(opts...)
```
alexcb, line 37: nice change!

tests/arg-set.earth
```diff
@@ -0,0 +1,38 @@
...
+
+ARG foo = sports
+SET foo = ${foo} car
+
+RUN echo $foo
+
+test:
+    ARG bar = baz
+    SET bar = ${bar}inga
+    RUN echo $bar
```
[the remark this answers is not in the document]
nelsam: Since this deals a lot with how `ARG`s are assigned, a lot of the tests are overriding the `ARG` values - so I had to do a bit more work to get the tests to use `test`. That said, I think I figured it out. I've updated all of the happy paths to use `test` rather than `echo/output_contains`.

alexcb (approved):
looks great!

variables/collection.go
```diff
@@ -208,17 +244,49 @@ func (c *Collection) DeclareArg(name string, defaultValue string, global bool, p
...
+	c.args().Add(name, finalValue, opts...)
 	if global {
-		c.globals().AddActive(name, finalValue)
+		c.globals().Add(name, finalValue, opts...)
 	}
 	c.effectiveCache = nil
 	return finalValue, finalDefaultValue, nil
 }
 
+// DeclareArg declares an arg. The effective value may be
```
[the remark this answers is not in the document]
nelsam: Interesting ... `DeclareArg` returns a different `defaultValue` depending on whether or not it was overridden. If it was overridden, the returned default will _not_ process shellouts. If it _was not_ overridden, the returned default _will_ process shellouts.

@alexcb Does the above make sense? Is this how it should work - with the returned default value changing based on whether or not it was overridden?

earthfile2llb/interpreter.go
```diff
@@ -171,19 +171,23 @@ func (i *Interpreter) handleStatement(ctx context.Context, stmt spec.Statement)
...
+	if stmt.Wait != nil {
 		return i.handleWait(ctx, *stmt.Wait)
-	} else if stmt.Try != nil {
+	}
+	if stmt.Try != nil {
 		return i.handleTry(ctx, *stmt.Try)
-	} else {
-		return i.errorf(stmt.SourceLocation, "unexpected statement type")
 	}
+	return i.errorf(stmt.SourceLocation, "unexpected statement type")
```
vladaionescu, line 190: I believe that the intention here is that only one of these is set at a time - hence the `else`s. Is this change intentional?

vladaionescu: Oh yeah - they end in return. That's fine then.

mwopfner:
Do we actually want to have `SET` for `ARG`  if #2808 gets implemented or do we only want to have `SET` for `LET`/`VAR`?

nelsam:
> Do we actually want to have `SET` for `ARG` if #2808 gets implemented or do we only want to have `SET` for `LET`/`VAR`?

The `LET/VAR` proposal was a very rough, quick thought I had while I was implementing `SET` and I honestly don't know if we plan to implement it, yet. I'll bring it up at our next sync.

PR #2805, opened 2023-03-20
Satellites doc updates

brandonSc:
Some things in the satellites doc page were out of date.

### DIFF
--- a/docs/cloud/satellites.md
+++ b/docs/cloud/satellites.md
@@ -67,11 +67,14 @@ Follow the steps in the [Earthly Cloud overview](./overview.md#getting-started) 
 
 ### 2. Purchase a Satellites Plan
 
-Satellites are now a paid feature and require a valid subscription to begin using. The subscription includes a 14-day free trial which can be canceled before any payment is made. Visit the [pricing page](https://earthly.dev/pricing) for more billing details.
+Satellites are a paid feature which require a subscription to begin using. The subscription includes a 14-day free trial which can be canceled before any payment is made. Visit the [pricing page](https://earthly.dev/pricing) for more billing details.
+
+You can start your free trial by using the checkout form below. Be sure to provide the name of your Earthly org from Step 1.
 
 [**Click here to start your subscription**](https://buy.stripe.com/8wM9Es4BT4Vvb4YbIJ)
 
-Once you've submitted your details on the checkout page, you will receive an email with further instructions. Please reply or [send us an email](mailto:support+satellite@earthly.dev) with your org name from step 1 in order for us to activate the Satellites feature on your account.
+Once you've submitted your details on the checkout page, you will receive an automated confirmation email.
+The Earthly team will send you another follow-up shortly after once the satellites feature has active on your account.
 
 ### 3. Ensure that you have the latest version of Earthly
 
@@ -156,12 +159,7 @@ For more information on managing satellites, see the [Managing Satellites page](
 
 ## Satellite specs
 
-Satellites are currently only available in one size, and it has the following specs:
-
-* 4 CPUs
-* 16 GB of RAM
-* 90 GB of cache storage
-* 5 Gib internet bandwidth
+Satellites are currently only available in different sizes and architectures. Please see the [Pricing Page](https://earthly.dev/pricing) for more details.
 
 ## Using Satellites in CI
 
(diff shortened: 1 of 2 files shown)


docs/cloud/satellites.md
```diff
@@ -67,11 +67,14 @@ Follow the steps in the [Earthly Cloud overview](./overview.md#getting-started)
 
 ### 2. Purchase a Satellites Plan
 
-Satellites are now a paid feature and require a valid subscription to begin using. The subscription includes a 14-day free trial which can be canceled before any payment is made. Visit the [pricing page](https://earthly.dev/pricing) for more billing details.
+Satellites are a paid feature which require a subscription to begin using. The subscription includes a 14-day free trial which can be canceled before any payment is made. Visit the [pricing page](https://earthly.dev/pricing) for more billing details.
+
+You can start your free trial by using the checkout form below. Be sure to provide the name of your Earthly org from Step 1.
```
vladaionescu, line 72: I see that this now includes a new field for the org. I should probably update the automated email in Zapier so that it no longer asks for this.

docs/cloud/satellites.md
```diff
@@ -156,12 +159,7 @@ For more information on managing satellites, see the [Managing Satellites page](
 
 ## Satellite specs
 
-Satellites are currently only available in one size, and it has the following specs:
-
-* 4 CPUs
-* 16 GB of RAM
-* 90 GB of cache storage
-* 5 Gib internet bandwidth
+Satellites are currently only available in different sizes and architectures. Please see the [Pricing Page](https://earthly.dev/pricing) for more details.
```
vladaionescu, line 162:
```suggestion
Satellites are available in various sizes and architectures. Please see the [Pricing Page](https://earthly.dev/pricing) for more details.
```

docs/guides/auth.md
```diff
@@ -118,6 +118,19 @@ git:
...
+You can run earthly with `--verbose`, which will provide debugging messages to help understand how a remote earthly reference is transformed into a git URL for cloning.
+
+You can additionally enable low-level git debugging in buildkit, by adding the following to your `~/.earthly/config.yml`:
+
+```yaml
+global:
+  buildkit_additional_args: [ '-e', 'BUILDKIT_DEBUG_GIT=1' ]
+```
+
+The buildkit logs can be displayed with `docker logs earthly-buildkitd`.
```
vladaionescu, line 132: This change seems like it belongs to another PR? Is this intentional?

docs/cloud/satellites.md
brandonSc: Yes, I was thinking that too.

docs/guides/auth.md
brandonSc: I think I scooped it from main by accident. But I don't see why this wouldn't be in the 0.7 docs. @alexcb is this okay with you?

brandonSc: I also tried bringing 0.7-docs up to date with main. I think I was used to the `next` branch. Apologies if I shouldn't have done that

vladaionescu: > I also tried bringing 0.7-docs up to date with main. I think I was used to the next branch. Apologies if I shouldn't have done that

We only do that on release. `main` might contain docs for unreleased features.

vladaionescu (approved):

docs/guides/auth.md
brandonSc: That makes sense. Sorry, I did that a bit impulsively. It looks like this is the only doc change from `main` that wasn't in `docs-0.7` yet, and doesn't correspond to an unreleased feature. 

Is the process for adding to these docs documented anywhere yet? Perhaps it can go in `CONTRIBUTING.md`. Happy to add that.

vladaionescu: Already there: https://github.com/earthly/earthly/blob/main/CONTRIBUTING.md#documentation ;-)

brandonSc: Heh I skimmed, but missed it. Should've used ctrl-f :)

brandonSc:
I snuck in a couple more satellite doc updates:
* Using the `EARTHLY_SATELLITE` environment var (as per user request)
* Encourage docker login to avoid rate limits

docs/guides/auth.md
alexcb: > I think I scooped it from main by accident. But I don't see why this wouldn't be in the 0.7 docs. @alexcb is this okay with you?

Seems like you just need to change this PR to be against `docs-0.7`; currently you're PR is attempting to merge `docs-0.7` (plus additional commits) into `main` -- if this is merged as-is (please dont!), github will squash commits together, which will then lead to a conflict when we merge `main` back into `docs-0.7`.

alexcb (approved):

PR #2806, opened 2023-03-21
Refactor TLS configuration

idodod:
This PR contains the following changes:
 - Rely only on configuration, remove flags.
 - use TLS by default when applicable.
 - Allow connection to satellites without TLS (for dev purposes)

The motivation for the above mentioned changes is to make the TLS settings a bit more straightforward.
It seems like the combination of configuration and flags added some complexity.
These are definitely breaking changes but since the removed flags were hidden, it should not have a significant effect on the majority of users.

In order to achieve a more dynamic changes with regards to TLS, instead of using flags/env vars, a user may pass a different config file - one that has tls settings enabled and one that does not.

also increase test timeout to address #2867


buildkitd/buildkitd.go
```diff
@@ -924,40 +921,44 @@ func makeTLSPath(path string, installationName string) (string, error) {
 }
 
 func addRequiredOpts(settings Settings, installationName string, opts ...client.ClientOpt) ([]client.ClientOpt, error) {
+	server, err := url.Parse(settings.BuildkitAddress)
```
idodod, line 924: didn't see a reason not to always validate first

cmd/earthly/build_cmd.go
```diff
@@ -225,14 +225,14 @@ func (app *earthlyApp) actionBuildImp(cliCtx *cli.Context, flagArgs, nonFlagArgs
...
 
-	err = app.initFrontend(cliCtx)
+	err = app.configureSatellite(cliCtx, cloudClient, gitCommitAuthor, gitConfigEmail)
 	if err != nil {
-		return err
+		return errors.Wrapf(err, "could not construct new buildkit client")
 	}
 
-	err = app.configureSatellite(cliCtx, cloudClient, gitCommitAuthor, gitConfigEmail)
+	err = app.initFrontend(cliCtx)
```
idodod, line 233: I changed the calling order here because initFrontend sets the majority of buildkitsettings properties according to a configueration that is not up to date.
More specifically, useTCP is determined according to the bkURL but in case of a satellite, the satellite host is only set in `configureSatellite`, so without changing the order, the value of UseTCP was set to false, which is wrong.

cmd/earthly/buildkit.go
```diff
@@ -110,31 +114,6 @@ func (app *earthlyApp) getBuildkitClient(cliCtx *cli.Context, cloudClient *cloud
 	return buildkitd.NewClient(cliCtx.Context, app.console, app.buildkitdImage, app.containerName, app.installationName, app.containerFrontend, Version, app.buildkitdSettings)
 }
 
-func (app *earthlyApp) handleTLSCertificateSettings(context *cli.Context) {
```
idodod, line 113: didn't see much point in keeping this function now that it's only passing values from the config to the buildkitsettings struct.

idodod:
Hi @rrjjvv ,
I wanted to get your attention to this PR because I know you recently made changes that are being reverted here.
To give you some context - we realized being able to set TLS settings in both a config and flags made things more complex, while the ability to change TLS settings per execution should still be possible by passing a different config file with different TLS settings.

Please let me know your thoughts.
Thanks!

idodod:
> > we realized being able to set TLS settings in both a config and flags made things more complex
> 
> Indeed. While my PR was relatively simple, it was surprisingly difficult to track down why my original issues were occurring and figuring out how to resolve it.
> 
> > changes that are being reverted here.
> 
> Fortunately (for you), I haven't had a chance to make use of my change (our workarounds are still in place), so based on your above description, I don't believe this will break me. (I wanted to get you a quick response, so haven't actually looked at the change; I'm going off of the PR description.)
> 
> > while the ability to change TLS settings per execution should still be possible by passing a different config file with different TLS settings.
> 
> Yeah... the only 'problem' is that we're limited to one config. If there was a concept of system and user configs that were merged, that would be one thing. But my configs are relatively large, so having to keep those 'base' settings in sync across TLS-specific configs is tedious and error-prone. So I do think there's still value in flags and/or environment variables, unless earthly supports multiple config sources. (I'm not suggesting it should... but it could come in hand.)
> 
> Though I'll be sad to see it go, from my perspective, I mainly care that the docs are accurate. I have a solution, and I'm pretty sure this PR won't break it. I lost a lot of time due to the inconsistencies, and my original PR was just as much to avoid myself or others running it into again. I appreciate you giving me a heads up!

I appreciate the feedback @rrjjvv , I'm happy this won't break your usage as things stand.
Re the config, I see your point. We were discussing the option of specifying tls settings by host, which I _think_ can help your use case(?), though it's probably low priority right now.

rrjjvv:
> We were discussing the option of specifying tls settings by host, which I think can help your use case(?), though it's probably low priority right now.

That actually wouldn't help in my _current_ scenario (not to say it might not be handy in the future).  In my case, it's whether to do TLS or not (in simple terms).

My current solution relies on the `earthly config` feature.  We have a baseline config (non-TLS), and if the decision is made to build remotely, at runtime we execute a series of `earthly config` commands to add the relevant TLS settings.  It works (and solves the problem of having to keep multiple config variations in sync), just verbose.

buildkitd/buildkitd.go
```diff
@@ -355,10 +355,7 @@ func Start(ctx context.Context, console conslogging.ConsoleLogger, image, contai
 			})
 		}
 
-		bkURL, err := url.Parse(settings.BuildkitAddress)
-		if err != nil {
-			panic("Buildkit address was not a URL when attempting to start buildkit")
-		}
+		bkURL, _ := url.Parse(settings.BuildkitAddress) // error is already handled in addRequiredOpts
```
nelsam, line 358: I would personally prefer to leave the error handling in. While this error may _currently_ be handled in `addRequiredOpts`, that may not always be the case - and if someone removes it from `addRequiredOpts` they probably won't know to update it here. This is a classic case of "comments are just lies waiting to happen."

buildkitd/buildkitd.go
```diff
@@ -924,40 +921,44 @@ func makeTLSPath(path string, installationName string) (string, error) {
...
 	if !settings.UseTCP || !settings.UseTLS {
 		return opts, nil
 	}
 
-	server, err := url.Parse(settings.BuildkitAddress)
-	if err != nil {
-		return []client.ClientOpt{}, errors.Wrap(err, "invalid buildkit url")
-	}
+	if settings.TLSCA == "" && settings.ClientTLSCert == "" && settings.ClientTLSKey == "" {
+		opts = append(opts, client.WithCredentials("", "", "", ""))
```
nelsam, line 942: nit: `opts` is never changed past this point, so it's generally more idiomatic to just return inside the `if` statement and remove the `else`.

```go
		return append(opts, client.WithCredentials("", "", "", ""))
	}
	// the rest of the logic
	return opts, nil
```

cmd/earthly/build_cmd.go
```diff
@@ -225,14 +225,14 @@ func (app *earthlyApp) actionBuildImp(cliCtx *cli.Context, flagArgs, nonFlagArgs
...
 	if err != nil {
-		return errors.Wrapf(err, "could not construct new buildkit client")
+		return err
```
nelsam, line 235: nit: I like to see wrapped errors to provide extra context about what failed. In this case, it may be as small as `errors.Wrap(err, "could not initialize frontend")`, but it still helps a little when tracking down the source of an error.

docs/ci-integration/remote-buildkit.md
```diff
@@ -77,12 +77,13 @@ This is the address of the remote daemon. It should look something like this: `t
...
+**`tls_disabled`**
 
-Set this to `true` when using TLS is desired.
+TLS will be enabled by default (unless using a local buildkit container).  
+Set this to `false` when using TLS is not desired.
 
 ### Local-Remote
 
-It is also possible to use the remote protocols (TCP and mTLS) locally, while still letting Earthly manage the daemon container. You can do this by enabling mTLS(`tls_enabled`).
+It is also possible to use the remote protocols (TCP and mTLS) locally, while still letting Earthly manage the daemon container.
```
nelsam, line 87: nit: this appears to have some trailing whitespace.

nelsam (approved):
Most of my comments are just on style and shouldn't block a merge. I do think we should still be checking the error from `url.Parse` even though we ostensibly check it elsewhere, but that's the only comment of note that I have.

cmd/earthly/build_cmd.go
idodod: I agree, this was just part of moving the order of the calls and their matching error handling if clauses. I can add a wrapping around this error

buildkitd/buildkitd.go
idodod: makes sense, will update, I mostly didn't like the panic but I suppose I can just adjust that instead (I'm guessing we used panic because it was unexpected given the previous err handling, but I could be wrong).

idodod:
> Most of my comments are just on style and shouldn't block a merge. I do think we should still be checking the error from `url.Parse` even though we ostensibly check it elsewhere, but that's the only comment of note that I have.

I appreciate the review @nelsam , I think I addressed all your comments.

buildkitd/buildkitd.go
```diff
@@ -924,24 +924,26 @@ func makeTLSPath(path string, installationName string) (string, error) {
 }
 
 func addRequiredOpts(settings Settings, installationName string, opts ...client.ClientOpt) ([]client.ClientOpt, error) {
+	server, err := url.Parse(settings.BuildkitAddress)
+	if err != nil {
+		return []client.ClientOpt{}, errors.Wrap(err, "invalid buildkit url")
```
alexcb, line 929: should we include the url here?
```suggestion
		return []client.ClientOpt{}, errors.Wrapf(err, "failed to parse buildkit url %s", settings.BuildkitAddress)
```

alexcb (approved):

rrjjvv:
@idodod After getting a chance to look at the actual changes... it appears this _will_ actually break me, FYI.  Specifically, the change from `tls_enabled` to `tls_disabled`.

What are the chances of allowing `tls_enabled` to remain (having it set `tls_disabled` as appropriate)?  Giving it a deprecation phase would be much, much nicer.  I don't think this will impact our end-users, but it will make upgrading our CI a non-trivial matter.  (With our rollout, it's not always as simple as "modify the config while you drop in the new binary"; our configs and the binaries that use them are not coupled by time or process.)

idodod:
changed the config value name back to `tls_enabled`, but set its default value to true.

buildkitd/buildkitd.go
```diff
@@ -928,24 +928,26 @@ func makeTLSPath(path string, installationName string) (string, error) {
...
 
 	if !settings.UseTCP || !settings.UseTLS {
 		return opts, nil
 	}
 
-	server, err := url.Parse(settings.BuildkitAddress)
-	if err != nil {
-		return []client.ClientOpt{}, errors.Wrap(err, "invalid buildkit url")
+	if settings.TLSCA == "" && settings.ClientTLSCert == "" && settings.ClientTLSKey == "" {
+		return append(opts, client.WithCredentials("", "", "", "")), nil
```
dchw, line 949: What happens if I set up TLS locally for some other buildkit instance, and then attempt to use a satellite? Doesn't this make it so we would use the generated certs for the instance (which wouldn't work)?

If so, doesn't that mean that we would have to toggle these settings to swap between a TLS buildkit and a satellite?

earthly-entrypoint.sh
```diff
@@ -46,7 +46,24 @@ if [ -z "$NO_BUILDKIT" ]; then
...
+
+    if [ ! -f /etc/cert.pem ]; then
+      ln -s /root/.earthly/certs/buildkit_cert.pem /etc/cert.pem
+    fi
+
+    if [ ! -f /etc/key.pem ]; then
+      ln -s /root/.earthly/certs/buildkit_key.pem /etc/key.pem
+    fi
+
+
```
dchw, line 64: Why are we using TLS in this container with the buildkit thats also running inside the container?

idodod: I added this because if tls is disabled, then connection to a satellite will fail or will require to enable it by passing 
`-e GLOBAL_CONFIG="{tls_enabled: true}"`.

I _think_ this is more seamless but I acknowledge it's somewhat redundant to use tls internally.

What are your thoughts?

dchw (approved):

PR #2807, opened 2023-03-21
earthly doc: output documentation for ARGs, ARTIFACTs, and IMAGEs

nelsam:
This updates `earthly doc` with additional output for `ARG`, `SAVE ARTIFACT`, and `SAVE IMAGE`.

NOTE: this is a draft because I still need to deal with global `ARG`s.

### DIFF
--- a/cmd/earthly/doc_cmds.go
+++ b/cmd/earthly/doc_cmds.go
@@ -1,0 +1,258 @@
+package main
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/earthly/earthly/ast/spec"
+	"github.com/earthly/earthly/buildcontext"
+	"github.com/earthly/earthly/domain"
+	"github.com/earthly/earthly/earthfile2llb"
+	"github.com/earthly/earthly/util/platutil"
+	gwclient "github.com/moby/buildkit/frontend/gateway/client"
+	"github.com/pkg/errors"
+	"github.com/urfave/cli/v2"
+)
+
+func (app *earthlyApp) actionDocumentTarget(cliCtx *cli.Context) error {
+	app.commandName = "docTarget"
+
+	if cliCtx.NArg() > 1 {
+		return errors.New("invalid number of arguments provided")
+	}
+
+	var tgtPath string
+	if cliCtx.NArg() > 0 {
+		tgtPath = cliCtx.Args().Get(0)
+		switch tgtPath[0] {
+		case '.', '/', '+':
+		default:
+			return errors.New("remote-paths are not currently supported - documentation targets must start with one of ['.', '/', '+']")
+		}
+	}
+
+	singleTgt := true
+	if !strings.ContainsRune(tgtPath, '+') {
+		tgtPath += "+base"
+		singleTgt = false
+	}
+
+	target, err := domain.ParseTarget(tgtPath)
+	if err != nil {
+		return errors.Errorf("unable to parse target [%v]", tgtPath)
+	}
+
+	gitLookup := buildcontext.NewGitLookup(app.console, app.sshAuthSock)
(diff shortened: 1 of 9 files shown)


vladaionescu:
Should there be a way to print the detail output for all targets? Like a flag people can use?

cmd/earthly/doc_cmds.go
```diff
@@ -0,0 +1,277 @@
...
+
+	singleTgt := true
+	if !strings.ContainsRune(tgtPath, '+') {
+		tgtPath += "+base"
+		singleTgt = false
+	}
+
+	target, err := domain.ParseTarget(tgtPath)
+	if err != nil {
+		return errors.Errorf("unable to parse target [%v]", tgtPath)
```
alexcb, line 43: maybe use quotes instead -- we already do that in some places.
```suggestion
		return errors.Errorf("unable to parse target %q", tgtPath)
```

earthfile2llb/earthfile2llb.go
```diff
@@ -350,3 +350,42 @@ func GetTargetArgs(ctx context.Context, resolver *buildcontext.Resolver, gwClien
...
+
+// ImageNames returns the parsed names of a SAVE IMAGE command.
+func ImageNames(ctx context.Context, cmd spec.Command) ([]string, error) {
+	var opts saveImageOpts
+	args, err := parseArgs("SAVE IMAGE", &opts, getArgsCopy(cmd))
+	if err != nil {
+		return nil, errors.Wrapf(err, "invalid SAVE IMAGE arguments %v", cmd.Args)
+	}
+	return args, nil
+}
```
vladaionescu, line 391: Since these API functions are growing in scope, I would perhaps move them to a new file within the same package, to avoid mixing too many things in here. Possibly even a new sub-package?

vladaionescu (approved):
This is great! Just a couple minor ideas.

nelsam:
> Should there be a way to print the detail output for all targets? Like a flag people can use?

🤔 the output gets extremely verbose very quickly when I do that. I can see it being used with `less` or `grep` though. Probably worth just biting the bullet, yeah.

earthfile2llb/earthfile2llb.go
nelsam: For the time being, I moved them to their own file. They rely pretty heavily on `parseArgs` and the like, and I don't feel like we're at the point where all of those functions should move to their own sub-package, yet.

nelsam:
@vladaionescu I've added a `--long` flag (to match `earthly ls --long`) that outputs full details of all targets. Can you just double-check that it's what you expect out of that flag? (re-requesting a review)

tests/Earthfile
```diff
@@ -1161,22 +1162,94 @@ help:
...
+      baz.txt -> out/baz.txt
+        out/baz.txt is a documented artifact that is saved locally.
+      bacon.txt -> out/eggs.txt
+        bacon.txt is also a documented artifact that is saved locally.
+    IMAGES:
+      baz
+        baz is a documented image.
+      bar
+      bacon, eggs
+        eggs is just one of the image names, and yet it should still be recognized for this multiple-name SAVE IMAGE format." output.txt
```
vladaionescu, line 1251: Looks great!

vladaionescu (approved):
:+1:

nelsam:
> > thinking the output gets extremely verbose very quickly when I do that. I can see it being used with less or grep though. Probably worth just biting the bullet, yeah.
> 
> Maybe we can add an option` --web` that opens the documentation as HTML in the default browser?

Yeah, I think that's a reasonable long-term goal. As is `pkg.earthly.dev` to document open source Earthfile libraries.

Those are much larger chunks of effort though (for example: we'd need to get some web design help on 'earthly doc --web'). It's probably best to propose them as new issues for users to vote on, since the effort is quite significant.

PR #2819, opened 2023-03-23
Modify color and location for log mesage+url

idodod:
A few things I wasn't sure about and would like to get your opinion on:
 - Should printing these messages be more opinionated (i.e create some special function to print log in a specific color, not expose `ColorPrintf`)?
 - Moving the location of `Streaming logs to` outside of the go routine, means that it might be printed before we start uploading the log, is that ok?
 - Is the color ok (see attached screenshots)?
 - Not directly related to this PR, but I noticed we have a private function `makeColor`, is it needed given `color.New` is available (the difference is the no color check in the latter)?

terminal - 
image
ci - 
image

### DIFF
--- a/cmd/earthly/build_cmd.go
+++ b/cmd/earthly/build_cmd.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"fmt"
+	"github.com/fatih/color"
 	"io"
 	"net/url"
 	"os"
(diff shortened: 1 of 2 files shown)


cmd/earthly/build_cmd.go
```diff
@@ -571,11 +572,14 @@ func (app *earthlyApp) actionBuildImp(cliCtx *cli.Context, flagArgs, nonFlagArgs
...
 					app.logbusSetup.StartLogStreamer(cliCtx.Context, cloudClient)
-					app.console.Printf("Streaming logs to %s\n", logstreamURL)
 				}
 			}
 		}
 	}()
+
+	if app.logstream && doLogstreamUpload {
+		app.console.ColorPrintf(color.New(color.FgHiYellow), "Streaming logs to %s\n\n", logstreamURL)
+	}
```
vladaionescu, line 582: Does this show up under `Init` instead of `Build` now by any chance?

vladaionescu (approved):

alexcb (approved):

cmd/earthly/build_cmd.go
idodod: yes, I had to remove it from the go routine, I just hope there aren't any significant cases of showing the message without it being available.

PR #2822, opened 2023-03-24
Add command line options for time/size based partial cache pruning

dchw:
This adds two command line options for pruning the cache.

* `--keep-duration`, which will drop all cache older than this duration. It expects a string parseable by `time.ParseDuration`.
* `--target-size`, which will drop the oldest items from the cache until the size is reached. It expects a string parseable by `humanize.ParseBytes` (which we already have in our `go.mod`).

I presume docs need to be added, I'll append those to this PR if these options look ok.

### DIFF
--- a/cmd/earthly/root_cmds.go
+++ b/cmd/earthly/root_cmds.go
@@ -174,12 +174,12 @@ func (app *earthlyApp) rootCmds() []*cli.Command {
 					Destination: &app.pruneReset,
 				},
 				&cli.StringFlag{
-					Name:        "target-size",
+					Name:        "size",
 					Usage:       "Prune cache to specified size, starting from oldest",
 					Destination: &app.pruneTargetSize,
 				},
 				&cli.DurationFlag{
-					Name:        "keep-duration",
+					Name:        "age",
 					Usage:       "Prune cache older than the specified duration",
 					Destination: &app.pruneKeepDuration,
 				},
@@ -791,10 +791,10 @@ func (app *earthlyApp) actionPrune(cliCtx *cli.Context) error {
 	}
 
 	sizeInBytes := uint64(0)
-	if len(app.pruneTargetSize) > 0 {
+	if app.pruneTargetSize != "" {
 		sizeInBytes, err = humanize.ParseBytes(app.pruneTargetSize)
 		if err != nil {
-			return errors.Wrap(err, "parse prune target size")
+			return errors.Wrapf(err, "parse prune target size (%s)", app.pruneTargetSize)
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


cmd/earthly/root_cmds.go
```diff
@@ -173,6 +173,16 @@ func (app *earthlyApp) rootCmds() []*cli.Command {
 					Usage:       "Reset cache entirely by wiping cache dir",
 					Destination: &app.pruneReset,
 				},
+				&cli.StringFlag{
+					Name:        "target-size",
+					Usage:       "Prune cache to specified size, starting from oldest",
+					Destination: &app.pruneTargetSize,
+				},
+				&cli.DurationFlag{
+					Name:        "keep-duration",
```
alexcb, line 182: this seems easier to type:
```suggestion
					Name:        "age",
```

alexcb (approved):
looks good; just a few simplifications.

cmd/earthly/root_cmds.go
```diff
@@ -173,6 +173,16 @@ func (app *earthlyApp) rootCmds() []*cli.Command {
 					Usage:       "Reset cache entirely by wiping cache dir",
 					Destination: &app.pruneReset,
 				},
+				&cli.StringFlag{
```
nelsam, line 176: I would rather create a flag value type for this size value:
```go
type byteSizeValue uint64

func (byteSizeValue) Create(val uint64, p *uint64, _ cli.NoConfig) cli.Value {
	*p = val
	return (*byteSizeValue)(p)
}

func (byteSizeValue) ToString(b uint64) string {
	return humanize.Bytes(b)
}

func (s *byteSizeValue) Set(s string) error {
	v, err := humanize.ParseBytes(s)
	if err != nil {
		return err
	}
	*s = byteSizeValue(v)
	return nil
}

func (s *byteSizeValue) Get() any { return uint64(*s) }

func (s *byteSizeValue) String() string { return humanize.Bytes(*s) }
```
And then to use it:
```go
&cli.FlagBase[uint64, cli.NoConfig, byteSizeValue]{
    Name: "size",
    Usage: "Prune cache to specified size, starting from oldest",
    Destination: &app.pruneTargetSize, // NOTE: pruneTargetSize would change to a uint64
}
```
This allows the `cli` library to know when the flag failed to parse, which _ideally_ means that it will be more portable and forward compatible and all that.

nelsam (approved):
I just have the one minor suggestion about CLI flag stuff. I don't think it's worth blocking a merge though.

cmd/earthly/root_cmds.go
dchw: Turns out you don't have to go that hard. They have a `GenericFlag` that works well enough with less code on our end.

PR #2825, opened 2023-03-27
Allow the user to self-serve themselves onto Earthly CI

vladaionescu:


docs/cloud/satellites.md
```diff
@@ -73,9 +73,6 @@ You can start your free trial by using the checkout form below. Be sure to provi
 
 [**Click here to start your subscription**](https://buy.stripe.com/8wM9Es4BT4Vvb4YbIJ)
 
-Once you've submitted your details on the checkout page, you will receive an automated confirmation email.
```
thtmnisamnstr, line 76: Is this fully automated now? Does the user not have to manually send an email to support for features to be enabled?

vladaionescu: Yes - there's a field in the checkout page that the user can fill out.

PR #2829, opened 2023-03-28
Allow overriding git branch

r-LaForge:
Fixes https://github.com/earthly/cloud-issues/issues/11

When triggering a build from CI we pass along a commit SHA because we are running the build against a specific commit.
However, there is a desire to pass along the branch if the build is triggered from activity on a branch (push / pr).
Specifically, we'd like the `EARTHLY_GIT_BRANCH` ARG to be populated and set the `EARTHLY_TARGET_TAG_DOCKER` ARG to the branch.

The plan is for our CI runners to pass this along for the appropriate trigger type.

I'm hoping to get a sanity check before I continue further.

Open to other ideas/suggestions on how to tackle this problem.


buildcontext/git.go
```diff
@@ -150,17 +150,18 @@ func (gr *gitResolver) resolveEarthProject(ctx context.Context, gwClient gwclien
...
-			CommitterTimestamp: rgp.committerTs,
-			AuthorTimestamp:    rgp.authorTs,
-			Author:             rgp.author,
-			CoAuthors:          rgp.coAuthors,
+			BaseDir:              "",
+			RelDir:               subDir,
+			RemoteURL:            gitURL,
+			Hash:                 rgp.hash,
+			ShortHash:            rgp.shortHash,
+			BranchOverrideTagArg: gr.gitBranchOverride != "",
```
r-LaForge, line 158: As per [the comment here](https://github.com/earthly/cloud-issues/issues/11#issuecomment-1467308267) we'd like to override the EARTHLY_TARGET_TAG and EARTHLY_TARGET_TAG_DOCKER ARGS if the action is triggered from a branch.

This felt like the cleanest way to pass it along, but open to other suggestions.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to [Earthly](https://github.com/earthly/earthly) will be doc
 
 ## Unreleased
 
+### Changed
+
+- The `VERSION --git-branch` feature flag has been removed since it was always included anyway.
```
r-LaForge, line 9: You'll see below in `variables/builtin.go` where the branch was set twice.

nelsam (approved):

PR #2830, opened 2023-03-28
introduce LET - a new command to declare non-arg variables

### LINKED ISSUE #2808: Proposal: Non-ARG variables
# Problem statement

As a developer writing complicated targets in an Earthfile
I want to define variables that are not controlled by user flags
So that I can write cleaner Earthfile code without causing confusion to CLI users of my targets.

# Description

With the advent of both `earthly doc` (which in the near future will document `ARG`s that are available to a target) and `SET`, we're coming up on a source of notable confusion:

1. Engineers often like to use variables to keep code clean
2. `ARG` is currently the only method of creating variables
3. All `ARG`s will show up as flags in `earthly doc` output
4. `SET` can be used to override CLI flags from end users

If I were writing an `Earthfile` using `VERSION --arg-scope-and-set 0.7`, there's a good chance that I'd have a few `ARG`s that I'm using for cleaner code, not for end users to actually use as flags. `SET` makes this even more tempting, because I don't have to risk a user overriding my variable from the CLI. For example:

```Earthfile
VERSION --arg-scope-and-set 0.7
FROM alpine

# test-golden runs a golden test
test-golden:
    ARG --required goldenFile
    ARG expected
    SET expected=$(cat $goldenFile)
    ARG actual
    SET actual=$(someIntegrationCommand)
    IF $actual != $expected
        RUN "golden test failed! diff:"
        RUN diff $actual $expected
        RUN exit 1
    END
```

In this scenario, the end user would see the following output from `earthly doc`:

```
TARGETS:
  +test-golden --goldenFile [--expected] [--actual]
      test-golden runs a golden test
```

When the end user calls the target using `earthly +test-golden --goldenFile=somefile.golden --expected=$(cat updatedOutput.txt)`, the user's `--expected` flag will be completely ignored, to the point that earthly will even use cached output from previous runs. If I were an end user, this would be quite confusing.

# Proposal

A new command, e.g. `LET` or `VAR`, which defines a variable that does not get overridden by user input and does not show up in `earthly doc` output, but otherwise works the same as `ARG`.

Thus, the example changes to:

```Earthfile
VERSION --arg-scope-and-set 0.7
FROM alpine

test-golden:
    ARG --required goldenFile
    LET expected=$(cat $goldenFile)
    LET actual=$(someIntegrationCommand)
    IF $actual != $expected
        RUN "golden test failed! diff:"
        RUN diff $actual $expected
        RUN exit 1
    END
```

Since I don't have to worry about end users trying to override my `expected` and `actual`, I no longer need to use `SET` to override CLI flags. And since `LET` does not impact `earthly doc` output, the end users will never think that they can pass `--expected` to the target and get it unexpectedly overridden.

Note that it is still on the author of the `Earthfile` to actually _use_ `LET` instead of `ARG` - this doesn't _prevent_ an engineer from creating confusing targets. It just provides a tool for engineers to author cleaner and clearer targets.

nelsam:
This introduces the LET command and adjusts the SET command so that it only works on variables declared with LET.

Variables declared with LET are considered to have a narrower scope than variables declared with ARG, so these variables may be redeclared in order to be assignable with LET. This allows, for example,  to promote an ARG to a local variable.

Resolves #2808


earthfile2llb/interpreter.go
```diff
@@ -1433,6 +1436,34 @@ func (i *Interpreter) handleArg(ctx context.Context, cmd spec.Command) error {
...
 
+func (i *Interpreter) handleLet(ctx context.Context, cmd spec.Command) error {
+	if i.pushOnlyAllowed {
+		return i.pushOnlyErr(cmd.SourceLocation)
+	}
+	var opts letOpts
+	argsCpy := getArgsCopy(cmd)
+	args, err := parseArgs("LET", &opts, argsCpy)
+	if err != nil {
+		return errors.Wrap(err, "failed to parse SET args")
```
r-LaForge, line 1447: very nit
```suggestion
		return errors.Wrap(err, "failed to parse LET args")
```

earthfile2llb/interpreter.go
```diff
@@ -1433,6 +1436,34 @@ func (i *Interpreter) handleArg(ctx context.Context, cmd spec.Command) error {
...
+	}
+	if len(args) != 3 || args[1] != "=" {
+		return hint.Wrap(errInvalidSyntax, "LET requires a variable assignment, e.g. LET foo = bar")
+	}
```
r-LaForge, line 1451: I'm just wondering what the reasoning is for disallowing a declaration without setting a value.
Since `LET` is pretty much a local, mutable, variable declaration it follows that one might expect `LET foo` to not through an error and perhaps default to empty.

earthfile2llb/interpreter.go
```diff
@@ -1433,6 +1436,34 @@ func (i *Interpreter) handleArg(ctx context.Context, cmd spec.Command) error {
...
+
+	key := args[0]
+	baseVal := args[2]
+	val, err := i.expandArgs(ctx, baseVal, true, false)
+	if err != nil {
+		return i.wrapError(err, cmd.SourceLocation, "failed to expand LET value %q", baseVal)
```
r-LaForge, line 1457: nit: If not appropriate, feel free to ignore or click "resolved".

My apologies as it's my first time looking at the interpreter code - how would this show up in a failed run?
Perhaps we should include the key here as well?

```suggestion
		return i.wrapError(err, cmd.SourceLocation, "failed to expand LET value %q for key %q", baseVal, key)
```

earthfile2llb/interpreter.go
```diff
@@ -1433,6 +1436,34 @@ func (i *Interpreter) handleArg(ctx context.Context, cmd spec.Command) error {
...
+	}
+
+	err = i.converter.Let(ctx, key, val)
+	if err != nil {
+		return i.wrapError(err, cmd.SourceLocation, "apply ARG")
```
r-LaForge, line 1462: nit
```suggestion
		return i.wrapError(err, cmd.SourceLocation, "apply LET")
```

tests/Earthfile
```diff
@@ -1020,29 +1021,33 @@ arg-redeclare-error:
...
+    DO +RUN_EARTHLY --earthfile=let-scope.earth --target=+base
+    DO +RUN_EARTHLY --earthfile=let-scope.earth --target=+let-overrides-arg
+    DO +RUN_EARTHLY --earthfile=let-scope.earth --target=+let-overrides-global-arg
+    DO +RUN_EARTHLY --earthfile=let-scope.earth --target=+arg-after-let-errors --should_fail=true --output_contains="Hint: 'bar' was already declared with LET and cannot be redeclared as an ARG"
+
+let-set:
+    DO +RUN_EARTHLY --earthfile=let-set.earth --target=+base
+    DO +RUN_EARTHLY --earthfile=let-set.earth --target="+test --expected=bazinga"
+    DO +RUN_EARTHLY --earthfile=let-set.earth --target="+test --bar=eggs --expected=eggsinga"
+    DO +RUN_EARTHLY --earthfile=let-set.earth --target="+test-global --foo=banana --topExpected='banana car' --expected='banana phone'"
```
r-LaForge, line 1042: 🍌

r-LaForge (approved):
This is my first time reviewing our parser code, but from what I can tell this looks good 👍  
I just had a few nits and 1 question.

earthfile2llb/interpreter.go
nelsam: I was basically trying to match the error messages from `ARG` here. The nice thing about `i.wrapError` is that it will add the file name, line number, and column number that the error occurred on - so it will be pretty obvious which `LET` line failed to parse. Including the key might be nice, but feels largely unnecessary since the line number is included already (and I'd like to stay consistent with `ARG`).

earthfile2llb/converter.go
```diff
@@ -2404,15 +2434,19 @@ func (c *Converter) checkAllowed(command cmdType) error {
...
 			return errors.New("the first command has to be FROM, FROM DOCKERFILE, LOCALLY, ARG, BUILD or IMPORT")
 		}
 	}
 
-	if command == setCmd && !c.ftrs.ArgScopeSet {
-		return errors.New("--arg-scope-and-set must be enabled in order to use SET")
+	switch command {
+	case setCmd, letCmd:
+		if !c.ftrs.ArgScopeSet {
+			return errors.New("--arg-scope-and-set must be enabled in order to use SET")
```
vladaionescu, line 2447:
```suggestion
			return errors.New("--arg-scope-and-set must be enabled in order to use LET and SET")
```

vladaionescu (approved):
Looks great! Just a minor comment

Issue #2761, opened 2023-03-08
tighter timeout when attempting to send analytics


alexcb:
There is a 30 second timeout on the initial connection to the earthly server, which results in earthly hanging even when the server isn't needed.

alexcb:
see #2760 for a reproduction case.

Once this is fixed, we should adjust the test to lower the `45` second test to a lower value (e.g. `5` seconds).

PR #2835, opened 2023-03-31
Add http connection timeout

idodod:
fix https://github.com/earthly/earthly/issues/2761

This change sets a timeout to establish a http connection to the API server.
Note that the issue this fixes did not seem to relate to analytics, even with analytics disabled, Earthly was stuck on trying to ping the API server.

Also note that this sets a timeout on connection errors such as Deadline exceeded, for which we do not retry.
Other types of connection errors do get retried and as a result the overall time might be bigger than the set timeout.

### DIFF
--- a/cloud/client.go
+++ b/cloud/client.go
@@ -56,6 +56,7 @@ type Client struct {
 	requestID                string
 	installationName         string
 	logstreamAddressOverride string
+	serverHTTPTimeout        int
 }
 
 type ClientOpt func(*Client)
@@ -67,16 +68,17 @@ func WithLogstreamGRPCAddressOverride(address string) ClientOpt {
 }
 
 // NewClient provides a new Earthly Cloud client
-func NewClient(httpAddr, grpcAddr string, useInsecure bool, agentSockPath, authCredsOverride, authJWTOverride, installationName, requestID string, warnFunc func(string, ...interface{}), opts ...ClientOpt) (*Client, error) {
+func NewClient(httpAddr, grpcAddr string, useInsecure bool, agentSockPath, authCredsOverride, authJWTOverride, installationName, requestID string, warnFunc func(string, ...interface{}), serverHTTPTimeout int, opts ...ClientOpt) (*Client, error) {
 	c := &Client{
 		httpAddr: httpAddr,
 		sshAgent: &lazySSHAgent{
 			sockPath: agentSockPath,
 		},
-		warnFunc:         warnFunc,
-		jum:              &protojson.UnmarshalOptions{DiscardUnknown: true},
-		installationName: installationName,
-		requestID:        requestID,
+		warnFunc:          warnFunc,
+		jum:               &protojson.UnmarshalOptions{DiscardUnknown: true},
+		installationName:  installationName,
+		requestID:         requestID,
+		serverHTTPTimeout: serverHTTPTimeout,
 	}
 	if authJWTOverride != "" {
 		c.authToken = authJWTOverride
--- a/cloud/client.go
+++ b/cloud/client.go
@@ -56,7 +56,7 @@ type Client struct {
 	requestID                string
 	installationName         string
 	logstreamAddressOverride string
-	serverHTTPTimeout        int
+	serverConnTimeout        int
 }
 
 type ClientOpt func(*Client)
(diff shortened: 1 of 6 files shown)


tests/Earthfile
```diff
@@ -1393,7 +1393,7 @@ test-works-without-earthly-server:
     " > /tmp/precommand && chmod +x /tmp/precommand
     DO +RUN_EARTHLY --verbose=false --pre_command=/tmp/precommand --post_command=" && date +%s > end-time" --earthfile=true.earth --target=+true
     RUN if grep "failed sending analytics" earthly.output; then echo "analytics should have failed silently, but didnt" && exit 1; fi
-    RUN export duration=$(echo "$(cat end-time) - $(cat start-time)" | bc) && test "$duration" -lt 45 || (echo "it took $duration seconds to run +true, which is too long" && exit 1)
+    RUN export duration=$(echo "$(cat end-time) - $(cat start-time)" | bc) && test "$duration" -lt 20 || (echo "it took $duration seconds to run +true, which is too long" && exit 1)
```
idodod, line 1396: this takes about 5-6 seconds on my local, but saw it failed with 12 seconds in the CI, not sure why.

cloud/client.go
```diff
@@ -56,6 +56,7 @@ type Client struct {
 	requestID                string
 	installationName         string
 	logstreamAddressOverride string
+	serverConnTimeout        int
```
alexcb, line 59: can we make this a duration here instead?
```suggestion
	serverConnTimeout        time.Duration
```

cloud/client.go
```diff
@@ -67,16 +68,17 @@ func WithLogstreamGRPCAddressOverride(address string) ClientOpt {
 }
 
 // NewClient provides a new Earthly Cloud client
-func NewClient(httpAddr, grpcAddr string, useInsecure bool, agentSockPath, authCredsOverride, authJWTOverride, installationName, requestID string, warnFunc func(string, ...interface{}), opts ...ClientOpt) (*Client, error) {
+func NewClient(httpAddr, grpcAddr string, useInsecure bool, agentSockPath, authCredsOverride, authJWTOverride, installationName, requestID string, warnFunc func(string, ...interface{}), serverConnTimeout int, opts ...ClientOpt) (*Client, error) {
```
alexcb, line 71: passing as a Duration makes it more clear (since you don't need to know what the unit is)
```suggestion
func NewClient(httpAddr, grpcAddr string, useInsecure bool, agentSockPath, authCredsOverride, authJWTOverride, installationName, requestID string, warnFunc func(string, ...interface{}), serverConnTimeout time.Duration, opts ...ClientOpt) (*Client, error) {
```

cmd/earthly/flags.go
```diff
@@ -217,6 +217,14 @@ func (app *earthlyApp) rootFlags() []cli.Flag {
 			Destination: &app.buildID,
 			Hidden:      true, // Internal.
 		},
+		&cli.IntFlag{
```
alexcb, line 220: Can we use a duration instead?
```suggestion
		&cli.DurationFlag{
```

tests/Earthfile
alexcb: woohoo! this is a great improvement.

alexcb (commented):
nice improvement; can you try to use `time.Duration` wherever possible? that way a user can give us a value like `500ms` or `10s`.

cloud/http.go
```diff
@@ -201,7 +201,7 @@ func (c *Client) doCallImp(ctx context.Context, r request, method, url, reqID st
 	client := &http.Client{
 		Transport: &http.Transport{
 			DialContext: (&net.Dialer{
-				Timeout: time.Duration(c.serverConnTimeout) * time.Second,
+				Timeout: c.serverConnTimeout * time.Second,
```
alexcb, line 204: should this be:
```suggestion
				Timeout: c.serverConnTimeout,
```
instead?

cmd/earthly/flags.go
```diff
@@ -217,9 +217,9 @@ func (app *earthlyApp) rootFlags() []cli.Flag {
 			Destination: &app.buildID,
 			Hidden:      true, // Internal.
 		},
-		&cli.IntFlag{
+		&cli.DurationFlag{
 			Name:        "server-conn-timeout",
-			Usage:       "The time to wait for the http client to establish a connection to the API server",
+			Usage:       "The time to wait in seconds for the http client to establish a connection to the API server",
```
alexcb, line 222: this usage seems out of date now.
```suggestion
			Usage:       "Earthly API server connection timeout value",
```

I have no idea if urfave adds on extra context regarding the supported unit suffixes, if not maybe add on something like

`values should end with ms, or s` (technically it could also be an `h` or longer :laughing:

alexcb (commented):
looking better, just a couple remaining issues.

alexcb (approved):

PR #2837, opened 2023-04-01
Include recently added keywords

rrjjvv:
Caveat: I'm not a vscode user.  But I am an intellij user, and these change work as intended, so I assume they do in vscode as well.

Some of them are experimental/beta, but they are all publicly documented.


rrjjvv:
Duplicate of #2836; somebody beat me to it and I didn't see it until after I created this one.

Issue #2834, opened 2023-03-30
Earthly does not fully qualify buildkitd image by default


mcheshkov:
Earthly does not qualify default buildkitd image, and thus does not work on default podman configuration.

On fresh install on Ubuntu 22.04 I've got this:
```
$ podman -v
podman version 3.4.4
$ earthly -v
earthly version v0.7.2 e6fed708700425fc2616bd1e7a0f202d414f5f23 linux/amd64; Ubuntu 22.04.2 LTS (Jammy Jellyfish)
$ earthly bootstrap
           buildkitd | Starting buildkit daemon as a podman container (earthly-buildkitd)...
           bootstrap | Warning: Bootstrapping buildkit failed: maybe start buildkitd: start: could not start buildkit: 1 error occurred:
           bootstrap | 	* command failed: podman run --privileged --env BUILDKIT_DEBUG=false --env BUILDKIT_TCP_TRANSPORT_ENABLED=true --env BUILDKIT_TLS_ENABLED=false --env BUILDKIT_MAX_PARALLELISM=20 --label dev.earthly.settingshash=a3f07bdb0e8c8209 --mount type=volume,source=earthly-cache,dst=/tmp/earthly --publish 127.0.0.1:8371:8371/tcp --publish 127.0.0.1:8372:8372/tcp -d --pull missing --name earthly-buildkitd --security-opt unmask=/sys/fs/cgroup earthly/buildkitd:v0.7.2: exit status 125: Error: short-name "earthly/buildkitd:v0.7.2" did not resolve to an alias and no unqualified-search registries are defined in "/etc/containers/registries.conf": exit status 125
           bootstrap | 
           bootstrap | Bootstrapping successful.
$ cat /etc/containers/registries.conf | grep unqualified-search-registries -B1
# # An array of host[:port] registries to try when pulling an unqualified image, in order.
# unqualified-search-registries = ["example.com"]
```

I believe #1239 is intended to fix it, but #1227 messed with it here:
https://github.com/earthly/earthly/blob/c140e9cb76240c3b572c51d16a1f888ca290727c/release/Earthfile#L99

vladaionescu:
Thanks for raising and narrowing down to the problem. We'll fix and investigate why this wasn't caught by our testing.

mcheshkov:
I've peeked into recent CI runs, looks like podman in CI is configured with unqualified search registries 

https://github.com/earthly/earthly/actions/runs/4556044358/jobs/8040989080
[...]

PR #2842, opened 2023-04-04
.github: remove podman registries

nelsam:
Since most users will not have 'docker.io' added to podman's search registries by default, we shouldn't have it in our CI. This removes the registry file that GHA adds so that we can prove that earthly's release fully qualifies the buildkit image.

This is the first part for solving #2834. We want to make sure that our release script fails when our buildkitd image isn't fully qualified before we fix the issue.


alexcb (approved):

alexcb:
> We want to make sure that our release script fails when our buildkitd image isn't fully qualified before we fix the issue.

It's also possible to trigger the staging release job by naming your branch with a `-staging-test` suffix; see https://github.com/earthly/earthly/blob/main/.github/workflows/ci-staging-deploy.yml#L7

nelsam:
> > We want to make sure that our release script fails when our buildkitd image isn't fully qualified before we fix the issue.
> 
> It's also possible to trigger the staging release job by naming your branch with a `-staging-test` suffix; see https://github.com/earthly/earthly/blob/main/.github/workflows/ci-staging-deploy.yml#L7

oh nice!

PR #2844, opened 2023-04-04
fully qualified earthly-buildkitd

### LINKED ISSUE #2834: Earthly does not fully qualify buildkitd image by default

nelsam:
Since most users will not have 'docker.io' added to podman's search registries by default, we need to fully qualify the earthly-buildkitd image by default.

Resolves #2834

### DIFF
--- a/.github/workflows/reusable-bootstrap-integrations.yml
+++ b/.github/workflows/reusable-bootstrap-integrations.yml
@@ -47,7 +47,7 @@ jobs:
         run: ${{inputs.SUDO}} apt-get purge docker-engine docker docker.io docker-ce docker-ce-cli ; ${{inputs.SUDO}} rm -rf /usr/bin/docker
         if: inputs.binary == 'podman'
       - name: Install Podman (with apt-get)
-        run: ${{inputs.SUDO}} apt-get update && ${{inputs.SUDO}} apt-get install -y podman
+        run: ${{inputs.SUDO}} apt-get update && ${{inputs.SUDO}} apt-get install -y podman && ${{inputs.SUDO}} rm -f /etc/containers/registries.conf
         if: inputs.binary == 'podman'
       - name: Podman debug info
         run: podman version && podman info && podman info --debug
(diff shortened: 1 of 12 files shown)


nelsam:
@alexcb two minor notes for your information in this PR:

1. The `Docker CI Ubuntu` failure is due to unquoted variables in a shell script. This is fixed in the `go1.20` PR, so maybe we should merge that one first? or not, if you're not too worried.
2. Since the latest earthly is broken on podman without registries, and since the `Build latest earthly using released earthly` step uses the same service that is being tested (podman, in this case), the podman tests can't pass until there's a `released earthly` with the fully qualified image in place. This feels ... brittle. But I don't know how we would go about fixing that, other than configuring earthly to use docker for building the latest earthly, then reconfigure it to use podman before the tests. Thoughts?

alexcb (approved):

PR #2850, opened 2023-04-06
Allow `COPY --if-exists` to work with target sources

### LINKED ISSUE #2541: COPY --if-exists +target/file works strange
version: `earthly version 0.6.30 743687991d227f2f7e2e6ba11a90f443b4f99c67 linux/amd64; Arch Linux`
___
### Problem:
```Earthfile
COPY --if-exists +get-file/file ./file
```
doesn't copy a file even if target `+get-file` saves it as artifact.

___
### Full example:
```Earthfile
VERSION 0.6
FROM alpine:latest

get-file:
  RUN echo "hi" > ./file
  SAVE ARTIFACT ./file ./file

main:
  COPY --if-exists +get-file/file ./file
  RUN --no-cache cat ./file
  # cat: can't open './file': No such file or directory
```

rrjjvv:
Fixes #2541 

Given source `+target/data`, the current code translates that to wildcard `[/]data`, which never matches.  This translates it to `[d]ata` (analogous to how it works for classical copies).

I'm not sure what I'm supposed to "carefully look for" in regards to the AST changes (or if I was even right to generate them); the activity on #1557 (a very similar issue/solution) was my guide.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 *.interp
 .antlr/
 .env
+.arg
 .vscode
 .idea
 .tmp-earthly-out*
--- a/ast/tests/if-exists.ast.json
+++ b/ast/tests/if-exists.ast.json
@@ -361,6 +361,57 @@
           }
         }
       ]
+    },
+    {
+      "name": "artifact-copy-exists",
+      "recipe": [
+        {
+          "command": {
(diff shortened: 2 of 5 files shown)


alexcb (approved):
nice work!

Issue #2823, opened 2023-03-24
`Could not detect digest for image ... Please update your buildkit installation` warning


alexcb:
```Earthfile
VERSION 0.7

db:
    FROM postgres:15.2-alpine
    ENV POSTGRES_PASSWORD=postgres
    SAVE IMAGE my/db

test:
    FROM earthly/dind:alpine
    WITH DOCKER \
        --load my/db=+db
        RUN docker images
    END
```

produces:

```
 earthly/dind:alpine | --> Load metadata earthly/dind:alpine linux/amd64
postgres:15.2-alpine | --> Load metadata postgres:15.2-alpine linux/amd64
                 +db | --> FROM postgres:15.2-alpine
                 +db | [----------] 100% resolve docker.io/library/postgres:15.2-alpine@sha256:1f86ede0903f60ecd2eb630b15803567324da7aa0d1f7bbc3a8f1fe5247a4592
              output | --> exporting outputs
              output | [----------] 100% exporting layers
              output | [----------] 100% exporting manifest sha256:7bf5e755353d2a225c7ecdbba8b8d6751d84392692e601af6ff6904d8caac2f7
              output | [----------] 100% exporting config sha256:1347e065b3b125bfbd8ffed262ccb337918fa84975d57145de4a5a08c41b0b34
              output | [          ]   0% transferring docker.io/my/db:latest
Warning: Could not detect digest for image my/db:latest. Please update your buildkit installation.
               +test | --> FROM earthly/dind:alpine
               +test | [----------] 100% resolve docker.io/earthly/dind:alpine@sha256:9b09804c68a2c68195b71e61a09616657b8f0a5ca01b4e029f710c7ca2396dc6
               +test | *cached* --> WITH DOCKER (install deps)
               +test | *cached* --> WITH DOCKER RUN --privileged docker images
              output | [----------] 100% transferring docker.io/my/db:latest
              output | --> exporting outputs
```

vladaionescu:
Last time this was happening it was a simple string interpolation issue like alpine:3.13 vs docker.io/alpine:3.13 around the code that maps the digest to the image. Maybe there are some leftover edge cases like that.

PR #2853, opened 2023-04-06
fix missing digest error for some docker.io images

alexcb:
Fixes the `Warning: Could not detect digest for image %s. Please update your buildkit installation.` error which prevented the digest from being loaded for image names of the form `docker.io/<user>/` (where as previously images were prefixed with `docker.io/library/<user>/`).


alexcb:
fixes #2823

alexcb:
It looks like `addRefToResult` gets passed an image name of the form `<user>/:<tag>`, then somewhere in the buildkit code, this is now being set as `docker.io/<user>/:<tag>`. I was hoping to pinpoint where this happens so we could cover all edge cases, but haven't found the spot. As a result I don't have a test for this fix.

vladaionescu (approved):
Good catch!

PR #2854, opened 2023-04-06
fix cleanup.sh endline

alexcb:


alexcb:
#2845 introduced a non-posix file, which is causing a linting error.

dchw (approved):

PR #2856, opened 2023-04-09
go.mod: bump to go 1.20

nelsam:
This updates earthly to go 1.20. There are a number of side effects of this change, including:

1. Bump the base image in our Earthfile to golang:1.20-alpine3.17.
    - This appears to also require us to install 'fuse-overlayfs' in our containers for tests to run properly. 
3. Bump onpar to v0.3.1
4. Bump hel to v0.4.4
5. Bump golangci to v1.52.2

Due to a breaking change in go 1.20, all onpar tests needed a new 'defer o.Run()' line after constructing the o.

Unrelated to go 1.20, there were a couple of other changes necessary due to recent changes in some of the code that we pull down without a specific version (i.e. tools that we just always install the "latest" version of):

1. examples/rust: bump to 1.68 since `cargo install cargo-chef` no longer works on 1.59
2. release/ami/install.sh: quote a variable that was previously unquoted


Earthfile
```diff
@@ -203,7 +203,7 @@ mocks:
 # unit-test runs unit tests (and some integration tests).
 unit-test:
     FROM +code
-    RUN apk add --no-cache --update podman
+    RUN apk add --no-cache --update podman fuse-overlayfs
```
alexcb, line 206: nice! I was wondering about this error I saw earlier, easy enough fix.

alexcb (approved):
:raised_hands:

PR #2859, opened 2023-04-10
Add AWS and MkDocs example; add devcontainer configuration

shepherdjerred:
This PR adds examples showing how to use Earthly with AWS OIDC/SSO, and MkDocs with live reloading. In addition, I added a basic devcontainer config with Docker + Earthly suitable for use with GitHub Codespaces.


examples/aws-sso/Earthfile
```diff
@@ -0,0 +1,85 @@
+VERSION 0.7
+FROM amazon/aws-cli
+ARG --global EARTHLY_CI
```
alexcb, line 3: should this default to `false` rather than the empty string?
```suggestion
ARG --global EARTHLY_CI=false
```

examples/aws-sso/Earthfile
```diff
@@ -0,0 +1,85 @@
...
+  END
+  RUN curl --location --fail --silent --show-error "https://awscli.amazonaws.com/awscli-exe-linux-$aws_architecture.zip" --output "awscliv2.zip" && \
+    unzip -q awscliv2.zip && \
+    ./aws/install && \
+    rm -rf awscliv2.zip aws
+
+LOGIN:
+  COMMAND
+  ARG --required sso_region
+  BUILD +login --sso_region=$sso_region
```
alexcb, line 84: is the `BUILD +login` required here? I see the `COPY` below references the same target. an explicit `BUILD` is typically only required for saving an artifact to the host (`AS LOCAL`) or pushing an image.

examples/aws-sso/Earthfile
```diff
@@ -0,0 +1,85 @@
...
+  IF [ "$EARTHLY_CI" = "false" ]
+    ARG --required sso_region
+    FROM +python
+    # https://github.com/benkehoe/aws-export-credentials
+    RUN pip install aws-export-credentials
+    BUILD +dev.login.open --sso_region=$sso_region
+    COPY +dev.login/aws /root/.aws
+    RUN aws-export-credentials --profile default --credentials-file-profile default
+    SAVE ARTIFACT /root/.aws/credentials /credentials
+  ELSE
```
alexcb, line 28: Should this validate `EARTHLY_CI` to be either `true` or `false`
```
ELSE IF [ "$EARTHLY_CI" = "true" ]

ELSE
  RUN echo "bad value for EARTHLY_CI" && exit 1
END
```

alexcb (commented):
Thanks for the additions!

Can you add an entry to the `+base` target of each example under https://github.com/earthly/earthly/blob/main/Earthfile#L718

i.e.
```
examples2:
  ...
  BUILD ./examples/aws-sso+base
  BUILD ./examples/mkdocs+base
```

that way we'll at least validate the examples are syntactically correct as part of our integration tests.

examples/aws-sso/Earthfile
shepherdjerred: I've had no problem with this being empty, but I have no problem setting it to false if you think that's better.

examples/aws-sso/Earthfile
shepherdjerred: +dev.login will save a local artifact `cache/` containing the AWS credentials, which prevents the need to log in again when the layer cache is invalidated, e.g. if you're changing the login code above you, or if one other base images get updated, you don't want to have to log in again. Because of this we need to use `BUILD` all the way down :/

examples/aws-sso/Earthfile
shepherdjerred: My assumption was that nobody would call this by setting `EARTHLY_CI` and would rather be using the `--ci` flag, which would mean `EARTHLY_CI` is always `true` or `false`. I can add explicit checking if you think it's useful.

examples/mkdocs/README.md
```diff
@@ -0,0 +1,7 @@
+# Material for MkDocs
+
+This example shows demonstrates [Material for MkDocs]() using Earthly, and includes a live reloading development enviorment with `mkdocs serve`
```
shepherdjerred, line 3: Oops forgot to put a link here

examples/aws-sso/.github/workflows/aws.yaml
```diff
@@ -0,0 +1,25 @@
...
+      - uses: actions/checkout@v3
+      - uses: earthly/actions-setup@v1.0.5
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v2.0.0
+        with:
+          role-to-assume: <your OIDC role ARN>
+          aws-region: <your AWS region>
+      - name: Start Development Instance
```
shepherdjerred, line 24: Forgot to change this when copy-pasting

shepherdjerred:
> Can you add an entry to the +base target of each example under

Which target should I call for the AWS example? Everything meaningful requires AWS creds.

Aside from that, I've addressed the feedback you gave!

examples/aws-sso/Earthfile
```diff
@@ -0,0 +1,87 @@
...
+    RUN mkdir -p /root/.aws/
+    RUN --secret AWS_ACCESS_KEY_ID --secret AWS_SECRET_ACCESS_KEY --secret AWS_SESSION_TOKEN printf "[default]\naws_access_key_id=$AWS_ACCESS_KEY_ID\naws_secret_access_key=$AWS_SECRET_ACCESS_KEY\naws_session_token=$AWS_SESSION_TOKEN\n" >> /root/.aws/credentials
+  ELSE
+    RUN echo "bad value for EARTHLY_CI" && exit 1
+  END
+  SAVE ARTIFACT /root/.aws/credentials /credentials
+
+dev.login.open:
+  ARG --required sso_region
+  LOCALLY
```
alexcb, line 38: since this is `LOCALLY`, can you include a test for deps? e.g.

```
RUN command -v python3 >/dev/null || (echo "python3 is needed locally" && exit 1)
```

just in case the user doesn't have python.

examples/aws-sso/Earthfile
alexcb: ah ha! I missed that part. Thanks for the explanation.

alexcb:
> Which target should I call for the AWS example

the `+base` target (all Earthfiles have an implicit base, which is at the top of the Earthfile).

```
  BUILD ./examples/aws-sso+base
```

examples/mkdocs/Earthfile
```diff
@@ -0,0 +1,56 @@
...
+
+image:
+  FROM +src
+  SAVE IMAGE docs
+
+dev:
+  LOCALLY
+  WITH DOCKER --load=+image
+    # Open a web browser once mkdocs is ready
+    RUN (curl --output /dev/null --silent --retry 30 --retry-delay 1 --retry-all-errors \
```
alexcb, line 46: might be good to include a similar check here:

```
RUN command -v python3 >/dev/null || (echo "python3 is needed locally" && exit 1)
RUN command -v curl >/dev/null || (echo "curl is needed locally" && exit 1)

```

brandonSc (approved):
Thanks for sharing this! I think AWS examples will be helpful for users who are interested in doing deployments in Earthly, especially with our new CI product.

alexcb:
Thanks for the PR; I rebased it, made a small Earthfile change to fix the failure, and merged via 19cc25b05a9052b868e9f23ef1710d5d24b1ad6e

PR #2864, opened 2023-04-11
bugfix: set TargetPlatform in ConvertOpt

alexcb:
Fixes regression introduced in 4b139f43db7de4bb801e5070d304a469ed4b758a which incorrectly set TargetPlatforms in the dockerui Config rather than keeping it in ConvertOpt. This resulted in implicitTargetPlatform being set to true, which caused the host platform (e.g. darwin) to be propigated to the target platform.


idodod (approved):
🎉

nelsam (approved):

PR #2874, opened 2023-04-14
Update Ruby on Rails example

idodod:
Seems like current version started failing so I updated the ruby and bundler versions


alexcb (approved):
might have broken to the recent dependabot update? thanks for fixing.

PR #2877, opened 2023-04-14
loosen the tests required for a pre-release

alexcb:
Due to flaky tests, our pre-release workflow is not running; this means we are no longer dogfooding our code in main via the ./earthly script.

If the code is merged into main; then we need to be dogfooding it -- even if it causes tests to fail -- otherwise we won't ever taste bad dogfood.


nelsam (approved):
Yep, if the code's in main and produces a build, it should be in use. Especially if it causes us pain.

PR #2882, opened 2023-04-17
variables: support builtins in --arg-scope-and-set

### LINKED ISSUE #2879: Builtin args not set when using `--arg-scope-and-set`
For example, change https://github.com/earthly/earthly/blob/6810b288a36d0b9853cdad293a4d85e195373970/tests/builtin-args.earth to use `VERSION --arg-scope-and-set 0.7` and the build will fail.

```text
  +builtin-args-test | --> RUN test "$EARTHLY_TARGET" == "+builtin-args-test"
  +builtin-args-test | ERROR Earthfile line 21:4
  +builtin-args-test |       The command
  +builtin-args-test |           RUN test "$EARTHLY_TARGET" == "+builtin-args-test"
  +builtin-args-test |       did not complete successfully. Exit code 1

============================= ❌ FAILURE [Build 🔧] ==============================
```

nelsam:
This also includes a very rudimentary first pass at unit tests for `variables.Collection`.

Resolves #2879

### DIFF
--- a/variables/collection_test.go
+++ b/variables/collection_test.go
@@ -1,0 +1,104 @@
+package variables_test
+
+import (
+	"testing"
+
+	"github.com/earthly/earthly/ast/spec"
+	"github.com/earthly/earthly/features"
+	"github.com/earthly/earthly/util/platutil"
+	"github.com/earthly/earthly/variables"
+	specs "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/poy/onpar"
+	"github.com/poy/onpar/expect"
+)
+
+func TestCollection(topT *testing.T) {
+	type testCtx struct {
+		expect   expect.Expectation
+		coll     *variables.Collection
+		features *features.Features
+	}
+	o := onpar.BeforeEach(onpar.New(topT), func(t *testing.T) testCtx {
+		expect := expect.New(t)
+		f, _, err := features.Get(&spec.Version{Args: []string{"0.7"}})
+		expect(err).To(not(haveOccurred()))
+		return testCtx{
+			expect:   expect,
+			features: f,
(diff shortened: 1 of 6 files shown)


variables/collection_test.go
```diff
@@ -0,0 +1,107 @@
...
+				},
+				Features: tc.features,
+			})
+			return tc
+		})
+
+		registerBaseSpecs(o)
+
+		o.Spec("non-ARG variables ignore builtin values", func(tc testCtx) {
+			// TODO: is this what we want? non-ARG variables are required to
```
alexcb, line 93: I think that's good; otherwise a user might be tempted to do something like:

```
LET EARTHLY_TARGET
SET EARTHLY_TARGET="this-shouldnt-work"
``

tests/builtin-args-scope.earth
alexcb, line 1: NIT: any chance we can use `sed` instead, e.g. https://github.com/earthly/earthly/blob/main/tests/shell-out/Earthfile#L38-L41 ? That'll reduce the chance of this file drifting from the existing tests/builtin-args.earth

alexcb (approved):
looks good once TODO comment is reworded.

variables/collection_test.go
nelsam: Thanks! I'm glad I wrote that `TODO`, I had forgotten that was a question I had  :sweat_smile:

tests/builtin-args-scope.earth
nelsam: It always feels weird to sed-replace a test like this. But your resoning makes sense. I just hope we remember to delete it when `--arg-scope-and-set` becomes the default (in `0.8`?)

PR #2888, opened 2023-04-20
Ensure certificates are generated when using podman

idodod:
Not sure if we _should_ use TLS for podman, but this is (hopefully) a way to keep the TLS default behavior aligned, whether podman is used or not.

The issue has been that we deduce whether to generate certificates according to the buildkit host, which for podman, is set later in the code, so I changed the condition to generate the certificates in case podman is used.

### DIFF
--- a/cmd/earthly/root_cmds.go
+++ b/cmd/earthly/root_cmds.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"fmt"
-	"github.com/earthly/earthly/util/containerutil"
 	"net/url"
 	"os"
 	"path"
@@ -24,6 +23,7 @@ import (
 	"github.com/earthly/earthly/domain"
 	"github.com/earthly/earthly/earthfile2llb"
 	"github.com/earthly/earthly/util/cliutil"
+	"github.com/earthly/earthly/util/containerutil"
 	"github.com/earthly/earthly/util/fileutil"
 	"github.com/earthly/earthly/util/termutil"
 )
--- a/cmd/earthly/root_cmds.go
+++ b/cmd/earthly/root_cmds.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"github.com/earthly/earthly/util/containerutil"
 	"net/url"
 	"os"
 	"path"
@@ -429,7 +430,7 @@ func (app *earthlyApp) bootstrap(cliCtx *cli.Context) error {
 		if err != nil {
 			return errors.Wrapf(err, "invalid buildkit_host: %s", app.buildkitHost)
 		}
-		if bkURL.Scheme == "tcp" && app.cfg.Global.TLSEnabled {
+		if (bkURL.Scheme == "tcp" || app.cfg.Global.ContainerFrontend == containerutil.FrontendPodmanShell) && app.cfg.Global.TLSEnabled {
 			root, err := cliutil.GetOrCreateEarthlyDir(app.installationName)
 			if err != nil {
 				return err


nelsam (changes requested):
This does solve the core problem after bootstrapping, but there's one other problem: in my existing `podman` setup (before re-bootstrapping), I get this unhelpful error:

```
Error: build new buildkitd client: add required client opts: caPath: path './certs/ca_cert.pem' does not exist
```

This isn't telling me that I need to re-run `earthly bootstrap`. Can we add a hint about it? I think you'll need to pass the `buildkit_frontend` to the `addRequiredOpts` function and then [add some extra error logic around here](https://github.com/earthly/earthly/blob/main/buildkitd/buildkitd.go#L951-L954):

```go
if bkFrontend == "podman" {
    return hint.Error(err, "podman now requires TLS certs - try re-running 'earthly bootstrap'")
}
```

idodod:
> ```go
> podman now requires TLS certs - try re-running 'earthly bootstrap'
> ```

Good point @nelsam,
I added the hint in the two paths that might be relevant for podman. You'd notice I didn't add the hint inside `addRequiredOpts` because that would have required passing the fe type to [PrintSatelliteInfo](https://github.com/earthly/earthly/blob/eddb58b6b08f0a90a08792768d54f33b412763b7/buildkitd/buildkitd.go#L983) as well, and I didn't really want to do that because it's not applicable for satellites. 
I can see how potentially adding another function that calls `addRequiredOpts` might get us to miss the hint in that flow, but I think I'm willing to compromise on that for now. What do you think?

Either way, do you mind checking the hint is printed as you expect?

nelsam:
> > ```go
> > podman now requires TLS certs - try re-running 'earthly bootstrap'
> > ```
> 
> Good point @nelsam, I added the hint in the two paths that might be relevant for podman. You'd notice I didn't add the hint inside `addRequiredOpts` because that would have required passing the fe type to [PrintSatelliteInfo](https://github.com/earthly/earthly/blob/eddb58b6b08f0a90a08792768d54f33b412763b7/buildkitd/buildkitd.go#L983) as well, and I didn't really want to do that because it's not applicable for satellites.

In that code path, I think it's reasonable to set a constant (e.g. `const SatelliteBuildkitFrontend = "docker"`) and just use that in `PrintSatelliteInfo`.

buildkitd/buildkitd.go
```diff
@@ -42,6 +43,9 @@ var (
 func NewClient(ctx context.Context, console conslogging.ConsoleLogger, image, containerName, installationName string, fe containerutil.ContainerFrontend, earthlyVersion string, settings Settings, opts ...client.ClientOpt) (*client.Client, error) {
 	opts, err := addRequiredOpts(settings, installationName, opts...)
 	if err != nil {
+		if fe.Config().Setting == containerutil.FrontendPodmanShell {
+			return nil, hint.Wrap(err, "podman now requires TLS certs - try re-running 'earthly bootstrap'")
+		}
 		return nil, errors.Wrap(err, "add required client opts")
```
nelsam, line 49:
```suggestion
		err = errors.Wrap(err, "add required client opts")
		if fe.Config().Setting == containerutil.FrontendPodmanShell {
			return nil, hint.Wrap(err, "podman now requires TLS certs - try re-running 'earthly bootstrap'")
		}
		return nil, err
```

buildkitd/buildkitd.go
```diff
@@ -96,6 +100,9 @@ func ResetCache(ctx context.Context, console conslogging.ConsoleLogger, image, c
 
 	opts, err := addRequiredOpts(settings, installationName, opts...)
 	if err != nil {
+		if fe.Config().Setting == containerutil.FrontendPodmanShell {
+			return hint.Wrap(err, "podman now requires TLS certs - try re-running 'earthly bootstrap'")
+		}
 		return errors.Wrap(err, "add required client opts")
```
nelsam, line 106:
```suggestion
		err = errors.Wrap(err, "add required client opts")
		if fe.Config().Setting == containerutil.FrontendPodmanShell {
			return hint.Wrap(err, "podman now requires TLS certs - try re-running 'earthly bootstrap'")
		}
		return err
```

nelsam (approved):
I think what you have mostly works, but we should probably keep the original `errors.Wrap` in there.

I would still _prefer_ to do this inside of `addRequiredOpts` because currently this hint will be issued for _any_ error in a podman config, even if it's not TLS-related. But I'm not super opposed to what you have here - it's pretty reasonable. Hints don't have to be 100% accurate.

---

One other alternative: add `var ErrMissingCert = errors.New("TLS certificate not found")` and then you can do:

```go
if errors.Is(err, ErrMissingCert) && fe.Config().Setting == containerutil.FrontendPodmanShell {
// ...
```

But as I said, what you have already is good enough.

buildkitd/buildkitd.go
idodod: thanks, I missed the fact I changed the wrapping.

idodod:
> I think what you have mostly works, but we should probably keep the original `errors.Wrap` in there.
> 
> I would still _prefer_ to do this inside of `addRequiredOpts` because currently this hint will be issued for _any_ error in a podman config, even if it's not TLS-related. But I'm not super opposed to what you have here - it's pretty reasonable. Hints don't have to be 100% accurate.
> 
> One other alternative: add `var ErrMissingCert = errors.New("TLS certificate not found")` and then you can do:
> 
> ```go
> if errors.Is(err, ErrMissingCert) && fe.Config().Setting == containerutil.FrontendPodmanShell {
> // ...
> ```
> 
> But as I said, what you have already is good enough.
> 
> [edit] you could even change the error in `makeTLSPath`, returning `errors.Wrapf(os.ErrNotExist, "path '%s' does not exist", path)` and then in your `if` statement use `if errors.Is(err, os.ErrNotExist)`.

Gave it another try, I figured maybe it's best to be more specific using a boolean, rather than pass the setting.

nelsam (approved):
lgtm :shipit:

Issue #2250, opened 2022-10-06
LOCALLY SAVE ARTIFACT does not invalidate cache


alexcb:
```Earthfile
VERSION 0.6
test:
    FROM alpine
    COPY +fromlocal/foo .
    RUN --no-cache cat foo

fromlocal:
    LOCALLY
    SAVE ARTIFACT some-file foo
```

```
$ echo hello > some-file
```

```
$ earthly +test 
 1. Init 🚀
————————————————————————————————————————————————————————————————————————————————

           buildkitd | Found buildkit daemon as docker container (earthly-buildkitd)

 2. Build 🔧
————————————————————————————————————————————————————————————————————————————————

              alpine | --> Load metadata linux/amd64
               +test | --> FROM alpine
               +test | [██████████] 100% resolve docker.io/library/alpine@sha256:bc41182d7ef5ffc53a40b044e725193bc10142a1243f395ee852a8d9730fc2ad
               +test | *cached* --> COPY +fromlocal/foo ./
               +test | --> RUN --no-cache cat foo
               +test | world                                    <------- cached from a different version of some-file
              output | --> exporting outputs

 3. Push ⏫ (disabled)
————————————————————————————————————————————————————————————————————————————————

To enable pushing use earthly --push

 4. Local Output 🎁
————————————————————————————————————————————————————————————————————————————————



========================== 🌍 Earthly Build  ✅ SUCCESS ==========================
```

brandonSc:
May be related to https://github.com/earthly/earthly/issues/2593

alexcb:
note that this may only occur when the file size of `some-file` stays constant (e.g. `hello\n` and `world\n` are both 6 bytes)

alexcb:
https://github.com/earthly/fsutil/blob/a1fafbb12147241e9e76ba8c1818983e063c0148/diff_containerd.go#L201 appears to skip comparing the contents when `differ == DiffMetadata`

alexcb:
under `client/llb/source.go`, there's the option to set two differs:

```
const (
	// DiffNone will do no file comparisons, all files in the Local source will
	// be retransmitted.
	DiffNone DiffType = pb.AttrLocalDifferNone
	// DiffMetadata will compare file metadata (size, modified time, mode, owner,
	// group, device and link name) to determine if the files in the Local source need
	// to be retransmitted.  This is the default behavior.
	DiffMetadata DiffType = pb.AttrLocalDifferMetadata
)
```

[...]

PR #2889, opened 2023-04-20
Acb/set differ

alexcb:


alexcb:
unfortunately this did not fix #2250

PR #2892, opened 2023-04-21
Initialize log streaming upfront if org/project envs set

mikejholly:
This PR will help ensure that logs are (almost) always sent to the log streaming service when streaming is enabled. It does so by looking for `EARTHLY_ORG` and `EARTHLY_PROJECT` and initializing the stream using those values when present. The remote end will verify that the currently logged-in user has permission to write logs for the given org/project combination.

### DIFF
--- a/cmd/earthly/build_cmd.go
+++ b/cmd/earthly/build_cmd.go
@@ -556,7 +556,7 @@ func (app *earthlyApp) actionBuildImp(cliCtx *cli.Context, flagArgs, nonFlagArgs
 			app.console.VerbosePrintf("Logbus: hit context deadline at %s (%s later)", now.Format(time.RFC3339Nano), now.Sub(beforeSelect))
 			return
 		case details := <-buildOpts.MainTargetDetailsFuture:
-			if app.logbusSetup.OrgAndProjectSet() {
+			if app.logbusSetup.LogStreamer.Started() {
 				app.console.VerbosePrintf("Organization and project already set via environmental")
 				return
 			}
@@ -578,7 +578,7 @@ func (app *earthlyApp) actionBuildImp(cliCtx *cli.Context, flagArgs, nonFlagArgs
 		}
 	}()
 
-	if app.logstream && doLogstreamUpload && !app.logbusSetup.OrgAndProjectSet() {
+	if app.logstream && doLogstreamUpload && !app.logbusSetup.LogStreamer.Started() {
 		app.console.ColorPrintf(color.New(color.FgHiYellow), "Streaming logs to %s\n\n", logstreamURL)
 	}
 	_, err = b.BuildTarget(cliCtx.Context, target, buildOpts)
--- a/cmd/earthly/build_cmd.go
+++ b/cmd/earthly/build_cmd.go
@@ -528,7 +528,6 @@ func (app *earthlyApp) actionBuildImp(cliCtx *cli.Context, flagArgs, nonFlagArgs
 		BuiltinArgs:                builtinArgs,
 		LocalArtifactWhiteList:     localArtifactWhiteList,
 		Logbus:                     app.logbus,
-		MainTargetDetailsFuture:    make(chan earthfile2llb.TargetDetails, 1),
 		Runner:                     runnerName,
 
 		// feature-flip the removal of builder.go code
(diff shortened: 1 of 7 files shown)


alexcb (commented):
seems reasonable; I can take a second look once the tests are closer to passing.

cmd/earthly/flags.go
```diff
@@ -323,6 +323,13 @@ func (app *earthlyApp) buildFlags() []cli.Flag {
 			Required:    false,
 			Destination: &app.orgName,
 		},
+		&cli.StringFlag{
+			Name:        "project",
+			EnvVars:     []string{"EARTHLY_PROJECT"},
+			Usage:       wrap("The name of the organization the satellite belongs to. ", "Required when using --satellite and user is a member of multiple organizations."),
```
vladaionescu, line 329: Update usage text? Seems that it was copy-pasted from `--org`.

cmd/earthly/flags.go
```diff
@@ -323,6 +323,13 @@ func (app *earthlyApp) buildFlags() []cli.Flag {
...
+			Required:    false,
+			Destination: &app.projectName,
+		},
```
vladaionescu, line 332: Should this flag be hidden? Is there a good use-case for the user to specify this themselves?

cmd/earthly/main.go
```diff
@@ -619,6 +620,25 @@ func (app *earthlyApp) run(ctx context.Context, args []string) int {
...
+		if cloudClient.IsLoggedIn(ctx) {
+			app.console.VerbosePrintf("Logbus: setting organization %q and project %q", app.orgName, app.projectName)
+			analytics.AddEarthfileProject(app.orgName, app.projectName)
+			app.logbusSetup.SetOrgAndProject(app.orgName, app.projectName)
+			app.logbusSetup.StartLogStreamer(ctx, cloudClient)
+			logstreamURL := fmt.Sprintf("%s/builds/%s", app.getCIHost(), app.logbusSetup.InitialManifest.GetBuildId())
+			app.console.ColorPrintf(color.New(color.FgHiYellow), "Streaming logs to %s\n\n", logstreamURL)
+		}
+	}
 	app.logbus.Run().SetStart(time.Now())
```
vladaionescu, line 642: I don't know if we have deltas for the start time - it might be better to initialize this field before the stream begins, such that it goes in the initial manifest.

logbus/setup/setup.go
```diff
@@ -68,6 +68,11 @@ func (bs *BusSetup) SetDefaultPlatform(platform string) {
 	bs.Formatter.SetDefaultPlatform(platform)
 }
 
+// OrgAndProjectSet returns true if the org and project have already been set.
+func (bs *BusSetup) OrgAndProjectSet() bool {
+	return bs.InitialManifest.GetOrgName() != "" && bs.InitialManifest.GetProjectName() != ""
+}
```
vladaionescu, line 74: The choice of these two specific fields seems very arbitrary to detect whether logstream has started. Should we instead have a `IsLogStreamerStarted` function?

cmd/earthly/build_cmd.go
```diff
@@ -553,16 +553,16 @@ func (app *earthlyApp) actionBuildImp(cliCtx *cli.Context, flagArgs, nonFlagArgs
...
-			app.console.VerbosePrintf(
-				"========== CONTEXT DONE BEFORE LOGSTREAMER STARTED AT %s (%s later) ==========",
-				now.Format(time.RFC3339Nano),
-				now.Sub(beforeSelect),
-			)
+			app.console.VerbosePrintf("Logbus: hit context deadline at %s (%s later)", now.Format(time.RFC3339Nano), now.Sub(beforeSelect))
 			return
 		case details := <-buildOpts.MainTargetDetailsFuture:
+			if app.logbusSetup.OrgAndProjectSet() {
+				app.console.VerbosePrintf("Organization and project already set via environmental")
```
vladaionescu, line 560: In this case, do we perhaps want to still verify that the org and project match?

vladaionescu (commented):
Looks pretty good! Just a few minor remarks...

cmd/earthly/flags.go
mikejholly: Good point. Updated.

cmd/earthly/flags.go
mikejholly: Updated.

cmd/earthly/main.go
mikejholly: Moved above the new block. Does that work?

vladaionescu: Yes, that should be good!

logbus/setup/setup.go
mikejholly: That's fair. It's not particularly representative of whether streaming has started. Updated.

cmd/earthly/build_cmd.go
mikejholly: I wasn't really sure of what the behavior ought to be. Should we fail the build or just print a message?

vladaionescu: I would fail the build in this case - I think it's better to fail early, rather than let the user get confused when the secrets don't work.

vladaionescu (approved):

cmd/earthly/flags.go
```diff
@@ -319,10 +319,18 @@ func (app *earthlyApp) buildFlags() []cli.Flag {
 		&cli.StringFlag{
 			Name:        "org",
 			EnvVars:     []string{"EARTHLY_ORG"},
-			Usage:       wrap("The name of the organization the satellite belongs to. ", "Required when using --satellite and user is a member of multiple organizations."),
+			Usage:       wrap("The name of the organization that the satellite belongs to. ", "Required when using --satellite and user is a member of multiple organizations."),
```
alexcb, line 322: should we set this to `Hidden: true,` to match the `project` flag?

I think this was overlooked in a previous PR.

mikejholly: I'm not sure, but I'm happy to update it if that's preferred.

alexcb: up to you... just something that jumped out.

alexcb (approved):

PR #2894, opened 2023-04-22
Earthfile: support -GO_GCFLAGS overrides for delve debugging

nelsam:
Since delve needs optimizations disabled in order to properly track things like variable and function names, line numbers, and so on, it needs binaries to be compiled with some flags in order to properly debug them. This allows those flags to be set and documents their use.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -51,7 +51,37 @@ For development purposes, you may use the built `earthly` binary to rebuild itse
     ```bash
     ./build/darwin/amd64/earthly +for-darwin-m1
     ```
-  
+
+## Delve
+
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -51,7 +51,37 @@ For development purposes, you may use the built `earthly` binary to rebuild itse
     ```bash
     ./build/darwin/amd64/earthly +for-darwin-m1
     ```
-  
+
+## Delve
+
+To use the delve debugger with the earthly binary, you need to disable optimizations in the 'go build' command. This is done using the -GO_GCFLAGS arg:
```
alexcb, line 57:
```suggestion
To use the [delve debugger](https://github.com/go-delve/delve) with the earthly binary, you need to disable optimizations in the 'go build' command. This is done using the -GO_GCFLAGS arg:
```

alexcb (approved):

PR #2895, opened 2023-04-22
bump buildkit to pull in host bind mount fixes

nelsam:


alexcb (approved):

nelsam:
Podman tests are still failing, but they're getting a lot farther now. We are at least building earthly in podman.

buildkitd/certificates.go
```diff
@@ -11,53 +11,160 @@ import (
...
+	return nil
+}
+
+func genCert(ca *certData, role, hostname, keyPath, certPath string) error {
+	key, err := parseTLSKey(keyPath)
+	if err != nil && !errors.Is(err, os.ErrNotExist) {
+		return errors.Wrapf(err, "could not parse %v TLS key", role)
+	}
+	if errors.Is(err, os.ErrNotExist) {
+		os.Remove(certPath)
```
alexcb, line 113: this feels a big dangerous, since a user might have typoed an existing key path, then their cert gets deleted.

what if instead we only generated a new key+cert when both keyPath and certPath don't exist?

alexcb (commented):
looks good; just one question.

alexcb (approved):

PR #2901, opened 2023-04-25
move command definitions to ast package

alexcb:
defines Earthfile command string values in a seperate subpackage to support future interpreter code changes (and reduce chances of typos).


nelsam (approved):
Looks good to me. As discussed in slack I have a mild preference for singular package names, rather than plural - so `command`>`commands` and `commandflag`>`commandflags`. But it's a mild preference and not necessary for merge.

PR #2904, opened 2023-04-26
fix podman tests

nelsam:
This is a follow-up to #2895, finally getting our podman tests green again. The two final problems:

1. We need to pass along the environment when using `sudo`, using `sudo -E`.
2. Due to changes in podman setup, we need to re-bootstrap after rebuilding earthly.

After those changes, we should :crossed_fingers: have green podman tests again.


alexcb (approved):
`sudo --excellent` :+1:

PR #2922, opened 2023-05-04
git-lfs support

alexcb:


alexcb:
```
$ ./build/linux/amd64/earthly github.com/earthly/lfs-test+large
 Init 🚀

 Build 🔧

             context | --> git context github.com/earthly/lfs-test+large
github.com/earthly/lfs-test+large | --> FROM alpine
github.com/earthly/lfs-test+large | [----------] 100% resolve docker.io/library/alpine@sha256:124c7d2707904eea7431fffe91522a01e5a861a624ee31d03372cc1d138a3126
github.com/earthly/lfs-test+large | --> COPY large-file .
github.com/earthly/lfs-test+large | --> RUN --no-cache cat large-file
github.com/earthly/lfs-test+large | pretend-this-is-really-big            <---- WOOHOO
```

and this file is in fact stored in LFS: https://github.com/earthly/lfs-test/blob/main/large-file

image

buildcontext/git.go
```diff
@@ -28,7 +28,7 @@ import (
 )
 
 const (
-	defaultGitImage = "alpine/git:v2.30.1"
+	defaultGitImage = "earthly/alpine-git-lfs:v2.36.3"
```
vladaionescu, line 31: How are we managing the release of this image? Does it go out together with other images like earthly/dind?

vladaionescu (approved):

buildcontext/git.go
alexcb: that would be a reasonable place to handle that; so far it's just a one-off I made via

```
VERSION 0.7

gitlfs:
  ARG VERSION=v2.36.3
  FROM alpine/git:$VERSION
  RUN apk add git-lfs
  SAVE IMAGE --push earthly/alpine-git-lfs:$VERSION
```

but obviously needs to be stored somewhere besides a github comment.

vladaionescu: Oh BTW this image needs to be cross-platform such that is also works natively on arm64 too. Let's establish a proper release practice for this image before merging this PR.

nelsam: Also fully qualified so it works on podman

alexcb: this change ended up not being needed, as remotely referenced targets are fetched via the git binary in the buildkit container.

nelsam (approved):

PR #2924, opened 2023-05-04
earthfile2llb: when entering a DO scope, ensure existing overrides are passed through

### LINKED ISSUE #2920: Global args get reset to defaults when chaining DO calls
In some scenarios, `DO +SOME_CMD` causes global ARGs that have been overridden at the CLI or in the `.arg` file to be reset to their default value. I haven't yet figured out what's going on - in one project, changing the `.arg` file fixed it. In another, I couldn't ever get it to fail. In a third, I couldn't ever get it to succeed.

```Earthfile
VERSION 0.7
FROM alpine

ARG --global foo = bar

BAZ:
    COMMAND
    RUN echo "in BAZ: $foo"

BAR:
    COMMAND
    RUN echo "in BAR: $foo"
    DO +BAZ

foo:
    RUN echo "in foo: $foo"
    DO +BAR
```

^^^ at leas right now, calling `earthly --no-cache +foo --foo=asdf` results in the following output:

```
$ earthly --no-cache +foo --foo=asdf
 Init 🚀

           buildkitd | Found buildkit daemon as podman container (earthly-dev-buildkitd)

 Build 🔧

               +base | --> FROM alpine
               +base | [          ]   0% resolve docker.io/library/alpine@sha256:124c7d2707904eea7431fffe91522a01e5a861a624ee31d03372cc1d138a3126                  [----------] 100% resolve docker.io/library/alpine@sha256:124c7d2707904eea7431fffe91522a01e5a861a624ee31d03372cc1d138a3126
               +base | [----------] 100% sha256:f56be85fc22e46face30e2c3de3f7fe7c15f8fd7c4e5add29d7f64b87abdaa09
               +base | [----------] 100% extracting sha256:f56be85fc22e46face30e2c3de3f7fe7c15f8fd7c4e5add29d7f64b87abdaa09
                +foo | foo=asdf
                +foo | --> RUN echo "in foo: $foo"
                +foo | in foo: asdf
                +foo | --> RUN echo "in BAR: $foo"
                +foo | in BAR: asdf
                +foo | --> RUN echo "in BAZ: $foo"
                +foo | in BAZ: bar

 Push Summary ⏫ (disabled)

 Local Output Summary 🎁



Shareable link: https://ci.earthly.dev/logs?logId=d34a1c6b-9ead-4b3e-95f6-7ddeedb9ce7c
```

Note the last echo: `in BAZ: bar`. This is not only ignoring my `--foo=asdf` flag at the CLI, it's also different from the previous echoes.

## Expected:

Global args would be consistent throughout the Earthfile.

## Actual:

Global args change their value between `DO` calls.

nelsam:
When calling `DO +SOME_CMD`, we were dropping the existing overrides. This meant that globals were (sometimes?) losing their overrides if any UDC called `DO +SOME_OTHER_CMD`.

This updates DO scopes to include existing overrides, while still letting the `DO` call's args take precedence.

Resolves #2920

### DIFF
--- a/tests/command-nested-global.earth
+++ b/tests/command-nested-global.earth
@@ -10,7 +10,8 @@ ARG --global foo=default
 
 SHARED:
     COMMAND
-    RUN test "$foo" == "expected"
+    # Note that foo has not yet been declared in the command, therefore we reference the globally declared arg
+    RUN test "$foo" == "expected" 
 
     # foo should be passed in as "baz" when DO +SHARED is called for the
     # purposes of this test. This ensures that DO args do not impact globals but
--- a/tests/command-nested-global.earth
+++ b/tests/command-nested-global.earth
@@ -1,0 +1,31 @@
+VERSION --arg-scope-and-set 0.7
+
+FROM alpine:3.15
+WORKDIR /test
+
+# foo should be overridden with the value "expected" to run this test.
+# This asserts that the overriding scope impacts the global scope through nested
+# DO calls.
+ARG --global foo=default
+
+SHARED:
+    COMMAND
+    RUN test "$foo" == "expected"
+
+    # foo should be passed in as "baz" when DO +SHARED is called for the
+    # purposes of this test. This ensures that DO args do not impact globals but
+    # do impact ARGs that are declared in the COMMAND.
+    ARG foo
+    RUN test "$foo" == "baz"
+
+CALLER:
+    COMMAND
+    RUN test "$foo" == "expected"
+
+    ARG foo = "bar"
+    RUN test "$foo" == "bar"
+    DO +SHARED --foo="baz"
+
+test-overriding:
+    RUN test "$foo" == "expected"
+    DO +CALLER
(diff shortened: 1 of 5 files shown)


nelsam:
@alexcb After struggling for a day to solve this in all contexts, I finally broke down and restricted it to `--arg-scope-and-set`. After all, that feature flag is focused on ARG scoping - and this bug is related to that, too.

The problem with solving this in the default case is that by default, `ARG`s try to look up their value from _any_ previously declared ARG of the same name, and previously declared `ARG`s take precedence over build flags.  This means that solving global args in the default case will break existing functionality. For example:

```Earthfile
VERSION 0.7
FROM alpine

ARG --global bar

BAR:
    COMMAND
    ARG bar
    RUN echo $bar

foo:
    ARG foo
    DO +BAR --bar=$foo
```

^^^ In the above code, the current logic will do what I would expect, so:

```
earthly +foo --foo=bacon --bar=eggs
```

^^^ That will print `bacon`.

However, without heavy restructuring of the `earthfile2llb.Converter`, solving the global args issue causes it to print `eggs` because previously declared variables take precedence over build args.

tests/command-nested-global.earth
```diff
@@ -0,0 +1,31 @@
...
+WORKDIR /test
+
+# foo should be overridden with the value "expected" to run this test.
+# This asserts that the overriding scope impacts the global scope through nested
+# DO calls.
+ARG --global foo=default
+
+SHARED:
+    COMMAND
+    RUN test "$foo" == "expected"
```
alexcb, line 13: I initially scratched my head as to why this would pass; I think it's worthwhile explicitly pointing this out:
```suggestion
    # Note that foo has not yet been declared in the command, therefore we reference the globally declared arg
    RUN test "$foo" == "expected" 
```

alexcb (approved):

PR #2927, opened 2023-05-04
fix: SAVE IMAGE --push flag was ignored

### LINKED ISSUE #2923: --push cmd line argument tries to push SAVE IMAGE step not marked with --push flag
Hi. I am using the most recent earthly v0.7.4

Consider the following minimal `Earthfile`:

```earthfile
VERSION 0.7
FROM alpine
WORKDIR /workdir

build:
    RUN echo hello > file.txt
    SAVE IMAGE hello:latest
```

The `earthly --push +build` will fail with the following error(even though `--push` IMO should not make any difference for `+build` target):

```
              output | [          ]   0% pushing layers
             ongoing | output (5 seconds ago)
              output | [----------] 100% pushing layers
              output | WARN: The command 'exporting outputs' failed: server message: insufficient_scope: authorization failed
             ongoing | Waiting on Buildkit... (Utilization: unknown other builds, 0/20 op load)
Share your logs with an Earthly account (experimental)! Register for one at https://ci.earthly.dev.
Error: build target: build main: failed to run Build function: failed to SAVE IMAGE: server message: insufficient_scope: authorization failed
```

The error says that the push failed. Earthly just tried to push `hello:latest` to dockerhub. If I prefix `hello:latest` with my dockerhub username, the push will succeed.

---

The reason why I assume the `--push` cmd line argument should not make any effect in this case, is because based on the [`earthly` cmd line reference](https://docs.earthly.dev/v/earthly-0.7/docs/earthly-command#earthly):

> `--push` ... Instructs Earthly to push __any docker images declared with the `--push`__ flag to remote docker registries and to run any `RUN --push` commands.

But my `SAVE IMAGE` step is not marked with `--push` flag.

The [`SAVE IMAGE` reference](https://docs.earthly.dev/v/earthly-0.6/docs/earthfile#save-image) also confirms my assumption:

> __The `--push` options marks the image to be pushed to an external registry__ after it has been loaded within the docker daemon available on the host.
> ... __The actual push is not executed by default. Add the `--push` flag to the earthly invocation to enable pushing.__

alexcb:
An image should only be pushed when the `SAVE IMAGE --push ...` command is used in combination with earthly being run with the --push flag on the cli; this fixes a regression where simply running `earthly --push` would cause all `SAVE IMAGE ...` commands to be pushed regardless of if the command had a `--push` or not.

Fixes #2923

### DIFF
--- a/earthfile2llb/converter.go
+++ b/earthfile2llb/converter.go
@@ -1022,7 +1022,7 @@ func (c *Converter) PopWaitBlock(ctx context.Context) error {
 }
 
 // SaveImage applies the earthly SAVE IMAGE command.
-func (c *Converter) SaveImage(ctx context.Context, imageNames []string, pushImages bool, insecurePush bool, cacheHint bool, cacheFrom []string, noManifestList bool) error {
+func (c *Converter) SaveImage(ctx context.Context, imageNames []string, hasPushFlag bool, insecurePush bool, cacheHint bool, cacheFrom []string, noManifestList bool) error {
 	err := c.checkAllowed(saveImageCmd)
 	if err != nil {
 		return err
@@ -1052,7 +1052,7 @@ func (c *Converter) SaveImage(ctx context.Context, imageNames []string, pushImag
 					State:               pcState,
 					Image:               c.mts.Final.MainImage.Clone(), // We can get away with this because no Image details can vary in a --push. This should be fixed before then.
 					DockerTag:           imageName,
-					Push:                pushImages,
+					Push:                hasPushFlag,
 					InsecurePush:        insecurePush,
 					CacheHint:           cacheHint,
 					HasPushDependencies: true,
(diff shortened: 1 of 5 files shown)


earthfile2llb/converter.go
```diff
@@ -1098,7 +1098,7 @@ func (c *Converter) SaveImage(ctx context.Context, imageNames []string, pushImag
 			c.mts.Final.SaveImages = append(c.mts.Final.SaveImages, si)
 		}
 
-		if pushImages && imageName != "" && c.opt.UseInlineCache {
+		if hasPushFlag && imageName != "" && c.opt.UseInlineCache {
```
vladaionescu, line 1101: These are just renames in this file, right?

earthfile2llb/save_image_wait_item.go
```diff
@@ -38,6 +39,6 @@ func (siwi *saveImageWaitItem) SetDoPush() {
 	siwi.mu.Lock()
 	defer siwi.mu.Unlock()
 	if siwi.si.DockerTag != "" {
-		siwi.push = true
+		siwi.doPush = siwi.allowPush
```
vladaionescu, line 42: This is the key change, right?

vladaionescu (approved):

earthfile2llb/converter.go
alexcb: correct -- just renaming the function arg to make it more clear that it represents a `SAVE IMAGE` vs `SAVE IMAGE --push`.

earthfile2llb/save_image_wait_item.go
alexcb: good eyes! yeah we were always setting it to `true` whenever a `SAVE IMAGE` was chained via a `BUILD`, and forgot if the `SAVE IMAGE` had a `hasPushFlag=true`.

PR #2932, opened 2023-05-05
Support changing size of satellites and CI runners

brandonSc:
The following command can now be used to alter the size of the satellite:
```
earthly satellite update --size <new-size> <existing-satellite>
```
This is important for CI runners, which currently have no other way to specify a size other than medium.

Changing size will also drop the existing cache. I will explain that in a separate docs PR.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,7 +15,7 @@ require (
 	github.com/docker/distribution v2.8.1+incompatible
 	github.com/docker/go-connections v0.4.0
 	github.com/dustin/go-humanize v1.0.1
-	github.com/earthly/cloud-api v1.0.1-0.20230505151542-8f55d3a49564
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -15,7 +15,7 @@ require (
 	github.com/docker/distribution v2.8.1+incompatible
 	github.com/docker/go-connections v0.4.0
 	github.com/dustin/go-humanize v1.0.1
-	github.com/earthly/cloud-api v1.0.1-0.20230316192913-d68000c20ba2
+	github.com/earthly/cloud-api v1.0.1-0.20230505151542-8f55d3a49564
```
brandonSc, line 18: I need to update this once cloud PR is merged.

alexcb (approved):

PR #2935, opened 2023-05-09
Compute a target hash for analytics purposes

vladaionescu:
Also, more correct repo hash, in case the target is remote.

### DIFF
--- a/analytics/analytics.go
+++ b/analytics/analytics.go
@@ -125,7 +125,7 @@ func isGitDir() bool {
 	return (err == nil)
 }
 
-func getRepo(target domain.Target) string {
+func getRepo(localRepo string, target domain.Target) string {
 	if target.Target != "" && target.IsRemote() {
 		repoAndPath := target.GitURL
 		// Note that this makes an assumption about the typical repo path structure
@@ -137,10 +137,10 @@ func getRepo(target domain.Target) string {
 		}
 		return strings.Join(repoAndPathSplit[0:3], "/")
 	}
-	return getLocalRepo()
+	return localRepo
 }
 
-func getTarget(repo string, target domain.Target) string {
+func getTarget(localRepo string, target domain.Target) string {
 	if target.Target == "" {
 		return ""
 	}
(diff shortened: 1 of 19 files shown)


analytics/analytics.go
```diff
@@ -112,20 +125,44 @@ func isGitDir() bool {
...
+		// Note that this makes an assumption about the typical repo path structure
+		// and the number of slashes. e.g. github.com/foo/bar.
+		// This does not work well for gitlab sometimes.
+		repoAndPathSplit := strings.SplitN(repoAndPath, "/", 4)
+		if len(repoAndPathSplit) < 3 {
+			return repoAndPath
+		}
+		return strings.Join(repoAndPathSplit[0:3], "/")
+	}
+	return localRepo
```
idodod, line 140: nit (inverting the condition):
```suggestion
	if target.Target == "" || !target.IsRemote() {
	    return localRepo
	}
        repoAndPath := target.GitURL
        // Note that this makes an assumption about the typical repo path structure
        // and the number of slashes. e.g. github.com/foo/bar.
        // This does not work well for gitlab sometimes.
        repoAndPathSplit := strings.SplitN(repoAndPath, "/", 4)
        if len(repoAndPathSplit) < 3 {
	        return repoAndPath
        }
        return strings.Join(repoAndPathSplit[0:3], "/")

```

buildcontext/git.go
```diff
@@ -176,7 +176,7 @@ func (gr *gitResolver) resolveGitProject(ctx context.Context, gwClient gwclient.
 	if err != nil {
 		return nil, "", "", errors.Wrap(err, "failed to get url for cloning")
 	}
-	analytics.Count("gitResolver.resolveEarthProject", analytics.RepoHashFromCloneURL(gitURL))
+	analytics.Count("gitResolver.resolveEarthProject", "")
```
idodod, line 179: Curious what does this mean (that's it's an empty string)?

idodod (approved):

buildcontext/git.go
vladaionescu: In this case, it just tracks usage of the remote reference feature, but without being specific.

PR #2938, opened 2023-05-10
Add EARTHLY_CI_RUNNER builtin arg to indicate a build within Earthly CI

idodod:
This will be used in analytics to detect usage of Earthly CI


variables/builtin.go
```diff
@@ -89,6 +90,14 @@ func BuiltinArgs(target domain.Target, platr *platutil.Resolver, gitMeta *gituti
 		// Ensure SOURCE_DATE_EPOCH is always available
 		ret.Add(arg.EarthlySourceDateEpoch, "0")
 	}
+
+	if ftrs.EarthlyCIRunnerArg {
+		value := strings.TrimSpace(os.Getenv(arg.EarthlyCIRunner))
```
vladaionescu, line 95: We typically avoid looking up env vars so deep in the code - we try to only do that in the main package.

You may need to propagate this all the way from main. Here's the rough path:

* Add a hidden flag here: https://github.com/earthly/earthly/blob/main/cmd/earthly/flags.go
* Add a field to this struct and set it to the flag value: https://github.com/earthly/earthly/blob/0857e35cefb1a0155638f47b3ac0558844b48c3f/cmd/earthly/build_cmd.go#L484
* Add a field to this struct and set it to the builder opt field value: https://github.com/earthly/earthly/blob/0857e35cefb1a0155638f47b3ac0558844b48c3f/builder/builder.go#L179
* Add a field to this struct and set it to the converter opt field value: https://github.com/earthly/earthly/blob/0857e35cefb1a0155638f47b3ac0558844b48c3f/earthfile2llb/converter.go#L127

CC @alexcb for a second opinion here, just in case I'm overthinking this.

vladaionescu (commented):
> This will be used in analytics to detect usage of Earthly CI

I don't see analytics-related changes. Is that meant to be done in a different PR?

BTW you don't necessarily need a builtin arg for the analytics change, right? I imagine that the builtin arg is a separate feature actually.

variables/builtin.go
alexcb: I agree it's a bit more plumbing, but I too would rather have them defined via the flags.

analytics/analytics.go
```diff
@@ -28,15 +29,16 @@ import (
...
+		"GITHUB_WORKFLOW":        "github-actions",
+		"CIRCLECI":               "circle-ci",
+		"JENKINS_HOME":           "jenkins",
+		"BUILDKITE":              "buildkite",
+		"DRONE_BRANCH":           "drone",
+		"TRAVIS":                 "travis",
+		"GITLAB_CI":              "gitlab",
+		"EARTHLY_IMAGE":          "earthly-image",
+		"AGENT_WORKDIR":          "jenkins", // https://github.com/jenkinsci/docker-agent/blob/master/11/alpine/Dockerfile#L35
+		reserved.EarthlyCIRunner: "earthly-ci",
```
alexcb, line 41: once the `urfave/cli` flags is used, I would remove this, then just check higher up (wherever `app` is available), if `app.EarthlyCIRunner` is `true`.

variables/builtin.go
idodod: sounds good

idodod:
> I don't see analytics-related changes. Is that meant to be done in a different PR?

This is already done here:
https://github.com/earthly/earthly/blob/3035810b4fcc441407d6406c183711fbd25edaf1/analytics/analytics.go#L282
So it's just a matter of properly detecting Earthly CI

> BTW you don't necessarily need a builtin arg for the analytics change, right? I imagine that the builtin arg is a separate feature actually.

I figured using a builtin arg would prevent the value from getting set but I realize now it's an overkill and actually wouldn't prevent anyone from setting the env var (or what would be now a flag), BUT... I think I'll keep it since I already did the work, unless there are any objections.

vladaionescu:
> I figured using a builtin arg would prevent the value from getting set but I realize now it's an overkill and actually wouldn't prevent anyone from setting the env var (or what would be now a flag), BUT... I think I'll keep it since I already did the work, unless there are any objections.

It's a good feature to have, indeed. BTW, CI=true is another very common env var in CIs. I wonder if we should also set that as a builtin arg too?

idodod:
> > I figured using a builtin arg would prevent the value from getting set but I realize now it's an overkill and actually wouldn't prevent anyone from setting the env var (or what would be now a flag), BUT... I think I'll keep it since I already did the work, unless there are any objections.
> 
> It's a good feature to have, indeed. BTW, CI=true is another very common env var in CIs. I wonder if we should also set that as a builtin arg too?

We currently reserve `EARTHLY_CI` if `--ci` flag/`EARTHLY_CI` env var is used.
Is that what you meant?
https://github.com/earthly/earthly/blob/3035810b4fcc441407d6406c183711fbd25edaf1/variables/builtin.go#L52

variables/builtin.go
idodod: @vladaionescu thanks for guidance!

PR #2961, opened 2023-05-22
Use Earthly to build Dockerfile without Earthfile

idodod:
This PR adds a new `docker-build` command, replacing the previously hidden, experimental command `docker`.

The way the command works is by generating a temporary Earthfile leveraging `FROM DOCKERFILE` command,  and executing it (in the same process/cli invocation).

For example the command:
`earthly docker-build -f Dockerfile -tag my-test-image:latest --target target1 --platform linux/arm64 . --arg1=foo --arg2=bar`

will generate and then execute:

```Earthfile
VERSION --use-docker-ignore 0.7
# This Earthfile was generated using docker-build command
docker:
	ARG arg1
	ARG arg2
	FROM DOCKERFILE \
		 --build-arg arg1=$arg1 \
		 --build-arg arg2=$arg2 \
		 --target target1 \
		 -f Dockerfile \
                  .
	SAVE IMAGE --push my-test-image:latest

build:
    BUILD --platform linux/arm64 +docker
```

In addition, if the buildcontext path contains a .dockerignore file, it will be copied as a temporary .earthlyignore file so that we honor the ignore rules in the temporary Earthfile.

The new command reuses the some useful flags such as `--sat`, `--no-sat`, `--push` from `build` command.


docker2earthly/convert.go
```diff
@@ -139,3 +140,38 @@ func Docker2Earthly(dockerfilePath, earthfilePath, imageTag string) error {
 	fmt.Fprintf(out, "\nbuild:\n    BUILD +subbuild%d\n", i)
 	return nil
 }
+
+func DockerWithEarthly(buildContextPath string, dockerfilePath, imageTag string, buildArgs []string, platforms []string, target string) error {
+	earthfilePath := filepath.Join(buildContextPath, "Earthfile")
+	defer os.Remove(earthfilePath)
```
nelsam, line 146: This `defer os.Remove` also seems wrong to me - this function writes the file and then removes it. I don't see a point where we actually run a build command before the `return` statement (and all of the `defer`red statements run immediately after the `return`).

nelsam (commented):
I think most of the actual errors stem from `defer` triggering perhaps earlier than you might be expecting.

I also think we have an opportunity in `docker2earthly` to make the API more friendly (because `docker2earthly.Docker2Earthly` stutters), but I'm still working on that comment and I don't think these comments really need to wait on that one.

vladaionescu (commented):
This looks good to me, in general!

tests/autocompletion/Earthfile
```diff
@@ -37,7 +38,7 @@ build
 config 
 debug 
 doc 
-docker 
+docker-build
```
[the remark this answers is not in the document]
nelsam: OH, I see, so it is *actually* in the output twice. Got it.

cmd/earthly/root_cmds.go
```diff
@@ -488,32 +500,76 @@ func symlinkEarthlyToEarth() error {
...
+
+	flagArgs, nonFlagArgs, err := variables.ParseFlagArgsWithNonFlags(cliCtx.Args().Slice())
+	if err != nil {
+		return errors.Wrapf(err, "parse args %s", strings.Join(cliCtx.Args().Slice(), " "))
+	}
+	if len(nonFlagArgs) == 0 {
+		_ = cli.ShowAppHelp(cliCtx)
+		return errors.Errorf(
+			"no build context path provided. Try %s docker <path>", cliCtx.App.Name)
+	} else if len(nonFlagArgs) != 1 {
```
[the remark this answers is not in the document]
nelsam: Yeah, no worries, I'm the same way :) Not with `else` statements, but with other silly easy-to-miss stylistic mistakes - like bad documentation, for instance. I never catch that when I'm writing code, only when I'm reviewing it.

Long-winded style opinion: The main reason you most likely won't catch unnecessary `else` statements in my PRs is that I haven't used _any_ `else` statements in a very long time. Another engineer once told me to think of an early return as "popping information off of a developer's mental stack," and that made me realize that logic that merges back into a parent scope (like an `if` that doesn't end in a `return`) is pretty hard to read - because I'm forcing the reader to figure out which variables in the nested scope escape back to the parent scope, and mental merges of variable state are hard.

Around that time, I decided to _try_ just never using `else` statements, see how long it was before I started to miss them. It's been about five years now, and I'm still following that rule. In fact, I've started to lean toward an "every logic branch must end in a return" type of rule. More often than not, it just forces me to keep functions smaller, break logic out into helper functions more frequently, and limit how deeply I nest my scope. None of those are really bad things (most of the  time).

scripts/tests/docker-build/integration.sh
```diff
@@ -0,0 +1,178 @@
+#!/usr/bin/env bash
+set -eu
+
+echo "running tests with $earthly"
+
+earthly=${earthly:=earthly}
+earthly=$(realpath "$earthly")
+frontend="${frontend:-$(which docker)}"
+tag=test-image:v1.2.3
+testdir=/tmp/earthtest
```
alexcb, line 10: I would make this more specific to this particular test.
```suggestion
testdir=/tmp/earthly-docker-build-test
```

cmd/earthly/root_cmds.go
```diff
@@ -754,3 +805,31 @@ func (app *earthlyApp) actionPrune(cliCtx *cli.Context) error {
...
+	if err != nil {
+		return errors.Wrapf(err, "failed to check if %q exists", dockerIgnorePath)
+	}
+
+	if !dockerIgnorePathExists {
+		return nil
+	}
+
+	earthlyIgnoreFilePath := filepath.Join(buildContextPath, ".earthlyignore")
+	err = copy.Copy(dockerIgnorePath, earthlyIgnoreFilePath)
```
alexcb, line 830: this feels odd -- as I user, I would not expect running `earthly docker-build ...` to produce any files in my current working directory.

cmd/earthly/root_cmds.go
```diff
@@ -488,33 +501,71 @@ func symlinkEarthlyToEarth() error {
...
+		return errors.Wrap(err, "docker-build: failed to wrap Dockerfile with an Earthfile")
+	}
+
+	earthfilePath := filepath.Join(buildContextPath, "Earthfile")
+	defer func() {
+		os.Remove(earthfilePath)
+		os.Remove(filepath.Join(buildContextPath, ".earthlyignore"))
+	}()
+
+	out, err := os.Create(earthfilePath)
```
alexcb, line 549: this too feels odd -- the description of `docker-build` says `Build a Dockerfile without an Earthfile`.

If we must create any files, we should do so in a temporary directory and not in the user's CWD (unless the user has explicitly requested the temporary files to be kept).

cmd/earthly/root_cmds.go
```diff
@@ -754,3 +805,31 @@ func (app *earthlyApp) actionPrune(cliCtx *cli.Context) error {
...
 }
+
+func handleDockerFiles(buildContextPath string) error {
+	earthfilePath := filepath.Join(buildContextPath, "Earthfile")
+	earthfilePathExists, err := fileutil.FileExists(earthfilePath)
+	if err != nil {
+		return errors.Wrapf(err, "failed to check if %q exists", earthfilePath)
+	}
+	if earthfilePathExists {
+		return errors.Errorf("earthfile already exists; please delete it if you wish to continue")
```
alexcb, line 816: does this mean it's impossible for a user to run `earthly docker-build ...` twice in a row (without having to delete the generated Earthfile)?

This would irritate me as a user, since it would slow down my dev workflow.

alexcb (changes requested):
Looking good, but I think we need to avoid creating any files in the user's current working directory unless explicitly requested.

idodod:
> Looking good, but I think we need to avoid creating any files in the user's current working directory unless explicitly requested.

Hey @alexcb,
The files (earthfile, and potentially a .earthlyignore file) are generated in the specified build context directory and removed at the end of execution. I can change it to a temporary directory instead

cmd/earthly/root_cmds.go
idodod: The files are removed in the end of execution, but I can switch it to use a temp dir instead

cmd/earthly/root_cmds.go
idodod: since the file is removed in the end it shouldn't be a problem. This check is to make sure the user doesn't accidentally override an existing Earthfile. 
Switching to a temp dir can avoid this check though.

cmd/earthly/root_cmds.go
idodod: I moved the temporary Earthfile to a temp dir (which is removed at the end of the execution).
If there is a .dockerignore file present in the build context,  a temporary .earthlyignore file is created in the build context dir - I was not able to get it to work if .earthlyignore is in the temp dir

cmd/earthly/root_cmds.go
alexcb: Is this still being produced in the CWD rather than the temp dir?

Could we somehow create this in the temp dir, and override https://github.com/earthly/earthly/blob/main/buildcontext/local.go#L94 for this particular case?

cmd/earthly/root_cmds.go
idodod: we not longer need to create .earthlyignore because FROM DOCKER would use existing .dockerignore files from the build context directory

cmd/earthly/root_cmds.go
idodod: removed, the command does not longer generate a `.earthlyignore` file

idodod:
@alexcb thanks again for pushing me on the temp files location - I love we were able to avoid looking for an existing Earthfile or generating files we don't have to.

docs/earthly-command/earthly-command.md
```diff
@@ -961,3 +961,48 @@ If you are logged into the CLI the url will contain a token used to link your OA
...
+Specifies the target to build in a multi-target Dockerfile.
+
+##### `--platform <platform>`
+
+Also available as an env var setting: `EARTHLY_PLATFORMS=<platform>`.
+
+Sets the platform to build for.
+
+###### Note
+Unlike a normal build command, it is possible to specify multiple platforms through this flag.
```
vladaionescu, line 1000: You might want to use this gitbook-specific syntax here:

```suggestion
{% hint style='info' %}
##### Note
Unlike a normal build command, it is possible to specify multiple platforms through this flag.
{% endhint %}
```

docs/earthly-command/earthly-command.md
```diff
@@ -961,3 +961,48 @@ If you are logged into the CLI the url will contain a token used to link your OA
...
+
+##### Satellites:
+
+Just like a regular build, `docker-build` can also be run on a satellite. For more information see the [Using Satellites guide](../cloud/satellites/using.md).
+
+##### Other Build Options:
+
+`docker-build` support all other build options. For more information see [build-options](#build-options).
```
vladaionescu, line 1008: These two remarks are out of place over here. Perhaps they should be simple paragraphs within the description instead.

docs/earthly-command/earthly-command.md
```diff
@@ -961,3 +961,48 @@ If you are logged into the CLI the url will contain a token used to link your OA
...
+Also available as an env var setting: `EARTHLY_PLATFORMS=<platform>`.
+
+Sets the platform to build for.
+
+###### Note
+Unlike a normal build command, it is possible to specify multiple platforms through this flag.
+
+##### Satellites:
+
+Just like a regular build, `docker-build` can also be run on a satellite. For more information see the [Using Satellites guide](../cloud/satellites/using.md).
```
vladaionescu, line 1004: I would show a very simple example of a satellite build:

```
earthly docker-build --sat my-satellite -t my-image:latest .
```

vladaionescu (commented):
Some minor remarks...

docs/earthly-command/earthly-command.md
idodod: updated

docs/earthly-command/earthly-command.md
idodod: moved to the description, added an example for using a satellite

docs/earthly-command/earthly-command.md
idodod: added

vladaionescu (approved):

docker2earthly/convert.go
```diff
@@ -139,3 +142,71 @@ func Docker2Earthly(dockerfilePath, earthfilePath, imageTag string) error {
...
+	BuildArgs    []string
+	Target       string
+	Dockerfile   string
+	BuildContext string
+	ImageTags    []string
+	Platforms    []string
+}
+
+// GenerateEarthfileContent returns an Earthfile content string which contains a target to build a docker image using FROM DOCKERFILE
+func GenerateEarthfileContent(buildContextPath string, dockerfilePath string, imageTags []string, buildArgs []string, platforms []string, target string) (string, error) {
```
nelsam, line 183: nit: `docker2earthly.GenerateEarthfileContent` is pretty verbose to my eyes. We're already in the `docker2earthly` package, so maybe just `Generate` gives us most of the same information in 1/3 of the characters?

nelsam (approved):
Some nitpicks, but lgtm

docs/earthly-command/earthly-command.md
```diff
@@ -961,3 +961,48 @@ If you are logged into the CLI the url will contain a token used to link your OA
...
+  ```
+  earthly [options] docker-build [--dockerfile <dockerfile-path>] [--tag=<image-tag>] [--target=<target-name>] [--platform <platform>] <build-context-dir> [--arg1=arg-value] 
+  ```
+
+#### Description
+
+The command `earthly docker-build` builds a docker image from a Dockerfile instead of an Earthfile.
+The <context-path> is the path where the Dockerfile build context exists. By default, it is assumed that a file named Dockerfile exists in that directory.
+
+#### Options
```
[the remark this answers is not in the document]
idodod: I think it's ok to mention it as part of the description

docker2earthly/convert.go
idodod: I went with `GenerateEarthfile`

docs/earthly-command/earthly-command.md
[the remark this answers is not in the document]
nelsam: Oh, right. I didn't see that in the description.

Minor note: line `977` will be part of the same paragraph as `976`, because markdown ignores single newlines. I just wanted to check that that's what you want. Otherwise :+1:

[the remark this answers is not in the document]
idodod: good eye! updated

PR #2964, opened 2023-05-23
docs: add information about --arg-scope-and-set

nelsam:
This documents the `LET` and `SET` commands featured under `--arg-scope-and-set`, as well as the changed shadowing/scoping behavior of `ARG` variables when this feature is enabled.

### DIFF
--- a/docs/earthfile/earthfile.md
+++ b/docs/earthfile/earthfile.md
@@ -542,6 +542,25 @@ FROM +docker-image --NAME=john
 
 For more information on how to use build args see the [build arguments and secrets guide](../guides/build-args.md). A number of builtin args are available and are pre-filled by Earthly. For more information see [builtin args](./builtin-args.md).
 
+{% hint style='info' %}
+##### Shadowing Variables
+
+By default, `ARG` scoping isn't intuitive. When an `ARG` statement is parsed, earthly will look up _any_ previous declaration of the `ARG` and use the previous value, ignoring any new default. So for example:
+
+```
+VERSION 0.7
+ARG --global foo = bar
+
+baz:
+    ARG foo = bacon
+    RUN echo $foo
+```
+
+would print `bar`.
+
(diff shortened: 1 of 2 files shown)


docs/earthfile/earthfile.md
```diff
@@ -542,6 +542,25 @@ FROM +docker-image --NAME=john
...
+ARG --global foo = bar
+
+baz:
+    ARG foo = bacon
+    RUN echo $foo
+```
+
+would print `bar`.
+
+Changing the `VERSION` line to `VERSION --arg-scope-and-set 0.7` changes this behavior, allowing local `ARG`s to shadow global `ARG`s and erroring if an `ARG` is redeclared in the same scope. This means that the above example would instead print `bacon`.
```
vladaionescu, line 561: This should state more clearly that this enables an experimental feature.

vladaionescu (commented):
Looks good! I left a couple of small suggestions.

We also have a guide on args, envs and secrets that should be updated. Although since this is experimental, perhaps it's best that we update the guide when this becomes GA.

docs/earthfile/earthfile.md
```diff
@@ -564,6 +583,57 @@ A global `ARG` is an arg that is made available to all targets in the Earthfile.
 
 Global args may only be declared in base targets.
 
+## LET (experimental)
```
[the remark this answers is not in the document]
nelsam: Fixed. I was leaning back and forth on this one - `LET` is so closely related to `ARG` that I felt it might make sense to keep the two together, even though it's  experimental.

But at least right now, with our docs page limitations, I think you're probably right - it belongs at the end.

docs/earthfile/earthfile.md
nelsam: Updated to mention that the flag is experimental. Thanks!

alexcb (approved):

PR #2981, opened 2023-05-24
test cloud client can be created without network

alexcb:
This creates a new +offline-test target, which is used to validate the cloud client can be created without performing any network I/O (which is important for autocompletion and cases where a user has set DO_NOT_TRACK).

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -1,4 +1,4 @@
-VERSION --shell-out-anywhere --use-copy-link 0.6
+VERSION --no-network 0.7
 
 FROM golang:1.20-alpine3.17
 
@@ -254,6 +254,10 @@ unit-test:
 chaos-test:
     FROM +code
     RUN go test -tags chaos ./...
+
+offline-test:
+    FROM +code
+    RUN --network=none go test -tags offline ./...
 
 # submodule-decouple-check checks that go submodules within earthly do not
 # depend on the core earthly project.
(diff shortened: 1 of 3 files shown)


cloud/imports_test.go
```diff
@@ -0,0 +1,12 @@
+//go:build offline
```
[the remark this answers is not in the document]
nelsam: Yeah, a lot of go tooling doesn't understand build tags very well. `gopls` is also still ignorant of tags, which makes editors pretty bad at syntax highlighting of tagged files.

Does the test itself need to be behind a build tag? Personally, I _usually_ only put tests behind a build tag if running them would cause problems - for example, tests that take a long time or require a network.

Earthfile
```diff
@@ -255,6 +255,10 @@ chaos-test:
     FROM +code
     RUN go test -tags chaos ./...
 
+offline-test:
+    FROM +code
+    RUN --network=none go test -tags offline ./...
```
nelsam, line 260: This could alternatively be `RUN --network=none go test ./cloud -run TestClientConnectionIsLazy`. Or we could use an onpar suite to allow multiple tests with the `-run` flag, e.g.

```go
func TestOffline(t *testing.T)  {
    o := onpar.New(t)
    defer  o.Run()
    o.Spec("ClientConnectionIsLazy", func(t *testing.T) {
         // ...
    })
}
```

Then `go test -run TestOffline` would run all of the offline subtests.

alexcb: Thanks for the idea -- I like it more than using a tag.

nelsam (approved):

nelsam:
I realized after I approved that this is still a draft :sweat_smile: feel free to request a re-review if/when needed.

PR #2985, opened 2023-05-26
introduces a call to list satellites by org name

alexcb:
additionally fixes a bug where an incorrect satellite would be displayed as selected when a member is part of two orgs which contain the same satellite name.

### DIFF
--- a/cloud/client.go
+++ b/cloud/client.go
@@ -59,6 +59,7 @@ type Client struct {
 	installationName         string
 	logstreamAddressOverride string
 	serverConnTimeout        time.Duration
+	orgIDCache               map[string]string // orgName -> orgID
 }
 
 type ClientOpt func(*Client)
@@ -81,6 +82,7 @@ func NewClient(httpAddr, grpcAddr string, useInsecure bool, agentSockPath, authC
 		installationName:  installationName,
 		requestID:         requestID,
 		serverConnTimeout: serverConnTimeout,
+		orgIDCache:        map[string]string{},
 	}
 	if authJWTOverride != "" {
 		c.authToken = authJWTOverride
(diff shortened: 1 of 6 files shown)


cloud/org.go
```diff
@@ -177,6 +178,9 @@ func (c *Client) CreateOrg(ctx context.Context, org string) error {
 
 // GetOrgID retrieves the org ID for a named org.
 func (c *Client) GetOrgID(ctx context.Context, orgName string) (string, error) {
```
nelsam, line 180: If all of the methods now take org names, can we unexport this method?

alexcb: unfortunately the satellite-proxy depends on an `satellite_org` metadata field being set to the ID; we need to set https://github.com/earthly/earthly/blob/acb/list-satellites-by-org-name/cmd/earthly/satellite_cmds.go#L659 which ends up being used in https://github.com/earthly/earthly/blob/acb/list-satellites-by-org-name/buildkitd/buildkitd.go#L950

alexcb: as a result we must query the org ID when a `--org ...` is passed via the cli: https://github.com/earthly/earthly/blob/acb/list-satellites-by-org-name/cmd/earthly/satellite_cmds.go#L382

cloud/client.go
```diff
@@ -59,6 +59,7 @@ type Client struct {
 	installationName         string
 	logstreamAddressOverride string
 	serverConnTimeout        time.Duration
+	orgIDCache               map[string]string // orgName -> orgID
```
nelsam, line 62: one quick check: is there any chance this will be accessed concurrently? `sync.Map` is not terribly hard to use and would prevent any data races in such a case. I doubt the performance impact is notable for our purposes. It might be worth using here.

nelsam (approved):
lgtm. Just one minor question regarding concurrent map access.

cloud/client.go
alexcb: good point! Fixed.

PR #2986, opened 2023-05-26
Check http errors via http code when listing secrets

idodod:
Using a new type of Error similar to gRPC error which contains the status code.
This is used to check of server error according to code instead of string.
Currently this is only used when listing secrets, though I think overtime we can use it for other methods (I didn't want to increase the scope of this PR).

Also wrapping a few other errors to make it easier to track the call stack.

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -56,7 +56,7 @@ code:
     END
     COPY ./ast/parser+parser/*.go ./ast/parser/
     COPY --dir analytics autocomplete buildcontext builder logbus cleanup cmd config conslogging debugger \
-        dockertar docker2earthly domain features outmon slog cloud states util variables ./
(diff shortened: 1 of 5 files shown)


nelsam:
Suggestion: since the rest of `earthly` doesn't know (and shouldn't care) whether `ListSecrets` is a gRPC call or an http call, what do you think about making a separate `type StatusCode int` in the `cloud` package and translate from both gRPC status and http status to cloud status?

Then when we get an error from the cloud client, the logic flow is `if cloud.Status(err) ==  cloud.StatusNotFound {` - which I think reads a lot clearer. It also means that if we change `ListSecrets` to a gRPC call in the future, we only have to change the `cloud.Client` - everything else stays the same, because they don't care whether the call is http or gRPC.

PR #3001, opened 2023-06-02
Use .dockerignore in FROM DOCKERFILE targets

idodod:
This PR adds support for using a `.dockerignore` file in target using `FROM DOCKERFILE` with <build context> from the host system.
The feature is enabled using `--use-docker-ignore` feature flag.
The `.dockerignore` file will not be used if the <build context> contains an `.earthlyignore` or `.earthignore` file.

### DIFF
--- a/buildcontext/excludes.go
+++ b/buildcontext/excludes.go
@@ -1,7 +1,6 @@
 package buildcontext
 
 import (
-	"fmt"
 	"os"
 	"path/filepath"
 
@@ -47,7 +46,6 @@ func readExcludes(dir string, noImplicitIgnore bool, excludeFromDockerIgnore boo
 	dockerExists := false
 	if excludeFromDockerIgnore {
 		dockerExists, err = fileutil.FileExists(dockerIgnoreFilePath)
-		fmt.Printf("docker exists = %v for %s\n", dockerExists, dockerIgnoreFilePath)
 		if err != nil {
 			return nil, errors.Wrapf(err, "failed to check if %s exists", dockerIgnoreFilePath)
 		}
--- a/buildcontext/excludes.go
+++ b/buildcontext/excludes.go
@@ -24,7 +24,7 @@ var ImplicitExcludes = []string{
 	earthlyIgnoreFile,
 }
 
-func readExcludes(dir string, noImplicitIgnore bool, excludeFromDockerIgnore bool) ([]string, error) {
+func readExcludes(dir string, noImplicitIgnore bool, useDockerIgnore bool) ([]string, error) {
 	var ignoreFile = earthIgnoreFile
 
 	//earthIgnoreFile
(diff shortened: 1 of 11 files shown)


docs/earthfile/features.md
```diff
@@ -57,6 +57,8 @@ to require version `0.X` (or later), and could be rewritten as `VERSION 0.X`.
 | `--use-pipelines` | 0.7 | Enable the `PIPELINE` and `TRIGGER` commands |
 | `--earthly-git-author-args` | 0.7 | Enable the `EARTHLY_GIT_AUTHOR` and `EARTHLY_GIT_CO_AUTHORS` args |
 | `--wait-block` | 0.7 | Enable the `WAIT` / `END` block commands |
+| `--earthly-ci-runner-arg` | 0.7 | Enable the `EARTHLY_CI_RUNNER` builtin ARG |
```
idodod, line 60: I forgot to add this one in a previous PR

features/features.go
```diff
@@ -261,3 +265,21 @@ func processNegativeFlags(ftrs *Features) {
 		ftrs.UseRegistryForWithDocker = false
 	}
 }
+
+// WithContext adds the current *Features into the given context and returns a new context.
+// Trying to add the *Features to the context more than once will result in an error.
+func (f *Features) WithContext(ctx context.Context) (context.Context, error) {
+	if _, ok := ctx.Value(ctxKey{}).(*Features); ok {
```
nelsam, line 272: we *should* just be able to do `if ctx.Value(ctxKey{}) != nil {`. since no other package can use the `ctxKey` type, no other package could assign a value - so doing a type assertion isn't necessary.

nelsam (approved):
_very_ minor nitpick, otherwise lgtm

buildcontext/excludes.go
```diff
@@ -23,7 +24,7 @@ var ImplicitExcludes = []string{
 	earthlyIgnoreFile,
 }
 
-func readExcludes(dir string, noImplicitIgnore bool) ([]string, error) {
+func readExcludes(dir string, noImplicitIgnore bool, excludeFromDockerIgnore bool) ([]string, error) {
```
alexcb, line 27: I think this would be more clear (and less chars)

```suggestion
func readExcludes(dir string, noImplicitIgnore bool, readDockerIgnore bool) ([]string, error) {
```

or alternatively `useDockerIngnore`.

alexcb (approved):
just one minor naming suggestion nit.

buildcontext/excludes.go
idodod: done

docs/earthfile/earthfile.md
```diff
@@ -883,6 +883,7 @@ Instructs Earthly to not overwrite the file creation timestamps with a constant.
 The `FROM DOCKERFILE` command initializes a new build environment, inheriting from an existing Dockerfile. This allows the use of Dockerfiles in Earthly builds.
 
 The `<context-path>` is the path where the Dockerfile build context exists. By default, it is assumed that a file named `Dockerfile` exists in that directory. The context path can be either a path on the host system, or an [artifact reference](../guides/target-ref.md#artifact-reference), pointing to a directory containing a `Dockerfile`.
+Additionally, when using a `<context-path>` from the host system, a `.dockerignore` in the directory root will be used to exclude files (unless `.earthlyignore` or `.earthignore` are present). Use `--use-docker-ignore` feature flag to enable.
```
vladaionescu, line 886: For clarity - many do not know how to use these feature flags.

```suggestion
Additionally, when using a `<context-path>` from the host system, a `.dockerignore` in the directory root will be used to exclude files (unless `.earthlyignore` or `.earthignore` are present). Use `VERSION --use-docker-ignore 0.7` to enable.
```

vladaionescu (approved):
Just one nit

PR #3003, opened 2023-06-05
earthfile2llb: respect ctx.Done() in withDockerRunRegistry

nelsam:
This watches for ctx.Done() when looping through the res.ResultChan channel from MultiImageSolver.SolveImages when handling a WITH DOCKER RUN command that uses the registry.

### DIFF
--- a/earthfile2llb/with_docker_run_reg.go
+++ b/earthfile2llb/with_docker_run_reg.go
@@ -152,10 +152,11 @@ func (w *withDockerRunRegistry) Run(ctx context.Context, args []string, opt With
 	})
 
 	// Wait for all images to build (channel will be closed when finished).
-	results := make([]*states.ImageResult, 0, len(imagesToBuild))
-	for result := range res.ResultChan {
-		results = append(results, result)
+	results, err := readImgResults(ctx, res.ResultChan)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


alexcb (approved):
nice find!

earthfile2llb/with_docker_run_reg.go
```diff
@@ -320,3 +321,18 @@ func (w *withDockerRunRegistry) load(ctx context.Context, opt DockerLoadOpt) (ch
 
 	return imageDefChan, nil
 }
+
+func readImgResults(ctx context.Context, ch chan *states.ImageResult) ([]*states.ImageResult, error) {
+	var results []*states.ImageResult
+	for {
+		select {
+		case result, ok := <-ch:
+			if !ok {
```
alexcb, line 330: observation: the expected (non-error) code flow requires this to be "not ok" :laughing:

alexcb: (this isn't even a nit, just an observation I thought was amusing)

nelsam: Hah, that is amusing. It's tempting to use `keepGoing` or something, but I think convention makes it reasonable to still use `ok` in spite of the ... odd flow.

PR #3008, opened 2023-06-08
Remove default size in satellite update

brandonSc:
This fixes a bug that was causing manual updates to revert the size to medium if the user previously had a different size set.


alexcb (approved):
looks like an easy fix!

alexcb:
can you update the CHANGELOG too @brandonSc ?

vladaionescu:
So this just forces the user to specify a size, right? Would it be possible to detect the size of the currently running satellite and use that?

alexcb (approved):

PR #3012, opened 2023-06-09
.github: create issue templates

nelsam:


nelsam:
I also thought about creating a pull request template, but I don't really see the point. Bug and proposal templates probably cover the important bases.

.github/ISSUE_TEMPLATE/bug_report.md
alexcb, line 1: might be good to include a reminder to look at logs under `docker logs earthly-buildkitd`, and to try running earthly with `--verbose`.

maybe even with `--debug`; however we don't really advertise `--debug`, but maybe we should, since we do mention it in slack periodically.

nelsam: Updated. I didn't want to have a specific section just for certain types of output, so I grouped all of that up under `Other Helpful Information`. Hopefully that's enough of a reminder for most people.

If we feel like we're still getting issues that are missing critical information, we can adjust over time.

alexcb (commented):
I like the Reproduction Steps and Other helpful Information steps, but I worry the other sections will be too verbose for obvious bugs.

For example, if there's a case where earthly panics; this becomes both too much for a user to type, and for us to read:

```
Title: panic() while doing X
Description: earthly panics when doing X;
Expected behavior: do X without panicing
Actual behavior: earthly panics when doing X
```

Instead I would much rather receive an issue such as:

```
Title: panic() while doing X

**Reproduction Step**

Here's a sample Earthfile ...
Then I ran earthly +x

and it broke: <stack trace>

**Other Helpful Information**
<earthly version>
<using docker>
<buildkit logs>
```

alexcb:
How about something more minimal like?

```
<!-- What happened? -->

<!-- Can you reproduce it? -->

<!-- Do you have a sample Earthfile? -->

<!-- What did you expect to happen (if non-obvious) -->

**Other Helpful Information**
<!-- Please include any additional information you might have. This may include:
- The logs from the `earthly-buildkitd` container, usually from running `docker logs earthly-buildkitd`.
- The output of the failing command.
- Output of the failing command with `--verbose ` enabled, e.g. `earthly --verbose +some-failing-target`.
- Stack trace.
    - If you `kill -SIGABRT` your earthly process, you will get a full stack
      trace. This is useful if the process is stuck or slow, so that we can get an
      idea of which functions might be stuck.
- In some rare circumstances: screenshots (although if the output is text, we
  prefer copy/pasted text over screenshots).
-->
```

nelsam:
I like the "more minimal" approach. The only counterpoint I have: I don't think I want to say `(if non-obvious)` on the "what did you expect to happen" question.

Reasoning:
- We often expect something different than our users do specifically because we (usually) know our product better than they do.
- Users will _sometimes_ (i.e. more often than I'm comfortable with) think that what they expected to happen is obvious.

Usually, this leads to confusion. The thing I'm a little worried about:

- a user submits a bug report, thinking that the correct behavior is obvious
- we see an error message that is obviously wrong, implement a fix, and submit it as a solution for the user's issue
- the user comes back saying that it doesn't solve their issue
- it turns out the user was complaining about correct (but non-obvious) earthly behavior, and we could have unblocked them when they submitted the issue rather than making them wait for a release

alexcb:
> a user submits a bug report, thinking that the correct behavior is obvious

That's fair; 

changing it to simply be this:

```
<!-- What did you expect to happen -->
```

would allow a user to just leave the section blank if they choose too -- and we won't see any empty header section as a result.

nelsam:
@alexcb I've updaed it. I still have a heading for `What should have happened?` but I've reduced the bug report to three headings total. I don't think it's too verbose.

PR #3024, opened 2023-06-14
.github: spread out multiple test targets across multiple workflows

nelsam:
This splits up some of our longest-running (and most likely to flake) GHA jobs into smaller chunks. This helps specifically in the case where a test flakes, requiring us to re-run the job - taking around 10 minutes each retry instead of 45.

Hopefully, this will also give us a clearer picture of which tests are flaking most often, so we know which ones need to be fixed.


brandonSc (approved):
Splitting up the tests into smaller ones seems like a good idea

Issue #3025, opened 2023-06-14
multiple attempts to authenticate to the same registry


alexcb:
A user reported that there's multiple lookup attempts (or maybe even an infinite loop) for the same registry when using a satellite

```
looking up /user/std/registry/123456789.dkr.ecr.us-east-2.amazonaws.com/cred_helper
no cred_helper for "/user/std/registry/123456789.dkr.ecr.us-east-2.amazonaws.com/cred_helper" exists, checking for username/password based config
looking up /user/std/registry/123456789.dkr.ecr.us-east-2.amazonaws.com/username
looking up /user/std/registry/123456789.dkr.ecr.us-east-2.amazonaws.com/password
looking up /user/std/registry/123456789.dkr.ecr.us-east-2.amazonaws.com/cred_helper
no cred_helper for "/user/std/registry/123456789.dkr.ecr.us-east-2.amazonaws.com/cred_helper" exists, checking for username/password based config
looking up /user/std/registry/123456789.dkr.ecr.us-east-2.amazonaws.com/username
looking up /user/std/registry/123456789.dkr.ecr.us-east-2.amazonaws.com/password
looking up /user/std/registry/123456789.dkr.ecr.us-east-2.amazonaws.com/cred_helper
no cred_helper for "/user/std/registry/123456789.dkr.ecr.us-east-2.amazonaws.com/cred_helper" exists, checking for username/password based config
```

This user does not have **any** registries listed under `earthly secrets ls /user/std/` (or viewable via earthly registry list)

possibly related to #3010

PR #3026, opened 2023-06-15
keep track of which authprovider responded

alexcb:
Once an authprovider responds for a specific host, only subsequent requests should be sent to that authprovider; similarlly if an authserver has responded with ErrAuthProviderNoResponse (indicating it doesnt have credentials) it shouldn't be requeried for credentials.


alexcb:
attempted fix for #3025 (but needs better testing)

nelsam (approved):
looks good! hopefully this helps with some of our unauthorized responses in CI.

PR #3031, opened 2023-06-16
docker-build command additions

idodod:
- Add \*beta\* to help menu
 - Add integration test that uses a satellite
 - Support passing multiple platforms using docker command style, e.g `--platform linux/amd64,linux/arm64`
 - Update documentation accordingly

### DIFF
--- a/docker2earthly/convert.go
+++ b/docker2earthly/convert.go
@@ -180,7 +180,7 @@ type earthfileTemplateArgs struct {
 }
 
 // GenerateEarthfile returns an Earthfile content string which contains a target to build a docker image using FROM DOCKERFILE
-func GenerateEarthfile(buildContextPath string, dockerfilePath string, imageTags []string, buildArgs []string, platformsStr string, target string) (string, error) {
+func GenerateEarthfile(buildContextPath string, dockerfilePath string, imageTags []string, buildArgs []string, platforms []string, target string) (string, error) {
 	t, err := template.New("earthfile").Parse(earthfileTemplate)
 	if err != nil {
 		return "", errors.Wrapf(err, "failed to parse Earthfile template")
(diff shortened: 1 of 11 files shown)


docker2earthly/convert.go
```diff
@@ -180,7 +180,7 @@ type earthfileTemplateArgs struct {
 }
 
 // GenerateEarthfile returns an Earthfile content string which contains a target to build a docker image using FROM DOCKERFILE
-func GenerateEarthfile(buildContextPath string, dockerfilePath string, imageTags []string, buildArgs []string, platforms []string, target string) (string, error) {
+func GenerateEarthfile(buildContextPath string, dockerfilePath string, imageTags []string, buildArgs []string, platformsStr string, target string) (string, error) {
```
alexcb, line 183: this feels inconsistent -- why should imageTags and buildArgs be passed as a slice, but platforms be passed as a comma separated string?

I think it would be better to expose `splitPlatforms` and have the caller of `GenerateEarthfile` pre-split the platforms.

alexcb: furthermore, it looks like it's also possible to pass `--platform` as separate flags, e.g. `docker build --platform linux/amd64 --platform linux/arm64` so we should be able to support this as well.

alexcb (commented):
I think we need to support both `docker build --platform linux/amd64 --platform linux/arm64` and `docker build --platform linux/amd64,linux/arm64` to maintain consistency with docker's cli.

docker2earthly/convert.go
idodod: @alexcb , updated, please let me know if that looks good to you.

alexcb: looks great! Plus you wrote a `TestSplitFlagString` :raised_hands:

alexcb (approved):

alexcb (approved):

PR #3048, opened 2023-06-20
check Added/Changed/etc headers are correctly placed

alexcb:
Changelog linter will now detect if an Added/Changed/Fixed/etc header is in the incorrect location.

### DIFF
--- a/release/changelogparser.py
+++ b/release/changelogparser.py
@@ -72,6 +72,8 @@ def parse_changelog(changelog_data):
     dash_found = False
     body = []
     ignore = False
+    prev_header_num = None
+    prev_header_title = None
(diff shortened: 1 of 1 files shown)


release/changelogparser.py
```diff
@@ -81,6 +83,8 @@ def save_version(version, release_date, body):
             if not title.endswith(' Changelog'):
                 raise MissingTitleError("expected title ending with Changelog", line_num)
             is_title_body = True
+            prev_header_num = num_headers
```
alexcb, line 86: duplicated due to the `continue`; sorry :(

idodod (approved):
🎉

PR #3049, opened 2023-06-21
update buildkit with longer lease duration

alexcb:


alexcb:
depends on https://github.com/earthly/buildkit/pull/16

PR #3059, opened 2023-06-29
authprovider: invalidate cache when AddProject is called

nelsam:
Part of the improvements in PR #3026 was a form of caching authproviders that don't have any credentials configured for a given host. This did not account for the project being set *after* the first attempt at loading credentials.

This commit updates the aggregated authprovider to invalidate its cache of per-host child auth providers when AddProject is called.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -24,7 +24,6 @@ import (
 	"github.com/earthly/earthly/util/gatewaycrafter"
 	"github.com/earthly/earthly/util/gwclientlogger"
 	"github.com/earthly/earthly/util/llbutil"
-	"github.com/earthly/earthly/util/llbutil/authprovider/cloudauth"
 	"github.com/earthly/earthly/util/llbutil/pllb"
 	"github.com/earthly/earthly/util/llbutil/secretprovider"
 	"github.com/earthly/earthly/util/platutil"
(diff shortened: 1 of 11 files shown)


builder/builder.go
```diff
@@ -106,7 +109,7 @@ type BuildOpt struct {
 	Logbus                     *logbus.Bus
 	MainTargetDetailsFunc      func(earthfile2llb.TargetDetails) error
 	Runner                     string
-	CloudStoredAuthProvider    cloudauth.ProjectBasedAuthProvider
+	ProjectAdder               ProjectAdder
```
idodod, line 112: wondering about the field name - are we sure the previous name (or some variation of authProvider) is not clearer?

idodod (approved):
LGTM!
I thinks renaming makes things a lot clearer.
I was just wondering about one specific field name that I wasn't sure about.

builder/builder.go
nelsam: Since the only method that was ever used was `AddProject`, I don't think that there was any point in declaring that it's an auth provider - because nothing really prevents us from using an entirely different type (as long as it has an `AddProject` method). I had to change the name because it's no longer `CloudStoredAuthProvider` - the type we're passing in is now  the `MultiAuthProvider`, so naming it something about `cloud` is misleading.

I don't really think `ProjectAdder` is very descriptive, but it's probably good enough, I think?

idodod: I didn't even notice that was the only method used, sgtm!

PR #3064, opened 2023-06-30
Allow use of personal orgs in satellite commands

brandonSc:


cmd/earthly/satellite_cmds.go
```diff
@@ -456,10 +456,10 @@ func (app *earthlyApp) actionSatelliteLaunch(cliCtx *cli.Context) error {
 		return err
 	}
 
-	if !cloud.ValidSatellitePlatform(platform) {
+	if platform != "" && !cloud.ValidSatellitePlatform(platform) {
 		return errors.Errorf("not a valid platform: %q", platform)
 	}
-	if !cloud.ValidSatelliteSize(size) {
+	if size != "" && !cloud.ValidSatelliteSize(size) {
```
brandonSc, line 462: This was a separate bug I'm sneaking in with using empty string as default size.

cloud/org.go
```diff
@@ -207,13 +209,21 @@ func (c *Client) GuessOrgMembership(ctx context.Context) (orgName, orgID string,
...
+	switch len(orgs) {
+	case 0:
 		return "", "", errors.New("not a member of any organizations - cloud features require you are a member of an organization")
+	case 1:
+		return orgs[0].Name, orgs[0].ID, nil
+	case 2:
+		for _, o := range orgs {
+			if !o.Personal {
+				return o.Name, o.ID, nil
+			}
```
vladaionescu, line 221: I know it would be backwards incompatible, but I would have thought that it defaults to personal generally (if we ignore the history about how we got here).

I can't think of a better solution, though, the alternative is pretty bad too, because builds just start failing all of a sudden due to the missing `--org` param.

So I guess what you have here makes sense. Can we perhaps display a warning when this happens? And say that we are deprecating this behavior in the future?

nelsam: Doesn't that mean we're deprecating it now? Deprecated behaviors are behaviors that still work but are planned to be removed in the future.

nelsam (changes requested):
Can we get some tests added?

If this is critically important to merge today, can we at least get a follow-up ticket (very high priority) to implement tests for the feature?

cloud/org.go
vladaionescu: Yes, correct :+1:

vladaionescu: Thinking a bit more about this - should we also prevent a personal org from being auto-chosen if n = 1? I think this makes things more consistent when going to n = 2.

brandonSc: I thought the personal org at n=1 is convenient at least, but agree it's a bit of a trivial choice on what makes sense when n>=2.

Are you suggesting then that in all cases, the `--org` flag or `EARTHLY_ORG` env var should be set? With the exception of keeping it compatible temporarily when n=2?

vladaionescu: I think so. Open to other ideas.

It's definitely annoying to have to be so verbose all the time, but it's the most clear behavior.

Perhaps the way to counteract this verbosity is to have a way to "select" an org too, similarly to how you can select satellites.

brandonSc:
> Can we get some tests added?
> 
> If this is critically important to merge today, can we at least get a follow-up ticket (very high priority) to implement tests for the feature?

I think there's a bigger problem where none of the cloud endpoints have tests here. I think we should discuss how we want to test this (e.g. mocking lib?) Perhaps that should be the follow-up ticket you suggested

cloud/org.go
brandonSc: I'm on board with the org select idea. We also don't auto-select a satellite when you have only one satellite.

That being said, I don't have time to work on org select now before free tier launches, so perhaps we just stick with the warning for compatibility and add org select later.

vladaionescu: SGTM!

brandonSc: Screen Shot 2023-06-30 at 3 18 48 PM

brandonSc:
Ticket for future testability of cloud functions: https://github.com/earthly/earthly/issues/3065

nelsam (approved):

PR #3066, opened 2023-06-30
Add support for new satellite info

dchw:
This includes a couple small items/fixes for Satellites.

* `satellite ls -a` now shows the last time the satellite was used, and also the cache retention duration.
* `satellite ls -a` no longer specifies white as the color, instead just tinting the already-used color. Tested on light and dark terminal themes.
* `satellite inspect` also shows when the satellite was last used in the users timezone, and also the cache retention duration.
* Add support for "destroying", "offline", and "creating" states when performing operations that wait for satellites to finish transitioning.
* Add more fun messages for satellite loading flavor text.


cmd/earthly/satellite_cmds.go
```diff
@@ -309,22 +310,22 @@ func (app *earthlyApp) printSatellitesTable(satellites []satelliteWithPipelineIn
 	}
 
 	t := tabwriter.NewWriter(os.Stdout, 1, 2, 2, ' ', 0)
-	headerRow := []string{" ", "NAME", "PLATFORM", "SIZE", "VERSION", "STATE"} // The leading space is for the selection marker, leave it alone
+	headerRow := []string{" ", "NAME", "PLATFORM", "SIZE", "VERSION", "STATE", "LAST USED", "CACHE"} // The leading space is for the selection marker, leave it alone
```
vladaionescu, line 313: `CACHE` is for cache retention right? I wonder if we even need to list it. Most people will be on a single plan and all satellites will have the same entry. I would just keep it under `inspect` only.

cmd/earthly/satellite_cmds.go
```diff
@@ -1050,6 +1091,10 @@ func getSatelliteLoadingMessages() []string {
 		"checking battery levels",
 		"calibrating transponders",
 		"modifying downlink frequency",
+		"reticulating splines",
+		"perturbing matrices",
+		"synthesizing gravity",
+		"iterating cellular automata",
```
vladaionescu, line 1097: LoL 😂

vladaionescu (approved):

cmd/earthly/satellite_cmds.go
dchw: I'll move it to appear on `-a` only, since it is still useful here in some contexts.

PR #3069, opened 2023-07-04
Implement `org select` command

dchw:
Implement `org select`. This provides a default org in the config file that can be used when either the CLI `--org` is not provided, or when the `EARTHLY_ORG` environment variable is not set.

Unify org selection across Satellites and other commands.

### DIFF
--- a/cmd/earthly/org_cmds.go
+++ b/cmd/earthly/org_cmds.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"github.com/earthly/earthly/config"
 	"os"
 	"strings"
 	"text/tabwriter"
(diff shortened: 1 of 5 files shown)


cmd/earthly/org_cmds.go
```diff
@@ -448,3 +455,53 @@ func (app *earthlyApp) actionOrgMemberRemove(cliCtx *cli.Context) error {
...
+	if cliCtx.NArg() != 1 {
+		return errors.New("invalid number of arguments provided")
+	}
+	org := cliCtx.Args().Get(0)
+	cloudClient, err := app.newCloudClient()
+	if err != nil {
+		return err
+	}
+
+	allOrgs, err := cloudClient.ListOrgs(cliCtx.Context)
```
alexcb, line 470: can `GetOrgID` be used instead? It performs the same `if o.Name == org` logic inside the function.

vladaionescu (approved):
Apart from Alex's comment, this looks good to me!

vladaionescu (commented):
Oh.. can you add a changelog entry and also edit the docs too (need to change the configuration reference page and the command reference page)? You can now edit the docs in the same PR.

cmd/earthly/org_cmds.go
dchw: done!

alexcb (approved):

PR #3075, opened 2023-07-07
Org select suggestion in warning

vladaionescu:


dchw (commented):
Probably better wording. LGTM

PR #3078, opened 2023-07-07
Simone/clojure example

Velutas:


nelsam (commented):
It looks like you might want to add this to the top level `Earthfile` [here](https://github.com/earthly/earthly/blob/main/Earthfile#L822-L845). Then our GHA tests should (I hope) run the example.

PR #3083, opened 2023-07-10
test adding from branch instead of fork

Velutas:


examples/clojure/Earthfile
```diff
@@ -0,0 +1,17 @@
+VERSION 0.7 
+FROM clojure:temurin-8-lein
+WORKDIR /var/app
+
+ARG --global CLOJURE_STANDALONE_ARTIFACT="target/uberjar/earthly-example-0.1.0-SNAPSHOT-standalone.jar"
```
alexcb, line 5: It's generally better to avoid using `ARG --global` since changing this value will change the cache key hash of all targets (which will bust the cache).

I wonder if this `ARG` is even needed anywhere? Instead you could use a wildcard, e.g.

```
VERSION 0.7

a:
  FROM alpine
  RUN mkdir -p target/uberjar/
  RUN echo "bogus data" > target/uberjar/earthly-example-0.1.0-SNAPSHOT-standalone.jar
  SAVE ARTIFACT target/uberjar/earthly-example-*-SNAPSHOT-standalone.jar earthly-example.jar

b:
  FROM alpine
  COPY +a/earthly-example.jar .
  RUN cat earthly-example.jar
```

Additionally, is there a way you can extract the version (e.g. `0.1.0`) from the jar? if so, you could then bake in the version into the docker tag, e.g.

```
ARG tag="$(./extract-version earthly-example.jar)"
SAVE IMAGE clojure-example:$tag
```

examples/clojure/Earthfile
```diff
@@ -0,0 +1,18 @@
...
+    RUN lein uberjar
+    SAVE ARTIFACT target/uberjar/earthly-example-*-standalone.jar earthly-example-standalone.jar
+
+docker:
+    FROM openjdk:22-slim-bullseye
+    RUN apt update && apt install zip -y 
+    COPY +build/earthly-example-standalone.jar earthly-example-standalone.jar
+    EXPOSE 8080
+    ENTRYPOINT java -Dserver.port=8080 -jar earthly-example-standalone.jar
+    COPY extractJarVersion extractJarVersion
```
alexcb, line 16: nit: can you move this script (and dependencies, e.g. `zip`) into the `build` target?

That way you can produce a smaller docker image (which won't include these helper scripts)

It could look something like:

```
build:
    RUN apt update && apt install zip -y 
    COPY earthly-example/ .
    RUN lein uberjar
    RUN ./extractJarVersion target/uberjar/earthly-example-*-standalone.jar > version
    SAVE ARTIFACT target/uberjar/earthly-example-*-standalone.jar earthly-example-standalone.jar
    SAVE ARTIFACT version

docker:
    FROM openjdk:22-slim-bullseye
    COPY +build/* .
    EXPOSE 8080
    ENTRYPOINT java -Dserver.port=8080 -jar earthly-example-standalone.jar
    SAVE IMAGE clojure-example:$(cat version)
```

examples/clojure/extractJarVersion
```diff
@@ -0,0 +1,3 @@
+# Outputs the clojure project version from a passed in .jar file
```
alexcb, line 1: nit: missing shell interrupter

```
#!/bin/sh
set -e
```

examples/clojure/extractJarVersion
```diff
@@ -0,0 +1,3 @@
...
+version=$(unzip -p $1 META-INF/MANIFEST.MF | grep Leiningen-Project-Version: | sed 's/^.*: //' | tr -d '\r')
+echo $version
```
alexcb, line 3: nit: we try to keep all our text files as POSIX files (which require a final newline). It makes it easier to use programs like `cat`.

examples/clojure/extractJarVersion
```diff
@@ -0,0 +1,3 @@
+# Outputs the clojure project version from a passed in .jar file
+version=$(unzip -p $1 META-INF/MANIFEST.MF | grep Leiningen-Project-Version: | sed 's/^.*: //' | tr -d '\r')
```
alexcb, line 2: it would be good to test version was correctly extracted. perhaps:

```
test -n "$version" || (echo >&2 "failed to extract version" && exit 1)
```

PR #3088, opened 2023-07-12
Use selected org in satellite cmds

brandonSc:
The selected org set in the config (via the new `org selected` command) was not being used in satellite commands.

### DIFF
--- a/cmd/earthly/org_cmds.go
+++ b/cmd/earthly/org_cmds.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"fmt"
-	"github.com/earthly/earthly/config"
 	"os"
 	"strings"
 	"text/tabwriter"
@@ -11,6 +10,7 @@ import (
 	"github.com/urfave/cli/v2"
 
 	"github.com/earthly/earthly/cloud"
+	"github.com/earthly/earthly/config"
 )
 
 func (app *earthlyApp) orgCmds() []*cli.Command {
@@ -120,12 +120,14 @@ func (app *earthlyApp) orgCmds() []*cli.Command {
 			Name:      "select",
 			Usage:     "Select the default organization *beta*",
 			UsageText: "earthly [options] org select <org-name>",
+			Aliases:   []string{"s"},
 			Action:    app.actionOrgSelect,
 		},
 		{
 			Name:      "unselect",
 			Usage:     "Unselects the default organization *beta*",
 			UsageText: "earthly [options] org select <org-name>",
+			Aliases:   []string{"uns"},
 			Action:    app.actionOrgUnselect,
 		},
 	}
--- a/cmd/earthly/satellite_cmds.go
+++ b/cmd/earthly/satellite_cmds.go
@@ -4,13 +4,13 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/dustin/go-humanize"
 	"math/rand"
 	"os"
 	"strings"
 	"text/tabwriter"
 	"time"
 
+	"github.com/dustin/go-humanize"
 	"github.com/fatih/color"
 	"github.com/pkg/errors"
 	"github.com/urfave/cli/v2"
@@ -404,6 +404,14 @@ func (app *earthlyApp) getSatelliteOrg(ctx context.Context, cloudClient *cloud.C
 		}
 		return app.orgName, orgID, nil
 	}
+	if orgName = app.cfg.Global.Org; orgName != "" {
+		g
+		orgID, err = cloudClient.GetOrgID(ctx, orgName)
+		if err != nil {
+			return "", "", errors.Wrapf(err, "failed resolving ID for org '%s'", orgName)
+		}
+		return orgName, orgID, nil
+	}
 	orgName, orgID, err = cloudClient.GuessOrgMembership(ctx)
 	if err != nil {
 		return "", "", errors.Wrap(err, "could not guess default org")
@@ -506,7 +514,7 @@ func (app *earthlyApp) actionSatelliteLaunch(cliCtx *cli.Context) error {
 
 	app.console.Printf("Launching Satellite %q with auto-updates set to run at %s (%s)\n",
 		app.satelliteName, localWindow, zone)
-	app.console.Printf("Please wait...\n")
+	app.console.Printf("This may take a few minutes...\n")
 
 	err = cloudClient.LaunchSatellite(cliCtx.Context, cloud.LaunchSatelliteOpt{
 		Name:                    app.satelliteName,
@@ -617,7 +625,7 @@ func (app *earthlyApp) actionSatelliteRemove(cliCtx *cli.Context) error {
 		return fmt.Errorf("could not find %q for deletion", app.satelliteName)
 	}
 
-	app.console.Printf("Destroying Satellite %q. This could take a moment...\n", app.satelliteName)
+	app.console.Printf("Destroying Satellite %q. This may take a few minutes...\n", app.satelliteName)
 	err = cloudClient.DeleteSatellite(cliCtx.Context, app.satelliteName, orgName)
 	if err != nil {
 		if errors.Is(err, context.Canceled) {
--- a/cmd/earthly/satellite_cmds.go
+++ b/cmd/earthly/satellite_cmds.go
@@ -404,12 +404,12 @@ func (app *earthlyApp) getSatelliteOrg(ctx context.Context, cloudClient *cloud.C
 		}
 		return app.orgName, orgID, nil
 	}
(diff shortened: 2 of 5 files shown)


cmd/earthly/org_cmds.go
```diff
@@ -120,12 +120,14 @@ func (app *earthlyApp) orgCmds() []*cli.Command {
...
 			Usage:     "Select the default organization *beta*",
 			UsageText: "earthly [options] org select <org-name>",
+			Aliases:   []string{"s"},
 			Action:    app.actionOrgSelect,
 		},
 		{
 			Name:      "unselect",
 			Usage:     "Unselects the default organization *beta*",
 			UsageText: "earthly [options] org select <org-name>",
+			Aliases:   []string{"uns"},
```
brandonSc, line 130: These aliases match the similar select/unselect commands for satellites

cmd/earthly/satellite_cmds.go
```diff
@@ -404,6 +404,13 @@ func (app *earthlyApp) getSatelliteOrg(ctx context.Context, cloudClient *cloud.C
 		}
 		return app.orgName, orgID, nil
 	}
+	if orgName = app.cfg.Global.Org; orgName != "" {
```
brandonSc, line 407: There's an `if app.OrgName != ""` block above, which means setting the `--org` flag or using the `EARTHLY_ORG` env var still takes precedence over the value set in the config file. If neither are set, we still run the old "guess" logic.

vladaionescu (approved):

cmd/earthly/satellite_cmds.go
alexcb: nit: can you move `orgName = app.cfg.Global.` out of the `if`, to make it more clear it's assigning to the existing `orgName`.

(yes it will immediately be replaced once `orgName, orgID, err = cloudClient.GuessOrgMembership(ctx)` is hit, but the code immediately following the if might hypothetically change in the future.)

alexcb (changes requested):
I agree with Sam's point regarding testing (which is easy enough to expand to cover the multiple ways we can select an org). Without it, we might accidentally break this in the future.

alexcb:
> It should be okay to add a couple lines to select the org, run sat ls or something, and then unselect the org near here: https://github.com/earthly/earthly/blob/main/scripts/tests/earthly-image.sh#L89-L93

This might not actually work, because this is being done inside a docker container (which won't persist any state (config files) between runs).

perhaps something like this could work:

```
"$FRONTEND" run --rm --privileged -e EARTHLY_TOKEN="${EARTHLY_TOKEN}" --entrypoint="" "${EARTHLY_IMAGE}" /bin/sh -c "earthly org select <org> && sat inspect core-test"
```

alexcb:
It might be better to make a completely new `satellite-ingration.sh` test; you can reuse the script setup parts from https://github.com/earthly/earthly/blob/main/scripts/tests/cloud-docker-credentials-integration.sh

scripts/tests/satellites-integration.sh
alexcb, line 1: how is this test being run? I don't see an entry for it under `.github/workflows/reusable-misc-tests.yml` (which is how `./scripts/tests/cloud-docker-credentials-integration.sh` is called)

scripts/tests/satellites-integration.sh
```diff
@@ -0,0 +1,19 @@
...
+if [ -z "${EARTHLY_TOKEN:-}" ]; then
+  echo "using EARTHLY_TOKEN from earthly secrets"
+  EARTHLY_TOKEN="$(earthly secrets --org earthly-technologies --project core get earthly-token-for-satellite-tests)"
+  export EARTHLY_TOKEN
+fi
+test -n "$EARTHLY_TOKEN" || (echo "error: EARTHLY_TOKEN is not set" && exit 1)
+
+"$FRONTEND" run --rm --privileged -e EARTHLY_TOKEN="${EARTHLY_TOKEN}" --entrypoint="" "${EARTHLY_IMAGE}" /bin/sh -c "earthly org select earthly-technologies"
+
+"$FRONTEND" run --rm --privileged -e EARTHLY_TOKEN="${EARTHLY_TOKEN}" --entrypoint="" "${EARTHLY_IMAGE}" /bin/sh -c "earthly sat select core-test"
```
vladaionescu, line 19: Did you mean to hook this test up to the CI?

vladaionescu (approved):

scripts/tests/satellites-integration.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+set -euo pipefail # don't use -x as it will leak the mirror credentials
+
+# to run this locally; in the root of the repo:
+#   ./earthly +earthly-docker && EARTHLY_IMAGE="earthly/earthly:dev-$(git rev-parse --abbrev-ref HEAD | sed 's/\//_/g')" scripts/tests/satellites-integration.sh
+
+FRONTEND=${FRONTEND:-docker}
+EARTHLY_IMAGE=${EARTHLY_IMAGE:-earthly/earthly:dev-main}
```
alexcb, line 8: instead of using the earthly image, I think it's better to just use the earthly cli directly. This is why I made a follow up comment: 

> It might be better to make a completely new satellite-ingration.sh test; you can reuse the script setup parts from https://github.com/earthly/earthly/blob/main/scripts/tests/cloud-docker-credentials-integration.sh

Using the earthly image test makes it difficult since state (config files) are not stored between runs. e.g.

`"$FRONTEND" run --rm --privileged -e EARTHLY_TOKEN="${EARTHLY_TOKEN}" --entrypoint="" "${EARTHLY_IMAGE}" /bin/sh -c "earthly org select earthly-technologies"` does not persist the select org to following commands, which doesn't provide the test coverage that's needed.

alexcb (changes requested):
Using the earthly image to test this seems like the wrong direction; we should use the earthly cli directly in satellites-integration.sh

scripts/tests/satellites-integration.sh
alexcb: here's a more complete example:

```
#!/usr/bin/env bash
set -eu

earthly=${earthly:=earthly}
earthly=$(realpath "$earthly")
echo "running tests with $earthly"

# prevent the self-update of earthly from running (this ensures no bogus data is printed to stdout,
# which would mess with the secrets data being fetched)
date +%s > /tmp/last-earthly-prerelease-check

# ensure earthly login works (and print out who gets logged in)
test -n "$EARTHLY_TOKEN"
"$earthly" account login

# test --org
"$earthly" sat --org <org> --sat <sat> inspect

# test EARTHLY_ORG env
EARTHLY_ORG=ORG "$earthly" sat --sat <sat> inspect

# test org select
"$earthly" org select <foo>
"$earthly" sat --sat <sat> inspect
```

scripts/tests/satellites-integration.sh
brandonSc: Yes, I forgot 😞

scripts/tests/satellites-integration.sh
brandonSc: Good point. It only seemed to be working beforehand because it was relying on the "org guessing" logic. Working on it.

alexcb: It might be a good idea to `2>&1 | tee output` then make sure `Auto-selecting the default org` doesn't show up in the output.

brandonSc: I tried adding the check for `Auto-selecting the default org` but it was getting a bit too verbose for my liking and we want to remove that auto-selection soon anyways. But the new tests pass (thank you for helping write those!) and they are hooked-in to CI.

scripts/tests/satellites-integration.sh
```diff
@@ -0,0 +1,30 @@
...
+
+# test --org
+"$earthly" sat --org earthly-technologies inspect core-test
+
+# test EARTHLY_ORG env
+EARTHLY_ORG=earthly-technologies "$earthly" sat inspect core-test
+
+# test org select
+"$earthly" org select earthly-technologies
+"$earthly" sat select core-test
```
alexcb, line 30: is the `select` intentional here? or should it be `inspect`?
```suggestion
"$earthly" sat inspect core-test
```

alexcb: We should actually test both; can you:

```
# test inspect can use the org name from the config file
"$earthly" org select earthly-technologies
"$earthly" sat inspect core-test

# test the sat select works correctly uses the config's org
"$earthly" sat select core-test
"$earthly" satellite ls | grep '^\* \+core-test'
```

to ensure that the core-test satellite is correctly displayed as being selected.

alexcb (commented):
looking good; just one more suggestion for the test.

PR #3090, opened 2023-07-12
Misleading issues when Docker is not running

brandonSc:
A user reported issues with misleading errors and hanging when trying to run a local build with no Docker daemon.

This now shows the following in that case:
Screen Shot 2023-07-12 at 3 21 43 PM

However, if using satellites without docker running, the connection still works:
Screen Shot 2023-07-12 at 3 22 21 PM

Same with remote buildkit:
Screen Shot 2023-07-12 at 3 23 20 PM

I didn't test Podman, however, I assume we should have a buildkit address set to something similar as with docker in that case.

### DIFF
--- a/cmd/earthly/main.go
+++ b/cmd/earthly/main.go
@@ -448,7 +448,7 @@ func (app *earthlyApp) parseFrontend(cliCtx *cli.Context, cfg *config.Config) er
 		app.containerFrontend = stub
 
 		if !app.verbose {
-			console.Printf("No frontend initialized. Use --verbose to see details\n")
+			console.Printf("No frontend initialized. Docker or Podman not detected or had errors. Use --verbose to see details\n")
 		}
 		console.VerbosePrintf("%s frontend initialization failed due to %s", app.cfg.Global.ContainerFrontend, origErr.Error())
 		return nil
--- a/cmd/earthly/satellite_cmds.go
+++ b/cmd/earthly/satellite_cmds.go
@@ -4,13 +4,13 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/dustin/go-humanize"
 	"math/rand"
 	"os"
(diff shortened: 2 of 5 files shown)


brandonSc:
Oops apparently I wasn't on a fresh branch like I thought I was.

cmd/earthly/main.go
```diff
@@ -448,7 +448,7 @@ func (app *earthlyApp) parseFrontend(cliCtx *cli.Context, cfg *config.Config) er
 		app.containerFrontend = stub
 
 		if !app.verbose {
-			console.Printf("No frontend initialized. Use --verbose to see details\n")
+			console.Printf("No frontend initialized. Docker or Podman not detected or had errors. Use --verbose to see details\n")
```
alexcb, line 451:
```suggestion
			console.Printf("Unable to detect Docker or Podman. Use --verbose to see details (or errors)\n")
```

cmd/earthly/satellite_cmds.go
```diff
@@ -404,6 +404,13 @@ func (app *earthlyApp) getSatelliteOrg(ctx context.Context, cloudClient *cloud.C
 		}
 		return app.orgName, orgID, nil
 	}
+	if orgName = app.cfg.Global.Org; orgName != "" {
+		orgID, err = cloudClient.GetOrgID(ctx, orgName)
+		if err != nil {
+			return "", "", errors.Wrapf(err, "failed resolving ID for org '%s'", orgName)
```
alexcb, line 410:
```suggestion
			return "", "", errors.Wrapf(err, "failed resolving ID for org %q", orgName)
```

alexcb:
> Oops apparently I wasn't on a fresh branch like I thought I was.

you can use a `git rebase main && git push --force-with-lease` instead. Either way can you take a look at the suggestions @brandonSc

Issue #3094, opened 2023-07-13
File corruption when `LOCALLY` doing a `COPY` to overwrite an existing file


rrjjvv:
**What went wrong?**
When doing a conventional artifact `COPY` to an existing (local) destination, the resulting file can become corrupted.

This is specific to instructions run under `LOCALLY`.  The same copy to a non-existent location has no issues.  Doing the equivalent overwrite via saving `AS LOCAL` has no issues.  It is only an artifact copy to an existing location that's problematic.  In the process of coming up with a minimal reproduction, I discovered it is further limited to scenarios where the destination file is larger than the source file.

This is trivial to reproduce (but far from trivial to get there).  I've only attempted to reproduce in `0.7.9`, but I believe it to be an old/existing issue.

```Earthfile
VERSION 0.7

data:
    FROM busybox
    RUN printf "hello" > hello.txt
    SAVE ARTIFACT hello.txt

test:
    LOCALLY
    RUN printf "yo"        > short.txt && \
        printf "greetings" > long.txt
    COPY +data/hello.txt short.txt
    COPY +data/hello.txt long.txt
    # Destination was smaller than the source, so no issue.  Exact size is not an issue either.
    RUN test "$(cat short.txt)" = "hello"
    # The problem (due to larger destination)
    RUN test "$(cat long.txt)" = "hello" || echo "where's my data?"
    RUN test "$(cat long.txt)" = "helloings" && echo "wut?
```

```
$ earthly -v
earthly version v0.7.9 cf170dbe5514ed18ae6cab778aafcde3594470bb linux/amd64; Ubuntu 22.04.2 LTS (Jammy Jellyfish)

$ earthly --no-cache +test
<snip>
               +test *local* | --> RUN printf "yo" > short.txt && printf "greetings" > long.txt
             busybox | --> Load metadata busybox linux/amd64
               +data | --> FROM busybox
               +data | [----------] 100% resolve docker.io/library/busybox@sha256:2376a0c12759aa1214ba83e771ff252c7b1663216b192fbe5e0fb364e952f85c
               +data | --> RUN printf "hello" > hello.txt
               +test | --> COPY +data/hello.txt short.txt
               +test | --> COPY +data/hello.txt long.txt
               +test *local* | --> RUN test "$(cat short.txt)" = "hello"
               +test *local* | --> RUN test "$(cat long.txt)" = "hello" || echo "where's my data?"
               +test *local* | where's my data?
[...]

PR #3095, opened 2023-07-13
Tests to verify correct (local) overwrite behavior

rrjjvv:
Verifies https://github.com/earthly/buildkit/pull/19, which in turn fixes https://github.com/earthly/earthly/issues/3094.

This ensures that an existing file is truncated prior to writing; otherwise, if the source (`X`) is smaller than the destination (`Y`), the destination contains all of `X` bytes at the beginning, and `Y-X` bytes of the original destination at the end.

### DIFF
--- a/tests/local/Earthfile
+++ b/tests/local/Earthfile
@@ -90,6 +90,30 @@ test-save-local-artifact-with-workdir:
     COPY +save-local-artifact-with-workdir/data .
     RUN cat data | grep '^ddd$'
 
+#################################################################################################################
+# tests for testing overwriting existing local files
+
+some-artifact:
+    RUN printf '4321' > four-bytes.txt
+    SAVE ARTIFACT four-bytes.txt
+
+test-local-overwrites:
+    LOCALLY
+    RUN rm -fr              /tmp/earthly-test-21f37581-de77-4f81-ac78-4a2a617f7eda && \
+        mkdir -p            /tmp/earthly-test-21f37581-de77-4f81-ac78-4a2a617f7eda && \
+        printf "12345678" > /tmp/earthly-test-21f37581-de77-4f81-ac78-4a2a617f7eda/eight.txt && \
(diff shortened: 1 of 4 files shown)


rrjjvv:
Hopefully I followed the correct procedure; let me know if I did anything wrong.  I assumed that I was *not* supposed to push the results of `+update-buildkit`.

Note that the test will fail without the buildkit changes:
```
./tests/local+test-local-overwrites *local* | WARN tests/local/Earthfile line 112:4: The command 'RUN set -e; for i in eight four two; do cat /tmp/earthly-test-21f37581-de77-4f81-ac78-4a2a617f7eda/${i}.txt | grep '^4321$'; done' failed: error calling LocalhostExec: exit code: 1
Error: build target: build main: failed to solve: tests/local/Earthfile line 112:4 apply RUN: unlazy force execution: error calling LocalhostExec: exit code: 1
in		github.com/rrjjvv/earthly/tests/local:local-copy-overwrite-fix+test-local-overwrites
```

Of course, they should pass when they're built against the changes in the other PR.

tests/local/Earthfile
```diff
@@ -90,6 +90,30 @@ test-save-local-artifact-with-workdir:
...
     RUN cat data | grep '^ddd$'
 
+#################################################################################################################
+# tests for testing overwriting existing local files
+
+some-artifact:
+    RUN printf '4321' > four-bytes.txt
+    SAVE ARTIFACT four-bytes.txt
+
+test-local-overwrites:
```
alexcb, line 100: can you add a `BUILD +test-local-overwrites` to the `all` target, which will then run this with our integration tests.

rrjjvv: whoops, and done

alexcb:
it's a bit of a multi-repo dance, but could you do a `./earthly +update-buildkit --BUILDKIT_GIT_SHA=cf348cf54d34478bb7c0ac62b0036802a6d59dd8` now that the buildkit change has been merged.

alexcb (approved):

PR #3098, opened 2023-07-13
Improve output for indefinite auth tokens

mikejholly:
```
$ ./build/linux/amd64/earthly account create-token --expiry never delete-me-4
created token "UEBvw7Yau1WvDBu9BJQhQNJGX7JE8pmrmjeAwXMfYxKTMvXGC8SwS1PbWWnRLe29" which will never expire; save this token somewhere, it can't be viewed again (only reset)

$ ./build/linux/amd64/earthly account create-token delete-me-5 
created token "ELdaUHXyXMbEJ86PLQgI6hsxzwqjg6DcWXiYyAqIZ7XjmqUvtKNrcVjBAwZTux1y" which will never expire; save this token somewhere, it can't be viewed again (only reset)

$ ./build/linux/amd64/earthly account list-tokens             
Token Name   Read/Write  Expiry
local        r           2023-06-14T23:38 UTC *expired*
delete-me    r           2024-07-12T22:12 UTC
local-2      rw          2023-09-13T21:43 UTC
delete-me-2  r           Never
local-3      rw          2023-09-28T18:08 UTC
delete-me-4  r           Never
delete-me-5  r           Never
```

### DIFF
--- a/cloud/account.go
+++ b/cloud/account.go
@@ -18,9 +18,10 @@ import (
 
 // TokenDetail contains token information
 type TokenDetail struct {
-	Name   string
-	Write  bool
-	Expiry time.Time
+	Name       string
+	Write      bool
+	Expiry     time.Time
+	Indefinite bool
 }
 
 func (c *Client) ListPublicKeys(ctx context.Context) ([]string, error) {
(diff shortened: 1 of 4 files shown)


idodod (approved):
Nice!

alexcb (approved):
code looks fine, but I think defaulting to non-expiry is a bad idea from a security perspective.

PR #3101, opened 2023-07-14
use acbgrep to make test failures more clear

alexcb:
`acbgrep` from https://github.com/alexcb/acbtest; is a wrapper around grep which will display a failure message if grep fails (rather than silently exiting with an error code).

When data is pipped in, it will also display the contents of the data that is being grepped -- this makes it easier to see the full output of earthly if grep fails to find an expected string.

### DIFF
--- a/scripts/acbtest/update.sh
+++ b/scripts/acbtest/update.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+set -e
+
+if ! which curl 2>/dev/null >&2; then
+  echo >&2 "curl is required; please install it."
+  exit 1
+fi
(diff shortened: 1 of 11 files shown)


scripts/acbtest/update.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+set -e
+
+if ! which curl 2>/dev/null >&2; then
+  echo >&2 "the md5sum command is required; please install it."
```
alexcb, line 5:
```suggestion
  echo >&2 "curl is required; please install it."
```

PR #3102, opened 2023-07-17
.github: ensure there's a binary for the staging deploy to use

nelsam:


.github/workflows/ci-staging-deploy.yml
```diff
@@ -9,8 +9,21 @@ on:
       - 'docs/**'
 
 jobs:
+  build-earthly:
+    # TODO: create the release binary, use it in tests, _then_ release it. It's
+    # kind of silly to builld earthly twice here and release it without any
```
alexcb, line 14: not sure I fully understand this comment; are you just pointing out this job doesn't depend on any of the GHA test jobs?

alexcb (approved):

.github/workflows/ci-staging-deploy.yml
nelsam: Good point, it's not super clear. The point is that the `ci-docker-ubuntu` tests build and release this same binary already. Ideally, this job would just be releasing the artifacts that were used in the `ci-docker-ubuntu` workflow.

Instead, we're re-building (and re-uploading) the same artifacts, then using them to build earthly again, and then the artifacts that we release aren't used in any kinds of tests.

I could clarify that in the comment (or just remove it and make a ticket).

alexcb: I'm in favour of removing the comment and creating a ticket.

If I'm not mistaken, the initial build is only for linux-amd64, but this build is for all platforms?

nelsam: For clarification's sake:

The `build-earthly` job builds just for `linux-amd64`. It's identical to the job being used in `ci-docker-ubuntu`, and they push to the same dockerhub repo (and the same tag).

The release job builds for all platforms (including `linux-amd64`).

My comment was about the fact that we re-build the same binary as `ci-docker-ubuntu`. It would be great if we could wait until that binary passes (some) tests before the staging release runs. It would be even better if the `linux-amd64` binary included in the release was the binary that passed the tests.

I'll make a ticket.

Issue #3099, opened 2023-07-13
`earthly account login` should not log you into a different user.


alexcb:
```
EARTHLY_TOKEN="$(earthly secrets --org earthly-technologies --project core get earthly-token-for-satellite-tests)" ./build/linux/amd64/earthly account login; test -n "$EARTHLY_TOKEN" || echo token is empty; ./build/linux/amd64/earthly org ls; ./build/linux/amd64/earthly account login; ./build/linux/amd64/earthly org ls
```

will first log you in using an EARTHLY_TOKEN:

```
Logged in as "other-service+earthly@earthly.dev" using read-only-token auth
Log sharing is enabled by default. If you would like to disable it, run:

	earthly config global.disable_log_sharing true
```

Then note that the token is empty:

```
token is empty
```

then as expected you can see your orgs:

```
user:other-service+earthly@earthly.dev  admin
earthly-technologies                    member
```

but if you rerun login (without the EARTHLY_TOKEN), even though you have a valid jwt for other-service+earthly@earthly.dev; it will override your login, whcih will fallback to using an ssh key:

```
Logged in as "alex@earthly.dev" using ssh auth
Log sharing is enabled by default. If you would like to disable it, run:

	earthly config global.disable_log_sharing true
```

which then shows other orgs (since i'm logged in with a different user now):

```
acb                   admin
earthly-technologies  admin
alex                  admin
alexcb                admin
hush-co               member
test123-vlad          member
alexcb-2362           admin
```

## What I expect

running `earthly account login` would keep you logged in to the current user if you are already logged in (even if the `EARTHLY_TOKEN` is no longer set)

alexcb:
This may have been fixed under v0.7.13 but it's not completely clear -- we should test this before closing it.

PR #3104, opened 2023-07-17
don't change the currently logged in user

alexcb:
if there's a valid auth.jwt, earthly should not re-login (potentially with different credentials).

### DIFF
--- a/cloud/account.go
+++ b/cloud/account.go
@@ -19,10 +19,10 @@ import (
 type AuthMethod string
 
 const (
-	SSHAuthMethod       AuthMethod = "ssh"
-	PasswordAuthMethod  AuthMethod = "password"
-	TokenAuthMethod     AuthMethod = "token"
-	CachedJWTAuthMethod AuthMethod = "cached jwt"
+	AuthMethodSSH       AuthMethod = "ssh"
+	AuthMethodPassword  AuthMethod = "password"
+	AuthMethodToken     AuthMethod = "token"
+	AuthMethodCachedJWT AuthMethod = "cached jwt"
 )
 
 // TokenDetail contains token information
@@ -163,7 +163,7 @@ func (c *Client) WhoAmI(ctx context.Context) (string, AuthMethod, bool, error) {
 	}
 	authMethod := c.lastAuthMethod
 	if authMethod == "" {
-		authMethod = CachedJWTAuthMethod
+		authMethod = AuthMethodCachedJWT
 	}
 	return email, authMethod, writeAccess, nil
 }
--- a/cloud/account.go
+++ b/cloud/account.go
@@ -16,6 +16,15 @@ import (
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
+type AuthMethod string
+
+const (
+	SSHAuthMethod       AuthMethod = "ssh"
+	PasswordAuthMethod  AuthMethod = "password"
+	TokenAuthMethod     AuthMethod = "token"
+	CachedJWTAuthMethod AuthMethod = "cached jwt"
+)
+
 // TokenDetail contains token information
 type TokenDetail struct {
 	Name       string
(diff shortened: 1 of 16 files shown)


cmd/earthly/account_cmds.go
```diff
@@ -535,6 +535,16 @@ func (app *earthlyApp) actionAccountLogin(cliCtx *cli.Context) error {
 		return err
 	}
 
+	loggedInEmail, authType, _, err := cloudClient.WhoAmI(cliCtx.Context)
+	if err == nil {
+		// already logged in, don't re-attempt a login
+		app.console.Printf("Logged in as %q using %s auth\n", loggedInEmail, authType)
```
alexcb, line 541: this is the most significant logic-related change.

alexcb:
fixes #3099

scripts/tests/secrets-integration.sh
```diff
@@ -58,7 +59,7 @@ echo testing that the ssh-agent only contains a single key
 test "$(ssh-add -l | wc -l)" = "1"
 
 echo "testing earthly account login works (and is using the earthly-manitou account)"
-"$earthly" account login 2>&1 | acbgrep 'other-service+earthly-manitou@earthly.dev'
+"$earthly" account login 2>&1 | acbgrep 'Logged in as "other-service+earthly@earthly.dev"'
```
alexcb, line 62: whoops.... I think this is wrong. this shouldn't have changed :facepalm:

scripts/tests/secrets-integration.sh
```diff
@@ -40,6 +41,7 @@ ID_RSA=$("$earthly" secrets --org earthly-technologies --project core get -n sec
 
 # now that we grabbed the manitou credentials, unset our token, to ensure that we're only testing using manitou's credentials
 unset EARTHLY_TOKEN
+"$earthly" account logout
```
alexcb, line 44: this is now needed to force the cached jwt to be discarded.

cloud/account.go
```diff
@@ -16,6 +16,15 @@ import (
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
+type AuthMethod string
+
+const (
+	SSHAuthMethod       AuthMethod = "ssh"
+	PasswordAuthMethod  AuthMethod = "password"
+	TokenAuthMethod     AuthMethod = "token"
+	CachedJWTAuthMethod AuthMethod = "cached jwt"
```
nelsam, line 25: super minor nit: I believe that idiomatic go puts the type first, e.g. [http methods](https://pkg.go.dev/net/http#pkg-constants). `AuthMethodSSH` and such are probably more idiomatic.

cloud/auth.go
```diff
@@ -23,23 +23,32 @@ import (
...
 	case c.email != "" && c.password != "":
+		authMethod = "password"
 		err = c.loginWithPassword(ctx)
 	case c.authCredToken != "":
+		authMethod = "token"
 		err = c.loginWithToken(ctx)
 	default:
+		authMethod = "ssh"
 		err = c.loginWithSSH(ctx)
 	}
```
nelsam, line 39: It feels like this is a good opportunity to move this to some form of `authMethod` function/method, e.g.

```go
func (c *Client) authMethod(ctx context.Context) (AuthMethod, error) {
    if c.email != "" && c.password != "" {
        return AuthMethodPassword, c.loginWithPassword(ctx)
    }
    if c.authCredToken != "" {
        return AuthMethodToken, c.loginWithToken(ctx)
    }
    return AuthMethodSSH, c.loginWithSSH(ctx)
}
```

nelsam (approved):
Minor style nits aside, lgtm

cloud/auth.go
alexcb: good idea!

alexcb: I also totally missed that `authMethod = "password"` wasn't changed to use the consts.

PR #3105, opened 2023-07-18
Simone/codespaces docker sat tutorial

Velutas:


docs/examples/codespaces.md
```diff
@@ -0,0 +1,111 @@
...
+
+This example was run using the `mcr.microsoft.com/devcontainers/universal:2` image.
+
+1. Update your devcontainer configuration in .devcontainer/devcontainer.json in the root of your project 
+    ````
+    // YOUR OTHER CONFIG HERE
+    ...
+    ...
+    "features": {
+        "ghcr.io/shepherdjerred/devcontainers-features/earthly:1": {
```
vladaionescu, line 85: What does this part do? It seems to be relying on a third party container?

docs/examples/codespaces.md
vladaionescu, line 1: Hey @Velutas - I was thinking it might be better for this to start as an example under the `examples` dir (under the root of the repo, not under `docs/`). That way we can include a fully working devcontainer configuration.

Although now that I think about it, Codespaces might expect that under `.devcontainer` in the root of the repo for it to work at all. If that's the case, I think the best option would be a completely new repo for this example that we could link from the documentation. Perhaps we can call it `codespaces-example`.

docs/examples/codespaces.md
Velutas: It's a feature which is a set of installation code for Codespaces. In this case, yes it is a third party that created the feature for Earthly installation. That's a good point to bring up since we don't have control of what goes in to that feature so I'll update the Example to use the official features + how to install Earthly on top of that instead

vladaionescu: I'm reading around about this "feature" concept for the first time :D

Perhaps the best user experience would be if we support a feature like this one ourselves, and recommend it officially.

I'll check with Jerred Shepherd on Slack whether he'd be open to share his work as open-source and we can credit him.

docs/examples/examples.md
```diff
@@ -37,6 +43,7 @@ Please note that these examples, although similar, are distinct from the ones us
 
 <!-- vale HouseStyle.Spelling = NO -->
 * [c](https://github.com/earthly/earthly/tree/main/examples/c)
+* [clojure](https://github.com/earthly/earthly/tree/main/examples/clojure)
```
Velutas, line 46: Adding as I noticed it was missing here.

docs/examples/examples.md
```diff
@@ -13,6 +13,12 @@ Examples of integrating Earthly into various CI systems can be found on the foll
 
 For more general information on CI systems not listed above, see the [CI integration guide](../ci-integration/overview.md).
 
+## Examples by dev environments
+
+Examples of how Earthly can be integrated into different dev environments
+
+* [codespaces](https://github.com/earthly/codespaces-example) - shows how Earthly can be used in Codespaces and Devcontainers
```
vladaionescu, line 20:
```suggestion
* [GitHub Codespaces](https://github.com/earthly/codespaces-example) - shows how Earthly can be used in GitHub Codespaces and Devcontainers
```

vladaionescu (commented):
Looks good overall! Some comments below...

There's a comment in there saying

```
<!-- NOTE: If you change this, please also change examples/README.md -->
```

Can you update that too? Unfortunately we have some duplication there that we need to maintain.

Velutas:
> Looks good overall! Some comments below...
> 
> There's a comment in there saying
> 
> ```
> 
> ```
> 
> Can you update that too? Unfortunately we have some duplication there that we need to maintain.

For the clojure example it is already there just needed to merge onto main to get the current version. I've updated now so that it will show up

nelsam (approved):

PR #3112, opened 2023-07-18
Selected org in config works with secrets

brandonSc:
The `org select` command now propagates to other earthly commands for `secrets`, `registry`, and `project`. 

The secrets needed some additional tweaking to support the selected org. That's because we always assumed that user-space secrets would be addressed and displayed with a `/user` prefix. Now that we can select a personal org, user-space secrets can be addressed and displayed without the `/user` prefix. The prefix is still supported though, and works even when a different org is selected.


cmd/earthly/secret_cmds.go
```diff
@@ -276,17 +285,55 @@ func (app *earthlyApp) fullSecretPath(path string) (string, error) {
...
-	if app.orgName == "" {
-		return "", fmt.Errorf("the --org flag is required")
+	orgName, projectName, isPersonal, err := app.getOrgAndProject(ctx, cloudClient)
+	if err != nil {
+		return "", err
 	}
-	if app.projectName == "" {
-		return "", fmt.Errorf("the --project flag is required")
+
+	// TODO this is a bit ugly, consider having getOrgAndProject return isPersonal bool
```
alexcb, line 293: is this comment still valid? I see `isPersonal` comes from a few lines up where `app.getOrgAndProject` is called.

cmd/earthly/secret_cmds.go
```diff
@@ -276,17 +285,55 @@ func (app *earthlyApp) fullSecretPath(path string) (string, error) {
...
+}
+
+func (app *earthlyApp) getOrgAndProject(ctx context.Context, client *cloud.Client) (org, project string, isPersonal bool, err error) {
+	if app.orgName != "" {
+		org = app.orgName
+	} else if app.cfg.Global.Org != "" {
+		org = app.cfg.Global.Org
+	}
+	if org == "" {
+		return org, project, isPersonal, errors.Errorf("the --org flag is required")
```
alexcb, line 315: should this error message be more generic? It's not clear if a user could set the org instead via `earthly org select`.

scripts/tests/secrets-integration.sh
```diff
@@ -75,9 +75,20 @@ EOF
...
 "$earthly" secrets --org manitou-org --project earthly-core-integration-test set my_test_file "secret-value"
-
 "$earthly" secrets --org manitou-org --project earthly-core-integration-test get my_test_file | acbgrep 'secret-value'
 
+# test set and get with org selected in config file
+"$earthly" org select manitou-org
+"$earthly" secrets --project earthly-core-integration-test get my_test_file | acbgrep 'secret-value'
+"$earthly" secrets --project earthly-core-integration-test set my_other_file "super-secret-value"
+"$earthly" secrets --project earthly-core-integration-test get my_other_file | acbgrep 'super-secret-value'
+
```
alexcb, line 85: can you also include some tests to validate the output of `earthly secrets ls`?

alexcb (commented):
looks good, but I think we should expand the tests to cover `earthly secrets ls`.

scripts/tests/secrets-integration.sh
brandonSc: Yep, will do

scripts/tests/secrets-integration.sh
```diff
@@ -77,17 +77,19 @@ EOF
 "$earthly" secrets --org manitou-org --project earthly-core-integration-test set my_test_file "secret-value"
 "$earthly" secrets --org manitou-org --project earthly-core-integration-test get my_test_file | acbgrep 'secret-value'
 
-# test set and get with org selected in config file
+# test secrets with org selected in config file
 "$earthly" org select manitou-org
 "$earthly" secrets --project earthly-core-integration-test get my_test_file | acbgrep 'secret-value'
 "$earthly" secrets --project earthly-core-integration-test set my_other_file "super-secret-value"
 "$earthly" secrets --project earthly-core-integration-test get my_other_file | acbgrep 'super-secret-value'
+"$earthly" secrets --project earthly-core-integration-test ls | acbgrep 'my_test_file'
```
alexcb, line 85: can you make this regex a bit tighter, e.g. `^my_test_file$`?

cmd/earthly/project_cmds.go
```diff
@@ -350,6 +350,8 @@ func (app *earthlyApp) projectOrgName(ctx context.Context, cloudClient *cloud.Cl
 
 	if app.orgName != "" {
 		return app.orgName, nil
+	} else if app.cfg.Global.Org != "" {
+		return app.cfg.Global.Org, nil
```
nelsam, line 354: the `else` is unnecessary here.
```suggestion
	}
	if app.cfg.Global.Org != "" {
		return app.cfg.Global.Org, nil
```

cmd/earthly/secret_cmds.go
```diff
@@ -138,10 +140,17 @@ func (app *earthlyApp) actionSecretsListV2(cliCtx *cli.Context) error {
...
 		display := secret.Path
-		if !strings.HasPrefix(display, "/user/") {
-			prefix := fmt.Sprintf("/%s/%s/", app.orgName, app.projectName)
+		if isPersonal && projectName == "" {
+			display = strings.TrimPrefix(display, "/user/")
+		} else {
+			prefix := fmt.Sprintf("/%s/%s/", orgName, projectName)
 			display = strings.TrimPrefix(display, prefix)
 		}
 		fmt.Println(display)
```
nelsam, line 156: this form of "initialize a variable, then use a series of logic branches to reassign it" suggests that it belongs in a helper function:

```go
func secretDisplay(personal bool, org, proj string, secret *Secret) string {
    if personal && proj == "" {
        return strings.TrimPrefix(secret.Path, "/user/")
    }
    return strings.TrimPrefix(display, fmt.Sprintf("/%s/%s/", org, proj))
}

// then in this code
fmt.Println(secretDisplay(isPersonal, orgName, projName, secret))
```

cmd/earthly/secret_cmds.go
```diff
@@ -276,17 +285,54 @@ func (app *earthlyApp) fullSecretPath(path string) (string, error) {
...
 	}
-	if app.projectName == "" {
-		return "", fmt.Errorf("the --project flag is required")
+
+	if isPersonal && projectName == "" && !strings.HasPrefix(path, "/user") {
+		if path == "/" {
+			return "/user", nil
+		} else {
+			return fmt.Sprintf("/user%s", path), nil
+		}
```
nelsam, line 298:
```suggestion
		}
		return fmt.Sprintf("/user%s", path), nil
```

cmd/earthly/secret_cmds.go
```diff
@@ -276,17 +285,54 @@ func (app *earthlyApp) fullSecretPath(path string) (string, error) {
...
+	//   measure.
+	return fmt.Sprintf("/%s/%s%s", orgName, projectName, path), nil
+}
+
+func (app *earthlyApp) getOrgAndProject(ctx context.Context, client *cloud.Client) (org, project string, isPersonal bool, err error) {
+	if app.orgName != "" {
+		org = app.orgName
+	} else if app.cfg.Global.Org != "" {
+		org = app.cfg.Global.Org
+	}
```
nelsam, line 312: Similar to the other declare/reassign block, this could use a helper. `app.OrgName()` or `app.org()` or something would work.

nelsam (commented):
Some readability comments to go along with Alex's test comments. It mostly looks good though.

nelsam (approved):
LGTM once the `Docker CI Ubuntu` tests pass

alexcb (approved):

PR #3115, opened 2023-07-19
print last accessed at field in list-tokens

idodod:


CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to [Earthly](https://github.com/earthly/earthly) will be doc
 
 ## Unreleased
 
+### Added
+- `earthly account list-tokens` would now show the last time a token was accessed
+
+### Fixed
+- Fixed a bug, where the command to create tokens with a set expiration failed.
```
idodod, line 11: this is from PR https://github.com/earthly/earthly/pull/3113

cmd/earthly/account_cmds.go
```diff
@@ -457,7 +457,11 @@ func (app *earthlyApp) actionAccountListTokens(cliCtx *cli.Context) error {
 				fmt.Fprintf(w, " *expired*")
 			}
 		}
-
+		lastAccessedAtStr := "\t"
+		if !token.LastAccessedAt.UTC().IsZero() {
```
alexcb, line 461: nit: can we rewrite this to be structured like the if/elses above in order to stay consistent.

additionally let's include a specific "Never" rather than an empty string, so it's possible to use tools like `awk`.

e.g.

```
if token.LastAccessedAt.UTC().IsZero() {
    fmt.Fprint(w, "\tNever")
} else {
    fmt.Fprint(w, "\t%s UTC", token.LastAccessedAt.UTC().Format("2006-01-02T15:04"))
}
```

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to [Earthly](https://github.com/earthly/earthly) will be doc
 
 ## Unreleased
 
+### Added
+- `earthly account list-tokens` would now show the last time a token was accessed
```
alexcb, line 8: a little more concise: 
```suggestion
- `earthly account list-tokens` now shows the last time a token was used
```

cmd/earthly/account_cmds.go
```diff
@@ -440,7 +440,7 @@ func (app *earthlyApp) actionAccountListTokens(cliCtx *cli.Context) error {
 	now := time.Now()
 
 	w := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', 0)
-	fmt.Fprintf(w, "Token Name\tRead/Write\tExpiry\n")
+	fmt.Fprintf(w, "Token Name\tRead/Write\tExpiry\tLast Accessed At\n")
```
alexcb, line 443:
```suggestion
	fmt.Fprintf(w, "Token Name\tRead/Write\tExpiry\tLast Used\n")
```

alexcb (commented):
looks good, just a few minor suggestions.

cmd/earthly/account_cmds.go
idodod: good idea!

yea re if/else I was trying to follow Sam's guideline to avoid else :) but I can see how being consistent is preferable in this case

PR #3116, opened 2023-07-19
Clarify satellite docs ahead of upcoming permissions change

dchw:


vladaionescu (approved):

alexcb (commented):
looks good; we'll need to make sure this goes into `docs-0.7` in order to be live.

Issue #3100, opened 2023-07-14
`earthly sat ls` doesn't show selected satellite when an org is selected with `earthly org select`


alexcb:
When running:
```
earthly org select <myorg>
earthly sat select <mysat>
earthly sat ls
```

the list of satellites no longer shows a `*` next to the selected satellite.

Note that the test was disabled under https://github.com/earthly/earthly/pull/3088/commits/0ef8599518fa485d7032c5b0485c5512a4464ed1#diff-107a6b6be0ee5c8602bc9a0b3adaf4d59650f65bb7e156055b7c9382d2ba8bc4R40 and needs to be re-enabled once fixed.

PR #3120, opened 2023-07-19
Fix issue with selected satellite not showing displaying in list

brandonSc:
Fixes  https://github.com/earthly/earthly/issues/3100


alexcb (approved):
thanks for the fix!

PR #3122, opened 2023-07-19
Org list behaves more similarly to satellite list

brandonSc:
An asterisk denotes the org that is currently selected in the config file. A hint also shows which org is the personal org.

Screen Shot 2023-07-19 at 3 24 19 PM


alexcb (approved):

scripts/tests/secrets-integration.sh
```diff
@@ -79,8 +79,12 @@ EOF
 "$earthly" secrets --org manitou-org --project earthly-core-integration-test set my_test_file "secret-value"
 "$earthly" secrets --org manitou-org --project earthly-core-integration-test get my_test_file | acbgrep 'secret-value'
 
-# test secrets with org selected in config file
+# test selecting org
 "$earthly" org select manitou-org
+"$earthly" org ls | acbgrep '^\* \+manitou-org'
```
alexcb, line 84: does this work? or do you need to disable color?

brandonSc: It seemed to work out of the box for this one. I think we are doing something fancy with colors in the `satellite ls` output to be able to gray-out those which come from CI runners.

alexcb: most bizarre! I thought you ran into other issues that were related to grep and color..... but if it works, it works.

PR #3123, opened 2023-07-20
check env values are not 0 or FALSE

alexcb:


cmd/earthly/main.go
```diff
@@ -287,13 +288,11 @@ func main() {
 	}
 
 	colorMode := conslogging.AutoColor
-	_, forceColor := os.LookupEnv("FORCE_COLOR")
-	if forceColor {
+	if envutil.IsEnvTrue("FORCE_COLOR") {
```
nelsam, line 291: This feels stuttery. `envutil.IsTrue` reads decently enough to me. Maybe an even better approach would be something like:

```go
package envutil

type EnvValue string

func (v EnvValue) Truthy() bool {
    switch v {
    case "", "false", "no", "0":
        return false
    default:
        return true
    }
}

func Get(v string) EnvValue {
    return EnvValue(os.GetEnv(v))
}
```

And then we could use it like

```go
if envutil.Get("FORCE_COLOR").Truthy() {
```

util/envutil/is_env_true.go
```diff
@@ -0,0 +1,18 @@
...
+// to something bash would interpret as true.
+func IsEnvTrue(k string) bool {
+	v, found := os.LookupEnv(k)
+	if !found {
+		return false
+	}
+	if v == "" || v == "0" || v == "false" || v == "FALSE" {
+		return false
+	}
+	return true
```
nelsam, line 17: `!found` and `""` are being treated the same here, so we can just use `Getenv`.

Also a switch allows us to avoid repeating `v ==` for each case.

```suggestion
	switch os.Getenv(k) {
	case "", "0", "false", "FALSE":
		return false
	default:
		return true
	}
```

nelsam (approved):
Honestly, it all seems fine. I have some nitpicky style/readability comments, but they're not a big deal.

cmd/earthly/main.go
alexcb: I like the `envutil.IsTrue` name.

I like that the functional approach is less chars than the OOP version.
```
if envutil.IsTrue("FORCE_COLOR") {
```

vs

```
if envutil.Get("FORCE_COLOR").Truthy() {
```

PR #3125, opened 2023-07-20
Simone/update and break apart main package and make functions testable

Velutas:


docs/earthly-command/earthly-command.md
```diff
@@ -182,7 +195,7 @@ Also used to be available as an env var setting: `GIT_URL_INSTEAD_OF=<git-instea
 This option is now obsolete. By default, `earthly` will automatically switch from ssh to HTTPS when no keys are found or the ssh-agent isn't running.
 Please use the [configuration file](../earthly-config/earthly-config.md) to override the default behavior.
 
-#### Build Options
+### Build Options
```
Velutas, line 198: Changed heading levels for consistency through the doc and with other docs

docs/earthly-command/earthly-command.md
```diff
@@ -324,22 +337,127 @@ These options can only be set via environment variables, and have no command lin
 | EARTHLY_TARGET_PADDING | `EARTHLY_TARGET_PADDING=n` will set the column to the width of `n` characters. If a name is longer than `n`, its path will be truncated and remaining extra length will cause the column to go ragged. |
 | EARTHLY_FULL_TARGET    | `EARTHLY_FULL_TARGET=1` will always print the full target name, and leave the target name column ragged.                                                                                                   |
 
-## earthly prune
+
+## earthly --version
```
Velutas, line 341: Moved earthly --version, earthly ls, earthly doc to the top of the doc as it felt awkward to have them halfway down

docs/earthly-command/earthly-command.md
```diff
@@ -427,106 +544,152 @@ git:
...
+### earthly account register
+
+#### Synopsis
 
-###### Synopsis
 
 * ```
-  earthly account register --email <email>
-  earthly account register --email <email> --token <email-verification-token> [--password <password>] [--public-key <public-key>] [--accept-terms-conditions-privacy]
+  # Register an account using your email
```
Velutas, line 553: Synopsis's updated for consistency throughout doc

cmd/earthly/root_cmds.go
```diff
@@ -35,15 +35,16 @@ func (app *earthlyApp) rootCmds() []*cli.Command {
 		{
 			Name:        "build",
 			Usage:       "Build an Earthly target",
-			Description: "Build an Earthly target",
+			Description: "Build an Earthly target.",
```
Velutas, line 38: Updated all usage strings to not have `.` for consistency and added them to descriptions also just for consistency across the help outputs for all commands to match.

Earthfile
```diff
@@ -106,6 +111,7 @@ lint-scripts-misc:
     # so -x is needed to let shellcheck read that file.
     RUN shellcheck -x shell_scripts/*
 
+# lint-scripts-auth-test runs the spellcheck package against shell scripts in the scripts/tests folder
```
alexcb, line 114:
```suggestion
# lint-scripts-auth-test runs the shellcheck package against shell scripts in the scripts/tests folder
```

Earthfile
```diff
@@ -92,6 +96,7 @@ lint-scripts-base:
 
     WORKDIR /shell_scripts
 
+# lint-scripts-misc runs the spellcheck package against different shell scripts
```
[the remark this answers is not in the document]
alexcb: the `+lint-scripts-misc` and `+lint-scripts-misc` are helper targets that are run via `+lint-scripts`; I wonder if documenting these helper targets is a good idea?

generally I only ever run `+lint-scripts` from the terminal -- perhaps we should only document that target to make `earthly doc` display help for targets users would typically invoke.

Earthfile
```diff
@@ -114,10 +120,12 @@ lint-scripts-auth-test:
     # flag to source setup.sh during analysis.
     RUN shellcheck -x test-*.sh
 
+# lint-scripts runs the spellcheck package against both misc and test shell scripts
```
alexcb, line 123: I would change this comment to describe what higher-level task the target attempts to perform rather than describing the implementation details.

If someone really wanted to know what other targets it called, they could just simply look at the target's code block and see it just calls two helper targets.

```suggestion
# lint-scripts runs the shellcheck package to detect potential errors in shell scripts
```

Earthfile
```diff
@@ -114,10 +120,12 @@ lint-scripts-auth-test:
...
 lint-scripts:
     BUILD +lint-scripts-auth-test
     BUILD +lint-scripts-misc
 
+# earthly-script-no-stdout validates the earthly script doesn't print anything to stdout (stderr only)
```
alexcb, line 128: this duplicates the comment below (and was likely written before `earthly docs` existed); can you remove the comment contained in the code block.

Additionally, I think it's valuable to reference it as `./earthly` because the term `earthly` can be rather generic.

```suggestion
# earthly-script-no-stdout validates the ./earthly script doesn't print anything to stdout (stderr only)
```

cmd/earthly/root_cmds.go
```diff
@@ -169,64 +170,76 @@ func (app *earthlyApp) rootCmds() []*cli.Command {
...
+					Name:    "reset",
+					EnvVars: []string{"EARTHLY_PRUNE_RESET"},
+					Usage: `Reset cache entirely by restarting BuildKit daemon and wiping cache dir.
+			This option is not available when using satellites.`,
 					Destination: &app.pruneReset,
 				},
+				&cli.DurationFlag{
+					Name: "age",
+					Usage: `Prune cache older than the specified duration passed in as a string.
+	Valid time units: ns, us, ms, s, n, h
```
alexcb, line 214: This seems overkill; I don't see a user-case for someone specifying a time based in nanoseconds.

I think `m, h, d` is good enough; (pretty sure it should be `m` and not `n` which looks like a typo).

```
Prune cache older than the specified duration passed in as a string; duration is specified with an integer value followed by a m, h, or d suffix which represents minutes, hours, or days respectively, e.g. 24h, or 1d
```
```

File: docs/earthly-command/earthly-command.md [hunk 10]
```diff
@@ -535,219 +698,359 @@ The token will expire in 1 year from creation date unless a different date is su
...
+##### `--write`
 
-###### Synopsis
+Grant write permissions in addition to read permissions
+
+##### `--expiry`
+
+Set token expiry date in the form YYYY-MM-DD or never
+
+Defaults to one (1) year.
```
alexcb, line 711: these tokens no longer default to expiring: https://github.com/earthly/earthly/pull/3098

docs/earthly-command/earthly-command.md
```diff
@@ -866,187 +1194,423 @@ Skips setting up the BuildKit container during bootstrapping. If needed, it will
...
+  earthly [options] satellite (launch|ls|inspect|select|unselect|rm)
   ```
 
 #### Description
 
-Prints version information about earthly.
+Launch and use a Satellite runner as remote backend for Earthly builds.
 
-## earthly ls
+Satellites can be used to optimize and share cache between multiple builds and users, as well as run builds in native architectures independent of where the Earthly client is invoked.
```
alexcb, line 1299: optimization seems like a rather strong word to use here, satellites make cache reuse easier -- since cache persists between builds, but there's no optimization process that makes the cache usage the absolute best perfect use. It's just a LRU cache :)

```suggestion
Satellites can be used to share cache between multiple builds and users, as well as run builds in native architectures independent of where the Earthly client is invoked.
```

docs/earthly-command/earthly-command.md
```diff
@@ -866,187 +1194,423 @@ Skips setting up the BuildKit container during bootstrapping. If needed, it will
...
+
+Read more about satellites here: https://docs.earthly.dev/earthly-cloud/satellites 
+
+Sign up for satellites here: https://cloud.earthly.dev/login
+
+#### Options
+
+##### `--org`
+
+The name of the organization the satellite belongs to. Required when user is a member of multiple.
```
alexcb, line 1309: This seems wrong; I thought we were changing this to make it always required -- which is why we introduced the `earthly org select` command.

alexcb (commented):
great work! just a few small things that jumped out.

Earthfile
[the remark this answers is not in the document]
alexcb: I wonder if it really makes sense to document every single target, e.g. there's a few other targets that simply exist to allow an Earthfile of a child directory to copy out of the parent directory -- I personally don't find documenting these helper targets all that useful.

If the helper target is difficult to understand, then a comment makes sense, but if the target is easy to understand and isn't intended to be used directly, these comments feel ceremonial. Additionally, having them displayed under `earthly doc` seems noisy.

side-tangent: should `earthly doc` contain a `-a` (or `--all`) which could display hidden helper target's comments?

@nelsam any thoughts on the pros and cons of documenting all of the underlying helper methods?

Earthfile
Velutas: Removing this comment

Earthfile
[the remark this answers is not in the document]
Velutas: That's true, if it not something that is ever used it's probably not worth it to include in the output. I'll remove those and some of the other helper functions

docs/earthly-command/earthly-command.md
Velutas: I copied this from the existing CLI help. I'll update both places as it was likely out of date

Earthfile
```diff
@@ -214,6 +209,11 @@ mocks:
         SAVE ARTIFACT $mockfile AS LOCAL $mockfile
     END
 
+# cli-unit-test runs the unit tests for the CLI commands in ./cmd/earthly
+cli-unit-test:
```
Velutas, line 213: These unit tests are also triggered from the +unit-test target but it's nice to be able to call them separately for speed

cmd/earthly/root_cmds.go
Velutas: Updated with tests for this for each of the files that have Usages and Descriptions

Earthfile
alexcb: I don't think this is needed since it should be possible to use `+unit-test --pkgname=./cmd/earthly/...` instead.

Velutas: Ahh, that is convenient :D I'll remove that

cmd/earthly/account_cmds_test.go
```diff
@@ -0,0 +1,46 @@
...
+			t:      t,
+			expect: expect.New(t),
+		}
+	})
+	defer o.Run()
+
+	o.Spec("Help command usage does not end with '.'", func(tt testCtx) {
+		ctx := context.TODO()
+		app := newEarthlyApp(ctx, conslogging.ConsoleLogger{})
+		lines := app.accountCmds()
```
alexcb, line 30: Rather than duplicate the various `account_cmds_test.go`, `project_cmds_test.go` files, is it possible to combine them in a single test function (and file)?

Take a look at how the autocompletion works under https://github.com/earthly/earthly/blob/main/cmd/earthly/autocomplete.go#L79

it does a similar call to `app := newEarthlyApp(ctx, conslogging.ConsoleLogger{})` but then recursively traverses all the subcommands under https://github.com/earthly/earthly/blob/main/autocomplete/complete.go#L354 and https://github.com/earthly/earthly/blob/main/autocomplete/complete.go#L490

Velutas: Yeah, I can combine them all into a single file. I had been thinking that if more tests are eventually added to the cli commands then the files for each part would then already be there but it makes sense to keep them altogether for now.

Velutas:
Still draft, just pushing

Earthfile
```diff
@@ -341,10 +341,10 @@ earthly:
     ARG GOCACHE=/go-cache
     RUN mkdir -p build
     RUN printf "$BUILD_TAGS" > ./build/tags && echo "$(cat ./build/tags)"
-    RUN printf '-X main.DefaultBuildkitdImage='"$DEFAULT_BUILDKITD_IMAGE" > ./build/ldflags && \
-        printf ' -X main.Version='"$VERSION" >> ./build/ldflags && \
-        printf ' -X main.GitSha='"$EARTHLY_GIT_HASH" >> ./build/ldflags && \
-        printf ' -X main.DefaultInstallationName='"$DEFAULT_INSTALLATION_NAME" >> ./build/ldflags && \
+    RUN printf '-X github.com/earthly/earthly/cmd/earthly/flag.DefaultBuildkitdImage='"$DEFAULT_BUILDKITD_IMAGE" > ./build/ldflags && \
```
alexcb, line 343: Can we avoid changing this?

They need to be kept in sync with https://github.com/Homebrew/homebrew-core/blob/master/Formula/earthly.rb#L25-L29 and it's easier if we don't have to submit a PR to homebrew.

as well as in our own homebrew tap, which is linked to below under the `# Important! ...` comment.

Earthfile
```diff
@@ -33,6 +33,10 @@ deps:
     SAVE ARTIFACT go.mod AS LOCAL go.mod
```
Velutas, line 33: Moving this to separate PR for clarity

docs/earthly-command/earthly-command.md
```diff
@@ -20,7 +20,7 @@
```
Velutas, line 20: Moving this to separate PR for clarity

alexcb (approved):
looks good!

When it comes time to merge this, it should be done via the cli:

```
git checkout main && get fetch origin && git reset --hard origin/main && git merge c66697338aef13849c343150823b8715b161a17c && git push
```

alexcb:
This was rebased against main and fast-forward merged in 47aad6fa30ee1ad7e35979794081aae6ed411027

PR #3126, opened 2023-07-21
Mention Earthly Satellites when not logged in

vladaionescu:
Screenshot 2023-07-21 at 4 20 33 PM

### DIFF
--- a/cmd/earthly/build_cmd.go
+++ b/cmd/earthly/build_cmd.go
@@ -233,9 +233,9 @@ func (app *earthlyApp) actionBuildImp(cliCtx *cli.Context, flagArgs, nonFlagArgs
 		} else {
 			defer func() { // Defer this to keep log upload code together
 				app.console.Printf(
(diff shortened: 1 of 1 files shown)


nelsam:
I think it might be nice to have a setting to disable this - e.g. if I've got an account but haven't set it up in my CI yet, I could envision this getting kind of annoying. `earthly config global.marketing disabled` or something?

What do you think, @vladaionescu?

vladaionescu:
I think that adds more complexity than is needed at this point. We could do that if it becomes a popular request.

Although I'm thinking of whether I should reduce this to a single line to reduce the annoyance.

nelsam:
:shrug: I don't think the message is too intrusive as is, I can just see it feeling like a watermark on my CI builds in my open source projects :/ granted, most CIs have secrets, but I kind of hate configuring secrets for my smaller open source projects. I know I would personally want an easy way to remove the "watermark" without needing to set up secrets and all that.

nelsam (approved):

PR #3136, opened 2023-07-24
Add a Zig Example

dchw:
I was working on learning it _anyways_, figured may as well make a simple example :D


alexcb (approved):
this looks cool! I see zig is designed with a goal of making it integrate with existing C codebases.... would be really cool to see it combined in an example... but only if you want to play around with that side of it :)

dchw:
It feels like go, but with a manual transmission

PR #3140, opened 2023-07-25
Remove AMI Bake

dchw:
Since this is moved internally now (and part of the Satellites release process), we can remove this release here. I think I got everything? I am unsure how to test the release outside of actually doing it, if there is a good way; let me know.


alexcb (approved):
thanks for the cleanup.

Issue #3135, opened 2023-07-22
Earthly is not respecting proxy variables, causing 5-second delay startup trying to talk to api.earthly.dev


gblewis1:
**What went wrong?**
Whenever I run earthly to build a target on my corporate network machine (Linux x86_64), it is spending about 5 seconds trying to ping the API during Init. I have traced this delay to the `func (c *Client) WhoAmI(ctx context.Context) (string, string, bool, error)` function, specifically inside `c.ping(ctx)`. The primary difference between the corporate network machine and an Internet-connected machine is that the corporate network blocks all outside IP access and forces http/https requests through a proxy. 

I have verified that when I take the machine off the corporate VPN and run earthly on an Internet network, the Init phase does not hang for 5 seconds here. It looks like earthly is not respecting my http/https proxy environment variables when trying to connect to https://api.earthly.dev.

**What should have happened?**
I would like to see a way to disable the call to api.earthly.dev, and also to have Earthly respect the proxy variables when trying to connect. I have not yet found any workarounds to this.

alexcb:
> I would like to see a way to disable the call to api.earthly.dev

This already exists, you need to set `EARTHLY_DISABLE_ANALYTICS=1` (or `DO_NOT_TRACK=1`); it's documented under `--help` and also in our CHANGELOG.

We even have a test for it: https://github.com/earthly/earthly/blob/main/tests/do-not-track/Earthfile#L25-L33

It does seem odd that the proxy envs aren't being used though -- I'll leave this issue open for users who require a proxy connection; thanks for reporting this.

gblewis1:
> > I would like to see a way to disable the call to api.earthly.dev
> 
> This already exists, you need to set `EARTHLY_DISABLE_ANALYTICS=1` (or `DO_NOT_TRACK=1`); it's documented under `--help` and also in our CHANGELOG.
> 
> We even have a test for it: https://github.com/earthly/earthly/blob/main/tests/do-not-track/Earthfile#L25-L33
> 
> It does seem odd that the proxy envs aren't being used though -- I'll leave this issue open for users who require a proxy connection; thanks for reporting this.

[...]

PR #3141, opened 2023-07-25
use HTTP_PROXY/HTTPS_PROXY envs

alexcb:


alexcb:
fixes #3135

nelsam (approved):
Looks good. Two questions (for follow-ups):

1. Should we have a test for this? It wouldn't be too hard to make a little `httptest.NewServer`, try to make an HTTP request with the cloud client with the proxy env set, and ensure the server receives a request.
2. Does our gRPC client support the proxy variables already? Do we need a follow-up PR to support these in gRPC land?

alexcb:
> Should we have a test for this? It wouldn't be too hard to make a little httptest.NewServer, try to make an HTTP request with the cloud client with the proxy env set, and ensure the server receives a request.

This feels a little overkill given `http.ProxyFromEnvironment` comes from the stdlib, I would personally rather see an integration test that stubs out our backend services, then we could run it with `--network=none`.

> Does our gRPC client support the proxy variables already?

fortunately it's supported by default: https://github.com/grpc/grpc-go/blob/master/Documentation/proxy.md

PR #3163, opened 2023-08-03
include dind base os version in tag

alexcb:
New dind image tags are:

    earthly/dind:alpine-3.18 (which is also earthly/dind:alpine)
    earthly/dind:ubuntu-20.04
    earthly/dind:ubuntu-23.04 (which is also earthly/dind:ubuntu)


nelsam (commented):
This looks fine, but there are a lot of test failures. It seems like some of them might be real.

alexcb:
These images ended up getting pushed :facepalm: which is why the tests are failing now.

The tests are not setup to rebuild the dind images each time.

i'll submit a fix for the tests in this PR too, but this has broken main

nelsam (approved):
Seems fine, and tests look to be fixed, so :crossed_fingers:

PR #3171, opened 2023-08-08
Fix missing web command URL test

mikejholly:


alexcb:
this appears to be the same as #3170 cc @Velutas

mikejholly:
> this appears to be the same as #3170 cc @Velutas

Cool. Feel free to merge either one once the tests pass.

PR #3176, opened 2023-08-09
Update dind docker hub readme

alexcb:
This additionally moves the docker hub readme update scripts to a separate workflow file.

### DIFF
--- a/.github/workflows/update-docker-hub-readmes.yml
+++ b/.github/workflows/update-docker-hub-readmes.yml
@@ -1,0 +1,43 @@
+name: Docker CI Ubuntu
+
+on:
+  push:
+    branches: [ main ]
+    paths: [ docs/** ]
+
+jobs:
+  update-docker-hub-readmes:
+    name: update dockerhub readmes
+    if: github.event_name == 'push'
+    runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


alexcb:
Note that I'm currently working on a follow up PR that will include the docker version in the tags. That follow up PR will update this readme.

.github/workflows/update-docker-hub-readmes.yml
```diff
@@ -0,0 +1,42 @@
+name: Docker CI Ubuntu
+
+on:
+  push:
+    branches: [ main ]
```
nelsam, line 5: Should we include a `paths` option so we only run this job if the `docs` folder changes? `paths: [ 'docs/**' ]` would probably do it.

It's not a big deal, but with the GHA problems we seem to have, being more selective about when we run jobs might be useful. :)

nelsam (approved):
It all looks good. The `paths` selector in the workflow is not really important.

.github/workflows/update-docker-hub-readmes.yml
alexcb: ah good call, i'll add that back in.

PR #3178, opened 2023-08-09
include docker version in dind tag

alexcb:
earthly/dind images will now include the docker version in the tag.

For example:

    alpine-3.18-docker-23.0.6-r4
    ubuntu-23.04-docker-24.0.5-1

Additionally we will tag images to include when they were produced.

For example:

    alpine-3.18-docker-23.0.6-r4-20230808225455
    ubuntu-23.04-docker-24.0.5-1-20230808225455

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -552,25 +552,32 @@ ci-release:
 
 # dind builds both the alpine and ubuntu dind containers for earthly
 dind:
-    ARG OS_IMAGE # e.g. alpine, ubuntu
-    ARG OS_VERSION # e.g. 3.18.0, 23.04
+    ARG --required OS_IMAGE # e.g. alpine, ubuntu
+    ARG --required OS_VERSION # e.g. 3.18.0, 23.04
+    ARG --required DOCKER_VERSION # e.g. 20.10.14
     FROM $OS_IMAGE:$OS_VERSION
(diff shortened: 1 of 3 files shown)


Earthfile
```diff
@@ -552,25 +552,32 @@ ci-release:
 
 # dind builds both the alpine and ubuntu dind containers for earthly
 dind:
-    ARG OS_IMAGE # e.g. alpine, ubuntu
-    ARG OS_VERSION # e.g. 3.18.0, 23.04
+    ARG --required OS_IMAGE # e.g. alpine, ubuntu
+    ARG --required OS_VERSION # e.g. 3.18.0, 23.04
+    ARG --required DOCKER_VERSION # e.g. 20.10.14
```
nelsam, line 557: For `earthly doc` sake, I think it might be useful to move these comments above:
```suggestion
    # OS_IMAGE is the base image to use, e.g. alpine, ubuntu
    ARG --required OS_IMAGE
    # OS_VERSION is the version of the base OS to use, e.g. 3.18.0, 23.04
    ARG --required OS_VERSION
    # DOCKER_VERSION is the version of docker to use, e.g. 20.10.14
    ARG --required DOCKER_VERSION
```

nelsam (approved):
LGTM! minor doc comment suggestion aside.

Earthfile
alexcb: good point!

PR #3186, opened 2023-08-14
Update rust example to use CACHE and FROM to speed up compilation times

Velutas:


examples/rust/Earthfile
```diff
@@ -16,13 +18,10 @@ build-cache:
     FROM +install-chef
     COPY +prepare-cache/recipe.json ./
     RUN cargo chef cook --release
-    SAVE ARTIFACT target
-    SAVE ARTIFACT $CARGO_HOME cargo_home
 
 build:
+    FROM +build-cache
```
Velutas, line 23: This is the biggest time improvement when testing

alexcb (commented):
something seems off with these `CACHE` commands,

I ran `earthly ./examples/rust+build`, then I applied:

```
diff --git a/examples/rust/Earthfile b/examples/rust/Earthfile
index 32475dba..10d7502a 100644
--- a/examples/rust/Earthfile
+++ b/examples/rust/Earthfile
@@ -5,6 +5,8 @@ WORKDIR /rustexample
 install-chef:
     CACHE cargo_home
     CACHE target
+    RUN ls -la cargo_home
+    RUN ls -la target
     RUN cargo install --debug cargo-chef
```

and when I re-ran this earthly, I got:

```
./examples/rust+install-chef | --> RUN ls -la cargo_home
./examples/rust+install-chef | total 8
./examples/rust+install-chef | drwxr-xr-x 2 root root 4096 Aug 14 23:48 .
./examples/rust+install-chef | drwxr-xr-x 1 root root 4096 Aug 14 23:51 ..
./examples/rust+install-chef | --> RUN ls -la target
./examples/rust+install-chef | total 8
./examples/rust+install-chef | drwxr-xr-x 2 root root 4096 Aug 14 23:48 .
./examples/rust+install-chef | drwxr-xr-x 1 root root 4096 Aug 14 23:51 ..
```

indicating these directories are empty -- is that expected?

Velutas:
> something seems off with these `CACHE` commands,
> 
> I ran `earthly ./examples/rust+build`, then I applied:
> ...
> ...
> indicating these directories are empty -- is that expected?

That's odd, there is a significant speed improvement when I include the CACHE there vs. when I do not for when the TOML file changes. I'll look into what is happening some more to ensure it is as expected

Velutas:
Checking again the CACHE was in the wrong location. Tested again and it is working as expected now

alexcb (approved):

PR #3193, opened 2023-08-16
states/image: remove redundant nil check

Juneezee:
From the Go docs:

> "If the map is nil, the number of iterations is 0." https://go.dev/ref/spec#For_range

Therefore, an additional nil check for before the loop is unnecessary. Example: https://go.dev/play/p/Cw_wnqYzemu


nelsam (approved):
lgtm! thanks for the contribution!

alexcb:
I tested this manually via `./build/linux/amd64/earthly -P +test`, looks good.

PR #3194, opened 2023-08-16
prevent frontend_test from panicing

alexcb:
Switch to using require instead of assert to avoid continuing past an error and then panicing on a nil pointer dereference.

### DIFF
--- a/util/containerutil/alias_test.go
+++ b/util/containerutil/alias_test.go
@@ -1,0 +1,19 @@
+package containerutil_test
+
+import (
+	"github.com/stretchr/testify/require"
+)
+
+var (
+	NoError  = require.NoError
+	Equal    = require.Equal
+	NotEqual = require.NotEqual
+	Nil      = require.Nil
+	NotNil   = require.NotNil
+	True     = require.True
+	False    = require.False
+	Error    = require.Error
+	Len      = require.Len
+	Contains = require.Contains
+	Empty    = require.Empty
+)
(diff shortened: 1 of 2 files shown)


util/containerutil/alias_test.go
```diff
@@ -0,0 +1,19 @@
+package containerutil_test
+
+import (
+	"github.com/stretchr/testify/require"
+)
+
+var (
+	NoError  = require.NoError
```
mikejholly, line 8: Why this style over `. "github.com/stretchr/testify/require"`, out of curiosity?

util/containerutil/alias_test.go
```diff
@@ -0,0 +1,19 @@
+package containerutil_test
+
+import (
+	"github.com/stretchr/testify/require"
```
mikejholly, line 4: I kind of prefer `require` over `assert` as those panics can be a hassle.

mikejholly (approved):

util/containerutil/alias_test.go
alexcb: it makes it possible to `grep` for the function, otherwise grepping for `Nil` won't match the `. "github.com/stretchr/testify/require"`.

We have a few older tests that still use the `. "github.com/stretchr/testify/require"` includes, which haven't been switched over.

util/containerutil/alias_test.go
alexcb: me too -- that's why this PR exists :)

util/containerutil/alias_test.go
mikejholly: Not sure what you mean. Wouldn't `Nil` still be present in the test file?

mikejholly: I see. Fair enough.

PR #3195, opened 2023-08-17
Enable Logstream by default

mikejholly:
This PR enables the Logstream integration by default. It includes a few test fixes where there were some differences in how logging was handled. Before we merged, I'd like to do some basic load testing of the new Logstream database sharding to make sure we can support the rollout.

### DIFF
--- a/.github/workflows/ci-docker-ubuntu-legacy-logs.yml
+++ b/.github/workflows/ci-docker-ubuntu-legacy-logs.yml
@@ -1,0 +1,203 @@
+name: Docker CI Ubuntu (Legacy Logs)
+
+on:
+  push:
+    branches: [ main ]
+    paths-ignore: [ docs/** ]
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  docker-tests-quick:
+    needs: build-earthly
+    uses: ./.github/workflows/reusable-test.yml
+    with:
+      TEST_TARGET: "+test-quick"
+      BUILT_EARTHLY_PATH: "./build/linux/amd64/earthly"
+      RUNS_ON: "ubuntu-latest"
+      BINARY: "docker"
+      SUDO: ""
+      LOGBUS: "--logstream false"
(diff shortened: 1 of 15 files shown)


alexcb (commented):
looks promising!

Is it possible to extend our tests to cover both logstream and non-logstream code?

I worry that we'll loose test coverage of our old non-logstream code, which could lead to regressions.

mikejholly:
@alexcb Would that mean running many of the tests twice? It's not super clear to me why there are output differences when Logstream is enabled. My assumption is that the output would just be captured and written to the LS service. However, it seems like a solid chunk of code is duplicated between the 2 implementations.

alexcb:
> @alexcb Would that mean running many of the tests twice?

Unfortunately yes, but we should limit it to a subset of the tests -- perhaps only the docker-based tests. Maybe it could be further limited to tests that look at the output of earthly (e.g. the various secrets-based tests which you had to change)

> It's not super clear to me why there are output differences when Logstream is enabled

The logstream code is completely different from the non-logstream code, and follows a different code path.

Take for instance, the old code under `solvermon/solvermon.go`, which has been reimplemented under `logbus/solvermon/solvermon.go`. 

This is why the `Repeating the output of the command that caused the failure` message changed to `Repeating the failure error...` when you enabled the feature.

> However, it seems like a solid chunk of code is duplicated between the 2 implementations

unfortunately yes! But until we delete the old code, we should try to support both, in order to:

1. maintain the stability of earthly for users who don't (yet) want to enable logbus, and 
2. make it easier for us in the event that we need to rollback `DefaultLogstreamUpload = true` to `false`. If this were to happen we should ideally just be able to submit a single-line change for this boolean, and not have to change anything related to tests.

Once we are confident logbus is working, then we'll want to delete the old code and tests that cover the old functionality.

mikejholly:
Closing in favor of #3199, where I'm incrementally enabling Logstream in tests.

Issue #3160, opened 2023-08-03
`earthly secret set --stdin foo` should display message saying waiting for EOF when running interactively


alexcb:
It's not obvious to all users that `earthly secret set --stdin foo` waits for an EOF rather than a newline.

If earthly is run via an interactive tty, we should display a message reminding users how the `--stdin` mode works.

(note that we must wait for an EOF rather than a newline because some secrets contain newlines)

vladaionescu:
Could we maybe have `--stdin` (uses new-line) and `--stdin-eof` (uses eof) as separate options? Although secrets could indeed contain new lines, it's pretty rare.

alexcb:
it might be more intuitive if we simply treated a missing value as a prompt to interactively ask for a value.

```
$ earthly secret set foo
secret: <cursor>
```

Then we could use the `promptPassword()` code from https://github.com/earthly/earthly/blob/567e70d88685732a4ffc136d2c02e4d109888680/cmd/earthly/account_cmds.go#L307-L324

alexcb:
I'm worried about breaking the existing functionality of `--stdin`; encryption keys (e.g. private keys used by ssh and gpg) tend to have new lines.

idodod:
Addressed in #3196

PR #3196, opened 2023-08-17
prompt for secret value when missing

alexcb:
When running `earthly secret set <path>`, rather than exiting with an error, we will prompt the user for a single-line secret value.

This allows a user to type in a single line followed by hitting the <enter> key.

Additionally, if `-stdin` is used from an interactive tty, earthly will now display a message saying it is waiting for an EOF (ctrl-d).

This additionally fixes a bug with isTTY, which was only checking stdout but could not detect if a command was being piped to stdin or not.


alexcb:
fixes #3160

idodod (approved):

PR #3214, opened 2023-08-29
use --pass-args in tests

alexcb:
Make use of the new `--pass-args` flag (implemented in bc3fe7499d5125a5484b06ee690f9aa1e71f5651) to simplify our tests.


idodod (approved):
I was looking forward for those red lines :)

alexcb:
> I was looking forward for those red lines :)

once we do a new release (which will contain the `--pass-arg` in the binary we initially use for the build), we'll able to remove a few more lines from the root Earthfile :)

Issue #3183, opened 2023-08-14
Misleading transferring output


vladaionescu:
At the end of the build, when outputting images locally the output says

```transferring docker.io/some/image```

This is misleading because the user might think that the image is being pushed to Docker Hub.

This is not what happens - internally Earthly simply uses a normalized version of the image name, with the registry docker.io, when none is specified. It uses this for the transfer from buildkit to the localhost Docker.

Ideally, this message is clearer, to prevent panicking users who do not intend to publish the image.

**Comment by taytzehao** [comment]
take

**Comment by mikejholly** [comment]
@vladaionescu what's the best way to reproduce this? I don't see the text output in a few test commands. Is this still the case with Logbus/Logstream?

![image](https://github.com/earthly/earthly/assets/332408/66331714-4eb1-423c-abbb-d44023d3a9e8)

This may be where the text is set in BuildKit, but I don't see this being output in Core. https://github.com/earthly/buildkit/blob/earthly-main/source/local/source.go#L244

**Comment by alexcb** [comment]
this last occurred under earthly v0.7.17; (18 was never fully released); then under v0.7.19 this no longer occurred.

under v0.7.19 logbus was feature flipped -- from the changelog:

```
### Changed
- Log sharing will now stream logs as your build is running (rather than uploading logs when build execution completes).
```

We can close this issue.

PR #3215, opened 2023-08-30
Modifying buildkit print statement

taytzehao:
Added a function to modify the print statement that originates from the buildkit. This allows us to replace or even reformat certain texts that may not be the clearest.

### DIFF
--- a/outmon/solvermon.go
+++ b/outmon/solvermon.go
@@ -317,6 +317,7 @@ func (sm *SolverMonitor) printProgress(vm *vertexMonitor, id string, progress in
 		if !vm.headerPrinted {
 			sm.printHeader(vm)
 		}
+		id = sm.modifyId(id)
 		vm.printProgress(id, progress, sm.verbose, sm.lastOutputWasProgress)
 		sm.lastOutputWasProgress = (progress != 100)
 		sm.lastOutputWasNoOutputUpdate = false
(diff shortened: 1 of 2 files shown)


alexcb:
relates to #3183

outmon/solvermon.go
```diff
@@ -413,3 +414,14 @@ func (sm *SolverMonitor) reprintFailure(errVertex *vertexMonitor, phaseText stri
 	}
 	errVertex.printError()
 }
+
+
+func (sm *SolverMonitor) modifyId(id string) string {
+	modifiedId := id
+	for old, new := range map[string]string{
+		"transferring docker.io/library":  "transferring locally to docker.io/library",
```
alexcb, line 422: this isn't generic, how will this help with other image names?

e.g.

```
./earthly ./examples/python+docker
```

produces

```
              output | [----------] 100% transferring docker.io/earthly/examples:python
```

outmon/solvermon.go
```diff
@@ -413,3 +414,14 @@ func (sm *SolverMonitor) reprintFailure(errVertex *vertexMonitor, phaseText stri
 	}
 	errVertex.printError()
 }
+
+
+func (sm *SolverMonitor) modifyId(id string) string {
+	modifiedId := id
+	for old, new := range map[string]string{
```
alexcb, line 421: redefining this map on each call seems inefficient, it should be global and created once via go's `init()` function.

alexcb (changes requested):
thanks for tackling this; just a few minor issues.

outmon/solvermon.go
taytzehao: I think there are multiple criterias we have to consider if we want to determine it is transfering locally or remotely. We need to determine if the user's docker is logged in, destination username is the same as the username of the user, then only we can determine if the image would be push locally or remotely @alexcb. Let me know your thoughts. 

Of course, this is assuming that `SAVE IMAGE` only works with docker hub and not other container image repository. If there are other repository type, we may need to identify the different types of syntax

alexcb: I agree -- it should work in a way that's generic to all these use cases.

vladaionescu (commented):
Note that the whole solvermon logic is being replaced by logstream very soon.

alexcb:
Closing this, due to this codepath no longer being relevant* since we have enabled the logstream (aka logbus) feature by default.

* It's only relevant to users who go out of their way to disable the logstream feature (to revert back to the old solvermon code). Long term we will delete solvermon code.

PR #3217, opened 2023-08-30
use --pass-args in root Earthfile

alexcb:
Now that earthly-v0.7.17 has been released with --pass-args, we can use it during bootstrapping our tests/builds.


idodod (approved):
❤️

PR #3204, opened 2023-08-23
Update buildkit to c7693ebc3b039d047ed3def2d1bcda90de27876e

vladaionescu:
Includes https://github.com/earthly/buildkit/pull/23

PR #3218, opened 2023-08-30
Revert "Update buildkit to c7693ebc3b039d047ed3def2d1bcda90de27876e"

mikejholly:
Reverts earthly/earthly#3204

@vladaionescu was able to fix the root cause which was the slow consumption of messages from BK.

### DIFF
--- a/buildkitd/Earthfile
+++ b/buildkitd/Earthfile
@@ -12,7 +12,7 @@ buildkitd:
             ARG BUILDKIT_BASE_IMAGE=$BUILDKIT_PROJECT+build
         END
     ELSE
-        ARG BUILDKIT_BASE_IMAGE=github.com/earthly/buildkit:6d9a29f49bc4238b668797d8f5e77783de695daf+build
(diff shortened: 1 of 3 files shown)


alexcb:
Is there a corresponding revert that needs to be done in earthly/buildkit?

We should have earthly's `main` branch point to the earthly/buildkit's `earthly-main` branch at all times.

mikejholly:
> Is there a corresponding revert that needs to be done in earthly/buildkit?
> 
> We should have earthly's `main` branch point to the earthly/buildkit's `earthly-main` branch at all times.

Will do.

vladaionescu:
> @vladaionescu was able to fix the root cause which was the slow consumption of messages from BK.

Not exactly the root cause of the problem in buildkit - but rather the root cause in our specific setup - a lot of log output causes buildkit to be overwhelmed with status updates that are not consumed fast enough by Earthly. By reducing duplication in our output, we improved the speed by which we consume these, so the bug no longer manifests.

The actual issue is still in there.

In any case, this specific buildkit change was causing a deadlock, so it's not a proper fix - hence the revert.

alexcb (approved):

Issue #3226, opened 2023-09-05
display warning when tls is disabled


alexcb:
We should display a warning on the console when tls is disabled.

In addition to warning that the connection is insecure, it would also produce better visibility for cases like #3225

PR #3231, opened 2023-09-05
Display warning when TLS is disabled

alexcb:
Earthly will now warn:

    TLS has been disabled; this should never be done when connecting to Earthly's production API

When attempting to connect to a satellite without TLS.

### DIFF
--- a/cmd/earthly/base/buildkit.go
+++ b/cmd/earthly/base/buildkit.go
@@ -55,6 +55,8 @@ func (cli *CLI) ConfigureSatellite(cliCtx *cli.Context, cloudClient *cloud.Clien
 		cli.Flags().BuildkitdSettings.TLSCA = ""
 		cli.Flags().BuildkitdSettings.ServerTLSCert = ""
 		cli.Flags().BuildkitdSettings.ServerTLSKey = ""
+	} else {
+		cli.Console().Warnf("TLS has been disabled; this should never be done when connecting to Earthly's production API\n")
 	}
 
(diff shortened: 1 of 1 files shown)


alexcb:
fixes #3226

mikejholly (approved):

cmd/earthly/base/buildkit.go
```diff
@@ -55,6 +55,8 @@ func (cli *CLI) ConfigureSatellite(cliCtx *cli.Context, cloudClient *cloud.Clien
 		cli.Flags().BuildkitdSettings.TLSCA = ""
 		cli.Flags().BuildkitdSettings.ServerTLSCert = ""
 		cli.Flags().BuildkitdSettings.ServerTLSKey = ""
+	} else {
+		cli.Console().Warnf("TLS has been disabled; this should never be done when connecting to Earthly's production API\n")
```
idodod, line 59: nit: I think we can tell whether the user is trying to connect to the production API so the warning does not have to be displayed in localdev for example.

idodod (approved):

cmd/earthly/base/buildkit.go
alexcb: I wanted to get a quick warning in place since there's been a user who struggled with this issue.

If this warning becomes an issue, we could address it at a later time, but I don't want this to become a distraction from our higher priority issues.

If you do know how to easily do this, feel free to open a PR (but I don't think it warrants spending much time on it)

PR #3232, opened 2023-09-06
Acb/fix buildkit test

alexcb:


mikejholly (approved):

.buildkite/tests.sh
```diff
@@ -75,21 +72,23 @@ do
...
-
 # Yes, there is a bug in the upstream YAML parser. Sorry about the jank here.
 # https://github.com/go-yaml/yaml/issues/423
 "$earthly" config global.buildkit_additional_config "'[registry.\"docker.io\"]
 
  mirrors = [\"registry-1.docker.io.mirror.corp.earthly.dev\"]'"
 
+# setup secrets
+set +x # dont echo secrets
+echo "DOCKERHUB_MIRROR_USER=$($earthly secret --org earthly-technologies --project core get -n dockerhub-mirror/user)" > .secret
```
idodod, line 83: just curious what's behind using .secret/.arg instead of passing the values in the command as before

alexcb: I'm trying to make it easier for outside contributors (who don't have access to any PROJECT secrets) to run our tests.

idodod: Thanks!
How does the `secret` command work/what is expected of the user to do?

alexcb:
```
$ earthly-v0.7.17 secret --help 2>&1 | grep get
   get                      *beta* Retrieve a secret from the secrets store
```

it prints out the secret.

I'll be updating the CONTRIBUTING guide once the tests are all working.

idodod: I know how `secret` generally works lol, I meant how are the permissions different for the user in that particular command compared to an earthly build command that reads the secrets

alexcb: I don't understand the question -- this PR in particular is only for buildkite -- which is only something we run. Outside contributors will not be running this script.

outside contributors will be able to run the `+test` target in the root earthfile, and can optionally set args/secrets to authenticate with a mirror.

PR #3233, opened 2023-09-06
stage2-setup login for both docker and podman

alexcb:
The stage2-setup script should login for both docker and podman.


alexcb:
merged via 647ec8271fc514f1f47424993e08a0a3b26b8058

PR #3237, opened 2023-09-08
exit if earthly secret get fails

alexcb:
If the `earthly secret get ...` command substitution fails, propigate the failure to the calling script (since a `set -e` or `-o pipefail` doesn't catch this case)

### DIFF
--- a/.buildkite/tests.sh
+++ b/.buildkite/tests.sh
@@ -80,8 +80,8 @@ done
 
 # setup secrets
 set +x # dont echo secrets
-echo "DOCKERHUB_MIRROR_USER=$($earthly secret --org earthly-technologies --project core get -n dockerhub-mirror/user)" > .secret
-echo "DOCKERHUB_MIRROR_PASS=$($earthly secret --org earthly-technologies --project core get -n dockerhub-mirror/pass)" >> .secret
+echo "DOCKERHUB_MIRROR_USER=$($earthly secret --org earthly-technologies --project core get -n dockerhub-mirror/user || kill $$)" > .secret
+echo "DOCKERHUB_MIRROR_PASS=$($earthly secret --org earthly-technologies --project core get -n dockerhub-mirror/pass || kill $$)" >> .secret
 # setup args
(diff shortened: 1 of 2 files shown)


mikejholly (approved):

.buildkite/tests.sh
```diff
@@ -80,8 +80,8 @@ done
 
 # setup secrets
 set +x # dont echo secrets
-echo "DOCKERHUB_MIRROR_USER=$($earthly secret --org earthly-technologies --project core get -n dockerhub-mirror/user)" > .secret
-echo "DOCKERHUB_MIRROR_PASS=$($earthly secret --org earthly-technologies --project core get -n dockerhub-mirror/pass)" >> .secret
+echo "DOCKERHUB_MIRROR_USER=$($earthly secret --org earthly-technologies --project core get -n dockerhub-mirror/user || kill $$)" > .secret
+echo "DOCKERHUB_MIRROR_PASS=$($earthly secret --org earthly-technologies --project core get -n dockerhub-mirror/pass || kill $$)" >> .secret
```
idodod, line 84: nit: 
just in case it might look cleaner (not sure what happens after kill)
could we instead check the status of  $? and exit on non zero?

idodod (approved):

.buildkite/tests.sh
alexcb: > could we instead check the status of $? and exit on non zero?

It could potentially be done with named pipes, but that seems rather overkill. I'm not sure I understand your suggestion -- it's unclear how this could be done via a command substitution.

In this case of the kill -- we'll see the error message from earthly written to stderr, followed by a "terminated" message.

idodod: sounds good.
my suggestion was to do:
```
if [ $? != 0] then
   exit 1
end if
```
but I agree it's likely an overkill and will make the script more verbose.

alexcb: It's still unclear how that would work with a command substitution.

An `exit 1` will not propagate to the caller.

PR #3240, opened 2023-09-08
display name of secret that is not found

alexcb:
When project-based secrets were introduced, the ErrNotFound type was changed, which prevented the secret lookup method from returning the correct error message which would contain **which** secret was not found.

Rather than displaying

    RUN ./some-command failed: not found

We will now show:

    RUN ./some-command failed: unable to lookup secret my_secret: not found


mikejholly (approved):
Nice one. I've found this error ambiguous in the past.

Issue #3200, opened 2023-08-21
Improve error messages


vladaionescu:
Some errors are difficult to parse by a human. Here are some examples...

Note that GH's representation doesn't do justice because there's no wrapping. In a terminal, the really long lines take up a LOT of real-estate space that looks very intimidating...

### Example 1

```
ERROR linkchecker/Earthfile line 46:6
  ./linkchecker+test *failed* |       The command
  ./linkchecker+test *failed* |           WITH DOCKER RUN --privileged docker-compose up --abort-on-container-exit
  ./linkchecker+test *failed* |       did not complete successfully. Exit code 1
Share your logs with an Earthly account (experimental)! Register for one at https://ci.earthly.dev/.
Error: build target: build main: failed to solve: linkchecker/Earthfile line 46:6 with docker run: unlazy force execution: process "/bin/sh -c EARTHLY_DOCKERD_DATA_ROOT=\"/var/earthly/dind/16a66f7cc7947118c7b820248d045d48bc25b32a7ef4c8d7d4a35b036f1f748c\" EARTHLY_DOCKER_LOAD_FILES=\"\" EARTHLY_IMAGES_WITH_DIGESTS=\"package-manager:docs-linkchecker@sha256:a24601ea37a1705c5099f8852612ce8301c6c3af125e965b102f51c90e6749a6 package-manager:linkchecker@sha256:33143bb82a1b0b39ada222c12418480f1f1d354f1b5c1ccafbb0ffafabd14f84\" EARTHLY_START_COMPOSE=\"true\" EARTHLY_COMPOSE_FILES=\"compose.yml\" EARTHLY_COMPOSE_SERVICES=\"docs\" DIND_COMMIT=42b1175eda071c0e9121e1d64345928384a93df1 DOCKER_TLS_CERTDIR=/certs DOCKER_VERSION=20.10.14 PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin build_number=development type=dev /var/earthly/dockerd-wrapper.sh execute /usr/bin/earth_debugger /bin/sh -c 'docker-compose up --abort-on-container-exit'" did not complete successfully: exit code: 1
in		github.com/rstudio/package-manager/docs/main/linkchecker:HEAD+test --build_number=development --type=dev

Error: Process completed with exit code 1.
```

##### Suggested improvement

This part is great:

```
ERROR linkchecker/Earthfile line 46:6
  ./linkchecker+test *failed* |       The command
[...]

PR #3241, opened 2023-09-09
Enable Logstream without upload support

mikejholly:


cmd/earthly/flag/global.go
```diff
@@ -25,6 +25,9 @@ const (
 
 	DefaultSecretFile = ".secret"
 	SecretFileFlag    = "secret-file-path"
+
+	DefaultLogstream       = true
+	DefaultLogstreamUpload = false
```
mikejholly, line 30: To be enabled in another PR.

logbus/bus.go
```diff
@@ -132,9 +132,11 @@ func (b *Bus) WriteDeltaManifest(dm *logstream.DeltaManifest) {
 			DeltaManifest: dm,
 		},
 	}
+
 	b.rawMu.Lock()
-	defer b.rawMu.Unlock()
 	b.rawHistory = append(b.rawHistory, delta)
+	b.rawMu.Unlock()
```
mikejholly, line 138: Lock the `append` only, not the cascading `Write`s.

logbus/command.go
```diff
@@ -78,23 +77,19 @@ func (c *Command) SetStart(start time.Time) {
...
 		HasProgress:    true,
 		Progress:       progress,
 	})
-	c.lastProgress = progress
+	c.lastProgress.Store(progress)
 }
 
 // SetCached sets the cached status of the command.
 func (c *Command) SetCached(cached bool) {
-	c.mu.Lock()
```
mikejholly, line 96: Not needed as there are several locks in the underlying methods.

idodod (approved):

scripts/tests/secrets-integration.sh
```diff
@@ -33,7 +33,7 @@ if NO_COLOR=0 "$earthly" account logout > output 2>&1; then
     echo "earthly account logout should have failed"
     exit 1
 fi
-diff output <(echo "Error: account logout has no effect when --auth-token (or the EARTHLY_TOKEN environment variable) is set")
+acbgrep "Error: account logout has no effect when --auth-token (or the EARTHLY_TOKEN environment variable) is set" output
```
alexcb, line 36: This should be changed back to the original test -- repeating all errors twice (in all cases) seems like a bad decision.

tests/remote-cache/Earthfile
```diff
@@ -36,13 +36,13 @@ test1:
     RUN nl=$(cat ./output | grep "execute-test1-run-before-copy" | wc -l) && cat /var/log/buildkitd.log && test "$nl" -eq 1
 
     # Cached - should only have 1 line output and not re-echo.
-    RUN nl=$(cat ./output | grep "execute-test1-run-after-copy" | wc -l) && test "$nl" -eq 1
+    RUN nl=$(cat ./output | grep "execute-test1-run-after-copy" | grep -v "*cached*" | wc -l) && test "$nl" -eq 1
```
alexcb, line 39: why did this have to change? Is the contents of previous cached lines now output?

logbus/bus.go
alexcb: If we really don't need to lock for accessing the rawSubs, then perhaps the `rawMu` should be renamed to `rawHistoryMu`?

I don't know the code well enough to know how thread-safety this change is.

alexcb (changes requested):
looks ok, except for changing all errors to always be repeated twice. This change goes against making our errors easier to understand #3200

logbus/bus.go
mikejholly: I think the name is suitable since it's used elsewhere: https://github.com/earthly/earthly/blob/main/logbus/bus.go#L80

tests/remote-cache/Earthfile
mikejholly: Here's the output without `--logstream=true`: 
```
+test1 | REGISTRY=172.17.0.4:5000
+test1 | *cached* --> RUN echo "execute-test1-run-before-copy"
+test1 | [----------] 100% RUN echo "execute-test1-run-before-copy"
+test1 | REGISTRY=172.17.0.4:5000
+test1 | --> COPY ./input ./
+test1 | REGISTRY=172.17.0.4:5000
+test1 | --> RUN echo "execute-test1-run-after-copy"
+test1 | execute-test1-run-after-copy
```

And with it disabled: 

```
+test1 | *cached* --> RUN echo "execute-test1-run-before-copy"
+test1 | REGISTRY=172.17.0.4:5000
+test1 | *cached* --> COPY ./input ./
+test1 | REGISTRY=172.17.0.4:5000
+test1 | *cached* --> RUN echo "execute-test1-run-after-copy"
```

It seems to be cached, but the `RUN` is being output twice for some reason.

scripts/tests/secrets-integration.sh
mikejholly: I'll take another look. I asked Vlad about the original intention. He said it seems fine for now. :shrug:

logbus/bus.go
mikejholly: The change should be thread-safe as there are locks on shared resources in the underlying methods.

tests/remote-cache/Earthfile
alexcb: I checked out this branch, and reverted this test to the copy from `main` and created a diff (left side is the non-logstream version of earthly, compliled from `main`, and the right side is earthly compiled from your branch).

image

The logstream version of earthly is now printing out two lines:

```
# cat ./output | grep "execute-test1-run-after-copy"
              +test1 | *cached* --> RUN echo "execute-test1-run-after-copy"
              +test1 | [----------] 100% RUN echo "execute-test1-run-after-copy"         <------ this being the new one
```

by adding the `grep -v "*cached*"`, it's being changed to validate 

```
+test1 | [----------] 100% RUN echo "execute-test1-run-after-copy" 
```

exists in the output, which breaks what this test is testing for.

instead, this should be changed to something like `grep -v '\[.*\] [0-9]\+%'` to ignore the progress-related output of the command.

mikejholly: Thanks for digging. Would it then make more sense to use `-eq 2` since we expect 2 lines here?

mikejholly: Actually, the uncached version expects 2 lines. I'll just go with your suggestion.

alexcb: Changing this to `2`, would then cause this test to fail when logstream is disabled -- this test should ideally pass when running both the legacy and logstream mode.

util/deltautil/apply.go
```diff
@@ -35,102 +50,116 @@ func WithDeltaManifest(m *pb.RunManifest, d *pb.Delta) (*pb.RunManifest, error)
...
+	if m.GetVersion() != 0 && m.GetVersion() != Version {
+		return fmt.Errorf("unsupported manifest version %d", m.GetVersion())
+	}
+	if d.GetVersion() != 0 && d.GetVersion() != Version {
+		return fmt.Errorf("unsupported delta version %d", d.GetVersion())
+	}
+	return nil
+}
+
+func setManifestFields(dm *pb.DeltaManifest, ret *pb.RunManifest) {
```
alexcb, line 68: nice change :)

mikejholly: Thanks!

alexcb (approved):
nice work hammering out all those bugs!

image

PR #3243, opened 2023-09-11
Distribute core-test load across shards

alexcb:
The earthly core-test satellite can get quite overloaded; randomly select 1 of 4 core-test-* instance to distribute load.

Also fixes some tests that were not correctly propagating through the `USE_SATELLITE` bool.


mikejholly (approved):

idodod (approved):
Nice to learn about the shell $RANDOM function after all these years :)

PR #3248, opened 2023-09-13
verbose print known_hosts to make it easier to debug git related commands

### LINKED ISSUE #3234: display known_hosts sent via llb for `GIT CLONE`, `FROM` , `COPY`, `BUILD` commands
To help debug failing commands such as `GIT CLONE github.com/user/privaterepo /code` or `FROM github.com/user/privaterepo+target` it would be nice to have `earthly --verbose` display the contents of all the known_hosts which are sent over to buildkit (as well as a total count of entries -- to make it easy for cases where there's 0 entries).

The context surrounding this issue is that a user was having buildkit fail with a ssh unknown fingerprint type of error, and it was not clear why, because the user had entries in their host's `known_hosts` file. Earthly should be sending this over with the llb request, so it would be nice to display those entries under the `--verbose` mode.

idodod:
Resolves #3234

### DIFF
--- a/buildcontext/gitlookup.go
+++ b/buildcontext/gitlookup.go
@@ -114,14 +114,14 @@ func (gl *GitLookup) DisableSSH() {
 
 func knownHostsToKeyScans(knownHosts string) []string {
 	knownHosts = strings.ReplaceAll(knownHosts, "\r\n", "\n")
-	var keyScans []string
+	foundKeyScans := make(map[string]bool)
 	for _, s := range strings.Split(knownHosts, "\n") {
 		s = strings.TrimSpace(s)
-		if s != "" && !strings.HasPrefix(s, "#") {
-			keyScans = append(keyScans, s)
+		if s != "" && !strings.HasPrefix(s, "#") && !foundKeyScans[s] {
+			foundKeyScans[s] = true
 		}
 	}
-	return keyScans
+	return maps.Keys(foundKeyScans)
 }
 
 // AddMatcher adds a new matcher for looking up git repos
--- a/buildcontext/gitlookup.go
+++ b/buildcontext/gitlookup.go
@@ -19,6 +19,7 @@ import (
 	"golang.org/x/crypto/ssh"
 	"golang.org/x/crypto/ssh/agent"
 	"golang.org/x/crypto/ssh/knownhosts"
+	"golang.org/x/exp/maps"
 
 	"github.com/earthly/earthly/conslogging"
 	"github.com/earthly/earthly/util/fileutil"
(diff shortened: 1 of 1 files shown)


buildcontext/gitlookup.go
```diff
@@ -306,22 +306,26 @@ func (gl *GitLookup) getHostKeyAlgorithms(hostname string) ([]string, []string,
...
+	foundKeys := make(map[string]bool)
 	for _, keyScans := range [][]string{
 		knownHostsKeyScans,
 		defaultKeyScans,
 	} {
 		for _, keyScan := range keyScans {
 			keyAlg, keyData, err := parseKeyScanIfHostMatches(keyScan, hostname)
-			switch err {
-			case nil:
-			case errKeyScanNoMatch:
```
alexcb, line 317: this is no longer being explicitly checked, when a `errKeyScanNoMatch` "error" occurs, we should still do the `continue`, but printing out a warning saying the key failed to parse is incorrect (since it did parse, it's just being omitted due to it not matching).

When it is skipped, can we do

```
gl.console.VerbosePrintf("ignoring key scan %q: due to host mismatch", keyScan)
```

buildcontext/gitlookup.go
```diff
@@ -306,22 +306,26 @@ func (gl *GitLookup) getHostKeyAlgorithms(hostname string) ([]string, []string,
...
-				gl.console.Warnf("failed to parse key scan %q: %s", keyScan, err)
+			if err != nil {
+				gl.console.Warnf("failed to parse key scan %q: %s; skipping", keyScan, err)
 				continue
 			}
 			foundAlgs[keyAlg] = true
-			keys = append(keys, fmt.Sprintf("%s %s %s", knownhosts.Normalize(hostname), keyAlg, keyData))
+			key := fmt.Sprintf("%s %s %s", knownhosts.Normalize(hostname), keyAlg, keyData)
+			if !foundKeys[key] {
+				gl.console.VerbosePrintf("found (normalized) key %s", key)
```
alexcb, line 325: nice use of map to reduce messages :)

alexcb (changes requested):
looking good, but we need to keep the check for `errKeyScanNoMatch`.

buildcontext/gitlookup.go
idodod: actually, I just realized, if I'm reducing it for the sake of the verbose print, shouldn't I reduce it for the sake of adding it to the keys slice 🤔 ?

alexcb: yeah.... maybe that dedup should be done inside `loadKnownHosts`..... that's a good point :)

alexcb (approved):

buildcontext/gitlookup.go
idodod: so I added a dedup step when loading from the file
but I kept trying to dedup again when iterating because some entries will have a hashed host and some won't
so dedup needs to happen after normalizing the host.

Issue #3247, opened 2023-09-12
Adding earthly metadata to created images


idelvall:
As of today, images generated by Earthly cannot be differentiated from others. It would be nice for earthly to add some metadata in the image, so downstream tools were be able to known when a image has been generated by Earthly and under which version.

jmgilman:
I would suggest labels, given that they are typically used for "finding" specific images. Additionally, it would also be nice to record the original target used to build the image. The goal being the ability to detect what images (if any) a particular target created.

alexcb:
I'm not sure adding extra `LABEL`s to a user's image without them requesting it is a good idea or not -- does docker do this in any places?

idelvall:
Feedback received! Let me know your thoughts on the PR

PR #3249, opened 2023-09-13
Add "dev.earthly.*" LABELS to saved images

idelvall:
Fixes https://github.com/earthly/earthly/issues/3247


earthfile2llb/interpreter.go
```diff
@@ -15,6 +15,7 @@ import (
 	"github.com/earthly/earthly/ast/hint"
 	"github.com/earthly/earthly/ast/spec"
 	"github.com/earthly/earthly/buildcontext"
+	"github.com/earthly/earthly/cmd/earthly/version"
```
alexcb, line 18: we shouldn't be importing any `cmd` packages from non-cmd packages -- this is to avoid getting into circular dependencies.

In this specific case `version` is definitely tied to the cmd, but it still feels a bit odd -- how about renaming the `github.com/earthly/earthly/cmd/earthly/version` package to `github.com/earthly/earthly/internal/version`.

earthfile2llb/interpreter.go
```diff
@@ -1127,6 +1128,18 @@ func (i *Interpreter) handleSaveImage(ctx context.Context, cmd spec.Command) err
 		fmt.Fprintf(os.Stderr, "Deprecation: using SAVE IMAGE with no arguments is no longer necessary and can be safely removed\n")
 		return nil
 	}
+
+	labels := make(map[string]string)
+	labels["dev.earthly.version"] = version.Version
+	labels["dev.earthly.git-sha"] = version.GitSha
+	labels["dev.earthly.built-by"] = version.BuiltBy
+	labels["dev.earthly.default-installation-name"] = version.DefaultInstallationName
```
alexcb, line 1136: What's the use-case for storing the default install name (and buildkitd image)? These values can be overridden at run time (or even could be using a remote buildkit instance, e.g. satellite, which could be a different version).

earthfile2llb/interpreter.go
```diff
@@ -1127,6 +1128,18 @@ func (i *Interpreter) handleSaveImage(ctx context.Context, cmd spec.Command) err
 		fmt.Fprintf(os.Stderr, "Deprecation: using SAVE IMAGE with no arguments is no longer necessary and can be safely removed\n")
 		return nil
 	}
+
+	labels := make(map[string]string)
```
alexcb, line 1132: you can define the map with key/values in it:

```
labels := map[string]string{
    "dev.earthly.version": version.Version,
    "dev.earthly.git-sha": version.GitSha,
    ...
    }
```

earthfile2llb/interpreter.go
```diff
@@ -1127,6 +1128,18 @@ func (i *Interpreter) handleSaveImage(ctx context.Context, cmd spec.Command) err
...
+	labels["dev.earthly.version"] = version.Version
+	labels["dev.earthly.git-sha"] = version.GitSha
+	labels["dev.earthly.built-by"] = version.BuiltBy
+	labels["dev.earthly.default-installation-name"] = version.DefaultInstallationName
+	labels["dev.earthly.default-buildkitd-image"] = version.DefaultBuildkitdImage
+	err = i.converter.Label(ctx, labels)
+	if err != nil {
+		return i.wrapError(err, cmd.SourceLocation, "save image (label)")
```
alexcb, line 1140: Our errors tend to start with a `failed to ...` message.

maybe something like this instead:
```suggestion
		return i.wrapError(err, cmd.SourceLocation, "failed to create dev.earthly labels during save image")
```

alexcb (changes requested):
looking good, just a few minor issues.

earthfile2llb/interpreter.go
idelvall: No use case, just added them there for completion. Wasn't aware they are mutable

idelvall:
Thanks Alex, it makes sense.
Everything should be done. 
BTW checked the case when multiple SAVE IMAGES are used in the same target, and looks like buildkit deduplicates labels, so working as expected

alexcb:
looking good!

Sorry one more request, can you add a test for this? You could add to any of the existing tests in `scripts/tests/export.sh`.

alexcb:
e.g. `"$frontend" inspect earthly-export-test-3:test | jq .[].Config.Labels ..........`

idelvall:
This should be it @alexcb

tests/Earthfile
```diff
@@ -1413,6 +1414,13 @@ test-init-golang:
     DO +RUN_EARTHLY --target "+build"
     DO +RUN_EARTHLY --target "+test"
 
+test-reserved-label:
+   DO +RUN_EARTHLY \
+       --earthfile=reserved-label.earth \
+       --target=+test1 \
+       --should_fail=true \
+       --output_contains="forbidden"
```
alexcb, line 1422: Is it possible to expand this to something more complete? 

This field is actually a regex. The quotes in the error message might make it difficult to include, but you can just use a `.` in place of the `"` to simplify the pattern.


```suggestion
       --output_contains="LABEL keys starting with .dev.earthly.. are forbidden"
```

scripts/tests/export.sh
```diff
@@ -211,4 +211,38 @@ EOF
...
+    LABEL foo=bar
+    SAVE IMAGE earthly-export-test-8b:test
+EOF
+
+"$earthly" prune --reset
+"$earthly" +test8
+
+label_count=$("$frontend" inspect earthly-export-test-8a:test | jq .[].Config.Labels | grep -c dev.earthly.)
+if [ "$label_count" -ne "3" ]; then
+    echo "Expected 3 dev.earthly labels on first image"
```
alexcb, line 238: it's helpful to see the actual value for cases where it fails.
```suggestion
    echo "Expected 3 dev.earthly labels on first image; got $label_count"
```

scripts/tests/export.sh
```diff
@@ -211,4 +211,38 @@ EOF
...
+    exit 1
+fi
+
+label_count=$("$frontend" inspect earthly-export-test-8b:test | jq .[].Config.Labels | grep -c dev.earthly.)
+if [ "$label_count" -ne "3" ]; then
+    echo "Expected 3 dev.earthly labels on second image"
```
alexcb, line 244:
```suggestion
    echo "Expected 3 dev.earthly labels on second image; got $label_count"
```

alexcb (approved):
looks good -- just left a few minor suggestions.

PR #3251, opened 2023-09-13
Logstream: refactor, simplify, resume

mikejholly:
This PR aims to improve the resiliency & memory usage of shipping log stream data to the server. 

It contains a few changes:
* Use a dynamically scaling buffered channel to eliminate blocking and pre-allocation of a large buffered channel.
* Adds support for the resumption of failed connections without replaying the entire history.
* Simplifies and unifies some code into a new `LogShipper`.
* Use a separate gRPC connection with custom rules for log streaming.
* Lots of tests to ensure it works.

A few notes on how the streaming works:
* Log deltas (manifest fields & logs) are sent from BK & Earthly to the custom buffered channel.
* The `StreamLogs` client method handles multiple stream attempts, including resumption.
* When a stream fails, a new one is started using a "resume" delta with a matching token.
* Any log messages that fail to send are included as the first item in the next attempt.
* If the build finishes and there are still logs to send, the code will spend 10s attempting to wrap up.


cloud/client.go
```diff
@@ -141,13 +157,34 @@ func (c *Client) getRequestID() string {
 	return uuid.NewString()
 }
 
-func logstreamClient(ctx context.Context, defaultConn grpc.ClientConnInterface, overrideAddr string, dialOpts ...grpc.DialOption) (logstream.LogStreamClient, error) {
-	if overrideAddr == "" {
-		return logstream.NewLogStreamClient(defaultConn), nil
+var serviceConfig = `{
```
alexcb, line 160: this is so scary..... but https://pkg.go.dev/google.golang.org/grpc#WithDefaultServiceConfig confirms this is what's expected..... there be dragons :dragon: :ship: :see_no_evil:

mikejholly: Yeah, I don't at all understand why there's only the string option for specifying the config. It should also be ok to express this in a struct.

cloud/logstream.go
```diff
@@ -5,69 +5,190 @@ import (
...
+		if last != nil {
+			first = append(first, last)
+		}
+		var err error
+		last, err = c.streamLogsAttempt(ctx, man.GetBuildId(), first, ch)
+		if err != nil {
+			if retryable(err) {
+				retry = true
+				if verbose {
+					errs = append(errs, err)
```
alexcb, line 37: this feels problematic, if the error is retryable, and on the next attempt it succeeds we will still have appended the initial error to `errs`, which will ultimately get returned below under:

```
	if len(errs) > 0 {
		return &multierror.Error{Errors: errs}
	}
```

alexcb: maybe we need to change:

```
if err != nil {
   ...
}
break // change this to a `return nil` instead?
```

cloud/logstream.go
```diff
@@ -5,69 +5,190 @@ import (
...
+				}
+				continue
```
alexcb, line 39: is there any exponential backoff in this retry loop?

logbus/bus.go
```diff
@@ -132,11 +132,9 @@ func (b *Bus) WriteDeltaManifest(dm *logstream.DeltaManifest) {
 			DeltaManifest: dm,
 		},
 	}
-
 	b.rawMu.Lock()
+	defer b.rawMu.Unlock()
```
alexcb, line 136: is this intentionally -- didn't you make this change fairly recently? Did it turn out to introduce a race condition?

logbus/ship/ship.go
```diff
@@ -0,0 +1,118 @@
...
+	defer t.Stop()
+	select {
+	case <-t.C:
+		l.cancel()
+	case <-l.done:
+		return
+	}
+}
+
+func (l *LogShipper) Err() error {
```
alexcb, line 61: this feels like a race, since it depends on the thread started via `Start()` to complete.

alexcb (commented):
looking good; just a couple issues/questions.

logbus/bus.go
mikejholly: Yes, it led to a race in certain circumstances. If the subscribers are added/removed frequently for instance. It ended up being less of a concern now that the `Write` operations don't block to the same degree.

logbus/ship/ship.go
mikejholly: Good point. It should use a lock.

cloud/logstream.go
mikejholly: Well, the idea was to use the `verbose` argument to help surface _all_ errors that were encountered during streaming, even those that were retryable. When `--verbose` is set, we get a nice printout of all of the errors (below). Without the flag, the recoverable error messages are suppressed. 

```
	* log streamer: 9 errors occurred:
	* rpc error: code = Unavailable desc = something broke
	* rpc error: code = Unavailable desc = something broke
	* rpc error: code = Unavailable desc = something broke
	...
```

cloud/logstream.go
```diff
@@ -5,69 +5,190 @@ import (
...
-func (c *Client) StreamLogs(ctx context.Context, buildID string, deltas Deltas) error {
-	streamClient, err := c.logstream.StreamLogs(c.withAuth(ctx), grpc_retry.Disable())
+func (c *Client) streamLogsAttempt(ctx context.Context, buildID string, first []*pb.Delta, ch <-chan *pb.Delta) (*pb.Delta, error) {
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	stream, err := c.logstream.StreamLogs(c.withAuth(ctx))
 	if err != nil {
-		return errors.Wrap(err, "failed to create log stream client")
+		return nil, errors.Wrap(err, "failed to create log stream client")
```
alexcb, line 58:
```suggestion
		return nil, errors.Wrap(err, "failed to stream logs")
```

cloud/logstream.go
mikejholly: Not at the moment. It's probably worth adding something simple to prevent spamming.

cloud/logstream.go
alexcb: I like the idea of `--verbose` displaying recovered errors.

I'm trying to understand how a caller would differentiate between a non-recoverable error and an error that was recovered, because simply checking `if err != nil` wouldn't work in this case.

What if instead we pass in a reference to `VerbosePrintf` or some console logger instead -- that way we would display these errors as they occur rather than waiting for all the errors to occur -- this could be especially helpful where we perform a `time.Sleep(...)` between retry attempts.

mikejholly: I suppose it would be helpful to differentiate between the 2 error types. I do like the idea of keeping a summary as it's difficult to catch the errors when they're printed inline.

We could use a new error type to capture whether the error was recoverable. In that case, the client can decide how to handle the errors. We could also remove the `verbose` argument from this function. That would be nice since error display isn't really the concern of this component.
```
// Create an error type that specifies whether the error was recoverable.
type StreamError {
  Recoverable bool
  Err err
}
```

logbus/setup/setup.go
```diff
@@ -122,40 +122,39 @@ func (bs *BusSetup) DumpManifestToFile(path string) error {
...
 	if bs.LogStreamer != nil {
-		manifestsWritten, logsWritten, err := bs.LogStreamer.Close()
-		if err != nil {
-			retErr = multierror.Append(retErr, errors.Wrap(err, "log streamer"))
+		bs.LogStreamer.Close()
+		if errs := bs.LogStreamer.Errs(); len(errs) > 0 {
+			multi := &multierror.Error{}
+			for _, err := range errs {
+				streamErr := &cloud.StreamError{}
+				if bs.verbose || (errors.As(err, &streamErr) && !streamErr.Recoverable) {
```
alexcb, line 151: What would happen if `errors.As(err, &streamErr)` fails? This feels like a potential for a different error type to end up here, which would then be ignored.

Currently the code shouldn't allow this to happen, but this feels fragile as a future code change could return a different error type without knowing this restriction.

We should either append it, or panic().

alexcb (commented):
looks good -- just one last issue.

alexcb (approved):

PR #3254, opened 2023-09-15
WITH DOCKER should switch to iptable-legacy in WSL

alexcb:
WSL does not support iptables-nft; for WITH DOCKER to work with kind; iptables must be switched to legacy.

### DIFF
--- a/buildkitd/dockerd-wrapper.sh
+++ b/buildkitd/dockerd-wrapper.sh
@@ -120,6 +120,13 @@ start_dockerd() {
     data_root=$(TMPDIR="$EARTHLY_DOCKERD_DATA_ROOT/" mktemp -d)
     echo "Starting dockerd with data root $data_root"
 
+    if uname -a | grep microsoft-standard-WSL >/dev/null; then
+        if iptables --version | grep nf_tables >/dev/null; then
+            echo "WARNING: WSL and iptables-nft may not work; attempting to switch to iptables-legacy"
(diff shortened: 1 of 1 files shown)


alexcb:
here's an example of the warning message:

image

(I'm testing via a remote connection, and can't copy/paste the terminal into my clipboard)

mikejholly (approved):

buildkitd/dockerd-wrapper.sh
```diff
@@ -120,6 +120,13 @@ start_dockerd() {
     data_root=$(TMPDIR="$EARTHLY_DOCKERD_DATA_ROOT/" mktemp -d)
     echo "Starting dockerd with data root $data_root"
 
+    if uname -a | grep microsoft-standard-WSL >/dev/null; then
+        if iptables --version | grep nf_tables >/dev/null; then
+            echo "WARNING: WSL and iptables-nft may not work; attempting to switch to iptables-legacy"
+            ln -sf "/sbin/iptables-legacy" /sbin/iptables
+        fi
+    fi
+
```
ingwarsw, line 129: @alexcb In latest versions of debian its also like that..
In most of my scripts I have..
```
    DO github.com/earthly/lib+INSTALL_DIND
    # Bit hacky hack to fix docker on new systems
    RUN update-alternatives --set iptables /usr/sbin/iptables-legacy
    DO +TEST_WITH_DOCKER --cmd="some"
```

Issue #3255, opened 2023-09-18
Satellite error (reading from server: read: connection timed out)


idelvall:
We have been experiencing this failure occasionally, when running in satellites over last days.
I've seen this happening both in GH actions as well as locally.
```
Init 🚀
————————————————————————————————————————————————————————————————————————————————

           satellite | satname is waking up. Please wait...
           satellite | ...amplifying transmission signal...
           satellite | ...testing antennas...
Shareable link: https://ci.earthly.dev/logs?logId=a9f1cc10-4427-4b05-b3ba-9244da7ffd25
Error: error reading from server: read tcp X.X.X.X:Y->100.21.239.443:443: read: connection timed out

Error: Process completed with exit code 1.
```

## Multiple jobs failing at the same time
The error is non-deterministic and happens rarely. 
The time window for reproducing it is narrow, since usually, when retrying it right away, it can't be reproduced, but if multiple GH jobs are running in parallel in the same satellite when it happens, the failure affects multiple of them.

For example, these two errors happened simultaneously in two different GH jobs. Notice the satellite state received was different, but still they failed:
```
Init 🚀
————————————————————————————————————————————————————————————————————————————————

           satellite | satname is waking up. Please wait...
           satellite | ...amplifying transmission signal...
           satellite | ...testing antennas...
Shareable link: https://ci.earthly.dev/logs?logId=a9f1cc10-4427-4b05-b3ba-9244da7ffd25
Error: error reading from server: read tcp X.X.X.X:Y->100.21.239.443:443: read: connection timed out

Error: Process completed with exit code 1.
```
```
Init 🚀
————————————————————————————————————————————————————————————————————————————————

           satellite | satname is starting. Please wait...
           satellite | ...perturbing matrices...
Shareable link: https://ci.earthly.dev/logs?logId=989e68d2-5577-4088-868a-649f0dec0852
Error: error reading from server: read tcp X.X.X.X:Y->100.21.239.443:443: read: connection timed out

Error: Process completed with exit code 1.
```

brandonSc:
Adding a simple retry mechanism to the earthly client will help to recover from these situations, however, more investigation still needs to be done for us to understand why it is disconnecting in the first place.

brandonSc:
> I am starting to think the problem the gRPC connection is terminated at that point, and the code times out later on when doing another call to the client.

[...]

PR #3260, opened 2023-09-18
retry stream

idelvall:
Mitigates https://github.com/earthly/earthly/issues/3255

### DIFF
--- a/cloud/satellite.go
+++ b/cloud/satellite.go
@@ -198,35 +198,58 @@ func (c *Client) ReserveSatellite(ctx context.Context, name, orgName, gitAuthor,
 		ctxTimeout, cancel := context.WithTimeout(ctx, 10*time.Minute)
 		defer cancel()
 		defer close(out)
-		stream, err := c.compute.ReserveSatellite(c.withAuth(ctxTimeout), &pb.ReserveSatelliteRequest{
-			OrgId:          orgID,
-			Name:           name,
-			CommitEmail:    gitAuthor,
-			GitConfigEmail: gitConfigEmail,
-			IsCi:           isCI,
-		})
-		if err != nil {
-			out <- SatelliteStatusUpdate{Err: errors.Wrap(err, "failed opening satellite reserve stream")}
-			return
-		}
-		for {
-			update, err := stream.Recv()
-			if err == io.EOF {
-				return
-			}
+		for i := 1; i <= 5; i++ {
+			stream, err := c.compute.ReserveSatellite(c.withRetryCount(c.withAuth(ctxTimeout), i), &pb.ReserveSatelliteRequest{
+				OrgId:          orgID,
+				Name:           name,
+				CommitEmail:    gitAuthor,
+				GitConfigEmail: gitConfigEmail,
+				IsCi:           isCI,
+			})
 			if err != nil {
-				out <- SatelliteStatusUpdate{Err: errors.Wrap(err, "failed receiving satellite reserve update")}
+				out <- SatelliteStatusUpdate{Err: errors.Wrap(err, "failed opening satellite reserve stream")}
 				return
 			}
-			status := satelliteStatus(update.Status)
-			if status == SatelliteStatusFailed {
-				out <- SatelliteStatusUpdate{Err: errors.New("satellite is in a failed state")}
-				return
+			var lastStatus string
+			for {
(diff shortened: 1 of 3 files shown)


cloud/satellite.go
```diff
@@ -198,37 +198,60 @@ func (c *Client) ReserveSatellite(ctx context.Context, name, orgName, gitAuthor,
...
 	}()
 	return out
 }
 
+func isFinalStatus(status string) bool {
+	switch status {
+	case SatelliteStatusOperational:
+		return true
+	case SatelliteStatusFailed:
+		return true
```
mikejholly, line 249:
```suggestion
	case SatelliteStatusOperational, case SatelliteStatusFailed:
		return true
```

Nit: cases can be combined.

cloud/satellite.go
```diff
@@ -279,31 +312,41 @@ func (c *Client) SleepSatellite(ctx context.Context, name, orgName string) (out
...
-		if err != nil {
-			out <- SatelliteStatusUpdate{Err: errors.Wrap(err, "failed opening satellite sleep stream")}
-			return
-		}
-		for {
-			update, err := stream.Recv()
-			if err == io.EOF {
-				return
-			}
+		for i := 1; i <= 5; i++ {
```
mikejholly, line 315: It seems like there's a bunch of duplicated code here. Can we create a reusable method that accepts the client call invocation as a function parameter?

cloud/satellite.go
```diff
@@ -240,30 +263,40 @@ func (c *Client) WakeSatellite(ctx context.Context, name, orgName string) (out c
...
-		if err != nil {
-			out <- SatelliteStatusUpdate{Err: errors.Wrap(err, "failed opening satellite wake stream")}
-			return
-		}
-		for {
-			update, err := stream.Recv()
-			if err == io.EOF {
-				return
-			}
+		for i := 1; i <= 5; i++ {
```
mikejholly, line 266: Can we use a constant since this retry count is repeated.

cloud/satellite.go
```diff
@@ -198,37 +198,60 @@ func (c *Client) ReserveSatellite(ctx context.Context, name, orgName, gitAuthor,
...
-		if err != nil {
-			out <- SatelliteStatusUpdate{Err: errors.Wrap(err, "failed opening satellite reserve stream")}
-			return
-		}
-		for {
-			update, err := stream.Recv()
-			if err == io.EOF {
-				return
-			}
+		for i := 1; i <= 5; i++ {
```
mikejholly, line 201: I see that there's another timeout above. But the hard-coded attempt count is less desirable than allowing the process to retry indefinitely within a max time window. Depending on the particular issue, the code may fail 5 times immediately or spend a while on each attempt. This library is pretty good and it can retry until a context times out: https://github.com/avast/retry-go

cloud/satellite.go
brandonSc: TBH I think we should just limit the patch to the reserve call anyway. This is a bit of a weird band-aid solution that I'm not sure we need to propagate everywhere. We could leave it in the one place that is affecting users and see if it improves things first.

cloud/satellite.go
brandonSc: I do like that library, but I'm also okay with the simple for-loop for now until we see if this is improving things or not. It may also be a temporary while we figure out what is going on with the unexpected hang-ups anyway.

brandonSc:
Added a commit @idelvall with an error log and removed the retries from sleep/wake calls for now since we aren't having issues there. Those calls are rarely invoked by users anyway (especially in their CI).

cloud/satellite.go
mikejholly: SGTM :+1:

brandonSc (approved):
LGTM but I am not 100% sure yet that this mitigates the error. I'll be trying to reproduce it tomorrow on staging.

I also think it's okay to merge if it works under normal circumstances and with the possibility that it might help. In cases where the issue occurs, the build fails anyway so we can't make things any worse.

idelvall:
Thanks Brandon, makes sense.

idelvall:
Just in case you missed it, this is why I think this will solve the problem: https://github.com/earthly/earthly/issues/3255#issuecomment-1724182400

alexcb (approved):
just one minor nit

PR #3266, opened 2023-09-20
Change how long console prefixes get printed

idodod:
Partially addresses https://github.com/earthly/earthly/issues/3200
 - Fix issue where target urls don't get shorter properly (a regression since v0.7.x).
 - Refactor `prettyprefix` function so it would cache prefixes instead of formatting every time anew.
 - Change GIT CLONE prefix to only print the git repo url once.
 - Shorten remote (git based) targets when used as prefix
 - Fix bug where credentials were wiped but only incidentally


conslogging/prefix_formatter.go
```diff
@@ -0,0 +1,128 @@
...
+		if curLen <= padding {
+			return fmt.Sprintf("%s%c", s, suffixChar)
+		}
+		return fmt.Sprintf("%s%c", truncateURL(s, padding, curLen), suffixChar)
+	})
+}
+
+func padStr(s string, padding int) string {
+	formatString := fmt.Sprintf("%%%vv", padding)
+	return fmt.Sprintf(formatString, s)
```
alexcb, line 90: this can be done via `*`:
```suggestion
	return fmt.Sprintf("%*s", padding, s)
```

alexcb (commented):
I like the refactor!

looks like there's some unit-tests failing; I can take another look once they are passing.

conslogging/conslogging.go
```diff
@@ -460,54 +459,8 @@ func (cl ConsoleLogger) color(c *color.Color) *color.Color {
...
 
-var (
-	// githubRegExp Matches :2dd88e53f2e59e96ec1f9215f24a3981e5565edf+ in a prefix.
-	// 	Prefix containing hash may resemble: g/e/hello-world:2dd88e53f2e59e96ec1f9215f24a3981e5565edf+base
-	//	Prefix must be exactly 40 characters
-	githubRegExp = regexp.MustCompile(`:[a-f0-9]{40}\+`)
-)
-
 func prettyPrefix(prefixPadding int, prefix string) string {
-	if prefixPadding <= NoPadding {
```
idodod, line 471: all of this was moved and refactored into prefix_formatter.go

earthfile2llb/converter.go
```diff
@@ -2395,7 +2395,7 @@ func (c *Converter) imageVertexPrefix(id string, platform platutil.Platform) str
 }
 
 func (c *Converter) vertexPrefixWithURL(url string) string {
-	return fmt.Sprintf("[%s(%s) %s] ", c.mts.Final.Target.String(), url, url)
```
idodod, line 2398: printing the url twice seemed redundant, I can put this back if there is a good reason I was unaware of

conslogging/conslogging_test.go
```diff
@@ -54,7 +54,7 @@ func Test_prettyPrefix(t *testing.T) {
 			name:          "simple target with path",
 			prefixPadding: DefaultPadding,
 			prefix:        "github.com/earthly/earthly+run",
-			expected:      "github.com/earthly/earthly+run",
```
idodod, line 57: this is a regression that was fixed, before earthly `v0.7.0` we used to shorten these urls, since `v0.7.0`
we only shorten addresses such as `github.com/earthly/earthly:80524f0d82a353b3444e83f056207e15f4d5447c+hello-world`

vladaionescu (approved):

CHANGELOG.md
```diff
@@ -10,6 +10,20 @@ All notable changes to [Earthly](https://github.com/earthly/earthly) will be doc
 ### Changed
 - `GIT CLONE` URLs will only be printed once as part of a prefix, e.g. `+my-clone-target(https://g/e/earthly) | --> GIT CLONE (--branch ) https://github.com/earthly/earthly`
 
+## v0.7.19 - 2023-09-20
```
alexcb, line 13: something looks wrong here -- why is the CHANGELOG being updated for a previously released version?

alexcb: not sure what happened here -- maybe a new commit was pushed, or some bug in github? CHANGELOG looks fine now.

alexcb (approved):

PR #3267, opened 2023-09-20
Fix request ID logging in gRPC streams 

brandonSc:
Previously, our gRPC streaming middleware was only intercepting errors that occurred at stream creation. We need to wrap the ClientStream and override its Recv and Send msg functions to intercept errors that occur _during_ the stream.

Screen Shot 2023-09-20 at 11 10 14 AM


cloud/grpc.go
```diff
@@ -70,23 +74,22 @@ func (c *Client) UnaryInterceptor(opts ...InterceptorOpt) grpc.UnaryClientInterc
 		}
 		ctx, err := c.reAuthIfExpired(ctx)
 		if err != nil {
-			return errors.Wrap(err, "failed refreshing expired token")
+			return errors.Wrapf(err, "failed refreshing expired token: %s", getReqID(ctx))
```
alexcb, line 77: nit: remove the `:`, otherwise it would be `failed refreshing expired token: 123: some-error-code`
```suggestion
			return errors.Wrapf(err, "failed refreshing expired token %s", getReqID(ctx))
```

alexcb: better yet, make it consistent with the change you made lower down in the code:

```
errors.Wrapf(err, "failed refreshing expired token {reqID: %s}", getReqID(ctx))
```

cloud/grpc.go
```diff
@@ -100,26 +103,72 @@ func (c *Client) StreamInterceptor() grpc.StreamClientInterceptor {
...
 		if err != nil {
-			return nil, errors.Wrap(err, "failed refreshing expired token")
+			return nil, errors.Wrapf(err, "failed refreshing expired token {reqID: %s}", getReqID(ctx))
 		}
 		newStreamer, err := streamer(ctx, desc, cc, method, opts...)
 		if err != nil {
 			s, ok := status.FromError(err)
 			if !ok {
-				return newStreamer, err
+				return nil, fmt.Errorf("%s {reqID: %s}", err.Error(), getReqID(ctx))
```
alexcb, line 112: why aren't you using `errors.Wrapf` like in the other error handling of this code?

cloud/grpc.go
```diff
@@ -100,26 +103,72 @@ func (c *Client) StreamInterceptor() grpc.StreamClientInterceptor {
...
+			return nil, status.Errorf(s.Code(), fmt.Sprintf("%s {reqID: %s}",
+				cleanStatusError(err.Error()), getReqID(ctx)))
+		}
+		return newWrappedStream(ctx, newStreamer), nil
+	}
+}
+
+// wrappedStream  wraps around the embedded grpc.ClientStream, and intercepts the RecvMsg and
+// SendMsg method call.
+type wrappedStream struct {
```
alexcb, line 130: `wrappedStream` seems too generic, maybe call it `requestIDErrorStreamWrapper` or something else?

alexcb (commented):
looks good; just a few small issues.

cloud/grpc.go
brandonSc: Couple reasons:
* I don't want to lose the request ID if someone prints `errors.Cause(err)` (which is exactly what we [usually do](https://github.com/earthly/earthly/blob/main/cmd/earthly/app/run.go#L252))
* I didn't have a meaningful message to add, and just wanted leave the original error unaltered while appending the request ID to it

TBH I don't think I should use `errors.Wrapf` above either

alexcb: > TBH I don't think I should use errors.Wrapf above either with the risk mention in the first point

sounds good -- we should be consistent in how we do it.

PR #3269, opened 2023-09-20
Nacho/retry reserve

idelvall:
This PR:
- Retries also initial stream connection
- Removes the "retrying in %d seconds" message, since the retry is performed after the TCP timeout (around 2.5 m in Linux), so the message is misleading for the user
- Increased the context timeout to give more room for retries


alexcb (approved):
looks good, but it's a bit confusing that this PR has 11 commits.

PR #3276, opened 2023-09-21
Fix broken links in docs

idodod:


docs/basics/part-5-importing.md
```diff
@@ -72,7 +72,7 @@ build:
 ```
 This code tells `FROM` that there is another Earthfile in  the `services/service-one` directory and that the Earthfile  contains a target called `+deps`. In this case, if we were to run `+build` Earthly is smart enough to go into the subdirectory, run the  `+deps` target in that Earthfile, and then use it as the base image for `+build`.
 
-We can also reference an Earthfile in another repo, which works in a similar way. If the reference does not begin with one of `/`, `./`, or `../`, then earthly treats it as a repository.  See [the reference](../earthfile#from) for details.
+We can also reference an Earthfile in another repo, which works in a similar way. If the reference does not begin with one of `/`, `./`, or `../`, then earthly treats it as a repository.  See [the reference](https://tinyurl.com/bdfys6ee) for details.
```
alexcb, line 75: this seems wrong -- the tinyurls are used to link to Earthfiles in github; where as this link should stay within the gitbook website.

idodod: I seemed to have encountered the same bug (gitbook adding `README.md` to the url path):
I used a relative path in the previous commit, you check it here (search for "this reference"): 
https://app.gitbook.com/o/-MB5s2W0v8S4xFU5ELti/s/yi347RZ02JXdgz9BJx5u/~/revisions/dYGQvGmUZ6scZkIrHqHi/basics/part-5-importing

alexcb: does `[the reference](../earthfile/earthfile.md#from)` work?

idodod: it does! good thinking!

alexcb (approved):

PR #3282, opened 2023-09-22
Docs: Update to guide on dropping satellite cache

brandonSc:


docs/cloud/satellites/managing.md
```diff
@@ -79,16 +79,18 @@ Currently selected: No
...
 
 There are two ways to clear the cache on satellite.
+One method may be faster than the other, depending on the size of your cache.
 
-### Recreating the Underlying Satellite Instance (often faster)
+### Recreating the Underlying Satellite Instance
 
 Running the `update` command with `--drop-cache` will relaunch the instance with an empty cache volume.
 Note that this operation can take a while, and the satellite may also receive any available updates during the process.
+The satellite must also be in a sleep state first, you can force a satellite to sleep by running `earthly satellite sleep`
```
vladaionescu, line 88: Could we just make it sleep automatically? That would be a nicer UX.

brandonSc: I like that idea. I think it would make sense for earthly cli to drive that, and perhaps ask the user if they want to force it asleep before proceeding

vladaionescu: It's better if it's not interactive.. maybe it fails, unless `--force` is used.

alexcb (commented):
I like Vlad's suggesting on having it auto-sleep.

(I'm mostly commenting here to prevent future daily notifications that this PR is still open)

docs/cloud/satellites/managing.md
alexcb: @brandonSc we ran into this use-case today where a user needed to do a `--drop-cache` on a shared satellite. Even chaining two commands together, e.g. `earthly satellite sleep my-sat && earthly satellite update --drop-cache my-sat` didn't always work as other users (or the CI) would cause a wake to occur before the second command was executed.

vladaionescu: Interesting... so a client-side-only solution would not be enough. We need the server to support this atomically.

PR #3290, opened 2023-09-25
Add CACHE mode support

idelvall:
Right now, we cannot set the permissions for a `CACHE` directory. This PR fixes that.
_(Also, realized that we were using `0` as the default value for cache mount mode, so changing that to the `0644` standard in this same PR)_

I've noticed that we are using "mode" and "chmod" across different commands. I can take this opportunity to do some clean-up and unify into a single term. Personally I think "chmod" is more expressive. Thoughts?

### DIFF
--- a/ast/commandflag/flags.go
+++ b/ast/commandflag/flags.go
@@ -138,5 +138,5 @@ type PipelineOpts struct {
 
 type CacheOpts struct {
 	Sharing string `long:"sharing" description:"The cache sharing mode: locked (default), shared, private"`
-	Mode    string `long:"mode" description:"Apply a mode to the cache folder" default:"0644"`
+	Mode    string `long:"chmod" description:"Apply a mode to the cache folder" default:"0644"`
 }
--- a/ast/commandflag/flags.go
+++ b/ast/commandflag/flags.go
@@ -138,4 +138,5 @@ type PipelineOpts struct {
 
 type CacheOpts struct {
 	Sharing string `long:"sharing" description:"The cache sharing mode: locked (default), shared, private"`
+	Mode    string `long:"mode" description:"Apply a mode to the cache folder" default:"0644"`
 }
--- a/docs/earthfile/earthfile.md
+++ b/docs/earthfile/earthfile.md
@@ -274,7 +274,7 @@ The `<mount-spec>` is defined as a series of comma-separated list of key-values.
 |-----------| --- | --- |
 | `type`    | The type of the mount. Currently only `cache`, `tmpfs`, and `secret` are allowed. | `type=cache` |
 | `target`  | The target path for the mount. | `target=/var/lib/data` |
-| `chmod`   | The permission of the mounted file, in octal format (the same format the chmod unix command line expects). | `chmod=0400` |
+| `mode`, `chmod`   | The permission of the mounted file, in octal format (the same format the chmod unix command line expects). | `chmod=0400` |
 | `id`      | The secret ID for the contents of the `target` file, only applicable for `type=secret`. | `id=my-password` |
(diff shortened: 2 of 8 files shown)


ast/commandflag/flags.go
```diff
@@ -138,4 +138,5 @@ type PipelineOpts struct {
 
 type CacheOpts struct {
 	Sharing string `long:"sharing" description:"The cache sharing mode: locked (default), shared, private"`
+	Mode    string `long:"mode" description:"Apply a mode to the cache folder" default:"0644"`
```
alexcb, line 141: can we call this `chmod` instead?

I don't see any existing instances of `mode` being used:

```
alex@surfperch:~/gh/earthly/earthly$ grep -RI 'long:"mode"' ast/commandflag/
alex@surfperch:~/gh/earthly/earthly$ grep -RI 'long:"chmod"' ast/commandflag/
ast/commandflag/flags.go:	Chmod           string   `long:"chmod" description:"Apply a mode to the copied files and directories"`
```

idelvall: `RUN --mount`:
image

idelvall: I also prefer `chmod`, going with it then.

alexcb: ah ha! here's the other instance.

I wonder if `mode` is too generic in the case of `CACHE --mode=...`. I don't think it's obvious that it's a file-permission-mode rather than a mode pertaining to the cache. where as `chmod` is the name of the command line utility for changing file modes, and seems easier to google for rather than just `mode`. 

In the case of the mount option `mode` of the `RUN` command -- it was picked to stay consistent with dockerfiles.

Alternatively maybe we should rename this to `mount-mode` (for both `CACHE` and `COPY` commands) -- that way it would also reflect the `RUN --mount .... mode=...` option. (and deprecate the old names)

@vladaionescu any thoughts on this rename?

idelvall: I was thinking on documenting `chmod` both for `CACHE` and `RUN --mount`, while silently supporting mode for the latter, so dockerfile portability is maintained

vladaionescu: I like @idelvall's latest suggestion.

idelvall:
This should be ready

docs/earthfile/earthfile.md
```diff
@@ -270,12 +270,12 @@ Mounts a file or directory in the context of the build environment.
...
-| --- | --- | --- |
-| `type` | The type of the mount. Currently only `cache`, `tmpfs`, and `secret` are allowed. | `type=cache` |
-| `target` | The target path for the mount. | `target=/var/lib/data` |
-| `mode` | The permission of the mounted file, in octal format (the same format the chmod unix command line expects). | `mode=0400` |
-| `id` | The secret ID for the contents of the `target` file, only applicable for `type=secret`. | `id=my-password` |
+| Key       | Description | Example |
+|-----------| --- | --- |
+| `type`    | The type of the mount. Currently only `cache`, `tmpfs`, and `secret` are allowed. | `type=cache` |
+| `target`  | The target path for the mount. | `target=/var/lib/data` |
+| `chmod`   | The permission of the mounted file, in octal format (the same format the chmod unix command line expects). | `chmod=0400` |
```
alexcb, line 277: we should document that both are supported:

```suggestion
| `mode`, `chmod`   | The permission of the mounted file, in octal format (the same format the chmod unix command line expects). | `chmod=0400` |
```

alexcb (commented):
looks good, just one minor suggestion regarding the docs.

alexcb (approved):

Issue #1129, opened 2021-07-29
Add shared cache option


xificurC:
Here's an updated proposal:

CACHE --id=my-shared-cache /target
RUN --mount type=cache,id=my-shared-cache,target=/target ...
The idea is that as long as the id matches, the cache would be shared across targets, regardless of where those targets exist: same Earthfile, or different Earthfile, completely different builds, etc. So presumably, even completely different projects could reuse the same cache mount.

---- original request below ----

real world use case: building a maven project. I want to reuse a .m2 cache in subsequent runs and across different images. This shouldn't break anything really and speed up my builds significantly. A `-v` flag like docker build has would do the trick, or any similar solution

vladaionescu:
This could be problematic for two reasons:
- The cache might grow indefinitely and it's up to the user to GC it regularly
- It can be a source of inconsistency

In addition, the implementation ideally should not be based on mounting a local directory, because it would not have a good equivalent with the future remote Earthly capabilities.

That said, there are some relevant use-cases for this to help with speeding up builds and it has been requested a few times.

xificurC:
Thanks for a quick response.

Running `mvn package` or any other maven command on a desktop reuses the `.m2` cache, for every project. If I put these projects in an `Earthfile`, I can't reproduce this behavior. Is that correct? If yes, this is a blocker for us right now.

> The cache might grow indefinitely and it's up to the user to GC it regularly

True, but isn't the problem the same with earthly caches? And since they aren't shared by default I can see it cluttering up the disk faster and making it harder to find the culprits, since they are spread across multiple project builds. Here I just have 1 `.m2` folder that in the direst case I can wipe out completely and let it refill only the necessary dependencies on the first run of each project.

> It can be a source of inconsistency

In the general case, yes.

> the implementation ideally should not be based on mounting a local directory, because it would not have a good equivalent with the future remote Earthly capabilities

Since I don't know what you mean I can't comment.

vladaionescu:
We've put together a proposal that might address the need for this.

https://github.com/earthly/earthly/issues/1399

Please leave feedback and/or 👍  over there if it serves your needs. Also, I'm curious about specific use-cases where it might not and sharing an actual cache between targets is strictly needed.

vladaionescu:
Some details above are a bit out of date. Here's an updated proposal:

```
CACHE --id=my-shared-cache /target
RUN --mount type=cache,id=my-shared-cache,target=/target ...
```

[...]

PR #3296, opened 2023-09-26
Add shared cache mount

idelvall:
Fixes https://github.com/earthly/earthly/issues/1129


mikejholly (approved):
Code looks good, but let's get @alexcb or @vladaionescu to review as well.

docs/earthfile/earthfile.md
```diff
@@ -270,13 +270,13 @@ Mounts a file or directory in the context of the build environment.
...
+| Key       | Description                                                                                                                                                         | Example |
+|-----------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------| --- |
+| `type`    | The type of the mount. Currently only `cache`, `tmpfs`, and `secret` are allowed.                                                                                   | `type=cache` |
+| `target`  | The target path for the mount.                                                                                                                                      | `target=/var/lib/data` |
+| `mode`, `chmod`   | The permission of the mounted file, in octal format (the same format the chmod unix command line expects).                                                          | `chmod=0400` |
+| `id`      | <ul><li>The cache ID for unsafe/cross-earthfiles caches, when `type=cache`</li><li>The secret ID for the contents of the `target` file, when`type=secret`</li></ul> | `id=my-password` |
```
alexcb, line 278: does gitbook support this? (I have no idea.... I couldn't find any existing examples of us doing this anywhere when I ran `grep -RI \<li docs/`)

tests/unsafe-cache.earth
```diff
@@ -0,0 +1,11 @@
...
+FROM alpine:3.18
+
+test:
+    RUN --mount=type=cache,target=/cache-test,id=cacheid echo "aaa">/cache-test/test.txt
+    BUILD +check
+
+check:
+    CACHE --id=cacheid /cache-test2
+    RUN ls -hal /cache-test2/test.txt
+    RUN cat /cache-test2/test.txt
```
alexcb, line 11: can we test the contents of `test.txt` is `aaa` ?

tests/unsafe-cache.earth
```diff
@@ -0,0 +1,11 @@
+VERSION 0.7
+FROM alpine:3.18
+
+test:
+    RUN --mount=type=cache,target=/cache-test,id=cacheid echo "aaa">/cache-test/test.txt
+    BUILD +check
```
alexcb, line 6: There's a race condition here due to `buildAsync`, which will scan the target looking for `BUILD`s that can be pre-executed.

You could do:

```
    RUN --mount=type=cache,target=/cache-test,id=cacheid echo "aaa">/cache-test/test.txt
    WAIT
        BUILD +check
    END
```

to prevent it.

alexcb: here's an example of the race at work:

```
VERSION 0.7
FROM alpine

test:
  RUN echo "in test" && sleep 10 && echo "still in test"
  BUILD +foo

foo:
  RUN echo "in foo"
```

then run

```
alex@surfperch:~/test/build-async$ earthly-v0.7.19 +test
 Init 🚀

           buildkitd | ...Done

 Build 🔧

               +base | --> FROM alpine
               +base | [----------] 100% FROM alpine
                +foo | --> RUN echo "in foo"
               +test | --> RUN echo "in test" && sleep 10 && echo "still in test"
               +test | in test
                +foo | in foo
               +test | still in test

 Push Summary ⏫ (disabled)

 Local Output Summary 🎁
```

tests/Earthfile
```diff
@@ -1473,6 +1474,12 @@ test-cache-mode:
       --should_fail=false \
       --output_contains="drw-r--r-- /cache-folder"
 
+test-unsafe-cache:
+  DO +RUN_EARTHLY \
+      --earthfile=unsafe-cache.earth \
+      --target=+test \
+      --should_fail=false
```
alexcb, line 1481: NIT: `--should_fail=false` is the default, and can be removed.

alexcb (commented):
looks good, just a couple minor issues.

ast/commandflag/flags.go
```diff
@@ -139,4 +139,5 @@ type PipelineOpts struct {
 type CacheOpts struct {
 	Sharing string `long:"sharing" description:"The cache sharing mode: locked (default), shared, private"`
 	Mode    string `long:"chmod" description:"Apply a mode to the cache folder" default:"0644"`
+	ID      string `long:"id" description:"Cache id, to reuse the same cache across different targets and Earthfiles"`
```
vladaionescu, line 142:
```suggestion
	ID      string `long:"id" description:"Cache ID, to reuse the same cache across different targets and Earthfiles"`
```

docs/earthfile/earthfile.md
vladaionescu, line 278: I don't think this is unsafe necessarily. I wouldn't label it like that. (I renamed this PR and the original issue accordingly)

```suggestion
| `id`      | <ul><li>The cache ID for a shared cache mount to be used across other targets or Earthfiles, when `type=cache`.</li><li>The secret ID for the contents of the `target` file, when `type=secret`</li></ul> | `id=my-shared-cache`, `id=my-password` |
```

docs/earthfile/earthfile.md
```diff
@@ -1395,6 +1395,11 @@ The sharing mode for the cache mount, from one of the following:
 The permission of the mounted folder, in octal format (the same format the chmod unix command line expects). 
 Default `--chmod 0644`
 
+
+##### `--id <cache-id>`
+
+The cache ID for unsafe/cross-earthfiles caches.
```
vladaionescu, line 1401:
```suggestion
The cache ID for a shared cache volume to be used across other targets or Earthfiles.
```

tests/unsafe-cache.earth
vladaionescu, line 1: Perhaps `shared-cache-mount.earth` is a better name?

tests/Earthfile
```diff
@@ -1473,6 +1474,12 @@ test-cache-mode:
...
+
```
vladaionescu, line 1482: Can you also add a test where the cache is used across two different Earthfiles?

idelvall: Do we have some tests using multiple Earthfiles right now @alexcb ?

tests/unsafe-cache.earth
idelvall: oh interesting, I thought commands after RUN would be executed after it. Good to know

tests/Earthfile
vladaionescu: The `import` test is perhaps an example: https://github.com/earthly/earthly/blob/main/tests/Earthfile#L22. When there are multiple files involved, we usually put them under their own dir under `test/`.

alexcb: I don't think `import` is a great example because it doesn't use the `DO +RUN_EARTHLY`, which requires the dev version of earthly to be used for running the tests rather than the earthly-in-earthly approach we tend to favour.

Instead I would suggest looking at:

`pass-args-test`, which copies two different files in:
https://github.com/earthly/earthly/blob/3a9567eb17b1d4f91b495b9b505500f3df067715/tests/Earthfile#L515-L518

or, `allow-privileged-import-test` is another example which performs an `echo` to produce a secondary Earthfile: 
https://github.com/earthly/earthly/blob/3a9567eb17b1d4f91b495b9b505500f3df067715/tests/Earthfile#L502C1-L509

That's not to say you can't combine these ideas together, and create a directory for your tests, here's a more complicated example that still uses earthly-in-earthly: https://github.com/earthly/earthly/blob/3a9567eb17b1d4f91b495b9b505500f3df067715/tests/locally-in-command/Earthfile#L12-L20

docs/earthfile/earthfile.md
idelvall: apparently it doesn't

alexcb (approved):

alexcb:
@idelvall we forgot to feature-flag this (unfortunately we didn't catch this during the PR review). Can you open up another PR that adds a new [feature flag](https://github.com/earthly/earthly/blob/main/features/features.go) which will only allow this cache ID to be global when an appropriate `VERSION --shared-cache-mount 0.7` is used? Feel free to pick a different feature flag name if you think of something better.

We need to do this before releasing this feature, otherwise users who require the old behaviour will be surprised. Likewise users who want to use this new feature might get confused as to why their Earthfile doesn't work if they use is between different released versions of earthly.

PR #3299, opened 2023-09-26
gRPC proxying of image registry requests

mikejholly:
This PR adds support for proxying image registry requests to the BK-embedded registry using a gRPC stream. A local proxy is used by the container frontend to "pull" an image from the local or remote BK via the gRPC proxy.

image

Associated with: https://github.com/earthly/buildkit/pull/25

### DIFF
--- a/buildkitd/Earthfile
+++ b/buildkitd/Earthfile
@@ -5,13 +5,12 @@ FROM alpine:3.18
 buildkitd:
     ARG BUILDKIT_PROJECT
     IF [ "$BUILDKIT_PROJECT" != "" ]
-        # Update to https://github.com/earthly/earthly/pull/3293#pullrequestreview-1643149385
-        #IF case "$BUILDKIT_PROJECT" in "../*") true;; *) false;; esac
+        IF case "$BUILDKIT_PROJECT" in ../*) true;; *) false;; esac
             # Assuming this is coming from the main Earthly Earthfile.
             ARG BUILDKIT_BASE_IMAGE=../$BUILDKIT_PROJECT+build
-        #ELSE
-        #    ARG BUILDKIT_BASE_IMAGE=$BUILDKIT_PROJECT+build
-        #END
+        ELSE
+            ARG BUILDKIT_BASE_IMAGE=$BUILDKIT_PROJECT+build
+        END
     ELSE
         ARG BUILDKIT_BASE_IMAGE=github.com/earthly/buildkit:40cd44f81962f99200f10df3a79d044f9df0aefe+build
     END
--- a/buildkitd/Earthfile
+++ b/buildkitd/Earthfile
@@ -12,7 +12,7 @@ buildkitd:
             ARG BUILDKIT_BASE_IMAGE=$BUILDKIT_PROJECT+build
         END
     ELSE
-        ARG BUILDKIT_BASE_IMAGE=github.com/earthly/buildkit:343404c5df1b9a13744c51d2ca951fd3175c73e2+build
+        ARG BUILDKIT_BASE_IMAGE=github.com/earthly/buildkit:af796f0cb8ff6cf53fdebd543ff1e7cd173c00a1+build
     END
     ARG EARTHLY_TARGET_TAG_DOCKER
     ARG TAG="dev-$EARTHLY_TARGET_TAG_DOCKER"
--- a/buildkitd/Earthfile
+++ b/buildkitd/Earthfile
@@ -12,7 +12,7 @@ buildkitd:
             ARG BUILDKIT_BASE_IMAGE=$BUILDKIT_PROJECT+build
         END
     ELSE
-        ARG BUILDKIT_BASE_IMAGE=github.com/earthly/buildkit:40cd44f81962f99200f10df3a79d044f9df0aefe+build
+        ARG BUILDKIT_BASE_IMAGE=github.com/earthly/buildkit:343404c5df1b9a13744c51d2ca951fd3175c73e2+build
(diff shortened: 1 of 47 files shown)


cmd/earthly/subcmd/build_cmd.go
```diff
@@ -579,14 +583,48 @@ func (a *Build) ActionBuildImp(cliCtx *cli.Context, flagArgs, nonFlagArgs []stri
...
+			err := regProxy.Serve(ln)
+			if err != nil && !errors.Is(err, http.ErrServerClosed) {
+				a.cli.Console().Warnf("Failed to serve registry proxy: %v", err)
+			}
+		}()
+
+		defer func() {
+			ctx := context.Background()
+			err := regProxy.Shutdown(ctx)
+			if err != nil {
```
vladaionescu, line 622: Maybe don't warn if the err is a cancellation.

vladaionescu (commented):
Looks good overall! Left some small comments below...

It would be good if we can add a test for this. We should be able to reuse [this script](https://github.com/earthly/earthly/blob/main/.github/workflows/reusable-export-test.yml) entirely. This is called from this [CI yaml](https://github.com/earthly/earthly/blob/main/.github/workflows/reusable-export-test.yml). I guess all we have to do is create a new CI job that runs this existing test with the right flag turned on.

cmd/earthly/subcmd/build_cmd.go
```diff
@@ -579,14 +583,48 @@ func (a *Build) ActionBuildImp(cliCtx *cli.Context, flagArgs, nonFlagArgs []stri
...
+
+		go func() {
+			err := regProxy.Serve(ln)
+			if err != nil && !errors.Is(err, http.ErrServerClosed) {
+				a.cli.Console().Warnf("Failed to serve registry proxy: %v", err)
+			}
+		}()
+
+		defer func() {
+			ctx := context.Background()
```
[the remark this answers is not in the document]
mikejholly: Thanks, I didn't notice that var.

features/features.go
```diff
@@ -60,6 +60,7 @@ type Features struct {
 	EarthlyCIRunnerArg         bool `long:"earthly-ci-runner-arg" description:"includes EARTHLY_CI_RUNNER ARG"`
 	UseDockerIgnore            bool `long:"use-docker-ignore" description:"fallback to .dockerignore incase .earthlyignore or .earthignore do not exist in a local \"FROM DOCKERFILE\" target"`
 	PassArgs                   bool `long:"pass-args" description:"Allow the use of the --pass-arg flag in FROM, BUILD, COPY, WITH DOCKER, and DO commands"`
+	UseRemoteRegistry          bool `long:"use-remote-registry" description:"Make use of the embedded BuildKit Docker registry for image pulls (rather than Tar transfer)"`
```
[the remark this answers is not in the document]
mikejholly: No, I added it initially, but it seems like a flag may be a better fit as the features aren't really available until we parse an Earthfile. Was looking for some feedback on the most appropriate strategy here.

cmd/earthly/subcmd/build_cmd.go
mikejholly: Good call. Thanks.

cmd/earthly/flag/global.go
```diff
@@ -510,5 +509,12 @@ func (global *Global) RootFlags(installName string, bkImage string) []cli.Flag {
 			Usage:       "A remote docker image tag use as explicit cache",
 			Destination: &global.RemoteCache,
 		},
+		&cli.BoolFlag{
+			Name:        "use-remote-registry",
+			EnvVars:     []string{"EARTHLY_USE_REMOTE_REGISTRY"},
+			Usage:       "Use remote registry with Docker pull",
+			Destination: &global.UseRemoteRegistry,
+			Hidden:      true, // Internal.
+		},
```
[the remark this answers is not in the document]
mikejholly: Good idea :+1:

buildkitd/Earthfile
```diff
@@ -12,7 +12,7 @@ buildkitd:
             ARG BUILDKIT_BASE_IMAGE=$BUILDKIT_PROJECT+build
         END
     ELSE
-        ARG BUILDKIT_BASE_IMAGE=github.com/earthly/buildkit:8df7c52e603ee6db6a471fcb7f6a85a407f611f4+build
+        ARG BUILDKIT_BASE_IMAGE=github.com/earthly/buildkit:40cd44f81962f99200f10df3a79d044f9df0aefe+build
```
alexcb, line 15: don't forget to change this to the rebased commit :)

features/features.go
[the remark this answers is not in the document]
vladaionescu: I think that the CLI flag makes sense for this. Usually, we use the features.go flags for syntax changes.

vladaionescu (approved):
👏 👏

buildkitd/Earthfile
alexcb: this still needs updating, earthly-main under buildkit is now at `af796f0cb8ff6cf53fdebd543ff1e7cd173c00a1`

alexcb (approved):
looks good, only nit is related to pining to the pre-merged commit of buildkit.

buildkitd/Earthfile
mikejholly: Thanks. Updated to `af796f0cb8ff6cf53fdebd543ff1e7cd173c00a1`

alexcb (approved):
thanks!

PR #3303, opened 2023-09-27
earthly debugger - avoid errors on exit

idodod:
Partially addresses #3200 (example 5)

Upon exiting the interactive earthly debugger, the user would often see errors such as:
``` 
+my-target | time="2023-02-08T18:44:39Z" level=error msg="failed to read from ptmx: read /dev/ptmx: input/output error"
```
Which can be confusing.
This PR prevent these kind of errors from occurring by properly closing the terminal/related io readers/writers.

In addition, moving some message printing as logs to be printed by the console.
Also updating related documentation of interactive mode with an up to date output.

### DIFF
--- a/cmd/debugger/main.go
+++ b/cmd/debugger/main.go
@@ -10,6 +10,8 @@ import (
 	"net"
 	"os"
 	"os/exec"
+	"sync"
+	"syscall"
 	"time"
 
 	"github.com/earthly/earthly/conslogging"
@@ -19,7 +21,7 @@ import (
 	"github.com/alessio/shellescape"
 	"github.com/creack/pty"
 	"github.com/fatih/color"
-	multierror "github.com/hashicorp/go-multierror"
+	"github.com/hashicorp/go-multierror"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
@@ -100,7 +102,7 @@ func sendFile(ctx context.Context, sockAddr, src, dst string) error {
 	defer func() {
 		err := conn.Close()
 		if err != nil {
-			log.Error(errors.Wrap(err, "error closing"))
+			log.Error(errors.Wrap(err, "earthly debugger: error closing"))
 		}
 	}()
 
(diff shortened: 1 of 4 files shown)


cmd/debugger/main.go
```diff
@@ -100,7 +102,7 @@ func sendFile(ctx context.Context, sockAddr, src, dst string) error {
 	defer func() {
 		err := conn.Close()
 		if err != nil {
-			log.Error(errors.Wrap(err, "error closing"))
+			log.Error(errors.Wrap(err, "earthly debugger: error closing"))
```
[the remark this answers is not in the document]
vladaionescu: Good idea

cmd/debugger/main.go
```diff
@@ -166,46 +168,57 @@ func interactiveMode(ctx context.Context, remoteConsoleAddr string, cmdBuilder f
 
 	ptmx, err := pty.Start(c)
 	if err != nil {
-		log.Error(errors.Wrap(err, "failed to start pty"))
+		console.VerbosePrintf("failed to start pty: %v\n", err)
```
vladaionescu, line 171: Should we add `earthly debugger: ` here too?

vladaionescu (commented):
Leaving the main review to @alexcb as he's more familiar with the debugger code.

alexcb:
I'm currently taking a deeper look.

> Upon exiting the interactive earthly debugger, the user would often see errors such as:

I always exit via ctrl-d, which never shows this error. I just noticed that if you type `exit 0` it reproduces it each time.

I'm going to checkout the branch and give it some more in-depth testing.

idodod:
> I'm currently taking a deeper look.
> 
> > Upon exiting the interactive earthly debugger, the user would often see errors such as:
> 
> I always exit via ctrl-d, which never shows this error. I just noticed that if you type `exit 0` it reproduces it each time.
> 
> I'm going to checkout the branch and give it some more in-depth testing.

For me, both CTRL-D and "exit" would result in the error, but I agree it was more common with the latter. This fix handles both cases.

cmd/debugger/main.go
vladaionescu: Oh yeah, good point.

alexcb (commented):
Ido and I paired on this via a huddle, I came up with df14b0779eabc92481915e8a87d88065253832ba as an alternative solution which Ido will apply the logging changes to.

idodod:
replaced by https://github.com/earthly/earthly/pull/3310

PR #3307, opened 2023-09-28
with this change, GH actions will only be able to run once for each branch/pr

idodod:
A common case is when a new commit is pushed to the branch - 
With this added configuration any running jobs for the previous commit will be stopped and the jobs will start again for the newest commit (same as they do today).

Note, the configuration is added only in workflow files that are triggered by a pull request.

[docs reference](https://docs.github.com/en/actions/using-jobs/using-concurrency#example-using-concurrency-to-cancel-any-in-progress-job-or-run)


alexcb (approved):
awesome!

PR #3310, opened 2023-09-28
interactive debugger ignore errors if command exits correctly

idodod:
Partially addresses #3200  (example 5)
 - Avoid printing log errors that occur due to debugger exiting
 - Make other log messages clearer that they are coming from the debugger, use console instead of Logger

Credit to @alexcb for solution.

### DIFF
--- a/cmd/debugger/main.go
+++ b/cmd/debugger/main.go
@@ -179,11 +179,11 @@ func interactiveMode(ctx context.Context, remoteConsoleAddr string, cmdBuilder f
 		return waitErrSet
 	}
 
-	logErrorIfNonCleanExit := func(console conslogging.ConsoleLogger, err error) {
+	logErrorIfNonCleanExit := func(err error) {
 		if hasCommandFinished() {
 			return
 		}
-		console.VerboseWarnf(err.Error())
+		conslogger.Warnf("%v\n", errors.Wrap(err, "failed to start pty"))
 	}
 
 	ptmx, err := pty.Start(c)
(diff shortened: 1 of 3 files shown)


cmd/debugger/main.go
```diff
@@ -349,7 +376,7 @@ func main() {
 
 			err := os.Setenv("TERM", debuggerSettings.Term)
 			if err != nil {
-				conslogger.Warnf("Failed to set term: %v", err)
+				conslogger.VerboseWarnf("Failed to set term: %v\n", err)
```
alexcb, line 379: Was this warning commonly occurring? I would think having `Setenv` fail should be extremely rare, and should always show up as a warning -- if the TERM can't be set, feature like resizing the terminal won't work.

cmd/debugger/main.go
```diff
@@ -320,7 +347,7 @@ func main() {
 		os.Exit(exitCode)
 	}
 
-	log.With("command", args).With("version", Version).Debug("running command")
+	log.With("command", args).With("version", Version).Debug("earthly-debugger: running command")
```
alexcb, line 350: switch this to consologger to be consistent?

cmd/debugger/main.go
```diff
@@ -297,17 +324,17 @@ func main() {
 
 		err := os.Setenv("TERM", debuggerSettings.Term)
 		if err != nil {
-			conslogger.Warnf("Failed to set term: %v", err)
+			conslogger.VerboseWarnf("Failed to set term: %v\n", err)
```
alexcb, line 327: similar to my comment below (apparently I started reviewing from the bottom up)...

> Was this warning commonly occurring? I would think having Setenv fail should be extremely rare, and should always show up as a warning -- if the TERM can't be set, feature like resizing the terminal won't work.

cmd/debugger/main.go
idodod: yeah, I thought it will be fine since it's debug level, but I suppose we can verbose print this one

cmd/debugger/main.go
alexcb: it's almost as if some of this code was copied and pasted in a previous commit :see_no_evil:

cmd/debugger/main.go
idodod: sounds good.

cmd/debugger/main.go
```diff
@@ -164,9 +165,30 @@ func interactiveMode(ctx context.Context, remoteConsoleAddr string, cmdBuilder f
...
+		if hasCommandFinished() {
+			return
+		}
+		console.VerboseWarnf(err.Error())
+	}
+
 	ptmx, err := pty.Start(c)
 	if err != nil {
-		log.Error(errors.Wrap(err, "failed to start pty"))
+		conslogger.VerboseWarnf("failed to start pty\n")
```
alexcb, line 191: if this fails, everything fails, it should stay as an error.

cmd/debugger/main.go
```diff
@@ -164,9 +165,30 @@ func interactiveMode(ctx context.Context, remoteConsoleAddr string, cmdBuilder f
...
+		waitErrMu.Lock()
+		defer waitErrMu.Unlock()
+		return waitErrSet
+	}
+
+	logErrorIfNonCleanExit := func(console conslogging.ConsoleLogger, err error) {
+		if hasCommandFinished() {
+			return
+		}
+		console.VerboseWarnf(err.Error())
```
alexcb, line 186:
```suggestion
		console.Warnf(err.Error())
```

alexcb (commented):
looking good (thanks for taking over my half-implemented idea); I like the change to the conslogger but feel we should be much more cautious about silencing actual errors/warnings.

Any errors that occur _before_ the wrapped command finishes are actual errors and need to be displayed to the user -- otherwise if an error does occur the user will experience a blank terminal that appears to have deadlocked (in the case where the wrapped command hasn't finished), or just silently exits which doesn't offer much insights as to what went wrong.

cmd/debugger/main.go
idodod: i agree, my thinking was that showing this as a warning (always) might confuse users with the build error they are experiencing and diagnosing a problem to use the interactive debugger can be further investigated in verbose mode.

on the other end we do have now `earthly debugger` as the prefix, so I don't feel strongly about the above.

alexcb (approved):

PR #3313, opened 2023-09-28
Add a new GH action and earthly target to merge main bracnh to docs-0.7

idodod:
As part of our release process, we merge the main branch to docs-0.7.
This PR attempts to automate this step - 
Upon a new GH release (or tag to be exact), a new GH action job will run an earthly target which git merges the branches.

Currently this is only in dry mode  - notice the actual git merge & push commands are flagged with `RUN --push`
and --push is not used in the earthly command


Earthfile
```diff
@@ -818,3 +818,25 @@ npm-update-all:
...
+
+# merge-main-to-docs merges the main branch into docs-0.7
+merge-main-to-docs:
+    FROM alpine/git
+    # Ok, if you have no choice
+    ARG git_url="git@github.com:earthly/earthly"
+    GIT CLONE "$git_url" earthly
+    RUN --mount=type=secret,id=littleredcorvette-id_rsa,mode=0400,target=/root/.ssh/id_rsa \
+          ssh-keyscan github.com >> ~/.ssh/known_hosts && \
+        rm -rf earthly &&\
```
alexcb, line 830: what's the reason for `rm -rf earthly`, can we just remove this entirely as well as remove the `GIT CLONE ...` command above?

Earthfile
```diff
@@ -818,3 +818,25 @@ npm-update-all:
...
         SAVE ARTIFACT --if-exists $nodepath/package-lock.json AS LOCAL $nodepath/package-lock.json
     END
+
+# merge-main-to-docs merges the main branch into docs-0.7
+merge-main-to-docs:
+    FROM alpine/git
+    # Ok, if you have no choice
+    ARG git_url="git@github.com:earthly/earthly"
+    GIT CLONE "$git_url" earthly
+    RUN --mount=type=secret,id=littleredcorvette-id_rsa,mode=0400,target=/root/.ssh/id_rsa \
```
alexcb, line 828: add a `--no-cache` option?

Earthfile
```diff
@@ -818,3 +818,25 @@ npm-update-all:
...
+          ssh-keyscan github.com >> ~/.ssh/known_hosts && \
+        rm -rf earthly &&\
+        git clone "$git_url" earthly
+    WORKDIR earthly
+    RUN git config --global user.name "littleredcorvette" && \
```
alexcb, line 833: move this above the `RUN --no-cache .... git clone ....` command to improve caching.

Earthfile
idodod: I actually meant to ask you about this, I was following our guidelines, wasn't sure if and how this is beneficial:
https://docs.earthly.dev/best-practices#git-clone-vs-run-git-clone

Earthfile
```diff
@@ -818,3 +818,25 @@ npm-update-all:
...
+        git config --global user.email "littleredcorvette@users.noreply.github.com"
+    RUN git checkout docs-0.7 && \
+    git merge --no-commit --no-ff main && \
```
alexcb, line 836: what's the reason for the `--no-ff`, I think a fast-forward merge is favoured if possible and we should just let git do whatever default it was (i.e. prefer a fast-forward, but fallback to a 3way merge if needed)

Earthfile
```diff
@@ -818,3 +818,25 @@ npm-update-all:
...
+    git status && \
+    git merge --abort
```
alexcb, line 838: ah ha..... is this for a dry-run sort of output? it was a bit confusing without a comment.

honestly maybe we can just remove it and use the single command you have below without a `git push` at the end for the initial trial?

If there's any conflicts it'll fail right?

idodod: yeah, I think if git merge has any conflicts it will return an error.
sure, I can merge the command into 

perhaps I leave the dry run mode we have for now just so we can see `git status` and when ready I can remove it?

Earthfile
idodod: not needed since the git_hash arg should invalidate the cache if necessary

Earthfile
idodod: discussed offline - going to keep for now and can remove when ready to get out of dry run mode

Earthfile
```diff
@@ -818,3 +818,28 @@ npm-update-all:
...
+    ARG git_url="git@github.com:$git_repo"
+    ARG to_branch="docs-0.7"
+    ARG from_branch="main"
+    RUN git config --global user.name "littleredcorvette" && \
+        git config --global user.email "littleredcorvette@users.noreply.github.com"
+    GIT CLONE "$git_url" earthly
+    WORKDIR earthly
+    ARG git_hash=$(git rev-parse HEAD)
+    RUN --mount=type=secret,id=littleredcorvette-id_rsa,mode=0400,target=/root/.ssh/id_rsa \
+          ssh-keyscan github.com >> ~/.ssh/known_hosts && \
```
alexcb, line 835: nit: store their actual keys instead? this is the equivalent of disabling host key checking.

alexcb (approved):

Earthfile
```diff
@@ -818,3 +818,28 @@ npm-update-all:
...
     END
+
+# merge-main-to-docs merges the main branch into docs-0.7
+merge-main-to-docs:
+    FROM alpine/git
+    ARG git_repo="earthly/earthly"
+    ARG git_url="git@github.com:$git_repo"
+    ARG to_branch="docs-0.7"
+    ARG from_branch="main"
+    DO github.com/earthly/lib/ssh:ido-add-known_hosts_udc+ADD_KNOWN_HOSTS --target_file=~/.ssh/known_hosts
```
idodod, line 829: I'll change the branch name (`ido-add-known_hosts_udc`) to a (new) tag after merging the branch and publishing a new release in earthly/lib

idelvall: using the commit hash would be even better, since tags can be changed (deleted and created over)

idelvall (approved):
LGTM once the UCD reference is updated

PR #3315, opened 2023-09-28
best practices docs: explain what ARG git_hash is doing

alexcb:


alexcb:
It took me quite some time to figure out what was happening here... I suspect some of our users might have a `WAT` moment surrounding this.

docs/best-practices/best-practices.md
```diff
@@ -231,6 +231,13 @@ WORKDIR my-proj
 RUN ls
 ```
 
+{% hint style='info' %}
+##### Note
+The final "if you have no choice" example contains an `ARG` shell-out, which will be set to the latest git sha; this is done so that if the git repository doesn't
+contain any new commits between multiple runs, the `ARG git_hash` value will stay constant, and will allow earthly to skip over the `RUN` command; where as a `RUN --no-cache`
```
vladaionescu, line 237:
```suggestion
contain any new commits between multiple runs, the `ARG git_hash` value will stay constant, and will allow earthly to skip over the `RUN` command; whereas a `RUN --no-cache`
```

vladaionescu (approved):
Minor typo

PR #3317, opened 2023-09-29
Registry proxy: Use lower-level TCP streaming

mikejholly:
Removes the HTTP-level code in favor of a lower-level TCP proxy approach. 

The tricky part here was that Docker (and most HTTP clients) leaves the TCP connection open. Hence, there's no way for the code to detect when to stop reading the initial request aside from a timeout. Apparently, the recommended approach is to set [`SetReadDeadline`](https://cs.opensource.google/go/go/+/refs/tags/go1.21.1:src/net/net.go;l=156) ahead of each read and appropriately handle any timeout errors. 

Overall, I think this is preferable to the original code as the HTTP code & header parsing is a bit brittle.

Works with: https://github.com/earthly/buildkit/pull/26

### DIFF
--- a/regproxy/proxy.go
+++ b/regproxy/proxy.go
@@ -2,7 +2,6 @@ package regproxy
 
 import (
 	"context"
-	"fmt"
 	"io"
 	"net"
 	"sync"
--- a/regproxy/proxy.go
+++ b/regproxy/proxy.go
@@ -2,9 +2,11 @@ package regproxy
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"net"
 	"sync"
+	"sync/atomic"
 
 	registry "github.com/moby/buildkit/api/services/registry"
 	"github.com/pkg/errors"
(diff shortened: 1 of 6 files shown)


regproxy/proxy.go
```diff
@@ -1,115 +1,143 @@
...
+}
+
+func copyWithDeadline(conn net.Conn, w io.Writer) (int64, error) {
+	var t int64
+	for {
+		err := conn.SetReadDeadline(time.Now().Add(readDeadline))
+		if err != nil {
+			return 0, err
 		}
+		buf := make([]byte, 1024)
```
alexcb, line 120: nit: 1kB is a bit small
```suggestion
		buf := make([]byte, 64*1024)
```

regproxy/proxy.go
```diff
@@ -1,115 +1,143 @@
...
+		n, err := conn.Read(buf)
+		if err != nil {
+			if errors.Is(err, io.EOF) || isNetTimeout(err) {
+				break
+			}
+			return 0, err
```
alexcb, line 126: nit: IO commands tend to return the number of successfully read/written bytes along with the error, to allow the caller to be more graceful in known how many bytes were processed before hand.

I'm aware the code that calls this function doesn't use it, but it future-proofs it in case we decide it's needed down the road.

```suggestion
			return t, err
```

regproxy/proxy.go
```diff
@@ -1,115 +1,143 @@
...
+		}
+		buf = buf[0:n]
+		n, err = w.Write(buf)
+		if err != nil {
+			return 0, err
```
alexcb, line 131:
```suggestion
			return t+n, err
```

regproxy/proxy.go
mikejholly: That's fine. However, there's not much point to streaming in this case as these requests are unlikely to be > 64K. This wouldn't be much different from using `io.ReadAll`.  I also noticed that `io.Copy` doesn't buffer at all by default.

PR #3320, opened 2023-09-29
Conditionally run GH actions jobs only if relevant files changed

idodod:
Adding a new reusable job that will detect whether "essential" files were changed.
Then other jobs will run conditionally only is essentials files (e.g. *.go files) are changed as part of a PR.

Note that according to documentation, skipped jobs that are used as required checks in Github do not block PRs
(see note [here](https://docs.github.com/en/actions/using-jobs/using-conditions-to-control-job-execution#overview)).

Since all the tests are dependent on `build-earthly` job, it was sufficient to conditionally execute the latter.
If it is skipped, all dependent jobs will be skipped as well. 

I tested these changes on my fork of earthly

### DIFF
--- a/.github/workflows/ci-docker-ubuntu.yml
+++ b/.github/workflows/ci-docker-ubuntu.yml
@@ -12,7 +12,18 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  # this job will output a boolean value to check whether files that require these tests to run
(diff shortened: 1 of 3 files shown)


mikejholly (approved):
Good idea :+1:

alexcb (approved):

PR #3334, opened 2023-10-03
changelog-parser: use a common base

alexcb:
The changelog-parser was changed in
7e7bd7c92be740354b1bb4b266f8c44d477351e4 to make use of the packaging library; however it was only ever installed in the base linter, and not the ./release+release-notes target, which caused that target to break.

This refactors the targets to use the same common base (which includes the packaging library).


mikejholly (approved):

alexcb:
force-merged since the lint-docs target has passed (which is the only test that uses this target)

PR #3340, opened 2023-10-04
Change how PRs are skipped so that required checks won't get stuck in pending

idodod:
The current configuration might causes jobs that are marked as required checks to get stuck in pending.
This PR aims to fix it.

See the readme added in this PR for more information.

Also fixing:
* workflows not running properly on push to main
* python tutorial tests

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -1,0 +1,51 @@
+# Github Actions
+
+Documentation for this repo Github actions configuration
+
+## Skipping PR Workflows
+### Motivation
+Some PR workflows (workflows triggered by PRs) in this repo might take substantial time to complete,  
+and it is not always necessary to run them all, especially in cases where the affected files are documents or any other file that don't affect our workflows or tests.  
+In addition, some of our tests use satellites and limiting the number of (concurrent) builds might help with their performance.
+
+### Github Builtin Support
+Github supports defining filters on workflows so that they won't get triggered, and there is a builtin filter that would do so according to affected files - [paths-ignore](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore).  
+The shortcoming of this filter is that when the workflow is marked as a `required check` and does not get triggered because of this filter,  
+the workflow will be stuck in `pending` and won't allow the PR to be merged.  
+Thus, using this filter is a good approach when handling workflows that are not required checks, but it's problematic otherwise.
+
+### Solution for required workflows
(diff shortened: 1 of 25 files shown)


mikejholly (approved):
Sounds reasonable.

But, I'm wondering if improving Earthly & test performance (caching, etc.) are the "correct" areas to improve. Maybe a bigger conversation for later.

idodod:
> Sounds reasonable.
> 
> But, I'm wondering if improving Earthly & test performance (caching, etc.) are the "correct" areas to improve. Maybe a bigger conversation for later.

Yeah, I would think that we would want to do both. This doesn't "hide" a slowness problem as most PRs will still take time, it should just allow us to merge PRs that only have to do with documentation or unrelated configuration faster.

.github/workflows/README.md
```diff
@@ -0,0 +1,51 @@
...
+So, in order to work around the above-mentioned issue, we need to avoid
+skipping the parent job. Instead, we evaluate the boolean as before, and pass it as  
+an input argument to the child job. As a matter of standardization, we call this input arg `CONTINUE`.
+This argument is then used to conditionally run the _child_ job instead of the _parent_ job,  
+and so if the child job is skipped, even if it is marked as a required check, it won't get stuck in `pending`.
+
+Lastly, because most jobs are dependent on (`needs` directive) `build-earthly` job, if the latter is skipped, by default - so will the dependent jobs.  
+As described before, that is something we need to avoid. To change that default configuration, we change
+the default condition of each dependent job to `$ {{ !failure() }}` which will allow the dependent jobs to run whether the dependency job was successful or if it was skipped.
+
```
alexcb, line 47: nit: remove trailing newlines (while still remaining posix compatible)

.github/workflows/reusable-wait-block-override.yml
```diff
@@ -15,10 +15,14 @@ on:
       RUNS_ON:
         required: true
         type: string
-
+      CONTINUE:
```
alexcb, line 18: can we pick a different variable name for this? `CONTINUE` makes me think we're inside a loop and will be re-running the loop.

what if instead we called this `PERFORM_TEST` instead? Or alternatively `SKIP_TEST`, which would then require us to also adjust the values we pass in.

alexcb (commented):
nice work! just one suggestion.

.github/workflows/reusable-wait-block-override.yml
idodod: will do!

alexcb (approved):
awesome!

PR #3341, opened 2023-10-04
Add builtin support for git references

### LINKED ISSUE #2735: EARTHLY_GIT_(REF|BRANCH) built in ARG
Would it be possible to include an `EARTHLY_GIT_(REF|BRANCH)` built-in ARG similar to to `EARTHLY_GIT_HASH`?

miles170:
Closes #2735.

The new ARG `EARTHLY_GIT_REFS` will contain the references to the current git commit, this ARG must be enabled with the `VERSION --git-refs` feature flag.


alexcb:
Thanks for the contribution, it looks promising!

We will need to look into why our github action tests aren't correctly getting our docker login credentials -- for some reason they aren't being used since this PR is from outside of our repo.

alexcb:
I opened #3346 (and in the process accidentally pushed a merge commit to your repo too) -- that PR was run from within earthly -- which has access to our secrets. unfortunately the test failed.

alexcb (changes requested):
I don't think running `git show-ref --heads` is appropriate in this case, as it includes references to other commits. Instead I think `git for-each-ref --contains HEAD` is what we're after, it outputs something like

```
$ git for-each-ref --contains HEAD
f590cea6f2d3f854b8e6943b1cff1b81e95b84d0 commit	refs/heads/issue-2735-git-ref
f590cea6f2d3f854b8e6943b1cff1b81e95b84d0 commit	refs/remotes/origin/issue-2735-git-ref
```

which will need to be parsed appropriately to only return the branch names (e.g. `issue-2735-git-ref`)

miles170:
> I opened #3346 (and in the process accidentally pushed a merge commit to your repo too) -- that PR was run from within earthly -- which has access to our secrets. unfortunately the test failed.

I have already updated the code to retrieve `EARTHLY_GIT_REFS` using `git for-each-ref --contains HEAD`. Furthermore, I have tested it locally and confirmed its success.

alexcb:
Thanks for the quick reply, I just pushed a fresh copy to #3346 along with a small fix, but it appears that it's failing on the empty-git-test,

maybe you can replicate it locally with:

```
./earthly +for-linux && ./build/linux/amd64/earthly -P ./tests+empty-git-test
```

alexcb (approved):

alexcb:
thanks for the contribution (and thanks for putting up with our github actions limitations surrounding secrets)

I force-merged this since it was tested via #3346 and also manually by me.

PR #3343, opened 2023-10-04
Implement registry capabilities check

mikejholly:
This PR moves the registry proxy start-up code into `Builder` and gates the feature using the new cap added here: https://github.com/earthly/buildkit/pull/27

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -80,6 +80,7 @@ type Opt struct {
 	DisableNoOutputUpdates                bool
 	ParallelConversion                    bool
 	Parallelism                           semutil.Semaphore
+	UseRemoteRegistry                     bool
 	LocalRegistryAddr                     string
 	FeatureFlagOverrides                  string
 	ContainerFrontend                     containerutil.ContainerFrontend
@@ -161,13 +162,13 @@ func (b *Builder) BuildTarget(ctx context.Context, target domain.Target, opt Bui
 
 func (b *Builder) startRegistryProxy(ctx context.Context, caps apicaps.CapSet) (func(), bool) {
 
+	if !b.opt.UseRemoteRegistry {
+		b.opt.Console.VerbosePrintf("Registry proxy not enabled")
+		return nil, false
+	}
+
 	if err := caps.Supports(pb.CapEarthlyRegistryProxy); err != nil {
-		var capErr *apicaps.CapError
-		if ok := errors.As(err, &capErr); ok && capErr.State != nil && !capErr.State.Enabled {
-			b.opt.Console.VerboseWarnf("Registry proxy is not supported by BuildKit: %s", err)
-		} else {
-			b.opt.Console.Warnf("Failed to check for proxy support: %s", err)
-		}
+		b.opt.Console.Warnf("Registry proxy is not supported by BuildKit: %s", err)
 		return nil, false
 	}
 
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"net"
 	"os"
 	"path/filepath"
 	"strings"
@@ -18,6 +19,7 @@ import (
 	"github.com/earthly/earthly/logbus"
 	"github.com/earthly/earthly/logbus/solvermon"
 	"github.com/earthly/earthly/outmon"
+	"github.com/earthly/earthly/regproxy"
 	"github.com/earthly/earthly/states"
 	"github.com/earthly/earthly/util/containerutil"
 	"github.com/earthly/earthly/util/dockerutil"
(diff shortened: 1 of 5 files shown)


builder/builder.go
```diff
@@ -174,12 +219,17 @@ func (b *Builder) convertAndBuild(ctx context.Context, target domain.Target, opt
...
 	var mts *states.MultiTarget
-	bf := func(childCtx context.Context, gwClient gwclient.Client) (*gwclient.Result, error) {
+	buildFn := func(childCtx context.Context, gwClient gwclient.Client) (*gwclient.Result, error) {
 		if opt.EnableGatewayClientLogging {
 			gwClient = gwclientlogger.New(gwClient)
 		}
 		var err error
 		caps := gwClient.BuildOpts().LLBCaps
+
+		if stopProxy, ok := b.startRegistryProxy(ctx, caps); ok {
```
mikejholly, line 229: This appears to be executed once.

builder/builder.go
```diff
@@ -174,12 +219,17 @@ func (b *Builder) convertAndBuild(ctx context.Context, target domain.Target, opt
 		dirIndex   = 0
 	)
 	var mts *states.MultiTarget
-	bf := func(childCtx context.Context, gwClient gwclient.Client) (*gwclient.Result, error) {
+	buildFn := func(childCtx context.Context, gwClient gwclient.Client) (*gwclient.Result, error) {
```
mikejholly, line 222: I wanted to clarify this a bit as `bf` is a little cryptic for people new to the repo.

alexcb: seems reasonable; can we call it `buildFunc` to be consistent with the name that buildkit gives it:

```
func (c *Client) Build(ctx context.Context, opt SolveOpt, product string, buildFunc gateway.BuildFunc, statusChan chan *SolveStatus) (*SolveResponse, error) {
```

builder/builder.go
```diff
@@ -155,6 +159,47 @@ func (b *Builder) BuildTarget(ctx context.Context, target domain.Target, opt Bui
 	return mts, nil
 }
 
+func (b *Builder) startRegistryProxy(ctx context.Context, caps apicaps.CapSet) (func(), bool) {
+
+	if err := caps.Supports(pb.CapEarthlyRegistryProxy); err != nil {
+		b.opt.Console.VerbosePrintf("Registry proxying not supported by BuildKit: %s", err)
```
idodod, line 165: nit: you might want to use VerboseWarnf
```suggestion
		b.opt.Console.VerboseWarnf("Registry proxying not supported by BuildKit: %v", err)
```

alexcb (commented):
looks good, just a couple minor things.

builder/builder.go
alexcb: I ran a quick test and confirmed this too.

builder/builder.go
```diff
@@ -155,6 +159,47 @@ func (b *Builder) BuildTarget(ctx context.Context, target domain.Target, opt Bui
 	return mts, nil
 }
 
+func (b *Builder) startRegistryProxy(ctx context.Context, caps apicaps.CapSet) (func(), bool) {
+
+	if err := caps.Supports(pb.CapEarthlyRegistryProxy); err != nil {
```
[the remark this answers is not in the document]
mikejholly: It appears that `Supports` will only ever return a `CapError`. https://github.com/earthly/buildkit/blob/earthly-main/util/apicaps/caps.go#L107-L123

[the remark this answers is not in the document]
mikejholly: I don't mind checking, but it doesn't seem especially necessary as `Supports` always returns the same error type. 

```golang
	if err := caps.Supports(pb.CapEarthlyRegistryProxy); err != nil {
		var capErr *apicaps.CapError
		if ok := errors.As(err, &capErr); ok && capErr.State != nil && !capErr.State.Enabled {
			b.opt.Console.VerboseWarnf("Registry proxy is not supported by BuildKit: %s", err)
		} else { // Not currently reachable. 
			b.opt.Console.Warnf("Failed to check for proxy support: %s", err)
		}
		return nil, false
	}
```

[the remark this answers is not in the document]
alexcb: ah interesting! I'm not sure why we ended up checking for other types..... seems fine then.

builder/builder.go
alexcb: I think I read this too quickly initially, `VerboseWarnf` is still flagged via verbose, (I read it simply as warnf). Either seems fine in my opinion.

builder/builder.go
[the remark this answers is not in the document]
mikejholly: I realized that I missed the flag check. Added back. 

I think a missing cap after that point probably warrants a `Warnf`.

alexcb (approved):

builder/builder.go
```diff
@@ -155,6 +160,52 @@ func (b *Builder) BuildTarget(ctx context.Context, target domain.Target, opt Bui
...
+			}
+		}
+		doneCh <- struct{}{}
+	}()
+
+	b.opt.LocalRegistryAddr = addr
+
+	return func() {
+		p.Close()
+		<-doneCh
```
vladaionescu, line 205: Should this listen to `ctx.Done()` too while waiting for `<-doneCh`?

mikejholly: That's probably worthwhile. `p.Close` will kill the server, but I don't think it's immediate.

vladaionescu (approved):

PR #3351, opened 2023-10-06
Move Logstream config code into separate func

mikejholly:
The `ActionBuildImp` method is quite long and difficult (for me) to parse. This piece of code seems like a good candidate to split out into a separate function. I also prefer to use guards over multiple levels of nested conditional.

### DIFF
--- a/cmd/earthly/subcmd/build_cmd.go
+++ b/cmd/earthly/subcmd/build_cmd.go
@@ -254,47 +254,9 @@ func (a *Build) ActionBuildImp(cliCtx *cli.Context, flagArgs, nonFlagArgs []stri
 		return err
 	}
 
-	// Default upload logs, unless explicitly configured
-	doLogstreamUpload := false
-	var logstreamURL string
-	if !a.cli.Cfg().Global.DisableLogSharing {
-		if cloudClient.IsLoggedIn(cliCtx.Context) {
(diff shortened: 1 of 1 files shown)


vladaionescu (approved):
This is pure refactoring, right? No functional changes, correct?

mikejholly:
> This is pure refactoring, right? No functional changes, correct?

Yep!

cmd/earthly/subcmd/build_cmd.go
```diff
@@ -254,47 +254,9 @@ func (a *Build) ActionBuildImp(cliCtx *cli.Context, flagArgs, nonFlagArgs []stri
...
-				a.cli.Console().Printf(
-					"🛰️ Reuse cache between CI runs with Earthly Satellites! " +
-						"2-20X faster than without cache. Generous free tier " +
-						"https://cloud.earthly.dev\n")
-			}()
-		}
-	}
+	// Determine if Logstream is enabled and create log sharing link in either case.
+	logstreamURL, doLogstreamUpload, printLinkFn := a.logShareLink(cliCtx.Context, cloudClient, target, cleanCollection)
+	defer printLinkFn() // Output log sharing link after build.
```
idodod, line 259: I really like this pattern of returning the func to call with defer.

idodod (approved):

PR #3355, opened 2023-10-06
Split up wait block CI jobs

mikejholly:
This PR splits up the `WAIT` override tests into groups similar to the default set, à la `docker-tests-no-qemu-slow` and friends. The existing combined test was taking about [42m](https://github.com/earthly/earthly/actions/runs/6435089157/job/17475879502), whereas the max of this new group was [19m](https://github.com/earthly/earthly/actions/runs/6435775702/job/17478524488?pr=3355) on the most recent run on this PR. Hopefully this helps with our workflows.


.github/workflows/reusable-wait-block-target.yml
```diff
@@ -19,6 +19,9 @@ on:
         required: false
         type: boolean
         default: false
+      TARGET_NAME:
```
idodod, line 22: we might need to give the job a unique name according to input.TARGET_NAME so it's easy to distinguish in the PR checks.
[Here](https://github.com/earthly/earthly/blob/f8b0db17a9a789ee2b22edbdd38684ac0622e7b1/.github/workflows/reusable-example.yml#L42) is an example where this is done.

mikejholly: Good idea. How about something like this? https://github.com/earthly/earthly/pull/3355/commits/4d04a40e5a7b65f7b52b84ceb0a0a4042580fefa

idodod (approved):

PR #3358, opened 2023-10-06
Auto-skip: Basic support for WAIT

mikejholly:
This PR adds AutoSkip support for `WAIT...END` blocks.

### DIFF
--- a/inputgraph/inputgraph.go
+++ b/inputgraph/inputgraph.go
@@ -195,8 +195,7 @@ func (l *loader) handleFor(ctx context.Context, forStmt spec.ForStatement) error
 }
 
 func (l *loader) handleWait(ctx context.Context, waitStmt spec.WaitStatement) error {
-	l.hasher.HashString("wait")
-	l.hasher.HashString(strings.Join(waitStmt.Args, " "))
+	l.hasher.HashStatement("WAIT", waitStmt.Args)
 	for _, stmt := range waitStmt.Body {
 		if err := l.handleStatement(ctx, stmt); err != nil {
 			return err
--- a/inputgraph/inputgraph.go
+++ b/inputgraph/inputgraph.go
@@ -137,7 +137,7 @@ func (l *loader) handlePipeline(ctx context.Context, cmd spec.Command) error {
 }
 
 func (l *loader) handleCommand(ctx context.Context, cmd spec.Command) error {
-	l.hasher.HashJSONMarshalled(cmd)
+	l.hashCommand(cmd)
 	switch cmd.Name {
 	case command.From:
 		return l.handleFrom(ctx, cmd)
@@ -164,7 +164,7 @@ func (l *loader) handleWith(ctx context.Context, with spec.WithStatement) error 
 }
 
 func (l *loader) handleWithDocker(ctx context.Context, cmd spec.Command) error {
-	l.hasher.HashJSONMarshalled(cmd) // special case since handleWithDocker doesn't get called from handleCommand
+	l.hashCommand(cmd) // special case since handleWithDocker doesn't get called from handleCommand
 	opts := commandflag.WithDockerOpts{}
 	_, err := parseArgs("WITH DOCKER", &opts, getArgsCopy(cmd))
 	if err != nil {
@@ -194,10 +194,11 @@ func (l *loader) handleFor(ctx context.Context, forStmt spec.ForStatement) error
 	return errors.Wrap(ErrUnableToDetermineHash, "for not supported")
 }
 
-func (l *loader) hashWaitStatement(waitStmt spec.WaitStatement) {
+func (l *loader) hashWaitStatement(w spec.WaitStatement) {
+	w.SourceLocation = nil
 	l.hasher.HashString("WAIT")
-	l.hasher.HashInt(len(waitStmt.Body))
-	l.hasher.HashJSONMarshalled(waitStmt.Args)
+	l.hasher.HashInt(len(w.Body))
+	l.hasher.HashJSONMarshalled(w.Args)
 }
 
 func (l *loader) handleWait(ctx context.Context, waitStmt spec.WaitStatement) error {
@@ -286,6 +287,11 @@ func (l *loader) loadTargetFromString(ctx context.Context, targetName string) er
 func (l *loader) hashVersion(v spec.Version) {
 	v.SourceLocation = nil
 	l.hasher.HashJSONMarshalled(v)
+}
+
+func (l *loader) hashCommand(cmd spec.Command) {
+	cmd.SourceLocation = nil
+	l.hasher.HashJSONMarshalled(cmd)
 }
 
 func (l *loader) findProject(ctx context.Context) (org, project string, err error) {
--- a/inputgraph/inputgraph.go
+++ b/inputgraph/inputgraph.go
@@ -195,7 +195,14 @@ func (l *loader) handleFor(ctx context.Context, forStmt spec.ForStatement) error
 }
 
 func (l *loader) handleWait(ctx context.Context, waitStmt spec.WaitStatement) error {
-	return errors.Wrap(ErrUnableToDetermineHash, "wait not supported")
+	l.hasher.HashString("wait")
+	l.hasher.HashString(strings.Join(waitStmt.Args, " "))
+	for _, stmt := range waitStmt.Body {
+		if err := l.handleStatement(ctx, stmt); err != nil {
+			return err
+		}
+	}
+	return nil
 }
 
 func (l *loader) handleTry(ctx context.Context, tryStmt spec.TryStatement) error {
@@ -351,9 +358,6 @@ func HashTarget(ctx context.Context, target domain.Target, conslog conslogging.C
 	err = loaderInst.load(ctx)
 	if err != nil {
 		return "", "", nil, err
-	}
-	if !loaderInst.isPipeline {
-		return "", "", nil, errors.Wrap(ErrUnableToDetermineHash, "target is not a pipeline")
 	}
 
 	return org, project, loaderInst.hasher.GetHash(), nil
--- a/inputgraph/inputgraph.go
+++ b/inputgraph/inputgraph.go
@@ -194,8 +194,14 @@ func (l *loader) handleFor(ctx context.Context, forStmt spec.ForStatement) error
 	return errors.Wrap(ErrUnableToDetermineHash, "for not supported")
 }
 
+func (l *loader) hashWaitStatement(waitStmt spec.WaitStatement) {
+	l.hasher.HashString("WAIT")
+	l.hasher.HashInt(len(waitStmt.Body))
+	l.hasher.HashJSONMarshalled(waitStmt.Args)
+}
+
 func (l *loader) handleWait(ctx context.Context, waitStmt spec.WaitStatement) error {
-	l.hasher.HashStatement("WAIT", waitStmt.Args, waitStmt.Body)
+	l.hashWaitStatement(waitStmt)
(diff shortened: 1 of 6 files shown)


inputgraph/inputgraph.go
```diff
@@ -195,7 +196,15 @@ func (l *loader) handleFor(ctx context.Context, forStmt spec.ForStatement) error
 }
 
 func (l *loader) handleWait(ctx context.Context, waitStmt spec.WaitStatement) error {
-	return errors.Wrap(ErrUnableToDetermineHash, "wait not supported")
+	l.hasher.HashString("wait")
+	sort.Strings(waitStmt.Args)
+	l.hasher.HashString(strings.Join(waitStmt.Args, " ")) // TODO: Is this necessary/helpful?
```
vladaionescu, line 201: It might be helpful in case `WAIT` gets args in the future.

vladaionescu (commented):
This looks good to me at first glance - I'll let @alexcb do the actual review.

inputgraph/inputgraph.go
alexcb: definitely useful -- I like the sorting of the Args -- that's actually something the existing code doesn't do a great job of.

a command like  `RUN --ssh --no-cache echo this last bits of the args matter`, will simply hash all args in the same order since we use a rather generic approach:

```
func (h *Hasher) HashCommand(cmd spec.Command) {
	dt, err := json.Marshal(cmd)
	if err != nil {
		panic(fmt.Sprintf("failed to hash command: %s", err)) // shouldn't happen
	}
	h.HashBytes(dt)
}
```

later on in the code it makes calls to `parseArgs` -- we could in theory figure out which of the Args represent flags (and can therefore be sorted), and which Args are part of the command can't be changed.

But this is all a rather small improvement to auto-skip in general, and could be pushed down the road.

alexcb (commented):
looking good so far! seems like the WAIT/END blocks didn't need too much extra work.

ultimately I think most of the work is needed in extending the tests under `tests/autoskip/...` to improve our test coverage.

inputgraph/inputgraph.go
vladaionescu: Hmm.. I missed this part. Couldn't sorting lead to strange behavior? `RUN foo bar` being hashed the same as `RUN bar foo`?

alexcb: in the generic sense (like `RUN`, or `COPY`) yes; ~~however it's unlikely that the ordering of the future args (flags) for `WAIT` would matter.~~

actually the order does matter since we could have future flags like `WAIT --key value` rather than `WAIT --key=value`.

best to remove the `sort.Strings(waitStmt.Args)`.

If the user changes any ordering of flags in their Earthfile, I think it's fine for us to report it's a difference -- I don't think it's a use-case which we'll encounter frequently.

inputgraph/inputgraph.go
```diff
@@ -195,7 +195,14 @@ func (l *loader) handleFor(ctx context.Context, forStmt spec.ForStatement) error
 }
 
 func (l *loader) handleWait(ctx context.Context, waitStmt spec.WaitStatement) error {
-	return errors.Wrap(ErrUnableToDetermineHash, "wait not supported")
+	l.hasher.HashString("wait")
```
vladaionescu, line 198: This is in-lieu of the `HashCommand` call, right? I wonder if we should separate this in a similar function and co-locate the functions? I'm thinking that if we make changes to one, we might want to change the other too, and it's easier to spot that if the functions are next to each other.

vladaionescu (approved):
Looks good to me - just a minor remark. I would also wait for @alexcb's review, since he's more familiar with this code.

inputgraph/inputgraph.go
alexcb: this might be a good idea, because we already have `HashCommand`, `HashVersion`; however those commands just take a `spec.Command`, or `spec.Version` which are both leaf nodes of the ast.

WAIT on the other hand has references to other `spec.Command`, which means we have to walk the tree, and would need to call back into inputgraph package.

I honestly think it's fine keeping this as-is.


However, I think the current code might have an issue, as it's not clear how the following different blocks:

```
WAIT
  RUN echo foo
END
```

and

```
WAIT
END
RUN echo foo
```

will produce a different hash.

I think we would an additional:

`HashString("end")` or `HashInt(len(waitStmt.Body))`

util/buildkitskipper/hasher/hasher.go
```diff
@@ -30,6 +31,11 @@ func (h *Hasher) GetHash() []byte {
 	return h.h.Sum(nil)
 }
 
+func (h *Hasher) HashStatement(name string, args []string) {
+	h.HashString(name)
+	h.HashString(strings.Join(args, " "))
```
alexcb, line 36: will `[]string{"a", "b"}` and `[]string{"a b"}` produce different hashes?

alexcb (changes requested):
looking good; just a few questions regarding how we're encoding data to hash in a way that accounts for the number of elements, and if whitespace is significant or not.

inputgraph/inputgraph.go
mikejholly: Updated with a new function called `HashStatement` in favor of specific functions for `HashWait`, `HashIf`, etc.

util/buildkitskipper/hasher/hasher.go
alexcb: > They seem to produce the same hash.

We need to avoid cases like this -- any changes that the user makes to their Earthfile that could possibly propagate down to their container, should result in a different hash.

> Shall I use a different separator?

This will lead to the same issue. instead you'll need to come up with a more robust way to hash them to different values. That's why we were using `json.Marshal` in the `HashCommand` function.

util/buildkitskipper/hasher/hasher.go
```diff
@@ -38,6 +45,10 @@ func (h *Hasher) HashCommand(cmd spec.Command) {
 	h.HashBytes(dt)
 }
 
+func (h *Hasher) HashInt(i int) {
+	h.h.Write([]byte(fmt.Sprintf("%d", i)))
```
alexcb, line 49: I think it's worthwhile that we include the datatype that we're hashing.

This will be a backwards-breaking change, but I think it's better to break things now rather than latter.

```suggestion
	h.h.Write([]byte(fmt.Sprintf("int:%d", i)))
```

and can you make a change to the HashString to include a `str:` prefix?

util/buildkitskipper/hasher/hasher.go
mikejholly: I was going to suggest a separator that cannot be used in the args, if one exists, but JSON is fine too. Updated.

util/buildkitskipper/hasher/hasher.go
mikejholly: Updated.

util/buildkitskipper/hasher/hasher.go
```diff
@@ -30,6 +31,16 @@ func (h *Hasher) GetHash() []byte {
 	return h.h.Sum(nil)
 }
 
+func (h *Hasher) HashStatement(name string, args []string, body spec.Block) {
```
alexcb, line 33: This function seems very specific in requiring a single body, I think we should rename it to reflect that it's not a general use function, maybe just call it:

```suggestion
func (h *Hasher) HashWaitStatement(waitStmt spec.WaitStatement) {
```

honestly pushing down the `spec` into the hasher package feels like a wrong direction, which goes back to the original thread that @vladaionescu started.

The fact that we're passing in a `spec.*` yet only processing half of it in this package, and the other half in the `inputgraph` package feels like we have incorrectly broken this logic up into too many packages.

util/buildkitskipper/hasher/hasher.go
alexcb: awesome, thanks!

util/buildkitskipper/hasher/hasher.go
alexcb: ultimately I think the `HashCommand`, `HashVersion`, and now `HashWaitStatement` all belong in the `inputgraph` package.

perhaps this package could contain a helper function like `HashJSONMarshalled` which would take `v any`.

mikejholly: Yes, that's pretty much why I started with the WAIT-specific logic in `inputgraph.go`. It seems less desirable to have `Hasher` become aware of the specific `spec` types. Updated with suggestions.

util/buildkitskipper/hasher/hasher.go
```diff
@@ -30,24 +28,20 @@ func (h *Hasher) GetHash() []byte {
...
 }
 
-func (h *Hasher) HashCommand(cmd spec.Command) {
-	dt, err := json.Marshal(cmd)
-	if err != nil {
-		panic(fmt.Sprintf("failed to hash command: %s", err)) // shouldn't happen
-	}
-	h.HashBytes(dt)
+func (h *Hasher) HashInt(i int) {
+	h.HashString(fmt.Sprintf("int:%d", i))
```
alexcb, line 32: something seems odd with this, as we're going to end up hashing `str:int:123`.

I think it makes more sense to call `h.h.Write([]byte(fmt.Sprintf("int:%d", i)))` directly, so we don't end up with cases where we could easily end up with hashing collisions.

e.g. currently we have a collision between `HashInt(123)` and `HashString("int:123")`.

It would be a good idea to look over the API of this package to see if there's an potentials for collisions.

inputgraph/inputgraph.go
```diff
@@ -271,6 +283,11 @@ func (l *loader) loadTargetFromString(ctx context.Context, targetName string) er
 	return loaderInst.load(ctx)
 }
 
+func (l *loader) hashVersion(v spec.Version) {
```
alexcb, line 286: this looks good, can we create a similar function for `hashCommand`

inputgraph/inputgraph.go
```diff
@@ -194,8 +194,20 @@ func (l *loader) handleFor(ctx context.Context, forStmt spec.ForStatement) error
 	return errors.Wrap(ErrUnableToDetermineHash, "for not supported")
 }
 
+func (l *loader) hashWaitStatement(waitStmt spec.WaitStatement) {
+	l.hasher.HashString("WAIT")
+	l.hasher.HashInt(len(waitStmt.Body))
+	l.hasher.HashJSONMarshalled(waitStmt.Args)
```
alexcb, line 200: we should set `v.SourceLocation = nil` here too

mikejholly: Updated.

inputgraph/inputgraph.go
mikejholly: Done.

util/buildkitskipper/hasher/hasher.go
mikejholly: Updated.

alexcb (commented):
This is looking good! Thanks for adjusting the hasher package.

I think there's still some hashes in the tests that need to be updated since the entire hashing algorithm has changed.

mikejholly:
> This is looking good! Thanks for adjusting the hasher package.
> 
> I think there's still some hashes in the tests that need to be updated since the entire hashing algorithm has changed.

I was confused by the fact that the tests are passing as is. I'll take a look at why.

https://cloud.earthly.dev/builds/6f60462e-6c2d-4d8c-8c52-96e4f646b399

mikejholly:
@alexcb it looks like the existing tests are only looking for (or not looking for) the hash of a file rather than the auto-skip hash value.

alexcb (approved):

PR #3372, opened 2023-10-12
Auto-skip: add support for loops

mikejholly:
Basic AutoSkip support for `FOR...IN`.

### DIFF
--- a/inputgraph/inputgraph.go
+++ b/inputgraph/inputgraph.go
@@ -206,7 +206,12 @@ func (l *loader) handleIf(ctx context.Context, ifStmt spec.IfStatement) error {
 }
 
 func (l *loader) handleFor(ctx context.Context, forStmt spec.ForStatement) error {
-	return errors.Wrap(ErrUnableToDetermineHash, "for not supported")
+	l.hashFor(forStmt)
+	err := l.loadBlock(ctx, forStmt.Body)
+	if err != nil {
+		return err
+	}
+	return nil
 }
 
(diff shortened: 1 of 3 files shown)


inputgraph/inputgraph.go
```diff
@@ -287,9 +289,14 @@ func (l *loader) hashVersion(v spec.Version) {
 	l.hasher.HashJSONMarshalled(v)
 }
 
-func (l *loader) hashCommand(cmd spec.Command) {
-	cmd.SourceLocation = nil
-	l.hasher.HashJSONMarshalled(cmd)
+func (l *loader) hashCommand(c spec.Command) {
+	c.SourceLocation = nil
```
alexcb, line 293: I see that in one of the `hash*` functions, you simply pass through the Args, e.g.

```
func (l *loader) hashFor(f spec.ForStatement) {
	l.hasher.HashString("FOR")
	l.hasher.HashJSONMarshalled(f.Args)
}
```

but in others you set `SourceLocation = nil` and hash the entire command.

Any thoughts on making this consistent? I personally like the idea of explicitly hashing the Name and Args (like you did in `hashFor` -- since it makes it more clear what we're hashing on.

alexcb (approved):
looks good -- just one question about if we want to make our hashing functions more consistent with each other.

inputgraph/inputgraph.go
mikejholly: It's likely better to be specific and reference individual fields. This prevents new fields from sneaking into the mix and so on.

alexcb (approved):

alexcb (approved):

PR #3373, opened 2023-10-12
Split examples target for disk use issue

mikejholly:
Existing `examples2` target seems to be using all available GHA runner disk.

### DIFF
--- a/.github/workflows/ci-docker-satellites.yml
+++ b/.github/workflows/ci-docker-satellites.yml
@@ -140,7 +140,7 @@ jobs:
       RUNS_ON: "ubuntu-latest"
       BINARY: "docker"
       SUDO: ""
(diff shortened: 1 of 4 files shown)


alexcb (approved):
thanks for tackling this!

idodod:
> FYI @alexcb not sure if we have examples set as required checks, if we do we might need to adjust them since the names slightly changed.

Actually, yes, seems like it's already happening for this PR:
image

alexcb:
> > FYI @alexcb not sure if we have examples set as required checks, if we do we might need to adjust them since the names slightly changed.
> 
> Actually, yes, seems like it's already happening for this PR: image

that's a good call, I'll update them. Note that this means any existing PRs will have to be rebased

PR #3379, opened 2023-10-13
Upgrade Docker version for +dind

mikejholly:
Fixes a test failure in `main`.

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -638,7 +638,7 @@ all-buildkitd:
         ./buildkitd+buildkitd --BUILDKIT_PROJECT="$BUILDKIT_PROJECT"
 
 dind-alpine:
-    BUILD +dind --OS_IMAGE=alpine --OS_VERSION=3.18 --DOCKER_VERSION=23.0.6-r5
+    BUILD +dind --OS_IMAGE=alpine --OS_VERSION=3.18 --DOCKER_VERSION=23.0.6-r6
 
 dind-ubuntu:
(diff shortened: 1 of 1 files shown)


alexcb (approved):

Earthfile
```diff
@@ -638,7 +638,7 @@ all-buildkitd:
         ./buildkitd+buildkitd --BUILDKIT_PROJECT="$BUILDKIT_PROJECT"
 
 dind-alpine:
-    BUILD +dind --OS_IMAGE=alpine --OS_VERSION=3.18 --DOCKER_VERSION=23.0.6-r5
+    BUILD +dind --OS_IMAGE=alpine --OS_VERSION=3.18 --DOCKER_VERSION=23.0.6-r6
```
idodod, line 641: alpine packages are a pain, we could use ranges, i.e.
```suggestion
    BUILD +dind --OS_IMAGE=alpine --OS_VERSION=3.18 --DOCKER_VERSION=~23.0.6-r6
```
but maybe that would go against the explicit image tag.

Alternatively I'll add Renovate as part of my work to try to catch those upgrades faster before packages disappear.

idodod (approved):

Earthfile
mikejholly: I tried to use `23.0.6` (without the release candidate) but it led to an error. I presume that versions without RC will stick around longer. :shrug:

idodod: the tilda _might_ make it work.

mikejholly: The `+dind` target seems to fail with the tilde. I'll take another look later.

idodod: no worries, maybe the tilda needs to be added in the script because the arg is used for the image tag.
But anyway I'm working on the script, UDC and everything related so I'll take a look.

PR #3385, opened 2023-10-14
Fix warning log

cupen:
The warning log.
```
buildkitd | Warning: Buildkit version (v0.7.19) is different from Earthly version (0.7.19)
```

It's replaced by log with version
```
buildkitd | Buildkit version: v0.7.19
```

### DIFF
--- a/buildkitd/buildkitd.go
+++ b/buildkitd/buildkitd.go
@@ -851,22 +851,23 @@ func printBuildkitInfo(bkCons conslogging.ConsoleLogger, info *client.Info, work
 	if isLocal {
 		printFun = bkCons.VerbosePrintf
 	}
-	if info.BuildkitVersion.Version != "unknown" {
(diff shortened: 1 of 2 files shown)


alexcb:
Thanks for the submission @cupen.

I'm surprised that you're getting a string without the `v` prefix.

Can you share some additional details about how you have installed earthly?

What's the output when you run `earthly --version`?

alexcb:
If you're using arch; this was fixed by the package maintainer under https://aur.archlinux.org/cgit/aur.git/commit/?h=earthly&id=4c72e7ac88d28257915da3b4e1d520fc523c78c1

and should not occur under v0.7.20.

cupen:
@alexcb Yes,  you are right.  I'm using arch, and it's ok after upgrade it to `v0.7.20` from aur.

PR #3388, opened 2023-10-16
Auto-skip: finish support for WITH DOCKER --load

mikejholly:
Extends auto-skip to handle SAVE IMAGE, and removes invalid "unable to handle WITH DOCKER --load with implicit image name" error.


inputgraph/inputgraph_test.go
```diff
@@ -0,0 +1,51 @@
...
+	ctx := context.Background()
+	cons := conslogging.New(os.Stderr, &sync.Mutex{}, conslogging.NoColor, 0, conslogging.Info)
+
+	org, project, hash, err := HashTarget(ctx, target, cons)
+	r.NoError(err)
+	r.Equal("earthly-technologies", org)
+	r.Equal("core", project)
+
+	hex := fmt.Sprintf("%x", hash)
+	r.Equal("b644ef18314c4d28c98a26ff21c593ac25a21c86", hex)
```
alexcb, line 31: can you expand the test to validate that making a change to the Earthfile produces a different hash?

e.g. if a user changed `SAVE IMAGE saved:latest` to `SAVE IMAGE saved:newer` it would produce a different hash value.

alexcb (commented):
looking good; just one suggestion to improve our test coverage.

meta-nit: can you edit the description of this PR to something more descriptive for when it gets merged and squashed. (for example reading `From what I can gather, ...` as a message when viewing `git log` on main seems a bit odd)

inputgraph/inputgraph_test.go
```diff
@@ -0,0 +1,107 @@
...
+	org, project, hash, err := HashTarget(ctx, target, cons)
+	r.NoError(err)
+	r.Equal("earthly-technologies", org)
+	r.Equal("core", project)
+
+	hex := fmt.Sprintf("%x", hash)
+	r.Equal("9d2903bc18c99831f4a299090abaf94d25d89321", hex)
+
+	path := "./testdata/with-docker/Earthfile"
+	err = replaceInFile(path, "saved:latest", "other:latest")
```
alexcb, line 36: can we make use of a temp directory instead? manipulating the file directly could lead to a difficult to debug case where the final restore doesn't get run due to an error or seg fault.

furthermore, it could be odd to have your text editor prompt you to reload `./testdata/with-docker/Earthfile"` if you happen to have it open.

alexcb:
I see you changed the order of your commit message, but it's still a bit difficult to read. 

I find concise messages are better -- the make it easier for future readers to understand -- both while debugging issues, and while writing our CHANGELOG updates.

Maybe try something like:

> Extends auto-skip to handle `SAVE IMAGE`, and removes invalid "unable to handle WITH DOCKER --load with implicit image name" error.

Issue #2377, opened 2022-11-08
WITH DOCKER is not executed in parallel in latest v0.6.28 (but it was in version v0.6.25)


jrodrigv:
Hi,

I have some targets meant for running tests projects in parallel that are using WITH DOCKER to run a Rabbimq service before running the tests. Using v0.6.25 I can see how  the test-executor instances are running fully in parallel loading the rabbitmq tar files in parallel etc. However when using v0.6.28 this is not the case anymore and the  test-executor instances  are running sequentially instead.  Is there a way to prevent this from happening?

```
parallel-testing:
    FROM +build
    WAIT
        FOR dir IN $(ls Tests/*Tests/*.csproj)
            COPY (+test-executor/TestResults --PROJECT=./$dir) ./TestResults
        END
        SAVE ARTIFACT ./TestResults testresults AS LOCAL earthly-artifacts/testresults
    END
```

```
   test-executor:
    FROM +build
    ARG PROJECT
    ARG ID
    COPY docker-compose-unit-tests-rabbitmq.yml .
    COPY coverlet.runsettings .

      WITH DOCKER \
     --compose docker-compose-unit-tests-rabbitmq.yml \
     --service rabbitmq
          RUN dotnet test ${PROJECT} --blame-hang-timeout 5m --no-build --collect:"XPlat Code Coverage" -p:CollectCoverage=true --settings coverlet.runsettings --logger trx --configuration $buildConfiguration --results-directory ./TestResults
      END
  
    SAVE ARTIFACT ./TestResults TestResults
```

vladaionescu:
Hi @jrodrigv - what's your `VERSION` declaration at the top of the Eartfhile?

jrodrigv:
Hi @vladaionescu - In both cases I was using:

`VERSION --wait-block 0.6`

vladaionescu:
I suspect it has something to do with either `WAIT` or this new piece of functionality that we released in v0.6.26:

> Loading Docker images as part of WITH DOCKER is now faster through the use of an embedded registry in Buildkit. This functionality was previously hidden (VERSION --use-registry-for-with-docker) and was only auto-enabled for Earthly Satellite users. It is now enabled by default for all builds

Can you try the following:

* On earthly v0.6.29 enable `VERSION --wait-block --no-use-registry-for-with-docker 0.6` and see if the issue goes away.
[...]

PR #3403, opened 2023-10-18
Fix WITH DOCKER is not executed in parallel

idelvall:
https://github.com/earthly/earthly/issues/2377


idelvall:
closing in favor of https://github.com/earthly/earthly/pull/3406

PR #3404, opened 2023-10-18
Auto-skip arg expansion

mikejholly:
This PR adds support for ARG passing and expansion to auto-skip. CLI args, `BUILD` args, `--pass-args`, and other command args are now used to generate the hash. Changing any of these will invalidate the auto-skip check.


cmd/earthly/subcmd/build_cmd.go
```diff
@@ -262,113 +265,6 @@ func (a *Build) ActionBuildImp(cliCtx *cli.Context, flagArgs, nonFlagArgs []stri
 	a.cli.Console().PrintPhaseHeader(builder.PhaseInit, false, "")
 	a.warnIfArgContainsBuildArg(flagArgs)
 
-	var skipDB bk.BuildkitSkipper
```
mikejholly, line 265: Refactored and moved. See below.

cmd/earthly/subcmd/build_cmd.go
```diff
@@ -416,6 +312,62 @@ func (a *Build) ActionBuildImp(cliCtx *cli.Context, flagArgs, nonFlagArgs []stri
...
+		return errors.Wrap(err, "build new buildkitd client")
+	}
+	defer bkClient.Close()
+
+	platr, err := a.platformResolver(cliCtx.Context, bkClient, target)
+	if err != nil {
+		return err
+	}
+
+	skipDB, targetHash, doSkip, err := a.initAutoSkip(cliCtx.Context, target, overridingVars, platr, cloudClient)
```
mikejholly, line 356: Some variables require the BK client, platform resolver, and Cloud client. I had to reorder some of the initialization here to make those variables available. The auto-skip status now appears under the `Init` section.

```
auto-skip | target +build-args (hash c541ff2117990e1fa9a2f61187b232b3582ce461) has already been run; exiting
```

cmd/earthly/subcmd/build_cmd.go
```diff
@@ -807,6 +754,112 @@ func receiveFileVersion2(ctx context.Context, conn io.ReadWriteCloser, localArti
 	return f.Close()
 }
 
+func (a *Build) runnerName(ctx context.Context) (string, bool, error) {
```
mikejholly, line 757: I split a few sections of code out into their own methods.

inputgraph/hash.go
```diff
@@ -0,0 +1,39 @@
...
+	"context"
+
+	"github.com/earthly/earthly/conslogging"
+	"github.com/earthly/earthly/domain"
+	"github.com/earthly/earthly/util/platutil"
+	"github.com/earthly/earthly/variables"
+)
+
+// HashOpt contains all of the options available to the hasher.
+type HashOpt struct {
```
mikejholly, line 13: I new option struct with more options pertinent to ARG handling.

inputgraph/loader.go
```diff
@@ -216,25 +230,17 @@ func (l *loader) expandDirs(dirs ...string) ([]string, error) {
...
-	m := map[string]struct{}{}
-	for _, v := range all {
-		m[v] = struct{}{}
-	}
+func (l *loader) expandArgs(ctx context.Context, args []string) []string {
 	ret := []string{}
-	for k := range m {
-		ret = append(ret, k)
+	for _, arg := range args {
+		expanded := l.varCollection.ExpandOld(arg) // TODO: correct func? need to shell out?
```
mikejholly, line 236: I'm using the `ExpandOld` method because `Expand` seemed to be for shell-out expansion. Is this correct?

inputgraph/loader.go
```diff
@@ -498,59 +516,52 @@ func (l *loader) findProject(ctx context.Context) (org, project string, err erro
...
+		CI:               l.ci,
+		Push:             l.push,
+		BuiltinArgs:      l.builtinArgs,
+		OverridingVars:   l.overridingVars,
+		EarthlyCIRunner:  l.earthlyCIRunner,
+		PlatformResolver: l.platformResolver,
+		GitMeta:          buildCtx.GitMetadata,
+		Features:         l.features,
+	}
+	l.varCollection = variables.NewCollection(collOpt)
```
mikejholly, line 549: The `variables.Collection` is being assembled here and reassembled for each child target. Does this look ok?

inputgraph/loader.go
```diff
@@ -26,76 +27,89 @@ var (
...
-	return argsCopy
-}
-
-func parseArgs(cmdName string, opts interface{}, args []string) ([]string, error) {
-	processed := stringutil.ProcessParamsAndQuotes(args)
-	return flagutil.ParseArgs(cmdName, opts, processed)
 }
 
 func (l *loader) handleFrom(ctx context.Context, cmd spec.Command) error {
+	cmd.Args = l.expandArgs(ctx, cmd.Args)
```
mikejholly, line 64: For the actual argument expansion, I'm using this method for each command. Does this look correct?

Could this approach lead to issues with shell-out behavior?

earthfile2llb/interpreter.go
```diff
@@ -1398,13 +1398,13 @@ func (i *Interpreter) handleEnv(ctx context.Context, cmd spec.Command) error {
...
-var errGlobalArgNotInBase = errors.New("global ARG can only be set in the base target")
+var ErrInvalidSyntax = errors.New("invalid syntax")
+var ErrRequiredArgHasDefault = errors.New("required ARG cannot have a default value")
+var ErrGlobalArgNotInBase = errors.New("global ARG can only be set in the base target")
 
-// parseArgArgs parses the ARG command's arguments
+// ParseArgArgs parses the ARG command's arguments
 // and returns the argOpts, key, value (or nil if missing), or error
-func parseArgArgs(ctx context.Context, cmd spec.Command, isBaseTarget bool, explicitGlobalFeature bool) (commandflag.ArgOpts, string, *string, error) {
+func ParseArgArgs(ctx context.Context, cmd spec.Command, isBaseTarget bool, explicitGlobalFeature bool) (commandflag.ArgOpts, string, *string, error) {
```
mikejholly, line 1407: Exported and used with auto-skip code.

inputgraph/loader.go
```diff
@@ -393,94 +423,82 @@ func (l *loader) loadBlock(ctx context.Context, b spec.Block) error {
...
+	if err != nil {
+		return nil, err
+	}
 
-func copyVisited(m map[string]struct{}) map[string]struct{} {
-	m2 := map[string]struct{}{}
-	for k := range m {
-		m2[k] = struct{}{}
+	// TODO: confirm these are set correctly.
+	overriding, err := variables.ParseCommandLineArgs(flagArgs)
```
mikejholly, line 438: Does this look correct? I'm looking to pass build args to the new loader instance. It seems to work.

cmd/earthly/subcmd/build_cmd.go
alexcb: that sounds like some real tight coupling, but unfortunately we need to perform the autoskip detection before creating the client -- since we want it to be as fast as possible. creating the client could be slow for reasons like having to start a new earthly-buildkitd container, or having a satellite wake up.

is it possible to have some sort of lazy client that could be used to skip any buildkit operations? For example, if any buildkit calls are made, it would initially return an error saying it can't be done at this stage. Then once the auto-skip logic has finished (and returned an error saying autoskip detection can't be done), we then connect this lazy client with the real buildkit client.

cmd/earthly/subcmd/build_cmd.go
alexcb: nice refactor!

I personally find `runnerName` to be a bit confusing (I had to look it up under the original code to see what it does), maybe we should rename it while we're refactoring this.

I'm not entirely sure what it's used for, but it appears to be some sort of short-hand name of the buildkit instance we're connecting to, which gets passed to logbus, and sent in a proto field that's called "Runner".

maybe we should call it `buildkitInstanceName` instead? or `logBusRunnerName`?

runner seems too vague and could get confused with our the "runner" process we used from our CI intrastructure that wraps the `earthly` binary.

earthfile2llb/interpreter.go
alexcb: I'm not sure if it's a great idea to have the inputgraph call into the earthfile2llb package; this function seems isolated from the rest of the interpreter, so maybe it can be moved into a new file under `util/flagutil/parse_arg_args.go` (or maybe even rename it to `ParseArgCommandArgs` to make it more clear since argargs is a mouthful).

inputgraph/loader.go
alexcb: I would use `Expand`, and pass a function that prevents shelling out, you can follow the pattern used in https://github.com/earthly/earthly/blob/410c690e4df488287225c713da16d817a3488932/earthfile2llb/converter.go#L1710

and use something like:

```
	return c.varCollection.Expand(word, func(cmd string) (string, error) {
		return "", errShellOutNotPermitted
	})
```

cmd/earthly/subcmd/build_cmd.go
mikejholly: The BK client is used to determine the platform.
https://github.com/earthly/earthly/blob/mh%2Fauto-skip-arg-passing/util/platutil/native.go#L27

We can remove it but we'll lose support for these variables:
https://github.com/earthly/earthly/blob/mh%2Fauto-skip-arg-passing/variables/builtin.go#L37-L41

Perhaps we should assume the platform or just remove support for these variables for the time being.

The lazy client idea could work but it sounds like AS would still depend on a connection if these variables are used.

earthfile2llb/interpreter.go
mikejholly: Sounds good. Moving some of these shared funcs to a separate package makes sense.

inputgraph/loader.go
alexcb: at a quick glance yes. it's unfortunate we need to duplicate this logic from the converter, but it's a necessary evil -- we're going to rely on test-coverage to prevent drift.

inputgraph/loader.go
mikejholly: I should clarify. What I'm unsure of is whether using `OverridingVars` is the expected way to set build args when using `variables.Collection`.

This seems to be the case based on some uses elsewhere:
https://github.com/earthly/earthly/blob/mh%2Fauto-skip-arg-passing/earthfile2llb/converter.go#L137
https://github.com/earthly/earthly/blob/mh%2Fauto-skip-arg-passing/variables/collection.go#L93

inputgraph/loader.go
mikejholly: Great, thanks for confirming.

cmd/earthly/subcmd/build_cmd.go
alexcb: good find regarding the connection to buildkit to determine these.... I think this is just something we can't support and we'll need to simply return a `ErrUnableToDetermineHash` when these are required.

maybe down the road we could allow a way to specify these values locally, but i think we can leave this out for our autoskip-mvp.

cmd/earthly/subcmd/build_cmd.go
mikejholly: The lingo is used in a few other places. I'm not sure that we should change it everywhere.

https://github.com/earthly/earthly/blob/mh%2Fauto-skip-arg-passing/earthfile2llb/earthfile2llb.go#L174
https://github.com/earthly/earthly/blob/mh%2Fauto-skip-arg-passing/earthfile2llb/converter.go#L2393

cmd/earthly/subcmd/build_cmd.go
mikejholly: SGTM. Makes sense.

inputgraph/loader.go
mikejholly: Thanks, I'll give that a shot.

inputgraph/loader.go
alexcb: I take back my "I think so!" comment, it's rather murky :sweat_smile: 

when the converter ends up in `buildTarget`, it calls prepBuildTarget: https://github.com/earthly/earthly/blob/410c690e4df488287225c713da16d817a3488932/earthfile2llb/converter.go#L1781

which then calls some similar variables functions:

https://github.com/earthly/earthly/blob/410c690e4df488287225c713da16d817a3488932/earthfile2llb/converter.go#L1740C2-L1748

note that it depends on if `--pass-args` is given, or if we're calling a target in the same Earthfile, or a different one.

which then gets stored under `OverridingVars`, which then gets referenced under https://github.com/earthly/earthly/blob/410c690e4df488287225c713da16d817a3488932/earthfile2llb/converter.go#L1791-L1831

there's a lot of logic here -- maybe we can move it out of the `earthfile2llb` package, and into a common util which both the inputgraph and earthfile2llb packages could call?

cmd/earthly/subcmd/build_cmd.go
alexcb: feel free to ignore it, I think a simple comment for the method would be good enough for now.

alexcb: maybe something like:
```
// runnerName returns the name of the logbus runner; which is a representation of what buildkit instance is being used, e.g. local:<hostname>, sat:<org>/<name>, or bk:<remote-address>
```

inputgraph/loader.go
mikejholly: Thanks. I ran into this code as well. I think it may be hard to extract given that the `Converter` code is [bound](https://github.com/earthly/earthly/blob/mh%2Fauto-skip-arg-passing/earthfile2llb/converter.go#L1800) to the `MultiTarget`. But I think the basic idea of propagating them is the same. I'll also have to account for `--pass-args` & `--build-args` I think. Anyhow, sounds like I'm on the right track!

cmd/earthly/subcmd/build_cmd.go
mikejholly: SGTM!

earthfile2llb/interpreter.go
mikejholly: I ended up moving a number of shared functions to `flagutil`. Since some deal with arguments rather than de facto flags, I wonder if we should create a separate module. Conversely, the `flagutil` module already contains a few functions that aren't strictly to do with flags.

alexcb (approved):
This looks great!

I checked it out, built it and tried it against some examples like

```
VERSION 0.7

target:
  FROM alpine
  COPY data .
  RUN cat data

test:
  FROM alpine
  ARG callthis
  BUILD "+$callthis"
```

and called it with `earthly --auto-skip --auto-skip-db-path=skip.db +test --callthis=target` and it correctly knew to go into `+target` and hash the contents of `data`, very nice work!

once small nit, the console output has changed, it used to look like

```
$ earthly --auto-skip --auto-skip-db-path=skip.db +simple
 Init 🚀

target +simple (hash 5647e69e15b84c1b52387b39c3c63a4bdf466a9c) has already been run; exiting
```

and now it looks like

```
$ /home/alex/gh/earthly/earthly/build/linux/amd64/earthly --auto-skip --auto-skip-db-path=skip.db +simple
 Init 🚀

           auto-skip | target +simple (hash 48397af8fc3696ece93f7d483e74583d340ed668) has already been run; exiting
```

it's probably fine for this PR. I think in the future we might want to see if we can remove the ` Init 🚀` header, and possibly revert this change, but it's very minor.

earthfile2llb/interpreter.go
alexcb: nice work on the refactor here, it's good to see we were able to make it more clear which parts were used in both inputgraph and earthlfile2llb packages.

mikejholly:
@alexcb 
> console output has changed, it used to look like

Sure, I can remove the console prefix. I added that alignment when it seemed that initializing the BK client first may have been necessary.

PR #3405, opened 2023-10-18
Support `d`  (for days) suffix in `prune --age` 

### LINKED ISSUE #3401: `earthly prune --age` does not support days suffix, despite help saying it does
**What went wrong?**

```console
$ earthly -v
earthly version v0.7.20 9600f376026c11d23eb43fc68ad716d6ae4cca2e linux/amd64; Debian GNU/Linux 

$ earthly prune --help
...
   --age value   Prune cache older than the specified duration passed in as a string;
                           duration is specified with an integer value followed by a m, h, or d suffix which represents minutes, hours, or days respectively, e.g. 24h, or 1d (default: 0s)

$ earthly prune --age 7d
...
Error: invalid value "7d" for flag -age: parse error
```

Ultimately the CLI parsing is using (via `urfave/cli.DurationFlag`) [`time.ParseDuration`](https://pkg.go.dev/time#ParseDuration) which says:

> Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h". 

**What should have happened?**

Code and help message should be consistent.

idodod:
Fixes #3401


CHANGELOG.md
```diff
@@ -5,9 +5,11 @@ All notable changes to [Earthly](https://github.com/earthly/earthly) will be doc
 ## Unreleased
 
 ### Added
-
 - The new ARG `EARTHLY_GIT_REFS` will contain the references to the current git commit, this ARG must be enabled with the `VERSION --git-refs` feature flag. [#3341](https://github.com/earthly/earthly/pull/3341)
 
+### Fixed
+- `prune --age` does not support `d` (for days) as a suffix contradictory to help message [#3401](https://github.com/earthly/earthly/issues/3401)
```
alexcb, line 11: I had to read this a few times before I understood what it meant, maybe something like this instead?
```suggestion
- `prune --age` did not support `d` (for days) suffix, even thought `earthly --help` said it did [#3401](https://github.com/earthly/earthly/issues/3401)  
```

cmd/earthly/subcmd/common.go
```diff
@@ -22,6 +25,36 @@ func (b *byteSizeValue) Set(s string) error {
 
 func (b *byteSizeValue) String() string { return humanize.Bytes(uint64(*b)) }
 
+// duration implements cli.GenericFlag methods to support time.Duration with days, e.g. 1d
+type duration struct {
```
alexcb, line 29: nit: is it possible to just do `type duration time.Duration` instead?

alexcb: also this feels like something that should exist in a seperate package, possibly under `util/timeutil` ?

cmd/earthly/subcmd/prune_cmds.go
```diff
@@ -55,11 +52,11 @@ func (a *Prune) Cmds() []*cli.Command {
 				This option is not available when using satellites.`,
 					Destination: &a.reset,
 				},
-				&cli.DurationFlag{
+				&cli.GenericFlag{
```
alexcb, line 55: neat! I didn't know this existed.

alexcb (commented):
looks good! just think it belongs in a util package instead.

cmd/earthly/subcmd/common.go
idodod: I'm not sure I'll be able to pass the value from `Set(value string)`, or am I missing something?

idodod: oh, actually looks like I could:
`*d = duration(value)`

alexcb: haha beat me to it, I just put this very simple example together:

```
package main

import (
	"fmt"
	"time"
)

type duration time.Duration

func (d *duration) Set(value string) {
	fmt.Printf("hello %s\n", value)
	*d = duration(time.Second)
}

func main() {
	var d duration
	fmt.Printf("before: %v\n", d)
	d.Set("world")
	fmt.Printf("after: %v\n", d)
}
```

idodod:
> looks good! just think it belongs in a util package instead.

I noticed we already use another (unexported) generic type in common.go, so that's why I put it there.
https://github.com/earthly/earthly/blob/335f6346302b4ac712a090ece641ba8991bae4f1/cmd/earthly/subcmd/common.go#L14

Should I move that type there as well then?
I was also thinking of placing it in `cmd/earthly/flag`

vladaionescu (approved):

util/flagutil/flag.go
alexcb, line 1: nit: break this up into two files, e.g. `byte_size_value.go` and `duration.go`.

util/flagutil/flag_test.go
alexcb, line 1: and correspondingly rename this to `duration_test.go`

alexcb (approved):
looks good! just a minor nit for further breaking up the source code.

PR #3406, opened 2023-10-19
Fix for limited parallelism when the target is the same but has different args

### LINKED ISSUE #2377: WITH DOCKER is not executed in parallel in latest v0.6.28 (but it was in version v0.6.25)

vladaionescu:
Fixes #2377

Note that I was able to simplify the reproduction case - it is actually not directly related to `WITH DOCKER`. A difference in the implementation of the new `WITH DOCKER` command caused a pre-existing issue to surface.

The previous `VisitedCollection` implementation attempts to wait for ongoing targets with the same name to complete before comparing the inputs of the target. This has the advantage of being more precise with which ARGs actually influence the outcome of the targets, ignoring overriding ARGs that end up being unused within those targets. But it also has the disadvantage that all targets with the same name (but different overriding args) execute sequentially.

This new implementation simplifies this greatly, by computing the target input hash upfront based on all overriding args, without knowing if there are any args that will end up being unused. The result is that we are able to run all targets with the same name but different args in parallel. But it also has the disadvantage that in some cases we would create duplicated LLBs for some targets when certain overriding args are different but they are unused. Buildkit will generally de-duplicate the LLB in these cases, although it's possible that there might be edge cases if the LLB construction is not consistent.


states/visited2.go
```diff
@@ -0,0 +1,64 @@
...
+		select {
+		case <-ctx.Done():
+			return nil, false, ctx.Err()
+		case <-sts.Done():
+			// Return existing.
+			return sts, true, nil
+		}
+	}
+	vc.visited[newKey] = newSts
+	vc.visitedList = append(vc.visitedList, newSts)
```
idelvall, line 61: What is the purpose of the list? Seems like we could remove it

states/visited2.go
```diff
@@ -0,0 +1,64 @@
...
+		return nil, false, err
+	}
+	newKey, err := newSts.targetInput.Hash()
+	if err != nil {
+		return nil, false, err
+	}
+	vc.mu.Lock()
+	sts, found := vc.visited[newKey]
+	if found {
+		vc.mu.Unlock()
```
idelvall, line 50: Looks like `newSts.targetInput.Hash()` is not consistent with `newSts.targetInput.Equals()`, that is, we are getting cache hits (same hash), that are not considered equal. Not sure yet of the consequences but worth mentioning.

Changing to this test file
```
VERSION 0.7

FROM earthly/dind:alpine

parallel:
  BUILD +test-executor
  BUILD +test-executor --INDEX=2
  BUILD +test-executor --INDEX=3
  BUILD +test-executor --INDEX=4
  BUILD +test-executor --INDEX=5

test-executor:
  ARG INDEX=1
  RUN --no-cache for i in $(seq 1 3); do  \
        echo "### $INDEX"; \
        sleep 0.3; \
      done;
  IF true
    RUN true
  END
```
and adding here this code:
```
if !sts.targetInput.Equals(newSts.targetInput) {
			nb, _ := json.Marshal(newSts.targetInput)
			b, _ := json.Marshal(sts.targetInput)
			fmt.Println(fmt.Sprintf("%s %s\ncreated:%s\nfound:  %s", target.String(), strings.Join(overridingVars.Sorted(), " "), string(nb), string(b)))
		}
```
outputs:
```
./tests/with-docker-parallel+test-executor DOCKERHUB_MIRROR_AUTH_FROM_CLOUD_SECRETS INDEX
created:{"targetCanonical":"github.com/earthly/earthly/tests/with-docker-parallel:vlad/issue#2377+test-executor","buildArgs":[{"name":"DOCKERHUB_MIRROR_AUTH_FROM_CLOUD_SECRETS","constantValue":"true","defaultConstant":""},{"name":"INDEX","constantValue":"2","defaultConstant":""}],"platform":"linux/amd64","allowPrivileged":true}
found:  {"targetCanonical":"github.com/earthly/earthly/tests/with-docker-parallel:vlad/issue#2377+test-executor","buildArgs":[{"name":"DOCKERHUB_MIRROR_AUTH_FROM_CLOUD_SECRETS","constantValue":"true","defaultConstant":""},{"name":"INDEX","constantValue":"2","defaultConstant":"1"}],"platform":"linux/amd64","allowPrivileged":true}
./tests/with-docker-parallel+test-executor DOCKERHUB_MIRROR_AUTH_FROM_CLOUD_SECRETS INDEX
created:{"targetCanonical":"github.com/earthly/earthly/tests/with-docker-parallel:vlad/issue#2377+test-executor","buildArgs":[{"name":"DOCKERHUB_MIRROR_AUTH_FROM_CLOUD_SECRETS","constantValue":"true","defaultConstant":""},{"name":"INDEX","constantValue":"3","defaultConstant":""}],"platform":"linux/amd64","allowPrivileged":true}
found:  {"targetCanonical":"github.com/earthly/earthly/tests/with-docker-parallel:vlad/issue#2377+test-executor","buildArgs":[{"name":"DOCKERHUB_MIRROR_AUTH_FROM_CLOUD_SECRETS","constantValue":"true","defaultConstant":""},{"name":"INDEX","constantValue":"3","defaultConstant":"1"}],"platform":"linux/amd64","allowPrivileged":true}
./tests/with-docker-parallel+test-executor DOCKERHUB_MIRROR_AUTH_FROM_CLOUD_SECRETS INDEX
created:{"targetCanonical":"github.com/earthly/earthly/tests/with-docker-parallel:vlad/issue#2377+test-executor","buildArgs":[{"name":"DOCKERHUB_MIRROR_AUTH_FROM_CLOUD_SECRETS","constantValue":"true","defaultConstant":""},{"name":"INDEX","constantValue":"4","defaultConstant":""}],"platform":"linux/amd64","allowPrivileged":true}
found:  {"targetCanonical":"github.com/earthly/earthly/tests/with-docker-parallel:vlad/issue#2377+test-executor","buildArgs":[{"name":"DOCKERHUB_MIRROR_AUTH_FROM_CLOUD_SECRETS","constantValue":"true","defaultConstant":""},{"name":"INDEX","constantValue":"4","defaultConstant":"1"}],"platform":"linux/amd64","allowPrivileged":true}
./t
```

states/visited2.go
```diff
@@ -0,0 +1,64 @@
...
+// All returns all visited items.
+func (vc *VisitedCollection2) All() []*SingleTarget {
+	vc.mu.Lock()
+	defer vc.mu.Unlock()
+	return append([]*SingleTarget{}, vc.visitedList...)
+}
+
+// Add adds a target to the collection, if it hasn't yet been visited. The returned sts is
+// either the previously visited one or a brand new one.
+// This function blocks if there is a visited target that is still running.
```
idelvall, line 36: >// This function blocks if there is a visited target that is still running.

This doesn't seem accurate, since it only blocks if there is cache hit to a running target?

states/visited2.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	"context"
+	"sync"
+
+	"github.com/earthly/earthly/domain"
+	"github.com/earthly/earthly/util/platutil"
+	"github.com/earthly/earthly/variables"
+)
+
+// VisitedCollections is a collection of visited targets.
```
idelvall, line 12:
```suggestion
// VisitedCollections2 is a collection of visited targets.
```

states/visited2.go
```diff
@@ -0,0 +1,64 @@
...
+type VisitedCollection2 struct {
+	mu      sync.Mutex
+	visited map[string]*SingleTarget // targetInputHash -> sts
+	// Same collection as above, but as a list, to make the ordering consistent.
+	visitedList []*SingleTarget
+}
+
+// NewVisitedCollection returns a collection of visited targets.
```
idelvall, line 20:
```suggestion
// NewVisitedCollection2 returns a collection of visited targets.
```

idelvall:
PR for feature flag: https://github.com/earthly/earthly/pull/3407

states/visited2.go
vladaionescu: Yes, correct - it only blocks if the target input hash matches.

states/visited2.go
vladaionescu: This should be ok. The only difference is the `DefaultValue` field, which we want to ignore anyway in the new implementation.

states/visited2.go
idelvall: What is the rationale behind this behavior?  
I mean why the first time visited it doesn't block but successive calls do?

idelvall:
Will polish tests next

states/visited2.go
idelvall: I would rather make `hash` and `equals` implementations consistent. 
What makes more sense, making `hash` dealing with the defaultConstant or `equals` ignoring it?

vladaionescu: Let me check if it's safe to drop it from equals - we really need hash to ignore it as part of this.

vladaionescu: Ok, I made these consistent now.

idelvall:
Add tests PR: https://github.com/earthly/earthly/pull/3413

idelvall:
We should run all the tests with this new visitor implementation, to be sure we didn't introduce any regressions. Notice that right now, the feature flag is disabled by default and hence the new visitor implementation is only used in the new test added

alexcb:
> We should run all the tests with this new visitor implementation, to be sure we didn't introduce any regressions. Notice that right now, the feature flag is disabled by default and hence the new visitor implementation is only used in the new test added

we do something similar with the global wait/end code under: https://github.com/earthly/earthly/blob/2a9f2b10beec627d7a43238a9adb6516a833416d/.github/workflows/reusable-wait-block-target.yml#L58-L61

perhaps a similar approach could be done?

vladaionescu:
> perhaps a similar approach could be done?

Would that add too much overhead to our CI stack? I think it's safe enough to let it go without extensive testing until we flip the flag in 0.8. I can see why this is necessary for the wait block though.

alexcb:
> > perhaps a similar approach could be done?
> 
> Would that add too much overhead to our CI stack? I think it's safe enough to let it go without extensive testing until we flip the flag in 0.8. I can see why this is necessary for the wait block though.

You could always limit it to a target like `+test-quick`

vladaionescu:
> You could always limit it to a target like +test-quick

I ended up enabling via `.earthly_version_flag_overrides`

states/visited2.go
vladaionescu: I ended up moving it out, so that I can use the hash for infinite recursion checking before waiting.

Earthfile
```diff
@@ -466,7 +466,7 @@ earthly-docker:
 # Otherwise, it will attempt to login to the docker hub mirror using the provided username and password
 earthly-integration-test-base:
     FROM +earthly-docker
-    RUN apk update && apk add pcre-tools curl python3 bash perl findutils expect yq
+    RUN apk update && apk add pcre-tools curl python3 bash perl findutils expect yq && apk add --upgrade sed
```
alexcb, line 469: I think it's worthwhile adding a comment why the `--upgrade` flag is needed here -- is it because the included version of sed is busybox, which doesn't work? (I had to run this target to come up with this guess)

alexcb (approved):
I _think_ it looks ok; I found the leaselock a bit hard to follow -- is there a reason we can't just use a global mutex instead for LOCALLY?

vladaionescu:
> I think it looks ok; I found the leaselock a bit hard to follow -- is there a reason we can't just use a global mutex instead for LOCALLY?

The complexity is unfortunately necessary, due to this case:

```Earthly
target1:
  LOCALLY
  BUILD +target2
  ...

target2:
  LOCALLY
  ...
```

You need some sort to re-entrant lock. The leaselock is something of that sort.

vladaionescu:
I ended up removing the locally lock as I noticed some dead-locking. I'll separate into a different PR since it's not related to these changes anyway (it was a pre-existing bug).

Earthfile
idelvall: Yes, that is exactly the reason

alexcb (approved):

idelvall (approved):
LGTM

PR #3407, opened 2023-10-19
Hides new target visited cache under a feature flag

idelvall:


states/visited-new.go
```diff
@@ -48,6 +51,11 @@ func (vc *VisitedCollection2) Add(ctx context.Context, target domain.Target, pla
 	sts, found := vc.visited[newKey]
 	if found {
 		vc.mu.Unlock()
+		if !sts.targetInput.Equals(newSts.targetInput) {
+			nb, _ := json.Marshal(newSts.targetInput)
+			b, _ := json.Marshal(sts.targetInput)
+			fmt.Println(fmt.Sprintf("%s %s\ncreated:%s\nfound:  %s", target.String(), strings.Join(overridingVars.Sorted(), " "), string(nb), string(b)))
+		}
```
vladaionescu, line 58: Just a reminder to remove this debug output. It's expected that the same hashes don't necessarily match completely, due to the `DefaultValue` field, which we want to ignore anyway.

states/visited.go
```diff
@@ -2,160 +2,13 @@ package states
...
-		} else {
-			if !bai.IsDefaultValue() {
-				return false, nil
-			}
-		}
-	}
-	return true, nil
+type VisitedCollection interface {
+	All() []*SingleTarget
+	Add(ctx context.Context, target domain.Target, platr *platutil.Resolver, allowPrivileged bool, overridingVars *variables.Scope, parentDepSub chan string) (*SingleTarget, bool, error)
```
vladaionescu, line 13: Mind copying the documentation of these functions + add documentation for the interface itself?

vladaionescu (approved):

PR #3411, opened 2023-10-19
vendor dependencies

### LINKED ISSUE #3408: Enhancement request: Consider using go-vendor
Hello!

I'm writing as a Earthly user but also as the maintainer of Earthly Gentoo package.

Currently in Gentoo if a go package does not provide vendored go deps we have to create it ourselves.
Please see: https://wiki.gentoo.org/wiki/Writing_go_Ebuilds#Packaging_the_dependencies

It would speed up greatly earthly package maintainership and testing if Earthly provided vendor in the repo.

For reference project using go-vendor see: https://github.com/containers/podman-tui

Thanks a lot in advance.

alexcb:
fixes #3408


mikejholly:
Does this now need to be run for every PR that changes `go.mod`?

alexcb:
> Does this now need to be run for every PR that changes `go.mod`?

unfortunately yes -- we should add some sort of check to make sure they stay in sync.

mikejholly:
Having to vendor deps is bit of a step backward. I wonder if this "ebuild" script can just run `ego mod vendor` itself [before the build step](https://gitweb.gentoo.org/repo/gentoo.git/tree/app-containers/earthly/earthly-0.7.23.ebuild#n51). `ego` looks like it's just a [simple wrapper](https://github.com/gentoo/gentoo/blob/master/eclass/go-module.eclass#L205) around `go`, so this seems like a possibility. If not, my follow-up question is whether it makes sense to take on the extra effort of vendoring for each `go.mod` change as it doesn't seem like it's saving that much effort on Gentoo release end. If we really need to vendor it would be nice to add `+vendor` to a pre-commit [Git hook](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks).

mikejholly:
Oh, it looks like the `go-module.eclass` file may already support "live vendoring" https://github.com/gentoo/gentoo/blob/master/eclass/go-module.eclass#L513.

alexcb:
@xgqt do the above alternatives that Mike suggested work for you?

xgqt:
@alexcb @mikejholly 

Live packages

> Oh, it looks like the `go-module.eclass` file may already support "live vendoring" https://github.com/gentoo/gentoo/blob/master/eclass/go-module.eclass#L513.

Live is a special mode for packages that declare `PROPERTIES="live"` and are meant to pull in the most recent changes of a given software project when requested by user. Live packages do no have specific version, most of the time the version id `9999` or a main-release branch, say 0.9999, 1.9999. Live packages that also use git/mercurial/that-haskell-thingy can be rebuilt automatically by portage whenever there are updates detected on the repo's HEAD.

Vendor pre-build

> I wonder if this "ebuild" script can just run ego mod vendor itself [before the build step](https://gitweb.gentoo.org/repo/gentoo.git/tree/app-containers/earthly/earthly-0.7.23.ebuild#n51).

We can not do that for package releases (and tags) as all sources have to have a consistent checksum ***and*** Portage prevents fetching of assets form network for the sake of reproducible builds, system consistency and user safety.

Exception are live ebuilds that are allowed to access net only in `src_fetch` (a phase for ensuring source availability before the build step).

Choice

>  If not, my follow-up question is whether it makes sense to take on the extra effort of vendoring for each go.mod change as it doesn't seem like it's saving that much effort on the Gentoo release end.

Some Go projects were already vendoring deps before any from-source distro dev asked them if they can. It is mostly up to you, the devs to decide if this is worth the effort.

If I understand correctly (as I do not code in Go) using go vendor will make the build reproducible and in turn could be worth the effort of vendoring.

Current status

Right now it is not a bug deal for me as I have written a simple tool in F# that can take care of creating a vendor tarball to easy my work a bit, so the amount of work I had to do becomes even less.

I think if using vendor would be more mostly unnecessary work for Go devs it can stay as-is now.

Sidenote: We do not use `git` in release builds at all, only for live package versions.

mikejholly:
Hey @xgqt, thanks for the extra context.

Is it possible to extend the `go-module.eclass` script to optionally run `go mod vendor` when releasing a specific version?

I'm not entirely against vendoring, but it is worth considering all of the costs & benefits. Vendoring can become a pain.

Pros
* Make some third-party release processes easier
* Less reliance on GitHub & other services during builds
* More repeatable builds (debatable with `go.sum`)

Cons
* Heavier repos (first commit is 1.4 million lines)
* Small extra development burden (can be automated)
* Noisy PRs where small changes can include 1000s of lines
* Forgetting to `go mod vendor` may break 3rd party release processes
* Make break Renovate workflow (looks like vendoring is supported)

@alexcb @vladaionescu @idodod Any thoughts?

xgqt:
@mikejholly 

> Is it possible to extend the go-module.eclass script to optionally run go mod vendor when releasing a specific version?

Yes, but actually no :) 

I think if you try running it and you do not already have any vendored deps in your build dir it will fail with "sandbox violation" since you have no access to net during build. We can only fetch some extra stuff that is ***not** recorded in ebuild Manifest files only for "live" packages.

alexcb:
These are the two cons that stick out to me:
- Heavier repos (first commit is 1.4 million lines)
- Noisy PRs where small changes can include 1000s of lines

I would be personally fine with closing this PR, and deciding not to vendor our code since @xgqt has a tool that vendors in code before packaging.

Issue #3381, opened 2023-10-13
Logstream retry back-off issue


mikejholly:
**What went wrong?**

Logstream retry attempts seem to be happening immediately (without backoff) in some cases. 

**What should have happened?**

Retries should backoff progressively. 9m worth of failures should only produce ~20 errors. We also may want to truncate the error list.

**Other Helpful Information**

![image](https://github.com/earthly/earthly/assets/332408/8bff3f20-641c-40f7-aba4-cd7d9af8eee5)

mikejholly:
This was fixed in Oct.

PR #3415, opened 2023-10-20
Logstream: fixes flaky test & improves error handling

mikejholly:
Addresses: https://github.com/earthly/earthly/issues/3381

This test was essentially stuck in an infinite retry loop as the backoff time was always 0. This PR includes some revised error handling and test tweaks to fix the retry issue. The error handling changes ensure errors are always reported when the CLI exits.

### DIFF
--- a/cloud/logstream.go
+++ b/cloud/logstream.go
@@ -24,39 +24,47 @@ func (s *StreamError) Error() string {
 	return s.Err.Error()
 }
 
-func (c *Client) StreamLogs(ctx context.Context, man *pb.RunManifest, ch <-chan *pb.Delta) []error {
+func (c *Client) StreamLogs(ctx context.Context, man *pb.RunManifest, ch <-chan *pb.Delta) <-chan error {
 	if man.GetResumeToken() == "" {
 		man.ResumeToken = stringutil.RandomAlphanumeric(40)
 	}
-	var (
-		errs   []error
-		retry  bool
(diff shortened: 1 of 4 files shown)


cloud/logstream.go
```diff
@@ -24,12 +24,14 @@ func (s *StreamError) Error() string {
 	return s.Err.Error()
 }
 
-func (c *Client) StreamLogs(ctx context.Context, man *pb.RunManifest, ch <-chan *pb.Delta) []error {
+func (c *Client) StreamLogs(ctx context.Context, man *pb.RunManifest, ch <-chan *pb.Delta, errCh chan<- error) {
```
vladaionescu, line 27: Just a pattern idea: What if this StreamLogs function was async. It would start off the goroutine within it, and it would construct errCh itself and return it immediately.

The caller can wait for errCh to close to know that the streaming is done.

It would make the API a little cleaner.

vladaionescu (approved):

alexcb (approved):

cloud/logstream.go
mikejholly: I like it. I'll take another look in the next week or so. I'll merge this for now as it fixes a flaky test.

mikejholly: Actually, it was a pretty straightforward change. Updated!

vladaionescu (approved):

PR #3417, opened 2023-10-21
WIP Fix for `LOCALLY` parallelism

vladaionescu:
`LOCALLY` targets should not execute in parallel. This PR fixes that by introducing a global lock.

Depends on https://github.com/earthly/earthly/pull/3406


vladaionescu:
It's a design principle - that things should be as repeatable as possible. We figured that kinda like how you rarely run `make` with the `-j` argument because it's hard to get parallelism right when the environment is shared, we should avoid parallelizing in LOCALLY cases in Earthly too.

It's probably not an important guardrail now that there is also `WAIT`, which allows the user to explicitly control parallelism where necessary. Plus nobody has asked for this feature, so we could also abandon it. CC @alexcb

Issue #3412, opened 2023-10-20
Registry proxy limitations with Docker Desktop


mikejholly:
**What existing functionality needs improvement?**
The remote registry proxy doesn't work smoothly with Mac. Because `docker pull` pulls from within the Docker Desktop VM, it needs to reference a special address, `host.docker.internal`, in order to communicate with the user's host machine. Only `localhost` can be used when pulling insecure images. Hence, a cumbersome configuration change is required to use `host.docker.internal`.

**Expected Behavior**
One workaround idea would be to create a minimal Docker container that would act as the proxy. This container would be started by Earthly and would enable local or remote `docker pull`.

**Open Questions**
* Should we use a single code path (with the container) for all platforms?

mikejholly:
We can use `socat` as a simple proxy to forward `localhost` to `docker.host.internal`.

Example: 
```
docker run --rm -p 127.0.0.1:50928:50928 alpine/socat:1.7.4.4 tcp-listen:50928,fork,reuseaddr tcp:host.docker.internal:50929
```

PR #3420, opened 2023-10-23
Registry proxy: support container for Mac

mikejholly:
Enables image registry proxy support for Mac via a side-car `socat` container. On Mac, `docker pull` (and friends) will make requests to `localhost`. These requests will be answered by the `socat` container and proxied through to the registry proxy server spawned by the CLI. That server then sends the request to BK (remote or local) via the gRPC tunnel. 

Depends on: https://github.com/earthly/buildkit/pull/30

https://github.com/earthly/earthly/issues/3412

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -873,7 +873,3 @@ check-broken-links:
     ELSE
         RUN --no-cache echo -e "${GREEN}No Broken Links were found${NOCOLOR}"
     END
-
-image-test:
-    FROM alpine
-    SAVE IMAGE alpine
--- a/Earthfile
+++ b/Earthfile
@@ -873,3 +873,7 @@ check-broken-links:
     ELSE
         RUN --no-cache echo -e "${GREEN}No Broken Links were found${NOCOLOR}"
     END
+
+image-test:
+    FROM alpine
+    SAVE IMAGE alpine
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -5,11 +5,14 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"net/http"
 	"os"
 	"path/filepath"
 	"runtime"
+	"strconv"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/earthly/earthly/buildcontext"
 	"github.com/earthly/earthly/buildcontext/provider"
@@ -174,7 +177,7 @@ func (b *Builder) startRegistryProxy(ctx context.Context, caps apicaps.CapSet) (
 		return nil, false
 	}
 
-	addr := "localhost:0" // Have the OS select a port
+	addr := "127.0.0.1:0" // Have the OS select a port
 
 	lnCfg := net.ListenConfig{}
 	ln, err := lnCfg.Listen(ctx, "tcp", addr)
@@ -186,8 +189,8 @@ func (b *Builder) startRegistryProxy(ctx context.Context, caps apicaps.CapSet) (
 	p := regproxy.NewRegistryProxy(ln, b.s.bkClient.RegistryClient())
 	go p.Serve(ctx)
 
-	port := ln.Addr().(*net.TCPAddr).Port
-	addr = fmt.Sprintf("localhost:%d", port)
+	registryPort := ln.Addr().(*net.TCPAddr).Port
+	addr = fmt.Sprintf("localhost:%d", registryPort)
 	b.opt.Console.VerbosePrintf("Starting local registry proxy: %s", addr)
 
 	doneCh := make(chan struct{})
@@ -204,35 +207,25 @@ func (b *Builder) startRegistryProxy(ctx context.Context, caps apicaps.CapSet) (
 	isDarwin := runtime.GOOS == "darwin"
 
 	if isDarwin {
-		b.opt.Console.VerbosePrintf("Starting support container for registry proxy")
-		addr = strings.ReplaceAll(addr, "localhost", "docker.host.internal")
-		err = b.opt.ContainerFrontend.ContainerRun(ctx, containerutil.ContainerRun{
-			NameOrID: "darwin-registry-proxy",
-			ImageRef: "alpine/socat:1.7.4.4",
-			Ports: []containerutil.Port{
-				{
-					IP:            "127.0.0.1",
-					HostPort:      port,
-					ContainerPort: port,
-					Protocol:      containerutil.ProtocolTCP,
-				},
-			},
-			ContainerArgs: []string{
-				fmt.Sprintf("tcp-listen:%d,fork,reuseaddr", port),
-				fmt.Sprintf("tcp:host.docker.internal:%d", port),
-			},
-		})
+		darwinProxyPort, err := b.startDarwinProxyHelper(ctx, registryPort)
 		if err != nil {
-			b.opt.Console.Warnf("Failed to start support container for registry proxy: %v", err)
+			b.opt.Console.Warnf("Darwin registry container: %v", err)
+			err := b.stopDarwinProxyHelper(ctx) // Clean up dangling container on failure.
+			if err != nil {
+				b.opt.Console.Warnf("Failed to stop registry proxy support container: %v", err)
+			}
 			return nil, false
 		}
+		b.opt.Console.VerbosePrintf("Started support container for registry proxy on port %d", darwinProxyPort)
+		addr = fmt.Sprintf("host.docker.internal:%d", darwinProxyPort)
 	}
 
 	b.opt.LocalRegistryAddr = addr
 
 	closeFn := func() {
+		fmt.Println("Closing registry proxy")
 		if isDarwin {
-			err = b.opt.ContainerFrontend.ContainerRemove(ctx, true, "darwin-registry-proxy")
+			err := b.stopDarwinProxyHelper(ctx)
 			if err != nil {
 				b.opt.Console.Warnf("Failed to stop registry proxy support container: %v", err)
 			}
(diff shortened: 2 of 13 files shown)


builder/builder.go
```diff
@@ -199,15 +204,143 @@ func (b *Builder) startRegistryProxy(ctx context.Context, caps apicaps.CapSet) (
...
+
+	return assignedPort, nil
+}
+
+func (b *Builder) stopDarwinProxyHelper(ctx context.Context) error {
+	err := b.opt.ContainerFrontend.ContainerRemove(ctx, true, darwinProxyContainerName)
+	if err != nil {
+		return errors.Wrap(err, "failed to stop support container")
+	}
+	return nil
```
vladaionescu, line 343: It's probably worth moving all this in its own package. Builder has gotten pretty big and is normally concerned with other things - like putting together the build function and interpreting the build results.

builder/builder.go
```diff
@@ -199,15 +204,143 @@ func (b *Builder) startRegistryProxy(ctx context.Context, caps apicaps.CapSet) (
...
+
+const darwinProxyContainerName = "darwin-registry-proxy"
+
+// startDarwinProxyHelper: Since Docker Desktop (Mac) containers run in a VM, a
+// special host name, host.docker.internal, is made available to access the
+// host's network. Docker can only pull insecurely from localhost, so we use a
+// socat container to proxy localhost:<port> request back out to the local
+// registry proxy created above.
+func (b *Builder) startDarwinProxyHelper(ctx context.Context, registryPort int) (int, error) {
+	infoMap, err := b.opt.ContainerFrontend.ContainerInfo(ctx, darwinProxyContainerName)
```
vladaionescu, line 251: This can present issues when running Earthly in parallel on the same system, right? We might need to come up with a unique name here.

builder/builder.go
```diff
@@ -199,15 +204,143 @@ func (b *Builder) startRegistryProxy(ctx context.Context, caps apicaps.CapSet) (
...
+			return nil, false
+		}
+		b.opt.Console.VerbosePrintf("Started support container for registry proxy on port %d", darwinProxyPort)
+		addr = fmt.Sprintf("127.0.0.1:%d", darwinProxyPort)
+	}
+
 	b.opt.LocalRegistryAddr = addr
 
-	return func() {
+	closeFn := func() {
```
vladaionescu, line 225: Just a question: is this function passed along to the cleanup collection somewhere? It might be useful to do, to help shut things down cleanly in case of cancellation.

mikejholly: At the moment it's called when the [build function exits](https://github.com/earthly/earthly/blob/mh%2Fregistry-proxy-mac/builder/builder.go#L372).

builder/builder.go
mikejholly: Good point. I'll use a dynamic name.

builder/builder.go
mikejholly: Sure, that makes sense.

builder/builder.go
vladaionescu: Ok, then that is probably good enough!

builder/builder.go
```diff
@@ -199,15 +204,142 @@ func (b *Builder) startRegistryProxy(ctx context.Context, caps apicaps.CapSet) (
...
+	if exists {
+		err := b.opt.ContainerFrontend.ContainerRemove(ctx, true, darwinProxyContainerName)
+		if err != nil {
+			return 0, errors.Wrap(err, "failed to remove existing container")
+		}
+	}
+
+	runCfg := containerutil.ContainerRun{
+		NameOrID: darwinProxyContainerName,
+		ImageRef: "alpine/socat:1.7.4.4",
```
alexcb, line 269: should this tag be override-able via a config value? I could see a case where a user has to pin to an older version of earthly; however something occurs to this specific tag and it gets removed from dockerhub. This would mean the user would have to pull this image from some other registry, or fake this tag locally to point to a different image.

builder/builder.go
```diff
@@ -199,15 +204,142 @@ func (b *Builder) startRegistryProxy(ctx context.Context, caps apicaps.CapSet) (
...
+	if len(info.Ports) == 0 {
+		return 0, errors.New("container port not set")
+	}
+
+	parts := strings.Split(info.Ports[0], ":")
+	if len(parts) != 3 {
+		return 0, errors.Errorf("invalid container port format: %s", info.Ports[0])
+	}
+
+	assignedPort, err := strconv.Atoi(parts[1])
```
alexcb, line 308: It's unlikely, but we should check len(parts) > 1 before indexing here.

builder/builder.go
```diff
@@ -199,15 +204,142 @@ func (b *Builder) startRegistryProxy(ctx context.Context, caps apicaps.CapSet) (
...
+	if err != nil {
+		return 0, errors.Wrap(err, "failed to parse container port")
+	}
+
+	ctx, cancel := context.WithTimeout(ctx, 10*time.Second)
```
alexcb, line 313: this should be a config value, for cases where there's an overloaded system (like a macstadium instance), and the user wants a greater timeout value.

builder/builder.go
```diff
@@ -199,15 +204,142 @@ func (b *Builder) startRegistryProxy(ctx context.Context, caps apicaps.CapSet) (
...
+	defer cancel()
+
+	// Wait for the proxy chain to resolve to the BK registry. The /v2/ path
+	// will return a 200 when ready.
+	for {
+		res, err := http.Get(fmt.Sprintf("http://localhost:%d/v2/", assignedPort))
```
alexcb, line 319: is it ok to use `localhost` here? i.e.. Is it ok to have it resolve to IPv6 instead of IPv4?

builder/builder.go
mikejholly: There's an explicit length check on L304.

util/containerutil/shell_shared.go
```diff
@@ -52,23 +74,7 @@ func (sf *shellFrontend) ContainerInfo(ctx context.Context, namesOrIDs ...string
 	}
 
 	// Anonymous struct to just pick out what we need
```
alexcb, line 76: this comment is out of date now.

builder/builder.go
alexcb: sorry, my bad I clearly miss read that as something else :facepalm:

mikejholly: No worries.

builder/builder.go
mikejholly: Updated to use `127.0.0.1` in all cases.

Earthfile
```diff
@@ -873,3 +873,7 @@ check-broken-links:
     ELSE
         RUN --no-cache echo -e "${GREEN}No Broken Links were found${NOCOLOR}"
     END
+
+image-test:
```
mikejholly, line 877: Will remove before merge.

vladaionescu (commented):
This looks good overall!

Some QA questions:
* Have you tested this with podman on Darwin? At the very least, we should fall back to the old tar-based output if this doesn't work for that.
* Have you tested this with a satellite from Darwin?

regproxy/controller.go
```diff
@@ -0,0 +1,227 @@
...
+	return containerPort, nil
+}
+
+func (c *Controller) stopOldDarwinProxies(ctx context.Context) error {
+	containers, err := c.containerFrontend.ContainerList(ctx)
+	if err != nil {
+		return err
+	}
+	for _, container := range containers {
+		if strings.HasPrefix(container.Name, darwinContainerPrefix) &&
```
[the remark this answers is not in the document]
mikejholly: That might be a bit better. The only issue is that labels aren't included in `ps` output and `ps --format json` output differs between Podman & Docker. I don't love the string parsing here so perhaps it should just use `ps` to find the IDs and `inspect` to load all values. Might be a bit slower...

config/config.go
```diff
@@ -18,6 +19,12 @@ const (
 	// DefaultLocalRegistryPort is the default user-facing port for the local registry used for exports.
 	DefaultLocalRegistryPort = 8371
 
+	// DefaultDarwinProxyImageTag is the alpine/socat tag used for the Docker Desktop registry proxy on Darwin.
+	DefaultDarwinProxyImageTag = "1.7.4.4"
```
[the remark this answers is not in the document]
mikejholly: I considered that but chose not to as the code requires a valid `socat` container. It's not a big concern, but perhaps it's a little less brittle to keep the image fixed. Thoughts?

regproxy/controller.go
```diff
@@ -0,0 +1,227 @@
...
+	addr := "127.0.0.1:0"
+
+	ln, err := (&net.ListenConfig{}).Listen(ctx, "tcp", addr)
+	if err != nil {
+		return 0, errors.Wrap(err, "failed to listen on open port")
+	}
+	defer ln.Close() // Immediately close the listener
+
+	return ln.Addr().(*net.TCPAddr).Port, nil
+}
```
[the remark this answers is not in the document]
mikejholly: I noticed that as well. We can use a file based lock to prevent the race across `earthly` process instances, but that's a bit combersome. Although, It does seem exceedingly unlikely that someone would hit this and the fix would just be a retry. Thoughts?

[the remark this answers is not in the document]
alexcb: specifically binding to port `0` asks the OS to pick an unused port.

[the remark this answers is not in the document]
mikejholly: Right, but 2 `earthly` builds could identify the same free port if started in parallel. A listens and closes, B listens and closes, both attempt to use port for container. Still, quite unlikely.

util/containerutil/shell_shared.go
```diff
@@ -35,6 +60,73 @@ func (sf *shellFrontend) IsAvailable(ctx context.Context) bool {
...
+var listRE = regexp.MustCompile(`\s{2,}`)
+
+func (sf *shellFrontend) ContainerList(ctx context.Context) ([]*ContainerInfo, error) {
+	output, err := sf.commandContextOutput(ctx, "ps")
+	if err != nil {
+		return nil, err
+	}
+	ret := []*ContainerInfo{}
+
+	// The Docker & Podman JSON output format differs, so we parse the standard output here.
```
alexcb, line 72: :smiling_face_with_tear:

mikejholly: https://github.com/earthly/earthly/pull/3420#discussion_r1372141520

Maybe this idea is preferrable?

alexcb: I'm unsure to be honest; just commenting on a lovely constraint you have ended up with; this seems easier than trying to open a PR in podman to fix the json it generates :)

regproxy/controller.go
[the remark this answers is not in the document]
mikejholly: > specifically binding to port `0` asks the OS to pick an unused port.

I was using `0` originaly for the container port, explicitly. But there's an issue with [this code](https://github.com/earthly/earthly/blob/main/util/containerutil/shell_shared.go#L179) on Mac. The empty string differs from `0` in Docker Destop.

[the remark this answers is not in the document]
alexcb: I missed the part that `ln.Close()` is being called.... I understand now.

Is there a way to have `docker run .....` dynamically pick the port, then we call `docker inspect` to figure out what it was assigned?

regproxy/controller.go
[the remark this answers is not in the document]
vladaionescu: SGTM! Prefix is fine

regproxy/controller.go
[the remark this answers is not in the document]
mikejholly: Yes, but see my previous comment: https://github.com/earthly/earthly/pull/3420#discussion_r1372163696

I switched to this method after encountering a difficult to identify bug.

config/config.go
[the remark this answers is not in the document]
vladaionescu: What I'm thinking is that if the author of `alpine/socat` one day decides to take this down altogether, some versions of Earthly will be just broken forever. It's good to have an escape hatch - even if the image referenced has to be of a certain specification.

Another reason is that sometimes people want to host these containers in their own registries to avoid rate limiting. They can do that only if we put the full name in the config.

We have a similar approach for this [git image](https://github.com/earthly/earthly/blob/main/config/config.go#L86) for the same reasons.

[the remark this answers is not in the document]
mikejholly: Ok, sounds good.

mikejholly:
@vladaionescu 
> Some QA questions

I can test with a Satellite in MacStadium but I may need some help with Podman.

mikejholly:
I've also confirmed that it will fallback to Tar transfer if the proxy test request falls (or any other part of the Darwin-specific bits fail).

alexcb (approved):
looks good provided all the macOS testing works out

vladaionescu (approved):
Everything working as far as I can see! 🎉 

This was not an easy one!

PR #3426, opened 2023-10-25
docs: warn git credentials are no longer stored

alexcb:
Warn users that as of v0.7.21; buildkit no longer stores the credentials in the git/config file.

### DIFF
--- a/docs/earthfile/earthfile.md
+++ b/docs/earthfile/earthfile.md
@@ -904,7 +904,17 @@ If you need to perform a full-depth clone of a repository, you can use the follo
 GIT CLONE <git-url> <dest-path>
 WORKDIR <dest-path>
 ARG git_hash=$(git rev-parse HEAD)
+RUN git remote set-url origin <git-url> # only required if using authentication
 RUN git fetch --unshallow
+```
+{% endhint %}
+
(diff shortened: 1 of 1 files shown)


mikejholly (approved):

docs/earthfile/earthfile.md
```diff
@@ -908,6 +908,15 @@ RUN git fetch --unshallow
...
 {% endhint %}
 
+{% hint style='warning' %}
+As of Earthly v0.7.21, git credentials are no longer stored in the `.git/config` file; this includes the username.
+This means any ssh-based or https-based fetches or pushes will no longer work unless you restore the configured url,
+which can be done with:
+```Dockerfile
+RUN git remote set-url origin <git-url>
+```
+{% endhint %}
```
vladaionescu, line 918: Can you also update the example above this hint? It includes a `git fetch --unshallow`, which no longer works as is.

alexcb: i was on the fence about adding this since it's only required if you're using authentication, but I added a comment for that.

vladaionescu: Ah ok :+1:

vladaionescu (approved):

PR #3432, opened 2023-10-27
Update docs references from earthly/dind:alpine to earthly/dind:alpine-3.18-docker-23.0.6-r4

idodod:


alexcb (commented):
thanks for updating these.

I think we should encourage users to pin to the more explicit tags (i.e. the ones that contain the docker version) IMO, to prevent surprises from upstream changes.

According to our docs at https://hub.docker.com/r/earthly/dind, the latest alpine version is `alpine-3.18-docker-23.0.6-r4`.

alexcb (approved):

PR #3437, opened 2023-10-27
Improve error outputs

idodod:
* Don't display  `_known` target as prefix for an error when the target and command ids are not populated
* Identify (some) param related errors (e.g target is missing) and use a dedicated logstream failure type (new)
* Fix for Interpreter errors that were not propagated properly over gRPC
* Display errors with a “Error: “ prefix more consistently (some errors were using it, some were not)
* Remove redundant verbose messages that were not different than errors that were already being printed
* Call SetFatalError on non build commands
* Identify and strip gRPC errors prefix using the underlying error type instead of using regex
* Support sending formatted strings with args in SetFatalError


cmd/earthly/app/run.go
```diff
@@ -304,18 +303,16 @@ func (app *EarthlyApp) run(ctx context.Context, args []string) int {
...
-			return 1
-		default:
-			if app.BaseCLI.CommandName() == "build" {
-				app.BaseCLI.Console().VerboseWarnf("Error: %v\n", err)
-				app.BaseCLI.Logbus().Run().SetFatalError(time.Now(), "", "", logstream.FailureType_FAILURE_TYPE_OTHER, err.Error())
+			app.BaseCLI.Console().Printf("Identified interperter error")
+			if ie.TargetID == "" {
+				app.BaseCLI.Logbus().Run().SetGenericFatalError(time.Now(), logstream.FailureType_FAILURE_TYPE_SYNTAX, errorWithPrefix(ie.Error()))
 			} else {
-				app.BaseCLI.Logbus().Run().SkipFatalError()
```
idodod, line 315: Hey @mikejholly ,
i paired with @vladaionescu on this part and we decided to remove `SkipFatalError()` and use `SetFatalError()`.
I was wondering if you can share why the former was needed in the first place. I'd like to make sure I didn't miss anything with this change.

util/params/paramerror.go
vladaionescu, line 1: :+1:

cmd/earthly/subcmd/build_cmd.go
```diff
@@ -132,22 +132,22 @@ func (a *Build) Action(cliCtx *cli.Context) error {
 		a.cli.Flags().Strict = true
 
 		if a.cli.Flags().InteractiveDebugging {
-			return errors.New("unable to use --ci flag in combination with --interactive flag")
+			return params.Errorf("unable to use --ci flag in combination with --interactive flag")
```
vladaionescu, line 135: It seems that we only cover the build command here. Should we now perhaps change all other commands too for consistency? That might be a bigger change.

cmd/earthly/app/run.go
mikejholly: The move to Logstream led to some errors being printed twice. I wanted to remove one, but that led to `No SetFatalError called appropriately` being called via the `defer`. Hence, the new "skip" method.

Example: 
```
            _unknown *failed* | build target: build main: failed to solve: Earthfile line 0:0 target asdfasdfasdf not found
            _unknown *failed* | in		github.com/earthly/earthly:mh/enable-logstream-no-upload+asdfasdfasdf
Error: build target: build main: failed to solve: Earthfile line 0:0 target asdfasdfasdf not found
in		github.com/earthly/earthly:mh/enable-logstream-no-upload+asdfasdfasdf
```

vladaionescu (commented):
Looks good overall! My main comment is that we should add the `Error: ` prefix in the formatter downstream, rather than have to repeat it in every error message.

cmd/earthly/app/run.go
```diff
@@ -162,9 +164,9 @@ func (app *EarthlyApp) run(ctx context.Context, args []string) int {
 
 	defer func() {
 		// Just in case this is forgotten somewhere else.
-		app.BaseCLI.Logbus().Run().SetFatalError(
-			time.Now(), "", "", logstream.FailureType_FAILURE_TYPE_OTHER,
-			"No SetFatalError called appropriately. This should never happen.")
+		app.BaseCLI.Logbus().Run().SetGenericFatalError(
+			time.Now(), logstream.FailureType_FAILURE_TYPE_OTHER,
+			"Error: No SetFatalError called appropriately. This should never happen.")
```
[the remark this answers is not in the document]
idodod: Honestly Idk that it's necessary at all, but I believe errors originating from ufave/cli has this prefix so I was looking to standardize that, but I'm ok if you think it's unnecessary or that it should be moved to the formatter.

cmd/earthly/app/run.go
idodod: ok, so sounds like this change should be ok?
I instead removed the call to `app.BaseCLI.Console().Warnf("Error: %v\n", err)`.

BTW, this error would now look like this:
```
View logs at https://cloud.earthly.dev/builds/28ee7b18-8116-4362-88d1-7d9bcf2ad34e
Error: /var/folders/v7/4fy0kdvj4hd_kvgqctg0_qwc0000gn/T/earthly-git870944423/Earthfile line 0:0 target asdfasdfasdf not found
in              github.com/earthly/earthly+asdfasdfasdf

```

cmd/earthly/subcmd/build_cmd.go
idodod: yes, I agree, I decided to make that change in a separate PR.
However, the difference for the user wouldn't be substantial anyway - 
Errors would be printed the same (or very similar), we just wouldn't be sending the new logstream failure enum type and wrapped messages would show the full error stack.

cmd/earthly/app/run.go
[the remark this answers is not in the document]
vladaionescu: I think it's better via the formatter. That way the error string can usually be just what comes out of `err.Error()`, and if we keep the Golang rule of starting the error text in lowercase, then it fits well with a `Error: ` prefix when showing to the user.

cmd/earthly/app/run.go
mikejholly: LGTM :+1:

vladaionescu (approved):

alexcb (approved):

PR #3445, opened 2023-10-31
Update the dind alpine version in tests and in code hint

idodod:


Earthfile
```diff
@@ -874,3 +880,10 @@ check-broken-links:
     ELSE
         RUN --no-cache echo -e "${GREEN}No Broken Links were found${NOCOLOR}"
     END
+
+# BUILD_AND_FROM will issue a FROM and a BUILD commands for the provided target
+BUILD_AND_FROM:
```
alexcb, line 885: most interesting! :)

inputgraph/hash_test.go
```diff
@@ -32,7 +32,7 @@ func TestHashTargetWithDocker(t *testing.T) {
 	r.Equal("core", project)
 
 	hex := fmt.Sprintf("%x", hash)
-	r.Equal("e4ccd01c9eeee569917873921a6e5e531c1e4108", hex)
+	r.Equal("8cf1c171937b425cdc926f17eb8ba99aca8d1fff", hex)
```
alexcb, line 35: is this an intentional change? I think @mikejholly fixed these in main earlier today.

alexcb: maybe try rebasing against main?

alexcb (approved):
looks good as along as the `docker-*` tests pass.

inputgraph/hash_test.go
idodod: I think the hash changed again because the FROM image changed

PR #3447, opened 2023-10-31
Auto-skip: support for TRY...FINALLY

mikejholly:
Fleshes out support for `TRY...FINALLY`.

### DIFF
--- a/inputgraph/loader.go
+++ b/inputgraph/loader.go
@@ -363,7 +363,7 @@ func (l *loader) handleIf(ctx context.Context, ifStmt spec.IfStatement) error {
 }
 
 func (l *loader) handleFor(ctx context.Context, forStmt spec.ForStatement) error {
-	l.hashFor(forStmt)
+	l.hashForStatement(forStmt)
 	err := l.loadBlock(ctx, forStmt.Body)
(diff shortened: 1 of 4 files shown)


vladaionescu (approved):

inputgraph/loader.go
```diff
@@ -382,7 +382,27 @@ func (l *loader) handleWait(ctx context.Context, waitStmt spec.WaitStatement) er
 }
 
 func (l *loader) handleTry(ctx context.Context, tryStmt spec.TryStatement) error {
-	return errors.New("try not supported")
+	l.hashTryStatement(tryStmt)
+	for _, stmt := range tryStmt.TryBody {
```
alexcb, line 386: I think you need to include the number of elements in each loop, otherwise how do you distinguish between

```
TRY
RUN echo hello
END
RUN echo world
```

and

```
TRY
RUN echo hello
RUN echo world
END
```

(this example assumes we'll eventually support multiple commands inside a TRY.

mikejholly: Good idea. It's probably worth doing that for other statement sets.

alexcb (approved):

PR #3448, opened 2023-10-31
split up tests-no-qemu-quick

alexcb:
The tests-no-qemu-quick no longer is, split it up into two seperate parts.

### DIFF
--- a/.github/workflows/ci-docker-satellites.yml
+++ b/.github/workflows/ci-docker-satellites.yml
@@ -60,11 +60,25 @@ jobs:
       EARTHLY_ORG: "earthly-technologies"
     secrets: inherit
 
-  docker-tests-no-qemu-quick:
+  docker-tests-no-qemu-quick1:
     needs: build-earthly
     uses: ./.github/workflows/reusable-test.yml
     with:
-      TEST_TARGET: "+test-no-qemu-quick"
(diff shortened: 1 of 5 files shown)


idodod (approved):

.github/workflows/ci-docker-satellites.yml
```diff
@@ -60,11 +60,25 @@ jobs:
       EARTHLY_ORG: "earthly-technologies"
     secrets: inherit
 
-  docker-tests-no-qemu-quick:
+  docker-tests-no-qemu-quick1:
```
mikejholly, line 63: Is `quick` still a suitable name? :stuck_out_tongue:

alexcb: hahah right? we do have a `-quick`, `-normal`, and `-slow` targets, maybe these should all just be renamed to `tests-no-qemu-group{1,2,3,...10}`.

the `-quick` target was simply the first one in our tests file so I think everyone just added new tests to it, to the point it was no longer quick.

mikejholly: Yeah, I wonder if we should group them by domain. Or just split them with `group` as you suggested. By domain may be a challenge as there are so many distinct groups under `./tests`.

PR #3450, opened 2023-11-01
lower the parallelism when using core-test sats

alexcb:


mikejholly:
Is this worth adding? It looks like most of the problem tests still failed and seemed to take a bit longer (one has nearly taken 1h).

.github/actions/stage2-setup/action.yml
```diff
@@ -98,6 +98,8 @@ runs:
         if [ "$SATELLITE_NAME" = "core-test" ]; then
           SATELLITE_NAME="$SATELLITE_NAME-$(( $RANDOM % 4 ))"
           echo "Using core-test satellite: $SATELLITE_NAME"
+          # limit converstion parallelism to prevent overloading the core-test satellite
+          ${{inputs.SUDO}} ${{inputs.BUILT_EARTHLY_PATH}} config global.conversion_parallelism 2
```
vladaionescu, line 102: Couldn't we just limit this on the satellite side?

alexcb: we've tried that too, might make sense to just lower the parallelism again on the satellite size.

alexcb:
lets just try lowering the satellite limit again

PR #3457, opened 2023-11-02
split up tests

alexcb:
Split up the "normal" tests into two groups, and rename the "quick" tests to "group".


alexcb:
This is needed for #3451

vladaionescu (approved):

PR #3459, opened 2023-11-03
show applicable error/warning for expired or soon to be expired auth token

idodod:
In case a valid but expired token is used to login/authenticate, the error message will indicate so for the user.
Example:
`Error: failed executing ping request: auth token expired`
In case the token expiration is 14 days or under, it will display a warning stating when the token will expire, and then continue.
Example:
`Warning: token will expire in 1 day 9 hours`


cloud/auth.go
```diff
@@ -562,6 +572,16 @@ func (c *Client) getSSHCredentials(challenge string, key *agent.Key) (credential
 	return credentials, nil
 }
 
+func (c *Client) maybeWarnAuthExpiration() {
+	if authExpirationWarningDisplayed || c.lastAuthMethodExpiry.IsZero() {
+		return
+	}
+	if diff := c.lastAuthMethodExpiry.Sub(time.Now().UTC()); diff >= 0 && diff <= authExpirationWarningGracePeriod {
```
alexcb, line 579: `diff` doesn't make it clear if it's the duration until or after expiry.

maybe call it something like `remaining` instead?

```suggestion
	if remaining := c.lastAuthMethodExpiry.Sub(time.Now().UTC()); remaining >= 0 && remaining <= authExpirationWarningGracePeriod {
```

alexcb (approved):

cloud/auth.go
```diff
@@ -14,11 +14,19 @@ import (
...
 	"github.com/earthly/earthly/util/fileutil"
+
+	"github.com/hako/durafmt"
 	"github.com/pkg/errors"
 	"golang.org/x/crypto/ssh"
 	"golang.org/x/crypto/ssh/agent"
 )
 
+const (
+	authExpirationWarningGracePeriod = time.Hour * 24 * 14
```
idodod, line 25: @vladaionescu I set the warning to 14 days or under. Are you good with that or should I set it to 30 days or something else?

vladaionescu: 14 days is reasonable!

idodod: thanks for the quick reply :)

PR #3465, opened 2023-11-07
Support FOR..IN variable expansion

mikejholly:
This PR adds support for `FOR...IN` variable expansion (example below). Shell-out is not yet supported.  

```
test-vars:
    ARG vals = \
        foo \
        bar \
        baz \
        bim
    FOR idx IN $vals
        RUN echo "hello $idx"
    END
```

### DIFF
--- a/inputgraph/loader.go
+++ b/inputgraph/loader.go
@@ -364,11 +364,52 @@ func (l *loader) handleIf(ctx context.Context, ifStmt spec.IfStatement) error {
 
 func (l *loader) handleFor(ctx context.Context, forStmt spec.ForStatement) error {
 	l.hashForStatement(forStmt)
-	err := l.loadBlock(ctx, forStmt.Body)
+
+	opts := commandflag.ForOpts{
+		Separators: "\n\t ",
+	}
+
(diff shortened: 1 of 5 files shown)


vladaionescu (approved):

inputgraph/loader.go
```diff
@@ -364,13 +364,54 @@ func (l *loader) handleIf(ctx context.Context, ifStmt spec.IfStatement) error {
 
 func (l *loader) handleFor(ctx context.Context, forStmt spec.ForStatement) error {
 	l.hashForStatement(forStmt)
-	err := l.loadBlock(ctx, forStmt.Body)
+
+	opts := commandflag.ForOpts{
+		Separators: "\n\t ",
```
alexcb, line 369: I don't like how this value is now pre-initialized in two locations,

Maybe create a function in commandflag along the lines of

```
func InitForOptDefauls(*fo *ForOpts) {
  fo.Separators = "\n\t"
}
```

alexcb (approved):
looks good, just one suggestion for de-duplicating magic default values

PR #3466, opened 2023-11-07
Updated docs on caching

vladaionescu:
# Before

Screenshot 2023-11-07 at 5 42 33 PM

The docs on caching were scattered and were generally out of date. Important constructs like cache mounts were deemed as "advanced", and there was no mention of the ability to use global cache mounts. Too much emphasis was assigned to remote caching, and in some places remote caching was even recommended as better than remote runners (not accurate anymore). We were getting tons of negative feedback about remote caching - in reality remote caching is just difficult to get right.

# After

Screenshot 2023-11-07 at 5 43 15 PM

Screenshot 2023-11-07 at 5 48 05 PM

The new structure strongly discourages remote caching in favor of remote runners. It also provides a single location for all cache-related information, to help new users avoid the usual pitfalls.


vladaionescu:
@thtmnisamnstr & @adamgordonbell - these updates should take care of a good chunk of the cache-related feedback we've been getting.

idodod (approved):
🥇

adamgordonbell:
I think this is an improvement for sure. Should we include mention of the best practices section or a reference to the SAVE IMAGE pr I haven't raised yet? 

Even people using remote runners might want to use a registry to force the caching of something and that currently isn't covered. It is an advanced topic though, so I could see leaving it out for now.

adamgordonbell (approved):

vladaionescu:
> Should we include mention of the best practices section or a reference to the SAVE IMAGE pr I haven't raised yet? I can add that in my PR if it makes sense.

I think so - please add that in your PR. I can see a small mention of it in `caching-in-earthfiles.md`, linking to the best practice section.

PR #3471, opened 2023-11-08
Cache key is not intuitive

vladaionescu:
We've been getting a ton of feedback lately that cache mounts are difficult to work with. A source of confusion is that the cache key by default is a digest of target + args. Everyone expects this to be just target.

This PR fixes the issue.

### DIFF
--- a/earthfile2llb/converter.go
+++ b/earthfile2llb/converter.go
@@ -1511,10 +1511,7 @@ func (c *Converter) Cache(ctx context.Context, mountTarget string, opts commandf
 		return err
 	}
 	c.nonSaveCommand()
-	key, err := cacheKeyTargetInput(c.targetInputActiveOnly())
-	if err != nil {
-		return err
-	}
+	key := cacheKey(c.target)
 	cacheID := path.Join("/run/cache", key, path.Clean(mountTarget))
 	if c.ftrs.GlobalCache && opts.ID != "" {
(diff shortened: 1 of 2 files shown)


vladaionescu:
I considered whether we need a feature flag for this. My take is that the old behavior was so poorly understood that nobody expected it to behave that way. I figure it's easier to not have to maintain a feature flag for this.

idodod (approved):

adamgordonbell:
Couldn't there be cases where the things being cached were arg specific ( maybe building things for a specific platform ) where this could cause issues?

vladaionescu:
> Couldn't there be cases where the things being cached were arg specific ( maybe building things for a specific platform ) where this could cause issues?

It could - there's now the `id` option for cache mounts to configure the key of the cache. That ID can be set to be platform-dependent, if the user wants. This merely changes the default value of that ID underneath.

This could be a reason to introduce a feature flag for this change, though. I have a feeling that nobody was expecting this to behave this way, which makes me want to treat this as a bug fix (no feature flag).

vladaionescu:
Adding @alexcb too for a review, for good measure.

alexcb (approved):
I think it's ok -- although I agree with @adamgordonbell that there might be a user out there that will have something platform-specific break -- but they'll be able to configure an ID that includes an ARG.

PR #3473, opened 2023-11-08
Adding SAVE IMAGE tip

adamgordonbell:


vladaionescu:
It might be better to put this on the `docs-0.7` branch. It'll go out straight to production once merged. We usually only put docs related to an unreleaased feature in `main`.

PR #3478, opened 2023-11-08
Auto-skip: handle --no-cache & --push on command line

mikejholly:
Includes the following auto-skip changes:
* Fail with messages when `--no-cache` and `--push` are passed to the CLI with `--auto-skip`
* ~Abort when `--no-cache` is used in a visited `RUN` command~
* Don't add log-sharing link when build is skipped
* Added back prefix for some auto-skip messages (only when build continues)

### DIFF
--- a/inputgraph/loader.go
+++ b/inputgraph/loader.go
@@ -53,20 +53,6 @@ func newLoader(ctx context.Context, opt HashOpt) *loader {
 	}
 }
 
-func (l *loader) handleRun(ctx context.Context, cmd spec.Command) error {
-	opts := commandflag.RunOpts{}
-	_, err := flagutil.ParseArgsCleaned(command.From, &opts, flagutil.GetArgsCopy(cmd))
-	if err != nil {
-		return err
-	}
-
-	if opts.NoCache {
-		return errors.New("--no-cache not supported")
-	}
-
-	return nil
-}
-
 func (l *loader) handleFrom(ctx context.Context, cmd spec.Command) error {
 	opts := commandflag.FromOpts{}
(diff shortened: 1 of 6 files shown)


inputgraph/loader.go
```diff
@@ -51,13 +47,26 @@ func newLoader(ctx context.Context, opt HashOpt) *loader {
...
+func (l *loader) handleRun(ctx context.Context, cmd spec.Command) error {
+	opts := commandflag.RunOpts{}
+	_, err := flagutil.ParseArgsCleaned(command.From, &opts, flagutil.GetArgsCopy(cmd))
+	if err != nil {
+		return err
+	}
+
+	if opts.NoCache {
+		return errors.New("--no-cache not supported")
+	}
```
vladaionescu, line 65: I believe that the behavior we want in this case is to skip executing, even if `--no-cache` is used in `RUN`. Same with `RUN --push`. This is because auto-skip is useful to skip deployment if there were no changes to the app being deployed.

This behavior could also back-fire though if you attempt to do a rollback. So we could also just start with saying it's not supported and change the behavior later once we've figured out the usability.

mikejholly: Ok, so is the idea that `--auto-skip` overrides the other settings? 

What about when they're specified on the command line? Is a warning ok then?

vladaionescu: On the command line, I wonder if it should be an error even because they contradict each other.

mikejholly: Right, the contradiction is what I'm referring to.

vladaionescu (approved):

PR #3480, opened 2023-11-09
Update docs on importing in Earthly

vladaionescu:
* Renamed "Target, artifact and command referencing" to "Importing" - simpler, more obvious what it is about
* Reordered guides in the TOC to put importing and other core Earthly features first
* Redid the cheat sheet / infographic
* Added a gentler introduction to the importing page
* Renamed "project ref" to "Earthfile ref" across the docs - the previous name was confusing because there are also Earthly Cloud projects which are different


vladaionescu:
Preview: https://app.gitbook.com/o/-MB5s2W0v8S4xFU5ELti/s/yi347RZ02JXdgz9BJx5u/~/revisions/ImoLRrMCd0RxlOkSwESZ/docs/guides/importing

alexcb (approved):
I like the `project` to `Earthfile` rename, much simpler!

adamgordonbell (approved):
I think its a good change! 
I wish the image you added could look just like that but be in colored text and HTML, but I assume we are limited with what we can do in gitbook.

vladaionescu:
> I assume we are limited with what we can do in gitbook.

Yeah... that's the issue, unfortunately.

PR #3482, opened 2023-11-10
Logbus: Don't pre-allocate 80k for each command

mikejholly:
While looking into BK's memory usage, I noticed some odd consumption (below) when reviewing Earthly's `pprof` info.

It turns out that the previously used `circbuf` library was [preallocating](https://github.com/armon/circbuf/blob/master/circbuf.go#L27) ~80K for each command. For large builds like ours, this can represent 100s of MBs (or even GBs).

I've added a custom circular buffer implementation that starts with an empty buffer and grows to a maximum size before writing circularly.  This is massively reducing memory usage for some of our tests, like `+test-no-qemu`.


vladaionescu (approved):

alexcb (approved):
nice find! Do you have a screenshot of the pprof after you made this change?

mikejholly:
> nice find! Do you have a screenshot of the pprof after you made this change?

I was going to add one but the memory usage for that code path was so reduced that it's no longer included in the graph.

alexcb:
> > nice find! Do you have a screenshot of the pprof after you made this change?
> 
> I was going to add one but the memory usage for that code path was so reduced that it's no longer included in the graph.

wow! that's amazing!

Issue #3425, opened 2023-10-25
Show a friendlier message when user plan does not allow to launch more satellites


idodod:
**What existing functionality needs improvement?**

When a user tries to launch a satellite without its plan allowing for any more satellites the following error is showin:
```bash
earthly sat launch my-satellite
Launching Satellite "my-satellite" with auto-updates set to run at 02:00 (BST)
This may take a few minutes...
Error: plan only allows 0 satellites in use at one time {reqID: ...}
            _unknown *failed* | failed to create satellite my-satellite: failed launching satellite: rpc error: code = PermissionDenied desc = rpc error: code = PermissionDenied desc = plan only allows 0 satellites in use at one time {reqID: ...}
```

**Expected Behavior**

We can instead show a friendlier message (as a start, don't show `_unknown *failed* |`)
and additionally specify the current plan and the total number of satellites that currently exists and the total allowed

idodod:
fixed in #3494

PR #3494, opened 2023-11-14
Add billing plan info during build or failure to build/launch satellite

idodod:
This PR will add the following:
 - During a build, the user will see the number of build mintues used thus far.
 - If a build command fails due to missing build minutes - it will display proper message to verify or upgrade the plan to get more minutes.
 - If a sat launch command fails due to number of satellites over the max, it will display a proper message to verify/upgrade/remove existing satellites/contact support (fix for https://github.com/earthly/earthly/issues/3425)
 - Also adding a new (experimental) command to view billing plan info (`earthly billing --org <my-org> view`)


cmd/earthly/subcmd/billing_cmds.go
```diff
@@ -0,0 +1,128 @@
...
+	a.cli.Console().Printf(planInfoText(items, 5))
+
+	return nil
+}
+
+func planInfoText(items []*tuple, padding int) string {
+	maxKeyLen := 0
+	for _, item := range items {
+		if len(item.key) > maxKeyLen {
+			maxKeyLen = len(item.key)
```
alexcb, line 103: can you use `text/tabwriter` instead of rolling your own?

We already use `tabwriter` in many places, here's one example: https://github.com/earthly/earthly/blob/72aca6373f4289eaf351f467e8c0202f76a0afe6/cmd/earthly/subcmd/satellite_cmds.go#L359

cmd/earthly/app/run.go
```diff
@@ -259,6 +244,49 @@ func (app *EarthlyApp) run(ctx context.Context, args []string) int {
...
+			matches, _ := stringutil.NamedGroupMatches(msg, maxSatellitesRegex)
+			if len(matches["msg"]) > 0 {
+				msg = matches["msg"][0]
+			}
+			msg = fmt.Sprintf("%s %s", stringutil.Title(billing.Plan().Tier), msg)
+			app.BaseCLI.Console().VerboseWarnf(err.Error())
+			app.BaseCLI.Logbus().Run().SetGenericFatalError(time.Now(), logstream.FailureType_FAILURE_TYPE_OTHER, msg)
+			switch billing.Plan().Tier {
+			case billingpb.BillingPlan_TIER_LIMITED_FREE_TIER:
+				app.BaseCLI.Console().HelpPrintf("Visit your organization settings to verify your account\nfor an option to launch more satellites: %s\nor consider removing one of your existing satellites (`earthly sat rm <satellite-name>`)", billing.GetBillingURL(app.BaseCLI.CIHost(), app.BaseCLI.Flags().OrgName))
```
vladaionescu, line 274: Add billing URL here too?

util/stringutil/enum.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+var titleCaser = cases.Title(language.English)
+var lowerCaser = cases.Lower(language.English)
+
+type caser interface {
+	String(s string) string
+}
+
+type Enum interface {
```
alexcb, line 20: maybe use `fmt.Stringer` instead? https://pkg.go.dev/fmt#Stringer

cmd/earthly/subcmd/billing_cmds.go
```diff
@@ -0,0 +1,128 @@
...
+	return []*cli.Command{
+		{
+			Name:        "billing",
+			Aliases:     []string{"bill"},
+			Description: `*experimental* View Earthly billing info.`,
+			Usage:       `*experimental* View Earthly billing info`,
+			UsageText:   "earthly billing (view)",
+			Flags: []cli.Flag{
+				&cli.StringFlag{
+					Name:        "org",
```
vladaionescu, line 35: This still works with `earthly org select` too, right?

alexcb (commented):
looking good, just a couple minor code-reuse things.

cmd/earthly/app/run.go
```diff
@@ -259,6 +244,49 @@ func (app *EarthlyApp) run(ctx context.Context, args []string) int {
...
+			}
+			return 1
+		case grpcErrOK && grpcErr.Code() == codes.PermissionDenied && requestIDRegex.MatchString(grpcErr.Message()):
+			msg := grpcErr.Message()
+			matches, _ := stringutil.NamedGroupMatches(msg, requestIDRegex)
+			if len(matches["msg"]) > 0 {
+				msg = matches["msg"][0]
+			}
+			app.BaseCLI.Console().VerboseWarnf(err.Error())
+			app.BaseCLI.Logbus().Run().SetGenericFatalError(time.Now(), logstream.FailureType_FAILURE_TYPE_OTHER, msg)
```
vladaionescu, line 288: Just a thought regarding these hints. A lot of this logic seems tightly related to Earthly Cloud. I wonder if it would have been better for the server to format these different messages (and doing the switch-case's over different plans), and the client would just display whatever the server says. It would make it easier to change in the future, and it would have kept coupling low.

Just a thought for the future - no need to change anything in this PR, since this might be a bigger change.

vladaionescu (approved):

cmd/earthly/app/run.go
idodod: I agree.
for example I tried to avoid words like, "upgrade and increase to 10 satellites" because I didn't know what the plan max might be in the future.

I would actually consider decoupling further by exposing APIs that:
1. return the default values for each plan as they are currently set (e.g. how many satellites can you use in a tier plan right now?)
2. and/or APIs to return the exact text by plan. because I don't think it's compute's job (in this case) to return the too much information AND that it should be ok for compute to change its error messages without breaking how errors are displayed in earthly

cmd/earthly/subcmd/billing_cmds.go
idodod: yep!

cmd/earthly/subcmd/billing_cmds.go
idodod: good call, I ran into this when I googled ways to align the text but I dismissed it thinking I have to print to stdout.
Will do.

util/stringutil/enum.go
idodod: So the way this works is by taking into account how an enum might look like.
Actually now that I think about it's more how a protobuf enum would look like.
maybe I should change this to:
```golang
type ProtoEnum interface {
   fmt.Stringer
}
```

Side note: this actually might be a nice protobuf plugin

cmd/earthly/app/run.go
vladaionescu: Oh - we are... I thought you missed one message but I just didn't see it. Ignore me :-)

util/stringutil/enum.go
alexcb: that still feels a bit odd to have it tied to the concept of an `Enum` -- the interface is simply stating that "in order to pass an object to me, it must contain a `String() string` method", nothing about that requirement is related to an Enum.

idodod: That's true, but I'm not sure how to reconcile it with the fact the stringifying takes place according to certain expectations about the String() returned value

alexcb: is it possible to simplify your methods to just:

```
func Title(s fmt.Stringer) string {
	return pretty(titleCaser, s)
}
```

or am I missing something?

idodod: discussed offline - agreed to change the interface to something more specific to an enum

PR #3496, opened 2023-11-15
Auto-skip: further support for shell-out

mikejholly:
Allows targets that use shell-out features to be auto-skipped unless the target name is to be dynamically resolved using the result of a shell-out command.

### DIFF
--- a/inputgraph/loader.go
+++ b/inputgraph/loader.go
@@ -222,18 +222,16 @@ func (l *loader) expandDirs(dirs ...string) ([]string, error) {
 			if err != nil {
 				return nil, errors.Wrap(err, "failed to read dir")
 			}
+			children := []string{}
 			for _, entry := range entries {
-				next := filepath.Join(dir, entry.Name())
-				if entry.IsDir() {
-					found, err := l.expandDirs(next)
-					if err != nil {
-						return nil, err
-					}
-					ret = append(ret, found...)
-				} else {
-					ret = append(ret, next)
-				}
+				child := filepath.Join(dir, entry.Name())
+				children = append(children, child)
 			}
+			found, err := l.expandDirs(children...)
+			if err != nil {
+				return nil, err
+			}
+			ret = append(ret, found...)
 		} else {
 			ret = append(ret, dir)
 		}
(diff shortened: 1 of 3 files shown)


vladaionescu (approved):

tests/autoskip/Earthfile
```diff
@@ -147,6 +148,16 @@ test-push:
     DO --pass-args +RUN_EARTHLY_ARGS --earthfile=simple.earth --target=+simple --should_fail=true --extra_args="--push" --output_contains="push cannot be used"
     DO --pass-args +RUN_EARTHLY_ARGS --earthfile=simple.earth --target=+simple --should_fail=true --extra_args="--push" --output_does_not_contain="target .* has already been run; exiting"
 
+test-shell-out:
+    DO --pass-args +RUN_EARTHLY_ARGS --earthfile=shell-out.earth --target=+shell-out --should_fail=false
```
alexcb, line 152:
```suggestion
    DO --pass-args +RUN_EARTHLY_ARGS --earthfile=shell-out.earth --target=+shell-out
```

alexcb (changes requested):
looking good, but what about this case?

```
test:
  ARG path=$(echo "data")
  WORKDIR /thetest
  COPY $path .
  RUN ls -la
```

this outputs:

```
           auto-skip | unable to calculate hash for +test: failed to stat file: stat $(echo "data"): no such file or directory
```

which seems like a bad user experience (compared to printing out a more clear error such as `dynamic copy path could not be resolved`

mikejholly:
@alexcb Updated. One caveat is that legitimate file names with a `$` in them will break auto-skip. Would checking for `$()` be more reliable?

alexcb:
> @alexcb Updated. One caveat is that legitimate file names with a `$` in them will break auto-skip. Would checking for `$()` be more reliable?

If it doesn't break anything I think it's a good idea to make it more strict (i.e. a `$(` as you suggested).

inputgraph/loader.go
```diff
@@ -119,9 +119,13 @@ func (l *loader) handleCopy(ctx context.Context, cmd spec.Command) error {
 	return nil
 }
 
+func hasShellExpr(s string) bool {
```
alexcb, line 122: nit: call this `mostLikelyHasShellExpre` :laughing:

alexcb (approved):

inputgraph/loader.go
mikejholly: Yeah, it's something of a best effort.  :slightly_smiling_face:

PR #3497, opened 2023-11-15
make sure selected org is used in all applicable commands when org is not explicitly set

idodod:
When using the selected org instead of explicitly setting the org in the command (or via an env var),
the org was not properly sent to analytics server.

This PR adds a new function that reads the org name from the command flag and falls back to the selected org if the former is not set.

In addition this also changes the `earthly registry` command to fall back to the selected org which it did not do before.

### DIFF
--- a/cmd/earthly/app/run.go
+++ b/cmd/earthly/app/run.go
@@ -18,7 +18,6 @@ import (
 
 	"github.com/earthly/earthly/analytics"
 	"github.com/earthly/earthly/buildkitd"
-	"github.com/earthly/earthly/cloud"
 	"github.com/earthly/earthly/cmd/earthly/common"
 	"github.com/earthly/earthly/cmd/earthly/helper"
 	"github.com/earthly/earthly/conslogging"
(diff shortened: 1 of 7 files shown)


cmd/earthly/app/run.go
```diff
@@ -141,26 +140,6 @@ func (app *EarthlyApp) run(ctx context.Context, args []string) int {
 	}()
 	defer app.BaseCLI.ExecuteDeferredFuncs()
 	app.BaseCLI.Logbus().Run().SetStart(time.Now())
-	// Initialize log streaming early if we're passed the organization and
-	// project names as environmental variables. This will allow nearly all
-	// initialization errors to be surfaced to the log streaming service. Access
-	// to this organization and project will be verified when the stream begins.
-
-	if app.BaseCLI.Flags().OrgName != "" && app.BaseCLI.Flags().ProjectName != "" && app.BaseCLI.Cfg() != nil && !app.BaseCLI.Cfg().Global.DisableLogSharing && app.BaseCLI.Flags().LogstreamUpload {
```
idodod, line 149: this is dead code, removing it as discussed on slack.

cmd/earthly/subcmd/build_cmd.go
```diff
@@ -590,15 +590,6 @@ func (a *Build) ActionBuildImp(cliCtx *cli.Context, flagArgs, nonFlagArgs []stri
 	// information to logbusSetup. This function will be called right at the
 	// beginning of the build within earthfile2llb.
 	buildOpts.MainTargetDetailsFunc = func(d earthfile2llb.TargetDetails) error {
-		if a.cli.LogbusSetup().LogStreamerStarted() {
```
idodod, line 593: this is also dead code since the only function that can turn this condition true is `StartLogStreamer` which is only called once in the codebase, a few lines below.

mikejholly (approved):

alexcb (approved):

PR #3498, opened 2023-11-15
Detect build executions from AWS CodeBuild for analytics

idodod:


analytics/analytics.go
```diff
@@ -31,15 +31,16 @@ func DetectCI(isEarthlyCIRunner bool) (string, bool) {
...
+		"GITHUB_WORKFLOW":    "github-actions",
+		"CIRCLECI":           "circle-ci",
+		"JENKINS_HOME":       "jenkins",
+		"BUILDKITE":          "buildkite",
+		"DRONE_BRANCH":       "drone",
+		"TRAVIS":             "travis",
+		"GITLAB_CI":          "gitlab",
+		"EARTHLY_IMAGE":      "earthly-image",
+		"AGENT_WORKDIR":      "jenkins", // https://github.com/jenkinsci/docker-agent/blob/master/11/alpine/Dockerfile#L35
+		"CODEBUILD_BUILD_ID": "aws codebuild",
```
alexcb, line 43: is it safe to use a space in the value? I have no idea, but it might pose problems in some (poorly written) downstream data consumers.
```suggestion
		"CODEBUILD_BUILD_ID": "aws-codebuild",
```

alexcb: maybe we should move this map to a global var, then we could have a unit test check that none of the values have spaces.

alexcb: either that or we yolo, and stick some unicode data here :)

idodod: No strong feelings, but using a space was a mindful decision :) - 
The only other two words ci name is "circle ci" and they use the name as CircleCI, so circle-ci made sense.
I suppose a dash could avoid problems :)

idodod: ha, I guess there is also "github-actions" 😆

alexcb: > I guess there is also "github-actions"

that's exactly why I suggested this -- to keep things consistent.

alexcb (approved):

analytics/analytics.go
idodod: so I guess I was mindful but not mindful enough

vladaionescu (approved):

PR #3500, opened 2023-11-15
Clarify errors when attempting to pass secrets to a build

idodod:
This PR will clarify error messages when there is a problem setting secrets in a build command.
It includes a refactor the the hintError so that hints will be printed more consistently when handling errors at the final of the binary execution.

I didn't want to change how all hints are printed as part of this PR (will take too long atm to confirm how these messages will be displayed) so I created deprecated methods to create hint errors so that other hints will still be displayed in the same way they have so far.

### DIFF
--- a/ast/hint/error.go
+++ b/ast/hint/error.go
@@ -5,12 +5,23 @@ import (
 	"strings"
 )
 
-type hintError struct {
+// TODO: once WrapWithDisplay & WrapfWithDisplay are out of use,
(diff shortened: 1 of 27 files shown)


ast/hint/error.go
```diff
@@ -28,17 +39,34 @@ func (e hintError) Error() string {
 	}
 }
 
-func (e hintError) Unwrap() error {
-	return e.err
+func (e Error) Hint() string {
```
vladaionescu, line 42: Same with the methods. Needs godoc

alexcb (approved):

PR #3504, opened 2023-11-16
Auto-skip: support for COPY --if-exists & FOR IN fix

mikejholly:
This PR adds auto-skip support for `COPY --if-exists` & fixes a bug with `FOR...IN` incorrectly including the index var & `IN`.

### DIFF
--- a/inputgraph/loader.go
+++ b/inputgraph/loader.go
@@ -111,7 +111,8 @@ func (l *loader) handleCopy(ctx context.Context, cmd spec.Command) error {
 
 	srcs := args[:len(args)-1]
 	for _, src := range srcs {
-		if err := l.handleCopySrc(ctx, src, opts.IsDirCopy); err != nil {
+		mustExist := !opts.IfExists
+		if err := l.handleCopySrc(ctx, src, mustExist); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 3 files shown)


inputgraph/loader.go
```diff
@@ -397,8 +404,13 @@ func (l *loader) handleFor(ctx context.Context, forStmt spec.ForStatement) error
...
 
 func flattenForArgs(args []string, seps string) []string {
+	// The first 2 args will be the index variable name and IN so we ignore the
+	// first 2 values. Example: "FOR idx IN vars..." etc.
+	if len(args) < 3 {
+		return nil
+	}
 	var ret []string
-	for _, arg := range args {
+	for _, arg := range args[2:] {
```
alexcb, line 413: does this mean changing `FOR foo IN ...` to `FOR bar IN` has no effect on the hash, or is it hashed elsewhere?

If it doesn't affect the hash, this seems wrong as it could change the behaviour of the command that's being executed.

mikejholly: The command's args are hashed beforehand. https://github.com/earthly/earthly/blob/mh%2Fauto-skip-if-exists/inputgraph/loader.go#L274

alexcb: ah perfect!

alexcb (approved):

PR #3508, opened 2023-11-18
Auto-skip: support for some remote targets

mikejholly:
Adds `BUILD`, `COPY`, `FROM` support for remote targets that use explicit SHAs or tags.

### DIFF
--- a/inputgraph/loader.go
+++ b/inputgraph/loader.go
@@ -70,11 +70,7 @@ func (l *loader) handleFrom(ctx context.Context, cmd spec.Command) error {
 	}
 
 	target, err := domain.ParseTarget(fromTarget)
-	if err != nil {
-		return errors.Wrap(err, "failed to parse target")
-	}
-
-	if target.IsRemote() {
+	if err == nil && target.IsRemote() {
 		if supportedRemoteTarget(target) {
 			l.hasher.HashString(target.StringCanonical())
 			return nil
@@ -104,11 +100,7 @@ func (l *loader) handleBuild(ctx context.Context, cmd spec.Command) error {
 	}
 
 	target, err := domain.ParseTarget(targetName)
-	if err != nil {
-		return errors.Wrap(err, "failed to parse target")
-	}
-
-	if target.IsRemote() {
+	if err == nil && target.IsRemote() {
 		if supportedRemoteTarget(target) {
 			l.hasher.HashString(target.StringCanonical())
 			return nil
--- a/inputgraph/loader.go
+++ b/inputgraph/loader.go
@@ -24,8 +24,8 @@ import (
 )
 
 var (
-	errCannotLoadRemoteTarget  = errors.New("remote targets not supported")
-	errUnsupportedRemoteTarget = errors.New("only remote targets referenced by a complete git SHA or tag are supported")
+	errUnsupportedRemoteTarget = errors.New("only remote targets referenced by a complete Git SHA or tag (e.g. tags/my-tag) are supported")
+	errCannotLoadRemoteTarget  = errors.New("cannot load remote target")
 )
 
 type loader struct {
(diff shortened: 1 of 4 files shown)


inputgraph/loader.go
```diff
@@ -188,6 +224,12 @@ func (l *loader) handleCopySrc(ctx context.Context, src string, mustExist bool)
 	return nil
 }
 
+var sha1RE = regexp.MustCompile("^[0-9a-f]{40}$")
+
+func validRemoteTarget(t domain.Target) bool {
```
vladaionescu, line 229: Minor suggestion: Maybe rename this to `supportedRemoteTarget`? The unsupported ones are still valid.

inputgraph/loader.go
```diff
@@ -22,7 +23,10 @@ import (
 	"github.com/pkg/errors"
 )
 
-var ErrRemoteNotSupported = fmt.Errorf("remote targets not supported")
+var (
+	errRemoteNotSupported  = errors.New("remote targets not supported")
```
vladaionescu, line 27: Is this error still needed?

vladaionescu (commented):
Note that there's also `FROM DOCKERFILE` and `WITH DOCKER --load`. But no need to do these now - just FYI.

inputgraph/loader.go
mikejholly: Not necessarily, but it is used below to guard the non-remote code path against passed-in remote targets.

mikejholly: I clarified the name a bit.

inputgraph/loader.go
```diff
@@ -22,7 +23,10 @@ import (
 	"github.com/pkg/errors"
 )
 
-var ErrRemoteNotSupported = fmt.Errorf("remote targets not supported")
+var (
+	errCannotLoadRemoteTarget  = errors.New("remote targets not supported")
+	errUnsupportedRemoteTarget = errors.New("only remote targets with full SHAs or tags are supported")
```
vladaionescu, line 28: I would clarify a little better that only tags explicitly referenced with the `tags/` prefix are supported.

```suggestion
	errUnsupportedRemoteTarget = errors.New("only remote targets with full SHAs or tags referenced explicitly as 'tags/<my-tag>' are supported")
```

vladaionescu (approved):
One last minor, but important thing...

inputgraph/loader.go
alexcb, line 28:
```suggestion
	errUnsupportedRemoteTarget = errors.New("only remote targets referenced by a complete git SHA or tag are supported")
```

idodod: maybe we can start using hinterror type from now on instead of having to go back later and make these clearer.
This way the user might see in the end:
```
Help: The remote target url is <url>, Please make sure your remote target is referenced in the form of `url/repo:full_sha` or  `url/repo/tags/<my-tag>`
```

vladaionescu: Yeah, that would be nice!

alexcb (approved):

inputgraph/loader.go
mikejholly: Ok, sounds good.

mikejholly: It's not obvious to me where this text belongs. Inline, before the log sharing link, at the very end?
image

idodod: oh, honestly I forgot that this is not a build error but more of a warning for why autoskip would not work, so not sure how this is going to be displayed.

vladaionescu: Since auto-skip is now opt-in, as opposed to default on for pipelines, I wonder if the auto-skip errors should just be fatal, and not just warnings.

I would imagine that most people who use auto-skip will want to make sure that the build's compatibility with auto-skip is enforced.

Maybe we can still offer a flag to ignore auto-skip errors, if someone still wants the previous behavior.

mikejholly: Sure, that seems reasonable.

mikejholly: On second thought, that change will require some test & output changes. Can we merge this PR first? I clarified the error message and added a tag example.

PR #3512, opened 2023-11-20
code-comment reminder for feature flags

alexcb:
A friendly reminder, stemming from the discussion in https://github.com/earthly/earthly/pull/3510#issuecomment-1819561503

### DIFF
--- a/ast/commandflag/flags.go
+++ b/ast/commandflag/flags.go
@@ -4,6 +4,10 @@ import (
 	"time"
 )
 
+// NOTE: Any new flags must be accompanied by the introduction of a new `VERSION` feature flag.
+// This applies to new features which do **not** break backwards compatibility, which is needed
+// to ensure an Earthfile that uses `VERSION 0.7` can be built by **any** of the earthly-v0.7.x binaries.
+
 type IfOpts struct {
 	Privileged bool     `long:"privileged" description:"Enable privileged mode"`
 	WithSSH    bool     `long:"ssh" description:"Make available the SSH agent of the host"`


alexcb:
side note: I think this is already captured under `Each Earthfile should list the current earthly version it depends on using the [VERSION](https://github.com/docs/earthfile#version) command` from our docs: https://docs.earthly.dev/docs/earthfile/features

but perhaps the docs could also be improved in a separate PR?

ast/commandflag/flags.go
```diff
@@ -4,6 +4,10 @@ import (
 	"time"
 )
 
+// NOTE: Any new flags must be accompanied by a `VERSION --feature-flag ...` change.
+// This applies to new features which do **not** break backwards compatibility, which is needed
+// to ensure an Earthfile such that uses `VERSION 0.7` can be built by **any** of the earthly-v0.7.x binaries.
```
mikejholly, line 9: > ensure an Earthfile ~such~ that uses

ast/commandflag/flags.go
```diff
@@ -4,6 +4,10 @@ import (
 	"time"
 )
 
+// NOTE: Any new flags must be accompanied by a `VERSION --feature-flag ...` change.
```
mikejholly, line 7: > Any new flags must be accompanied by a `VERSION --feature-flag ...` change.

What sort of change? Do you mean a changelog entry?

alexcb: does 

```
// NOTE: Any new command flags must be accompanied by the introduction of a new `VERSION` feature flag.`
```

 read better?

This is not related to CHANGELOG entries (but yes those are important too, but seems out of scope of this PR since a CHANGELOG entry should be added for all changes/fixes/etc, we could improve docs for CHANGELOG requirements if you think that's important)

mikejholly (approved):

PR #3514, opened 2023-11-21
Auto-skip: improved shell expression parsing

mikejholly:
A minor improvement over the previous string contents check.

### DIFF
--- a/inputgraph/loader.go
+++ b/inputgraph/loader.go
@@ -1,6 +1,7 @@
 package inputgraph
 
 import (
+	"bufio"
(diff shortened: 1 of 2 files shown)


alexcb (approved):
looks fine; should we also prevent false positives from `echo '$(echo this) is not a process substitution'`?

alexcb:
what about `echo "the date is $(date) :)"` ?

mikejholly:
@alexcb This check is for target names in particular, so I'm not sure that we need to cover all of these cases.

PR #3519, opened 2023-11-21
Introduce a new hint error to display at the end of a failed execution

idodod:
In a recent previous PR the existing hint.Error was refactored in order to allow for hints to be displayed at the end of a failed execution in a more consistent way compared to other help messages.

In this PR that change was slightly reverted so that it would be easier to remove the deprecated hint.Error
(`WrapWithDisplay` is reverted back to `Wrap`)

So instead, this PR introduced a new hint.Error, which if used, will show the hint at the end of an execution like other help messages. In addition the mechanism was improved such that even hints going through gRPC will be properly identified.

Finally, this new hint error is used in cases where a cloud secret could not be found.

### DIFF
--- a/ast/antlrhandler/error_listener.go
+++ b/ast/antlrhandler/error_listener.go
@@ -111,5 +111,5 @@ func (rel *ReturnErrorListener) SyntaxError(recognizer antlr.Recognizer, offendi
 			}
 		}
 	}
-	rel.Errs = append(rel.Errs, hint.WrapWithDisplay(finalErr, hints[0], hints[1:]...))
+	rel.Errs = append(rel.Errs, hint.Wrap(finalErr, hints[0], hints[1:]...))
 }
--- a/ast/hint/error.go
+++ b/ast/hint/error.go
@@ -5,25 +5,13 @@ import (
 	"strings"
 )
 
-// TODO: once WrapWithDisplay & WrapfWithDisplay are out of use,
-// remove displayHints field and let earthly/cmd/earthly/app/run.go use the Hint() value instead of printing it
-// as part of the Error() function
-
-// Error is an error that includes hints to be displayed after the error.
-type Error struct {
+type hintError struct {
 	err   error
 	hints []string
-	// Indicate whether to display the hint as part of the Error() function
-	// or expose the hints via Hint().
-	// Once the deprecated functions are removed, this flag can be removed as well.
-	displayHints bool
 }
 
-// Error returns the error string with or without the provided hints, depending on how the error was created
-func (e Error) Error() string {
-	if !e.displayHints {
-		return e.err.Error()
-	}
+// Error returns the error string
+func (e hintError) Error() string {
 	switch len(e.hints) {
 	case 0:
 		return e.err.Error()
@@ -41,35 +29,14 @@ func (e Error) Error() string {
 	}
 }
 
-// Hint returns all hints in a single string separated by a new line
-func (e Error) Hint() string {
-	if e.displayHints {
-		// if true we can leave this empty so that it is not displayed twice in cmd/earthly/app/run.go
-		return ""
-	}
-	return strings.Join(e.hints, "\n")
-}
-
-// Wrap wraps up an error with hints, to help display hints to a user about what
-// might fix the problem.
+// Wrap wraps up an error with hints, to help display hints to a user about what might fix the problem.
+// Deprecated: use utils/hint.Wrap instead
 func Wrap(err error, firstHint string, extraHints ...string) error {
-	return Error{err: err, hints: append([]string{firstHint}, extraHints...)}
-}
-
-// WrapWithDisplay wraps up an error with hints, to help display hints to a user about what
-// might fix the problem.
-// Deprecated: use Wrap instead after verifying hint is properly displayed
-func WrapWithDisplay(err error, firstHint string, extraHints ...string) error {
-	return Error{err: err, displayHints: true, hints: append([]string{firstHint}, extraHints...)}
+	return hintError{err: err, hints: append([]string{firstHint}, extraHints...)}
 }
 
 // Wrapf wraps an error with a single hint with formatting arguments.
+// Deprecated: use utils/hint.Wrapf instead
 func Wrapf(err error, hintf string, args ...any) error {
 	return Wrap(err, fmt.Sprintf(hintf, args...))
-}
-
-// WrapfWithDisplay wraps an error with a single hint with formatting arguments.
-// Deprecated: use Wrapf instead after verifying hint is properly displayed
-func WrapfWithDisplay(err error, hintf string, args ...any) error {
-	return WrapWithDisplay(err, fmt.Sprintf(hintf, args...))
 }
--- a/cmd/earthly/app/run.go
+++ b/cmd/earthly/app/run.go
@@ -179,10 +179,10 @@ func (app *EarthlyApp) run(ctx context.Context, args []string) int {
 		}
 
 		grpcErr, grpcErrOK := grpcerrors.AsGRPCStatus(err)
+		hintErr, hintErrOK := getHintErr(err, grpcErr)
 		var paramsErr *params.Error
-		var hintErr *hint.Error
 		switch {
-		case getHintErr(err, grpcErr, &hintErr):
+		case hintErrOK:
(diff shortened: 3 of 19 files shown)


ast/antlrhandler/error_listener.go
```diff
@@ -111,5 +111,5 @@ func (rel *ReturnErrorListener) SyntaxError(recognizer antlr.Recognizer, offendi
 			}
 		}
 	}
-	rel.Errs = append(rel.Errs, hint.WrapWithDisplay(finalErr, hints[0], hints[1:]...))
+	rel.Errs = append(rel.Errs, hint.Wrap(finalErr, hints[0], hints[1:]...))
```
idodod, line 114: example of a revert - behavior is not changed

ast/hint/error.go
idodod, line 1: this file is changed back to what it was a few PRs ago, but adding comments deprecating the exported functions.

cmd/earthly/app/run.go
```diff
@@ -179,13 +180,11 @@ func (app *EarthlyApp) run(ctx context.Context, args []string) int {
...
 		var paramsErr *params.Error
-		var hintErr hint.Error
+		var hintErr *hint.Error
 		switch {
-		case errors.As(err, &hintErr):
-			app.BaseCLI.Logbus().Run().SetGenericFatalError(time.Now(), logstream.FailureType_FAILURE_TYPE_OTHER, hintErr.Error())
-			if hintErr.Hint() != "" {
-				app.BaseCLI.Console().HelpPrintf(hintErr.Hint())
-			}
+		case getHintErr(err, grpcErr, &hintErr):
```
idodod, line 185: the new helper function will also identify a hintError wrapped in a gRPC error

logbus/solvermon/vertexmon.go
```diff
@@ -73,6 +74,10 @@ func (vm *vertexMonitor) parseError() {
 		internalStr = " internal"
 	}
 	switch {
+	case reHint.MatchString(errString):
```
idodod, line 77: this will remove the Hint string from the error so it is only displayed in the help message at the end and not as part of the error message. This only applies to the new hint.Error and not the deprecated one

util/hint/hinterror.go
```diff
@@ -0,0 +1,74 @@
...
+	"errors"
+	"fmt"
+	"regexp"
+	"strings"
+
+	"github.com/earthly/earthly/util/stringutil"
+)
+
+// note this regex should be updated in case the error format changes in Error
+var regex = regexp.MustCompile(`(?P<error>.+?):Hint: (?P<hint>(?s).+)`)
```
mikejholly, line 13: Nit: `regex` is a bit vague. Should the name be clarified?

mikejholly (approved):
Nice! This is exactly what I was looking for with some recent work.

cmd/earthly/app/run.go
```diff
@@ -384,3 +383,24 @@ func (app *EarthlyApp) printCrashLogs(ctx context.Context) {
 func errorWithPrefix(err string) string {
 	return fmt.Sprintf("Error: %s", err)
 }
+
+func getHintErr(err error, grpcError *status.Status, target **hint.Error) bool {
```
alexcb, line 387: the double pointer here seems complicated; is there a reason we can't always return a new `hint.Error` instead?

e.g.

```
func getHintErr(err error, grpcError *status.Status) (hint.Error, bool) {
```

or

```
func getHintErr(err error, grpcError *status.Status) (*hint.Error, bool) {
```

?

alexcb: or maybe even just

```
func getHintErr(err error) (hint.Error, bool) {
```

and have it call `grpcErr, grpcErrOK := grpcerrors.AsGRPCStatus(err)` inside the function?

idodod: Yeah, it's so I can use it in the switch case condition (where it's called.
I could refactor and call it out of the condition.

side note: I was thinking of refactoring this whole switch case to make it much easier to maintain/read (hopefully), 
that should make these kind of considerations go away.

idodod: > and have it call grpcErr, grpcErrOK := grpcerrors.AsGRPCStatus(err) inside the function?

This is already called outside the function so I didn't want to call it twice.
This is also something that should be improved with the refactor I mentioned.

logbus/solvermon/vertexmon.go
```diff
@@ -73,6 +74,10 @@ func (vm *vertexMonitor) parseError() {
...
+		matches, _ := stringutil.NamedGroupMatches(errString, reHint)
+		errString = matches["msg"][0]
```
idodod, line 79: @alexcb , @mikejholly 
Any idea why the linter is throwing a warning for this line:
```
logbus/solvermon/vertexmon.go:79:3: ineffectual assignment to errString (ineffassign)
```

mikejholly: It's probably not smart enough to detect the `fallthrough` below. I think these can be ignored using a special comment.

alexcb: is `errString` being used anywhere after? e.g.

```
func() int {
  x := 1
  y := x * 2
  x = 10  /// this would be an example of a meaningless assignment
  return y
}

Comment (idodod) [reply logbus/solvermon/vertexmon.go:79]: right, but it seems clear that it is (unless I'm missing something)

File: util/hint/hinterror.go (hunk 4 above)
Comment (idodod) [reply util/hint/hinterror.go]: done

File: cmd/earthly/app/run.go (hunk 5 above)
Comment (idodod) [reply cmd/earthly/app/run.go]: done

File: logbus/solvermon/vertexmon.go (hunk 6 above)
Comment (idodod) [reply logbus/solvermon/vertexmon.go]: added a comment to ignore

**Review by mikejholly — APPROVED** [review]

File: logbus/solvermon/vertexmon.go (hunk 6 above)
Comment (alexcb) [reply logbus/solvermon/vertexmon.go]: I think this illustrates what's going on:

```
func main() {
	s := "a test"
	switch {
	case strings.Contains(s, "a"):
		s = "greetings"
		fallthrough
	case strings.Contains(s, "o"):
		fmt.Printf("this was hit; %s\n", s)
	}
}
```

this will still print `this was hit; greetings` since the `fallthrough` skips evaluating `strings.Contains(s, "o")`, but rather jumps to the block of the switch and executes the Printf.


in your specific case it turns out to be executing:

```
		matches, _ := stringutil.NamedGroupMatches(errString, reHint)
		errString = matches["msg"][0] //nolint:ineffassign,staticcheck
		vm.isCanceled = true
		vm.errorStr = "WARN: Canceled"
		return
```

which doesn't make use of `errString` after it's assigned to.

I think you should be able to delete this line rather than disable the linter.

Comment (mikejholly) [reply logbus/solvermon/vertexmon.go]: Oh, that's interesting. I didn't know that `fallthrough` skips conditionals. 

https://go.dev/play/p/e1lldiHKKP2

File: cmd/earthly/app/run.go [hunk 7]
```diff
@@ -384,3 +383,15 @@ func (app *EarthlyApp) printCrashLogs(ctx context.Context) {
...
 	return fmt.Sprintf("Error: %s", err)
 }
+
+func getHintErr(err error, grpcError *status.Status) (*hint.Error, bool) {
+	if res := new(hint.Error); errors.As(err, &res) {
+		return res, true
+	}
+	if grpcError != nil {
+		if res, ok := hint.FromError(errors.New(grpcError.Message())); ok {
+			return res, ok
```
alexcb, line 393: nit: since ok can only ever be true here, I would just return `true` to reduce cognitive load on future readers.
```suggestion
			return res, true
```

logbus/solvermon/vertexmon.go
alexcb: yeah.... this could very well be a bug if @idodod is expecting the rest of the switch cases to be evaluated here.

If that is the case, then this logic should be done before the switch statement.

idodod: yeah, exactly, I was expecting for the next condition to be checked. Thanks for clarifying. 

> If that is the case, then this logic should be done before the switch statement.

I will do just that. Thanks!

idodod: ok, fixed! I learned something today!

cmd/earthly/app/run.go
idodod: good point!
I realized I can just return from the function as is

alexcb (approved):

PR #3520, opened 2023-11-21
Support for self-hosted satellites

brandonSc:
Allows typical satellite commands and builds to work with new self-hosted satellites. 

Note that currently TLS is not supported, and must be disabled via `~/.earthly/config` for the connection to work. This is temporary and will require a follow-up PR once the backend supports TLS.


alexcb (commented):
just commenting on this to prevent github notification spamming us (since this was later marked as a draft); feel free to request a review once it's ready.

cmd/earthly/subcmd/build_cmd.go
```diff
@@ -923,7 +923,13 @@ func (a *Build) logShareLink(ctx context.Context, cloudClient *cloud.Client, tar
 	return logstreamURL, true, printLinkFn
 }
 
-func (a *Build) maybePrintBuildMinutesInfo(orgName string) {
+func (a *Build) maybePrintBuildMinutesInfo() {
+	orgName := a.cli.OrgName()
+	settings := a.cli.Flags().BuildkitdSettings
+	if settings.SatelliteName == "" || settings.SatelliteIsSelfHosted {
```
idodod, line 929: can we use [IsUsingSatellite](https://github.com/earthly/earthly/blob/2928bb846c10bb86aad04fcd0003d89782072690/cmd/earthly/subcmd/cli.go#L35) to make is more consistent?

cmd/earthly/subcmd/build_cmd.go
```diff
@@ -926,10 +926,10 @@ func (a *Build) logShareLink(ctx context.Context, cloudClient *cloud.Client, tar
 	return logstreamURL, true, printLinkFn
 }
 
-func (a *Build) maybePrintBuildMinutesInfo() {
+func (a *Build) maybePrintBuildMinutesInfo(cliCtx *cli.Context) {
 	orgName := a.cli.OrgName()
 	settings := a.cli.Flags().BuildkitdSettings
-	if settings.SatelliteName == "" || settings.SatelliteIsSelfHosted {
+	if a.cli.IsUsingSatellite(cliCtx) || settings.SatelliteIsSelfHosted {
```
idodod, line 932:
```suggestion
	if a.!cli.IsUsingSatellite(cliCtx) || settings.SatelliteIsSelfHosted {
```
right?

brandonSc: Yes 🥲

buildkitd/buildkitd.go
```diff
@@ -84,7 +84,11 @@ func NewClient(ctx context.Context, console conslogging.ConsoleLogger, image, co
 		remoteConsole := console
 		if settings.SatelliteName != "" {
 			remoteConsole = console.WithPrefix("satellite")
-			remoteConsole.Printf("Connecting to %s...", settings.SatelliteDisplayName)
+			selfHostedInfo := ""
+			if settings.SatelliteIsSelfHosted {
+				selfHostedInfo = fmt.Sprintf(" (self-hosted at %s)", settings.BuildkitAddress)
+			}
+			remoteConsole.Printf("Connecting to %s...%s", settings.SatelliteDisplayName, selfHostedInfo)
```
alexcb, line 91: this line includes `selfHostedInfo`; however doesn't fall under the `if settings.SatelliteIsSelfHosted {` block; is this intentional?

i.e. could this line be printed in non-self-hosted cases too?

alexcb, line 91: having `...` in the middle seems a bit odd in the case of self-hosted; I would just use an `else` in this case, e.g.

```
			if settings.SatelliteIsSelfHosted {
				remoteConsole.Printf("Connecting to %s (hosted at %s)...", settings.SatelliteDisplayName, selfHostedInfo)
			} else {
				remoteConsole.Printf("Connecting to %s...", settings.SatelliteDisplayName)
			}
```

or use a `strings.Builder` if you need greater control.

buildkitd/settings.go
```diff
@@ -11,33 +11,34 @@ import (
...
+	UseTCP                bool
+	UseTLS                bool
+	VolumeName            string
+	IPTables              string
+	MaxParallelism        int
+	SatelliteName         string `hash:"ignore"`
+	SatelliteDisplayName  string `hash:"ignore"`
+	SatelliteOrgID        string `hash:"ignore"`
+	SatelliteToken        string `hash:"ignore"`
+	SatelliteIsSelfHosted bool   `hash:"ignore"`
```
alexcb, line 38: it would be nice if we could be more consistent with the naming here, in other parts of the code we have `!satellite.IsManaged`. 

standardising on a single term (i.e. `IsManaged` or `IsSelfHosted`) makes it easier to search (and understand) the codebase.

alexcb (commented):
looking good, just a couple minor suggestions.

buildkitd/settings.go
brandonSc: Yeah fair point, I'll make those consistent

alexcb (approved):

PR #3523, opened 2023-11-22
Auto-skip: basic support for evaluating IF conditions

mikejholly:
This PR adds support for basic `IF` statements to auto-skip. This will allow the feature to ignore code paths that will not be visited due to a local `ARG` comparison or other simple check.

Some examples: 

```
[ true ]
[ false ]
[ -n "foo" ]
[ -n "" ]
[ -z "" ]
[ -z "foo" ]
[ "foo" == "bar" ]
[ "foo" = "bar" ]
[ "foo" != "bar" ]
[ "foo" > "bar" ]
[ "foo" < "bar" ]
[ "2" -eq "2" ]
[ "2" -le "2" ]
[ "4" -le "2" ]
[ true ] && [ false ]
[ true ] && [ -z "" ]
[ true ] && [ -z "" ] || [ false ]
[ 2 -gt 1 ] || [ "foo" == "bar" ] && [ false ]
[ 2 -gt 0 ] || [ "foo" == "foo" ] && [ true ]
[ 2 -lt 0 ] && [ "foo" == "foo" ] || [ true ]
```

### DIFF
--- a/inputgraph/loader.go
+++ b/inputgraph/loader.go
@@ -8,6 +8,7 @@ import (
 	"path/filepath"
 	"regexp"
 	"sort"
+	"strconv"
 	"strings"
 
 	"github.com/earthly/earthly/ast/command"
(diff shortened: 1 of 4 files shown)


vladaionescu (approved):
Neat!

PR #3524, opened 2023-11-22
Auto-skip: exit on hashing failure

mikejholly:
This PR tweaks the behavior of auto-skip failures such that hashing problems will force the CLI to exit. 

Before:
```
$ ./build/linux/amd64/earthly --auto-skip +code      
 Init 🚀

           auto-skip | unable to calculate hash for +code: circular dependency detected; +code already called

Streaming logs to https://cloud.earthly.dev/builds/a3d9a6de-08fc-4967-a103-db3e13440bf4

 Build 🔧

golang:1.21-alpine3.17 | --> Load metadata golang:1.21-alpine3.17 linux/amd64
...
```

After:
```
$ ./build/linux/amd64/earthly --auto-skip +code
 Init 🚀

Error: unable to calculate hash for +code: circular dependency detected; +code already called
```

### DIFF
--- a/cmd/earthly/subcmd/build_cmd.go
+++ b/cmd/earthly/subcmd/build_cmd.go
@@ -852,8 +852,7 @@ func (a *Build) initAutoSkip(ctx context.Context, target domain.Target, overridi
 		EarthlyCIRunner: a.cli.Flags().EarthlyCIRunner,
 	})
 	if err != nil {
-		console.Warnf("unable to calculate hash for %s: %s", target.String(), err.Error())
-		return nil, nil, false, nil
+		return nil, nil, false, errors.Wrapf(err, "unable to calculate hash for %s", target)
 	}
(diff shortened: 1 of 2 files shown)


vladaionescu (approved):
Looks good!

CC @idodod - in case he has opinion about the way we display this error. Wondering if "unable to calculate hash" is perhaps too cryptic for new users. Any ideas for an improved error message?

mikejholly:
Good point. Maybe something like "auto-skip identifier" instead of "hash".

idodod:
> Good point. Maybe something like "auto-skip identifier" instead of "hash".

Yeah, identifier sounds much better. Are we planning to release it with some documentation/tochas/examples?
perhaps we can add a hint with a link to that?

mikejholly:
I'll update the lingo in another PR.

PR #3529, opened 2023-11-23
Auto-skip: fixes arg parsing issues

mikejholly:
Fixes 2 issues that were blocking auto-skip on `+test` & `+for-linux`. 
* Problems parsing args with complex quoting/parens
* Arg expansion for ${$VAL#foo}. Example: https://github.com/earthly/earthly/blob/main/Earthfile#L366

The above targets now work!

### DIFF
--- a/util/shell/lex.go
+++ b/util/shell/lex.go
@@ -6,7 +6,6 @@ package shell
 import (
 	"bytes"
 	"fmt"
-	"regexp"
 	"strings"
 	"text/scanner"
 	"unicode"
@@ -523,11 +522,13 @@ func (sw *shellWord) processDollarCurlyBracket() (string, error) {
 		if !found && sw.skipUnsetEnv {
 			return fmt.Sprintf("${%s%%%s}", name, word), nil
 		}
-		re, err := regexp.Compile(word + "$")
-		if err != nil {
-			return "", errors.Wrap(err, "failed to compile regexp")
+		if len(word) > len(newValue) {
+			return newValue, nil
 		}
-		return re.ReplaceAllString(newValue, ""), nil
+		if newValue[len(newValue)-len(word):] == word {
+			return newValue[:len(newValue)-len(word)], nil
+		}
+		return newValue, nil
 	case '#':
 		word, _, err := sw.processStopOn('}')
 		if err != nil {
@@ -549,11 +550,13 @@ func (sw *shellWord) processDollarCurlyBracket() (string, error) {
 		if !found && sw.skipUnsetEnv {
 			return fmt.Sprintf("${%s#%s}", name, word), nil
 		}
-		re, err := regexp.Compile("^" + word)
-		if err != nil {
-			return "", errors.Wrap(err, "failed to compile regexp")
+		if len(word) > len(newValue) {
+			return newValue, nil
 		}
-		return re.ReplaceAllString(newValue, ""), nil
+		if newValue[0:len(word)] == word {
+			return newValue[len(word):], nil
+		}
+		return newValue, nil
 	case ':':
 		// Special ${xx:...} format processing
 		// Yes it allows for recursive $'s in the ... spot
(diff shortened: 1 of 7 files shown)


util/shell/lex.go
```diff
@@ -501,6 +501,35 @@ func (sw *shellWord) processDollarCurlyBracket() (string, error) {
 			return "", errors.Errorf("%s: %s", name, message)
 		}
 		return newValue, nil
+	case '#':
```
alexcb, line 504: If I'm not mistaken, this case appears to be identical to the above `case '?'`, except for the returned error message.

If so, can you combine them rather than duplicate this code?

alexcb (commented):
good find. Can you expand our tests in `lex_test.go` to cover this?

util/shell/lex.go
mikejholly: There's one subtle difference [here](https://github.com/earthly/earthly/pull/3529/files#diff-bc232306ec1e9b099303c5304a7ffd0d3783f0d39aa9430d1c424a1330cd7614R524).

But I can probably create a shared helper function if it's worthwhile.

alexcb: I totally read that as `fmt.Errorf` :sweat_smile: 

I don't think a helper function is needed, couldn't you just use the value of `ch`?

e.g.

`fmt.Sprintf("${%s#%s}", name, word)` -> `fmt.Sprintf("${%s%s%s}", name, ch, word)`

mikejholly: Oh, good point. Will do.

mikejholly: Actually, the previous code was incorrect. I misunderstood. Updated with correct behavior.

util/shell/lex.go
```diff
@@ -501,6 +502,58 @@ func (sw *shellWord) processDollarCurlyBracket() (string, error) {
...
+			found = true
+		case errEnvNotFound:
+			break
+		default:
+			return "", err
+		}
+		if !found && sw.skipUnsetEnv {
+			return fmt.Sprintf("${%s%%%s}", name, word), nil
+		}
+		re, err := regexp.Compile(word + "$")
```
alexcb, line 526: this confuses me -- `word` isn't a regular expression is it? what if a users tries something like `${foo%.}`?

mikejholly: Good point. I'll try with non-regex or do some escaping.

alexcb: I spoke too soon, apparently the bourne shell accepts _some patterns_ for example:

```
$ export foo=abc
$ echo "here with ${foo#[a-z]}"
here with bc
$ echo "here with ${foo#.}"        
here with abc
```

mikejholly: Updated with vanilla string trimming.

util/shell/lex_test.go
```diff
@@ -48,6 +48,26 @@ func TestShellParserMandatoryEnvVars(t *testing.T) {
 	require.Contains(t, err.Error(), "message herex")
 }
 
+func TestShellParserHash(t *testing.T) {
```
alexcb, line 51: can you create a test for the case where the replacement doesn't occur?

e.g. `"${foo#greetings}"` -> `hello world`

util/shell/lex_test.go
```diff
@@ -48,6 +48,26 @@ func TestShellParserMandatoryEnvVars(t *testing.T) {
...
+	shlex := NewLex('\\')
+	setEnvs := map[string]string{"VAR": "hello world"}
+	word := "goodbye${VAR#hello}"
+
+	newWord, err := shlex.ProcessWordWithMap(word, setEnvs, nil)
+	require.NoError(t, err)
+	require.Equal(t, "goodbye world", newWord)
+}
+
+func TestShellParserPercent(t *testing.T) {
```
alexcb, line 61: similar request for testing the case where the replacement doesn't occur.

util/shell/lex_test.go
mikejholly: Sure, sounds good.

alexcb (approved):
thanks for expanding the tests.

mikejholly:
Updated: `PROJECT` is no longer needed when using local DBs.

alexcb (approved):

PR #3533, opened 2023-11-25
Rename `COMMAND` to `FUNCTION`

idodod:
This PR adds another Earthly command called `FUNCTION` which will replace the command `COMMAND`.
For now `COMMAND` can still be used, but it will show a deprecation message explaining to use a feature flag (`--use-function-keyword`) and rename `COMMAND` TO `FUNCTION`.
When the flag is off => `FUNCTION` can't be used.
When the flag is on => `COMMAND` can't be used.

I'd appreciate a review, specifically on the lexer and parser files where I simply copied the existing configuration for `COMMAND` as `FUNCTION`


idelvall:
Changes LGMT but a comment:
I think it is a good practice to run the exact same code as before, when the feature flag is disabled. That way, the feature flag helps not only to guarantee backwards compatibility but also to dogfood the changes without the risk of affecting users.

idodod:
> Changes LGMT but a comment: I think it is a good practice to run the exact same code as before, when the feature flag is disabled. That way, the feature flag helps not only to guarantee backwards compatibility but also to dogfood the changes without the risk of affecting users.

I think I generally followed that practice unless I misunderstood what you mean by "exact same code" or missed anything. Might commenting in specific areas of the code?

alexcb (approved):

vladaionescu (approved):

PR #3536, opened 2023-11-27
Auto-skip: clarify error message language

mikejholly:
The idea to make the error messages a little friendlier came out of https://github.com/earthly/earthly/pull/3524. 

This PR replaces instances of `hash` in auto-skip messages with "auto-skip ID". Other suggestions are welcome!


vladaionescu:
I don't really have a strong opinion yet about this wording. They both seem similarly difficult to understand to a new user. Perhaps I should review auto-skip end-to-end to form a better opinion / come up with better suggestions.

PR #3547, opened 2023-11-28
prevent deprecation message by switching to FUNCTION keyword

idodod:
This was the last place that couldn't be updated until PR #3533 was merged.


alexcb (approved):

mikejholly (approved):
LGTM. Should we update more/all of these?

idodod:
> LGTM. Should we update more/all of these?

They were updated as part of the original PR that introduced the new name. This one couldn't be enabled because the earthly that built earthly didn't recognize the new feature flag :)

PR #3548, opened 2023-11-28
Use the correct field to assign the value of a slice flag

idodod:
I noticed that in some cases using `Value` field instead of `Destination` can cause a bug in how `urfave/cli` parses the flags and arguments.
According to [documentation](https://cli.urfave.org/v2/examples/flags/#placeholder-values) `Destination` is the correct field to use.

### DIFF
--- a/cmd/earthly/subcmd/build_flags.go
+++ b/cmd/earthly/subcmd/build_flags.go
@@ -9,17 +9,17 @@ import (
 func (a *Build) buildFlags() []cli.Flag {
 	return []cli.Flag{
 		&cli.StringSliceFlag{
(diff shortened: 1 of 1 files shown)


idelvall (approved):
IIUC Value and Destination have two different goals. The first sets the default value, the second the destination variable to store the input. 
I don't understand how it was working before

idodod:
> IIUC Value and Destination have two different goals. The first sets the default value, the second the destination variable to store the input. I don't understand how it was working before

I'm not sure, but I think there was a point in time Destination did not exist and Value doubled as both default and destination.

PR #3553, opened 2023-11-30
Better CI detection

vladaionescu:
Add detection for:

* Zuul
* Semaphore
* Azure DevOps Pipelines
* Bitbucket Pipelines

Also, add another catch-all heuristic for git detached head state (typically used in CI).

### DIFF
--- a/analytics/analytics.go
+++ b/analytics/analytics.go
@@ -31,16 +31,20 @@ func DetectCI(isEarthlyCIRunner bool) (string, bool) {
 		return "earthly-ci", true
 	}
 	for k, v := range map[string]string{
-		"GITHUB_WORKFLOW":    "github-actions",
-		"CIRCLECI":           "circle-ci",
-		"JENKINS_HOME":       "jenkins",
-		"BUILDKITE":          "buildkite",
-		"DRONE_BRANCH":       "drone",
-		"TRAVIS":             "travis",
-		"GITLAB_CI":          "gitlab",
-		"EARTHLY_IMAGE":      "earthly-image",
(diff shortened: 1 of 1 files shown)


analytics/analytics.go
```diff
@@ -62,9 +66,31 @@ func DetectCI(isEarthlyCIRunner bool) (string, bool) {
 		}
 	}
 
+	// detached head is often used in CIs. This is another heuristic.
```
alexcb, line 69: I often work in detached head when reviewing PRs, I would take this with a grain of salt.

alexcb (approved):

analytics/analytics.go
vladaionescu: Indeed. Let's see what the data looks like.

analytics/analytics.go
```diff
@@ -31,16 +31,20 @@ func DetectCI(isEarthlyCIRunner bool) (string, bool) {
...
+		"GITHUB_WORKFLOW":        "github-actions",
+		"CIRCLECI":               "circle-ci",
+		"JENKINS_HOME":           "jenkins",
+		"BUILDKITE":              "buildkite",
+		"DRONE_BRANCH":           "drone",
+		"TRAVIS":                 "travis",
+		"GITLAB_CI":              "gitlab",
+		"EARTHLY_IMAGE":          "earthly-image",
+		"AGENT_WORKDIR":          "jenkins", // https://github.com/jenkinsci/docker-agent/blob/master/11/alpine/Dockerfile#L35
+		"zuul_execution_branch":  "zuul",    // https://opendev.org/zuul/zuul/src/branch/master/zuul/executor/server.py#L3259
```
alexcb, line 43: nit: link to an immutable URL, i.e. use a git commit hash rather than master

idodod (approved):

PR #3554, opened 2023-11-30
seperate potential VERSION 0.8 features

alexcb:
Move the `no-use-registry-for-with-docker` and `earthly-ci-runner-arg` features to a seperate section for features that will never be enabled by default; where as the other unreleased features are likely to be promoted to enabled for `VERSION 0.8`.

### DIFF
--- a/features/features.go
+++ b/features/features.go
@@ -19,6 +19,11 @@ import (
 // Features is used to denote which features to flip on or off; this is for use in maintaining
 // backwards compatibility
 type Features struct {
+
+	// Never enabled by default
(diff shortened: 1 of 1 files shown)


vladaionescu (approved):
`--try` will also not go in 0.8, but it is planned for some distant future.

idodod (approved):

alexcb:
> --try will also not go in 0.8, but it is planned for some distant future.

I wasn't entirely sure about that one, but it makes sense since we only _currently_ support `SAVE ARTIFACT`, I suspect we'll need `RUN` before we're willing to take it out of experimental status?

PR #3561, opened 2023-11-30
track data for inconsistent graph state errors

alexcb:
buildkit will now display a list of previously used edges in the actives map when an inconsistent graph state error occurs.


alexcb:
depends on https://github.com/earthly/buildkit/pull/37

idelvall (approved):

PR #3563, opened 2023-12-01
Docs: Featured guide for Rust + Earthly lib

vladaionescu:
[Preview](https://app.gitbook.com/o/-MB5s2W0v8S4xFU5ELti/s/yi347RZ02JXdgz9BJx5u/~/revisions/ukKlu2hiWz3fhUwzf1wK/)

Rust is populated in more depth so far. The rest are placeholders pointing to existing examples.

This also adds a page about Earthly lib.


vladaionescu:
@thtmnisamnstr / @idelvall - thoughts on this format of having all languages in the new TOC menu, and only having a real page for Rust? I'm not decided whether it'll put people off because some of the other examples aren't very polished, or if it'll help us because it showcases the new rust example.

A separate question is how to best develop reference documentation for lib? It seems that the documentation of the existing functions isn't consistent. Rust is the only one more clearly documented, but ideally, we would have that documentation in our official docs too, not just in GitHub.

CC @idodod & @alexcb - in case you have an opinion.

idelvall:
First, I would avoid duplicating content, since the function docs are expected to change at a rapid pace based on feedback. Instead I propose:
- Defining commons structure/format to follow in all libraries
- Using remote references (maybe plus some tweaks) to bring the lib READMEs into these gitbook docs

vladaionescu:
@idelvall - I agree in principle, but we need something fast for now.

idodod:
> @thtmnisamnstr / @idelvall - thoughts on this format of having all languages in the new TOC menu, and only having a real page for Rust? I'm not decided whether it'll put people off because some of the other examples aren't very polished, or if it'll help us because it showcases the new rust example.
> 
> A separate question is how to best develop reference documentation for lib? It seems that the documentation of the existing functions isn't consistent. Rust is the only one more clearly documented, but ideally, we would have that documentation in our official docs too, not just in GitHub.
> 
> CC @idodod & @alexcb - in case you have an opinion.

Maybe we can add an explicit message about other languages (e.g. "Coming Soon")?

Since we don't have that many util functions, I think we could write something about each, but I understand if we have the time to wait for that.

thtmnisamnstr (approved):

idelvall (approved):

docs/lang-guides/node.md
```diff
@@ -0,0 +1,4 @@
+# NodeJS
+
+For a complete Earthfile example on how to use NodeJS in Earthly, visit the [js example directory on GitHub](https://github.com/earthly/earthly/tree/main/examples/js) or the [typescript-node example directory on GitHub](https://github.com/earthly/earthly/tree/main/examples/typescript-node).
+
```
alexcb, line 4: nit: trailing newline, which will cause issues with the `+lint-newline-ending` test, which unfortunately wasn't triggered on this PR.

alexcb (approved):

Issue #3499, opened 2023-11-15
Per-site git options should support a custom `GIT_SSH_COMMAND`


weaversam8:
I encountered an issue cloning a repository with an RSA key in my ssh-agent, because the alpine git image used for cloning doesn't allow rsa keys by default. I fix this on my local machine with a `PubkeyAcceptedKeyTypes +ssh-rsa` in my SSH config for that site.

It would be ideal if you could specify a value for the `GIT_SSH_COMMAND` environment variable, to allow the user to specify custom SSH options like this. It could look something like this:

```
git:
    git.example.com:
        auth: ssh
        user: git
        ssh_command: "ssh -o 'PubKeyAcceptedKeyTypes +ssh-rsa'"
```

and then the value of the `ssh_command` YAML key could be passed into the git image as `GIT_SSH_COMMAND`.

alexcb:
https://github.com/earthly/buildkit/blob/b4cfdf184379a5398e207ef8a8e39c5f8bc0b486/util/gitutil/git_cli.go#L279 is the relevant section under buildkit which would have to be changed; such a contribution would make sense upstream in the official moby buildkit code-base.

For this particular use-case; you should really update your server since ssh-rsa is no longer considered secure :stuck_out_tongue_closed_eyes:

However, PRs would be welcome :smile:

weaversam8:
@alexcb could you explain a bit more about the high level architecture here? I've been working on a PR but am a little bit stumped by the size of the earthly and buildkit codebases together. My main questions:

- What's calling `util/gitutil/git_cli.go` from the Earthly side?
- Are you sure it's not [`client/llb/source.go`][llb] that needs to be changed in Buildkit? If so, what does this Git library do in Earthly?

[llb]: https://github.com/earthly/buildkit/blob/b4cfdf184379a5398e207ef8a8e39c5f8bc0b486/client/llb/source.go#L242

alexcb:
there are some architecture docs under https://github.com/earthly/earthly/tree/main/docs-internals which might help (which also links to the upstream buildkit docs)

I don't think the `earthly` cli calls the `git` binary directly, but rather always submits a request to buildkit by constructing a `llb.Git` instruction.

Unfortunately I can't provide much more help on this. Please reconsider using a secure algorithm for verifying your key signatures; `sha-1` is no longer secure.

weaversam8:
#3567 and earthly/buildkit#38 are now open to close this issue.

alexcb:
[...]

PR #3567, opened 2023-12-01
Support custom ssh_command option

weaversam8:
Details are in the original issue #3499.

- Depends on a buildkit update in earthly/buildkit#38
- Adds `ssh_command` option to the Git site config
- Updates Earthfile to support building a buildkit fork with a repo name other than "buildkit"

This PR will need to be updated once earthly/buildkit#38 is merged.


buildcontext/gitlookup.go
```diff
@@ -615,26 +617,33 @@ func parseGitProtocol(remote string) (string, int) {
...
 	if len(path) > n {
 		subPath = path[n+1:]
 		path = path[:n]
 	}
 	host := path[:strings.IndexByte(path, '/')]
 	gitPath := m.prefix + match[(strings.IndexByte(match, '/')+1):] + m.suffix
 
+	sshCommand = ""
+	if m.sshCommand != "" {
+		sshCommand = m.sshCommand
```
alexcb, line 641: very cool to see this was done in a way that allows for custom commands for particular hosts.

alexcb:
Thanks for the PR! I wonder if we could create a new test for this under `tests/git-ssh-server`?

Maybe you could have the server start on a diffferent port, which we already support under https://github.com/earthly/earthly/blob/1b5ec2314eace4c991ef10c31b59a37b9da5b37a/tests/git-ssh-server/setup/Earthfile#L49

you could make a test that's very similar to https://github.com/earthly/earthly/blob/1b5ec2314eace4c991ef10c31b59a37b9da5b37a/tests/git-ssh-server/Earthfile#L338
which will start the server on port 2244

however instead of configuring the port via 

```
earthly --config \$earthly_config config 'git.\"git.example.com\"' '{\"auth\": \"ssh\", \"user\": \"git\", \"port\": 2244, \"strict_host_key_checking\": false}'
```

you could instead config it using the ssh command directly, maybe something like:

```
earthly --config \$earthly_config config 'git.\"git.example.com\"' '{\"auth\": \"ssh\", \"user\": \"git\", \"ssh_command\": "ssh -F /dev/null -p 2244\", \"strict_host_key_checking\": false}'
```

alexcb:
I opened a draft PR under https://github.com/earthly/earthly/pull/3569 to trigger our tests (since they require secrets that are only available to users of the earthly org).

alexcb:
@weaversam8 pointed out (via slack) that there's other locations in the code where the port value is used, so that won't work.

I manually tested it with:

```
git:
  github.com:
    auth: "ssh"
    user: "git"
    ssh_command: "ssh -c badcipher"
```

which correctly shows the `ssh_command` is correctly passed through, since it attempts to use the erroneous cipher:

```
Error: build target: build main: failed to solve: resolve build context for target github.com/earthly/earthly+base: read git-hash: failed to load cache key: error fetching default branch for repository git@github.com:earthly/earthly.git:  exit status 128

git ls-remote --symref git@github.com:earthly/earthly.git HEAD
Unknown cipher type 'badcipher'
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

It seems like we could skip having to stand up an ssh server entirely for this test, and just validate that earthly returns with an error that contains the string `Unknown cipher type 'badcipher'`.

weaversam8:
@alexcb - I just pushed a commit above with a test with your idea (thanks!)

The test is currently failing - it looks like the `GIT CLONE` Earthfile instruction isn't processing the ssh_command properly, so I'll need to fix that. That'll probably be Monday or Tuesday.

alexcb:
I think the test could be simplified -- there shouldn't be a need to startup the sshd server.

instead, you can stick this in `tests/Earthfile`:

```
# test-ssh-command-config tests that we can set a custom ssh_command
# The test intentionally sets a bad cipher type, then tests that buildkit fails, and the failure message
# contains the underlying ssh failure due to attempting to use an invalid cipher
test-ssh-command-config:
    RUN ssh-keygen -b 3072 -t rsa -f /root/dont-steal-my-key -q -N "" -C "rsa-testkey"
    RUN echo "#!/bin/sh
set -ex

# needed to get past the \"no SSH key default forwarded from the client\" error
eval \$(ssh-agent)
ssh-add /root/dont-steal-my-key

earthly --config \$earthly_config config git \"{github.com: {auth: ssh, user: git, strict_host_key_checking: false, ssh_command: ssh -c badcipher}}\"
earthly --config \$earthly_config github.com/earthly/earthly+base 2>&1 | tee output.txt;
" >/tmp/run-earthly-with-bad-ssh-cipher && chmod +x /tmp/run-earthly-with-bad-ssh-cipher
    DO +RUN_EARTHLY --exec_cmd=/tmp/run-earthly-with-bad-ssh-cipher
    RUN cat output.txt | acbgrep "Unknown cipher type 'badcipher'"
```

plus add a `BUILD +test-ssh-command-config` under `ga-no-qemu-group4`.

weaversam8:
@alexcb - That's a smart idea - I left the test in place for now since your test didn't test the `GIT CLONE` option, which was also currently broken. It is true that I don't need to start the server, but I figured it was easier to keep the pattern in the `tests/git-ssh-server` since it already contained a test Earthfile with a `GIT CLONE` instruction.

My latest commit fixes this failing test, so it should be ready to go. Are you okay leaving the more complex test case (since it covers `GIT CLONE`?) Or would you prefer to write a simpler test case that also covers `GIT CLONE` (we'd need to have it load an earthfile somehow.)

alexcb:
I would prefer extending the `badcipher` test, since it'll be faster to execute than the sshd-based test. Our integration tests are quite heavy, so any chance to avoid making them even heavier adds to a better internal developer experience.

I would create two tests in this case:

`test-ssh-command-config-via-remote-target` (which is just renaming my `test-ssh-command-config` test)

and 

`test-ssh-command-config-via-git-clone-earthfile-command`, which is a similar test:

```
test-ssh-command-config-via-git-clone-earthfile-command:
    RUN ssh-keygen -b 3072 -t rsa -f /root/dont-steal-my-key -q -N "" -C "rsa-testkey"

    # or alternatively do a COPY some-test.earthfile Earthfile
    RUN echo "VERSION 0.7
test:
    FROM alpine
    GIT CLONE ...... # the command we are testing
" > Earthfile

    RUN echo "#!/bin/sh
set -ex

# needed to get past the \"no SSH key default forwarded from the client\" error
eval \$(ssh-agent)
ssh-add /root/dont-steal-my-key

earthly --config \$earthly_config config git \"{github.com: {auth: ssh, user: git, strict_host_key_checking: false, ssh_command: ssh -c badcipher}}\"
earthly --config \$earthly_config +test 2>&1 | tee output.txt;
" >/tmp/run-earthly-with-bad-ssh-cipher && chmod +x /tmp/run-earthly-with-bad-ssh-cipher
    DO +RUN_EARTHLY --exec_cmd=/tmp/run-earthly-with-bad-ssh-cipher
    RUN cat output.txt | acbgrep "Unknown cipher type 'badcipher'" # this might have to be changed if the error is output slightly differently.
```

tests/Earthfile
```diff
@@ -1510,6 +1512,47 @@ test-exec-stats:
...
+earthly --config \$earthly_config config git \"{github.com: {auth: ssh, user: git, strict_host_key_checking: false, ssh_command: ssh -c badcipher}}\"
+earthly --config \$earthly_config github.com/earthly/earthly+base 2>&1 | tee output.txt;
+" >/tmp/run-earthly-with-bad-ssh-cipher && chmod +x /tmp/run-earthly-with-bad-ssh-cipher
+    DO +RUN_EARTHLY --exec_cmd=/tmp/run-earthly-with-bad-ssh-cipher
+    RUN cat output.txt | acbgrep "Unknown cipher type 'badcipher'"
+
+test-ssh-command-config-via-git-clone-earthfile-command:
+    RUN ssh-keygen -b 3072 -t rsa -f /root/dont-steal-my-key -q -N "" -C "rsa-testkey"
+
+    # or alternatively do a COPY some-test.earthfile Earthfile
```
alexcb, line 1536: can you remove this comment? this was added to help communicate my idea, not for the final version.

tests/Earthfile
```diff
@@ -1510,6 +1512,47 @@ test-exec-stats:
...
+
+# needed to get past the \"no SSH key default forwarded from the client\" error
+eval \$(ssh-agent)
+ssh-add /root/dont-steal-my-key
+
+earthly --config \$earthly_config config git \"{github.com: {auth: ssh, user: git, strict_host_key_checking: false, ssh_command: ssh -c badcipher}}\"
+earthly --config \$earthly_config +test 2>&1 | tee output.txt;
+" >/tmp/run-earthly-with-bad-ssh-cipher && chmod +x /tmp/run-earthly-with-bad-ssh-cipher
+    DO +RUN_EARTHLY --exec_cmd=/tmp/run-earthly-with-bad-ssh-cipher
+    RUN cat output.txt | acbgrep "Unknown cipher type 'badcipher'" # this might have to be changed if the error is output slightly differently.
```
alexcb, line 1554: has this been tested? if so please remove the comment.

weaversam8: yes, I ran the target to test it. my mistake, removed.

tests/Earthfile
weaversam8: ah, I missed this, thanks. removed.

alexcb (approved):
looks great, thanks for tackling this!

We had a buildkit update which resulted in a merge conflict, I opened #3598 which resolves the conflicts, I'll tackle anything else that's needed before we can merge it.

alexcb:
This has been merged via #3598; thanks for the contribution.

PR #3579, opened 2023-12-08
Auto-skip: IMPORT support

mikejholly:
I noticed this was missing when I tried to enable auto-skip in Cloud. I'm surprised I hadn't noticed, but here it is!


vladaionescu (commented):
When the import is a remote Earthfile, does it apply the same logic as other remote cases? (requires tag or hash to be pinned?)

tests/autoskip/imported/Earthfile
```diff
@@ -0,0 +1,5 @@
+VERSION 0.7
+
+test:
+    FROM alpine
+    RUN echo "hi"
```
alexcb, line 5: I think it would be a good idea to extend this to perform a `COPY some-file .`

then verify that editing this file will bust the cache.

mikejholly: That should be covered elsewhere, but I suppose it doesn't hurt to do something similar here.

alexcb: yeah my thinking is to be absolutely sure the imported referencing is making its way into the imported Earthfile.

mikejholly:
@vladaionescu updated with remote import handling.

tests/autoskip/imported/Earthfile
mikejholly: Updated.

vladaionescu (approved):

PR #3583, opened 2023-12-09
Auto-skip: Fix issue with COPY target resolution

mikejholly:
The existing code was not properly handling IMPORT aliases.

### DIFF
--- a/inputgraph/loader.go
+++ b/inputgraph/loader.go
@@ -238,7 +238,7 @@ func (l *loader) handleCopySrc(ctx context.Context, src string, mustExist bool) 
 		return errUnsupportedRemoteTarget
 	}
 
-	targetName := artifactSrc.Target.LocalPath + "+" + artifactSrc.Target.Target
+	targetName := artifactSrc.Target.String()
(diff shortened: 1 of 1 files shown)


inputgraph/loader.go
```diff
@@ -367,9 +367,7 @@ func (l *loader) handleImport(ctx context.Context, cmd spec.Command) error {
 		alias = cmd.Args[2]
 	}
 
-	isGlobal := l.target.Target == "base"
-
-	err := l.varCollection.Imports().Add(cmd.Args[0], alias, isGlobal, false, false)
+	err := l.varCollection.Imports().Add(cmd.Args[0], alias, false, false, false)
```
mikejholly, line 370: Once I fixed the above issue, the use of global imports here led to duplicate errors. I'm not 100% clear on how global imports are meant to behave. Are they global to a file or global across the entire execution context (all files)?

vladaionescu: Global imports are global within the file only.

vladaionescu (approved):

PR #3585, opened 2023-12-11
Show a hint to use interactive flag on relevant build failures

idodod:
If a build fails as a result of RUN failing with an exit code != 0, earthly will print a help message to use the interactive flag.
Example:
image

Note: this will only be printed if the interactive flag is not already in use.

### DIFF
--- a/cmd/earthly/app/run.go
+++ b/cmd/earthly/app/run.go
@@ -209,7 +209,7 @@ func (app *EarthlyApp) run(ctx context.Context, args []string) int {
 				err.Error())
 			if !app.BaseCLI.Flags().InteractiveDebugging && len(args) > 0 {
 				args[0] = args[0] + " --interactive"
-				app.BaseCLI.Console().HelpPrintf("To debug your build, please use --interactive flag: %q\n", strings.Join(args, " "))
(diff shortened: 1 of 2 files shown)


cmd/earthly/app/run.go
```diff
@@ -207,6 +207,10 @@ func (app *EarthlyApp) run(ctx context.Context, args []string) int {
 			// This SetFatalError is a catch-all just in case that hasn't happened.
 			app.BaseCLI.Logbus().Run().SetGenericFatalError(time.Now(), logstream.FailureType_FAILURE_TYPE_OTHER,
 				err.Error())
+			if !app.BaseCLI.Flags().InteractiveDebugging && len(args) > 0 {
+				args[0] = args[0] + " --interactive"
+				app.BaseCLI.Console().HelpPrintf("To debug your build, please use --interactive flag: %q\n", strings.Join(args, " "))
```
vladaionescu, line 212:
```suggestion
				app.BaseCLI.Console().HelpPrintf("To debug your build, you can use the --interactive (-i) flag to drop into a shell of the failing RUN step: %q\n", strings.Join(args, " "))
```

vladaionescu (commented):
Minor wording adjustments

PR #3586, opened 2023-12-11
Limit the number of warnings shown when using COMMAND keyword

idodod:
The previous version was too noisy.
This changes it to show 3 times max and only once per file.

### DIFF
--- a/earthfile2llb/earthfile2llb.go
+++ b/earthfile2llb/earthfile2llb.go
@@ -180,7 +180,7 @@ type ConvertOpt struct {
 	// ProjectAdder is a callback that is used to discover PROJECT <org>/<project> values
 	ProjectAdder ProjectAdder
 
-	// FilesWithCommandRenameWarning keeps count of the files for which the COMMAND => FUNCTION warning was displayed
+	// FilesWithCommandRenameWarning keeps track of the files for which the COMMAND => FUNCTION warning was displayed
 	// this can be removed in VERSION 0.8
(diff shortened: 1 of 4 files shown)


earthfile2llb/interpreter.go
```diff
@@ -32,6 +32,9 @@ import (
 
 var errCannotAsync = errors.New("cannot run async operation")
 
+var filesWithCommandRenameWarning = make(map[string]bool)
+var maxCommandRenameWarnings = 3
```
vladaionescu, line 36: We generally avoid global variables, unless we really have no choice. It might be better to store the map in `ConvertOpts` from earthfile2llb.go? The maxCommandRenameWarnings can stay here, but it should be a `const`.

earthfile2llb/earthfile2llb.go
```diff
@@ -179,6 +179,10 @@ type ConvertOpt struct {
 
 	// ProjectAdder is a callback that is used to discover PROJECT <org>/<project> values
 	ProjectAdder ProjectAdder
+
+	// FilesWithCommandRenameWarning keeps count of the files for which the COMMAND => FUNCTION warning was displayed
```
vladaionescu, line 183:
```suggestion
	// FilesWithCommandRenameWarning keeps track of the files for which the COMMAND => FUNCTION warning was displayed
```

vladaionescu (approved):

mikejholly (approved):
Nice. This was a bit noisy without the limit.

Issue #3470, opened 2023-11-08
IMPORTing project references that do not exist fails silently


mitre-tm:
**What went wrong?**

[IMPORT](https://docs.earthly.dev/docs/earthfile#import)ing project references that do not exist should produce an error or at least a warning, but it does not.

Using the following Earthfile:
```earthly
VERSION 0.7

IMPORT ./does-not-exist

foo:
  FROM does-not-exist+base
```

Here is the result:
```sh
$ earthly --no-sat +base
 Init 🚀
————————————————————————————————————————————————————————————————————————————————

           buildkitd | Found buildkit daemon as docker container (earthly-buildkitd)

Streaming logs to https://cloud.earthly.dev/builds/beebe77a-2a1f-49ef-a4c2-e664464ad82c

 Build 🔧
————————————————————————————————————————————————————————————————————————————————

              output | --> exporting outputs

 Push Summary ⏫ (disabled)
————————————————————————————————————————————————————————————————————————————————

To enable pushing use earthly --push

 Local Output Summary 🎁
————————————————————————————————————————————————————————————————————————————————



========================== 🌍 Earthly Build  ✅ SUCCESS ==========================
```
It only fails lazily:
```sh
$ earthly --no-sat +foo
 Init 🚀
————————————————————————————————————————————————————————————————————————————————

           buildkitd | Found buildkit daemon as docker container (earthly-buildkitd)

Streaming logs to https://cloud.earthly.dev/builds/ef0d1967-6013-4f59-bd12-7ae4415928d4

 Build 🔧
————————————————————————————————————————————————————————————————————————————————

View logs at https://cloud.earthly.dev/builds/ef0d1967-6013-4f59-bd12-7ae4415928d4
            _unknown *failed* | build target: build main: failed to solve: Earthfile line 6:2 apply FROM does-not-exist+base: apply build does-not-exist+base: earthfile2llb for does-not-exist+base: resolve build context for target does-not-exist+base: No Earthfile nor build.earth file found for target does-not-exist+base
            _unknown *failed* | in		+foo
```

**What should have happened?**

It should produce an error or at least a warning, but it does not. If this is intentional, it should be documented.

**Other Helpful Information**

PR #3592, opened 2023-12-12
Check if an import path is valid when it's a local path

idodod:
Addresses #3470 

When the import path is a local path, check that it's a valid existing path or display a helpful error otherwise.

Also modify existing tests for consistency.

### DIFF
--- a/domain/importtracker.go
+++ b/domain/importtracker.go
@@ -2,12 +2,26 @@ package domain
 
 import (
 	"fmt"
+	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/earthly/earthly/conslogging"
+	"github.com/earthly/earthly/util/hint"
 
 	"github.com/pkg/errors"
 )
+
+type pathResult int
+
+const ( // iota is reset to 0
+	notExist pathResult = iota // c0 == 0
+	file                = iota // c1 == 1
+	dir                 = iota // c2 == 2
+)
+
+// used in unit tests
+type pathResultFunc func(path string) pathResult
 
 // ImportTrackerVal is used to resolve imports
 type ImportTrackerVal struct {
@@ -17,9 +31,10 @@ type ImportTrackerVal struct {
 
 // ImportTracker is a resolver which also takes into account imports.
 type ImportTracker struct {
-	local   map[string]ImportTrackerVal // local name -> import details
-	global  map[string]ImportTrackerVal // local name -> import details
-	console conslogging.ConsoleLogger
+	local          map[string]ImportTrackerVal // local name -> import details
+	global         map[string]ImportTrackerVal // local name -> import details
+	console        conslogging.ConsoleLogger
+	pathResultFunc pathResultFunc // to help with unit testing
 }
 
 // NewImportTracker creates a new import resolver.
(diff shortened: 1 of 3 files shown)


features/features_test.go
```diff
@@ -120,6 +120,7 @@ func TestAvailableFlags(t *testing.T) {
 		{"arg-scope-and-set", "ArgScopeSet"},
 		{"earthly-ci-runner-arg", "EarthlyCIRunnerArg"},
 		{"use-docker-ignore", "UseDockerIgnore"},
+		{"use-function-keyword", "UseFunctionKeyword"},
```
idodod, line 123: unrelated, but forgotten in previous PR.

domain/importtracker_test.go
idodod, line 1: The diff isn't great here, it might be easier to view commit by commit:
First commit => Add more tests
Second commit =>  Change existing tests to include a name and run in parallel

domain/importtracker.go
```diff
@@ -158,3 +177,40 @@ func (ir *ImportTracker) Deref(ref Reference) (resolvedRef Reference, allowPrivi
...
+		if filepath.Base(path) == earthlyFileName {
+			return hint.Wrapf(errors.Errorf("path %q does not contain an %s", path, earthlyFileName), "The path %q ends with an %q which is a directory.\nDid you mean to create an %q as a file instead?", path, earthlyFileName, earthlyFileName)
+		}
+		return hint.Wrapf(errors.Errorf("path %q does not contain an %s", path, earthlyFileName), "Verify the path %q contains an %s", path, earthlyFileName)
+	}
+	if res == dir {
+		return hint.Wrapf(errors.Errorf("path %q does contains an %s which is not a file", path, earthlyFileName), "The local IMPORT path %q contains an %q directory and not a file", path, earthlyFileName)
+	}
+	return nil
+}
```
vladaionescu, line 204: Does this work when this is in a remote context? E.g. I call the target `github.com/foo/bar+target` and that Earthfile imports `./buz`.

idodod: Good question.
So it seems to not have a negative or positive impact.
Some options I tested inside the remote Earthfile:
* If the local IMPORT exists - it works.
* If the local IMPORT is an absolute path - it returns an errors (~ absolute paths are not supported in the context of remote ref)
* if the local import is a relative path that does not exist - returns an error (but not as pretty as this PR tries to make it, because it's checked when the path is resolved)

vladaionescu: Interesting... it might be nice to throw an error for that case too. But that could be a separate PR.

domain/importtracker.go
```diff
@@ -158,3 +177,40 @@ func (ir *ImportTracker) Deref(ref Reference) (resolvedRef Reference, allowPrivi
...
+	}
+	res = pathResultF(filepath.Join(path, earthlyFileName))
+	if res == notExist {
+		if filepath.Base(path) == earthlyFileName {
+			return hint.Wrapf(errors.Errorf("path %q does not contain an %s", path, earthlyFileName), "The path %q ends with an %q which is a directory.\nDid you mean to create an %q as a file instead?", path, earthlyFileName, earthlyFileName)
+		}
+		return hint.Wrapf(errors.Errorf("path %q does not contain an %s", path, earthlyFileName), "Verify the path %q contains an %s", path, earthlyFileName)
+	}
+	if res == dir {
+		return hint.Wrapf(errors.Errorf("path %q does contains an %s which is not a file", path, earthlyFileName), "The local IMPORT path %q contains an %q directory and not a file", path, earthlyFileName)
```
vladaionescu, line 201:
```suggestion
		return hint.Wrapf(errors.Errorf("path %q contains an %s which is not a file", path, earthlyFileName), "The local IMPORT path %q contains an %q directory and not a file", path, earthlyFileName)
```

vladaionescu (approved):

idodod:
This actually requires some more work to account for the path of the target's Earthfile that is being executed

PR #3595, opened 2023-12-12
Auto-skip: prune subcommand

mikejholly:
Adds a new `prune-auto-skip` that supports the following behavior:

```bash
earthly prune-auto-skip --prefix './path'
earthly prune-auto-skip --prefix './path+target'
earthly prune-auto-skip --prefix '+target'
earthly prune-auto-skip --prefix './p+target' # all sub-directories that start with 'p' 
earthly prune-auto-skip --prefix './p'
```

We also have support for pruning all hashes for org/project, but that's not yet exposed here.


vladaionescu:
Some questions:

* When using a prefix like `./path`, does it resolve to the canonical form of the Earthfile? Like `github.com/foo/bar/path` underneath?
* It's unintuitive that this case would prune even `./project+target`
  ```bash
  earthly prune-auto-skip --prefix './p+target' # all sub-directories that start with 'p' 
  ```
  Maybe "prefix" is not the best word here. How about separating things out a bit more via --path, --deep, and --target to distinguish between going in deep vs not?
  ```bash
  earthly prune-auto-skip --path github.com/foo/bar --deep                      # Clears auto-skip for the entire repo
  earthly prune-auto-skip --path github.com/foo/bar                             # Clears auto-skip for the top-level targets only
  earthly prune-auto-skip --path github.com/foo/bar --target +my-target         # Clears auto-skip for the top-level target called my-target
  earthly prune-auto-skip --path github.com/foo/bar --deep --target +my-target  # Clears auto-skip for all targets called my-target in the entire repo
  ```

mikejholly:
* No, it just uses the local path. This can be expanded to the full, canonical path.
* Sure. Either that or use `--path` for exact matches and `--path-prefix` for all matches.

mikejholly:
@vladaionescu updated to use `--path`, `--target`, & `--deep`.

mikejholly:
Examples:
```bash
$ earthly-dev prune-auto-skip --target "+for-linux" --path "github.com/foo" --deep
Deleted 0 matching auto-skip hashes
$ earthly-dev prune-auto-skip --target "+for-linux" --path "github.com/earthly/cloud" --deep  
Deleted 0 matching auto-skip hashes
$ earthly-dev prune-auto-skip --target "+for-linux" --path "github.com/earthly/earhtly" --deep
Deleted 0 matching auto-skip hashes
$ earthly-dev prune-auto-skip --target "+for-linux" --path "github.com/earthly/earthly" --deep
Deleted 1 matching auto-skip hash
```

cloud/askv.go
```diff
@@ -2,31 +2,53 @@ package cloud
...
 	}
 	return nil
 }
+
+func (c *Client) AutoSkipPrune(ctx context.Context, org, project, pathPrefix, target string, deep bool) (int, error) {
+	target = strings.TrimPrefix(target, "+")
+
+	req := &askv.PruneTargetRequest{
+		OrgName:       org,
+		ProjectName:   project,
```
vladaionescu, line 42: I'm noticing that the project is specified here - isn't this supposed to be global within the organization?

mikejholly: I'm not sure. I assumed we'd want to scope pruning to the project if passed via the flag or env. But that's not essential. If the value is empty in that request, it will be ignored.

vladaionescu: Shall we just explicitly pass it as empty string then? Just to make sure the hashes don't get sent to different places depending on whether `EARTHLY_PROJECT` is set or not.

I think that we've not yet figured out what to do with PROJECT, as that was a concept only needed for Earthly CI. I'd keep it global to the org until we get more clarity on that front.

mikejholly:
Updated. Project is no longer sent up and the code will fallback to `PROJECT` for the org name if it's not set as a flag or env.

vladaionescu (approved):
Nice! 👏

PR #3598, opened 2023-12-13
Acb/custom ssh command

### LINKED ISSUE #3499: Per-site git options should support a custom `GIT_SSH_COMMAND`

alexcb:
fixes #3499

- Adds ssh_command option to the Git site config
- Updates Earthfile to support building a buildkit fork with a repo name other than "buildkit"

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -74,9 +74,10 @@ update-buildkit:
     ARG BUILDKIT_GIT_SHA
     ARG BUILDKIT_GIT_BRANCH=earthly-main
     ARG BUILDKIT_GIT_ORG=earthly
+    ARG BUILDKIT_GIT_REPO=buildkit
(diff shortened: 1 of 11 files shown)


alexcb:
This PR is based on #3567 (which had a merge conflict)

alexcb:
depends on https://github.com/earthly/buildkit/pull/39

mikejholly (approved):

PR #3608, opened 2023-12-15
circleci tests: reduce amounts of tests

alexcb:
circleci appears to be struggling with the +test target; this introduces a new top-level +smoke-test target for use with circleci.


idodod (approved):

.circleci/config.yml
```diff
@@ -24,10 +24,7 @@ jobs:
           name: Run tests
           command: |-
               ./build/linux/amd64/earthly --ci -P \
-              +test --DOCKERHUB_MIRROR_AUTH_FROM_CLOUD_SECRETS=true
-      - run:
-          name: Execute fail test
-          command: "! ./build/linux/amd64/earthly --ci +test-fail"
```
vladaionescu, line 30: Should we leave this one in?

mikejholly (approved):

.circleci/config.yml
alexcb: it's covered under githubactions already:

```
.github/workflows/reusable-test.yml:        run: "! ${{inputs.SUDO}} ${{ inputs.BUILT_EARTHLY_PATH }} ${{inputs.EXTRA_ARGS}} --ci ./tests/fail+test-fail"
```

having it tested in circleci doesn't provide any extra coverage.

PR #3611, opened 2023-12-15
Fix commit email collection

idelvall:
This change tolerates some metadata commands failing as long as `Author` is returned.


vladaionescu (approved):

idelvall:
Made some tests and seems like the culprit in our case was `detectGitTags` failing
https://github.com/earthly/lib/actions/runs/7226500441/job/19692213282#step:3:29

PR #3613, opened 2023-12-16
Fix hanging when process killed by user

mikejholly:
Control-C was leading to a hung process. The log streaming context was being canceled which led to a deadlock with writing new messages. 
image

### DIFF
--- a/cmd/earthly/app/run.go
+++ b/cmd/earthly/app/run.go
@@ -4,9 +4,12 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"os/signal"
 	"regexp"
 	"strconv"
 	"strings"
+	"sync/atomic"
+	"syscall"
 	"time"
 
 	billingpb "github.com/earthly/cloud-api/billing"
@@ -155,6 +158,18 @@ func (app *EarthlyApp) run(ctx context.Context, args []string) int {
 		app.BaseCLI.Logbus().Run().SetGenericFatalError(
 			time.Now(), logstream.FailureType_FAILURE_TYPE_OTHER,
 			"Error: No SetFatalError called appropriately. This should never happen.")
+	}()
+
+	sigChan := make(chan os.Signal, 1)
+	signal.Notify(sigChan, syscall.SIGINT, syscall.SIGINT)
+
+	userKilled := atomic.Bool{}
+
+	go func() {
+		for range sigChan {
+			userKilled.Store(true)
+			return
+		}
 	}()
 
 	err := app.BaseCLI.App().RunContext(ctx, args)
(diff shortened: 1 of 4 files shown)


cmd/earthly/app/run.go
```diff
@@ -157,6 +160,18 @@ func (app *EarthlyApp) run(ctx context.Context, args []string) int {
 			"Error: No SetFatalError called appropriately. This should never happen.")
 	}()
 
+	sigChan := make(chan os.Signal, 1)
+	signal.Notify(sigChan, syscall.SIGINT, syscall.SIGINT)
```
idodod, line 164: did you mean to add syscal.SIGINT twice?

logbus/ship/ship.go
```diff
@@ -53,7 +52,7 @@ func (l *LogShipper) Write(delta *pb.Delta) {
 // Start the log streaming process and begin writing logs to the server.
 func (l *LogShipper) Start(ctx context.Context) {
 	go func() {
-		ctx, l.cancel = context.WithCancel(ctx)
+		ctx, l.cancel = context.WithCancel(context.Background())
```
idodod, line 55: jsut curious is this because ctx is already canceled?

cmd/earthly/app/run.go
mikejholly: No. Thanks.

logbus/ship/ship.go
mikejholly: When the parent context was canceled, the child was also canceled. This killed the log streaming process and led to a deadlock.

alexcb (approved):

cmd/earthly/app/run.go
```diff
@@ -157,6 +160,18 @@ func (app *EarthlyApp) run(ctx context.Context, args []string) int {
...
 	}()
 
+	sigChan := make(chan os.Signal, 1)
+	signal.Notify(sigChan, syscall.SIGINT, syscall.SIGTERM)
+
+	userKilled := atomic.Bool{}
+
+	go func() {
+		for range sigChan {
+			userKilled.Store(true)
```
vladaionescu, line 170: There could be a race condition, if this doesn't execute quickly enough.

Maybe it's better to store this here https://github.com/earthly/earthly/blob/main/cmd/earthly/main.go#L61 instead. It would also deduplicate the need for `signal.Notify`.

mikejholly: Updated. The signal is now provided to this code in a thread-safe way. 

I removed the warning line as it's somewhat duplicated, wasn't working, & is writing a log message after everything is closed. Ok, to remove?

We already have these messages.

```
^CReceived signal interrupt. Cleaning up before exiting...
Canceled
View logs at https://cloud.earthly.dev/builds/821b3352-5a00-42b0-aa3f-73e835367da2
```

vladaionescu (approved):
Looks great!

PR #3622, opened 2023-12-18
Support mTLS for Self-Hosted Satellites

brandonSc:
Uses a CA key/pair from the cloud backend to generate a new client-side certs/key pair for use in an mTLS connection with a self-hosted satellite.


buildkitd/certificates.go
```diff
@@ -247,3 +249,58 @@ func savePEM(path, typ string, bytes []byte) error {
...
+	earthlyCertPath := filepath.Join(dir, "earthly_cert.pem")
+	earthlyKeyPath := filepath.Join(dir, "earthly_key.pem")
+
+	settings.TLSCA = caCertPath
+	settings.ClientTLSCert = earthlyCertPath
+	settings.ClientTLSKey = earthlyKeyPath
+	settings.ServerTLSCert = ""
+	settings.ServerTLSKey = ""
+
+	if err := os.RemoveAll(dir); err != nil {
```
idodod, line 270: will this error if the previous certificates don't exist?

config/config.go
```diff
@@ -52,6 +53,9 @@ const (
 	// DefaultServerTLSKey is the default path to use when looking for the Buildkit TLS key
 	DefaultServerTLSKey = "./certs/buildkit_key.pem"
 
+	// DefaultSatelliteCertsDir is the default path to use for saving TLS certs and keys for self-hosted satellites
+	DefaultSatelliteCertsDir = "./certs"
```
idodod, line 57: do you mind checking if this is set in the right full path when specifying a different installation_dir (basically that the concatenation occurs after installation_dir and the cfg are set) ?

idodod (approved):
LGTM

buildkitd/certificates.go
brandonSc: It works when the dir doesn't exist (in fact they commented that it returns nil in that case in the src)

buildkitd/certificates.go
```diff
@@ -247,3 +249,58 @@ func savePEM(path, typ string, bytes []byte) error {
 
 	return nil
 }
+
+// ConfigureSatelliteTLS uses the CA cert and key associate with the satellite
+// to generate a new certificate/key pair for use in client-side mTLS.
+// The certificates are configured in the settings for a new buildkit client.
+// The certificates are also stored on disk in: `~/.earthly/certs/<org-id>/<sat-name>`.
+func ConfigureSatelliteTLS(sat *cloud.SatelliteInstance, settings *Settings, cfg *config.Config) error {
+	dir := filepath.Join(cfg.Global.SatelliteCertsDir, sat.Org, sat.Name)
```
alexcb, line 258: using the same directory (as specified by a config file) makes this code susceptible to race conditions (where two instances of earthly are running in parallel).

can we use a temp directory instead?

buildkitd/certificates.go
alexcb: this isn't needed now that it's a temp dir; however, we should have a deferred call to remove the temp dir before earthly exits.

I wouldn't worry about the edge-case of someone doing a hard-kill of earthly, the temp dir will eventually get cleaned by the OS..... I hope at least :)

go.mod
```diff
@@ -18,7 +18,7 @@ require (
 	github.com/docker/distribution v2.8.2+incompatible
 	github.com/docker/go-connections v0.4.0
 	github.com/dustin/go-humanize v1.0.1
-	github.com/earthly/cloud-api v1.0.1-0.20231214223040-20dabccd7dce
+	github.com/earthly/cloud-api v1.0.1-0.20231220174410-2c53dd70eda1
```
brandonSc, line 21: TODO update once server PR is merged

buildkitd/certificates.go
brandonSc: Okay I'll look into the deferred approach.  I'm also wondering now if the path contain a UUID, so two different earthly processes don't try to write/remove certs for the same satellite (they would each generated different client certs)

brandonSc: I've gone with the approach of having the new `ConfigureSatelliteTLS` function return a cleanup func that can be deferred by the caller. 

I have this cleanup func also returned through `GetBuildkitClient`, which is used in many places like debug commands. I think its a bit ugly to return both a cleanupTLS function and a client that needs to be closed here. It could be more graceful if that cleanup happened automatically as part of the buildkit client's `Close()`, but that seemed like too much refactoring to be worth it.

WDYT @alexcb ?

buildkitd/certificates.go
```diff
@@ -247,3 +250,60 @@ func savePEM(path, typ string, bytes []byte) error {
...
+	settings.TLSDir = dir
+	settings.TLSCA = caRootPath
+	settings.ClientTLSCert = earthlyCertPath
+	settings.ClientTLSKey = earthlyKeyPath
+	settings.ServerTLSCert = ""
+	settings.ServerTLSKey = ""
+
+	if err := os.MkdirAll(dir, 0755); err != nil {
+		return nil, err
+	}
```
alexcb, line 275: after this point there's a leak where the dir isn't cleaned up if any errors are returned.

maybe something like:

```
cleanup := func() { _ = os.RemoveAll(dir) }
defer func() {
    if err != nil {
        cleanup()
    }
}

...

return cleanup, nil // the final return in this function
```

alexcb (commented):
looking good -- just one potential cleanup not being called on errors leak

buildkitd/certificates.go
alexcb: The solution you have looks good (just left one minor suggestion)

buildkitd/certificates.go
brandonSc: Good point. I overlooked that.

alexcb (approved):

Issue #3628, opened 2023-12-19
unable to use `--use-remote-registry` in combination with `--use-inline-cache`


alexcb:
```
earthly-v0.7.23 --use-inline-cache --use-remote-registry ./buildkitd+buildkitd
```

fails with:

```
....
./buildkitd+buildkitd | DOCKERHUB_MIRROR_AUTH_FROM_CLOUD_SECRETS=true
./buildkitd+buildkitd | *cached* --> COPY ./oom-adjust.sh.template /bin/oom-adjust.sh.template
./buildkitd+buildkitd | DOCKERHUB_MIRROR_AUTH_FROM_CLOUD_SECRETS=true
./buildkitd+buildkitd | *cached* --> COPY ./runc-ps /bin/runc-ps
              output | --> exporting outputs
              output | [----------] 100% exporting outputs
              output | --> exporting outputs
              output | [----------] 100% exporting outputs
Error: pull ping error: pull ping response: rpc error: code = Unknown desc = image pull: 1 error occurred:
	* command failed: docker pull 127.0.0.1:34661/sess-9rxhsyo1hmjwbj1cbwommxdem/pullping:img-1: exit status 1: Error response from daemon: Get "http://127.0.0.1:34661/v2/": dial tcp 127.0.0.1:34661: connect: connection refused: exit status 1
```

_Originally posted by @alexcb in https://github.com/earthly/earthly/issues/3624#issuecomment-1861903278_

PR #3624, opened 2023-12-18
always use remote registry

alexcb:

PR #3629, opened 2023-12-19
Enable remote registry proxy everwhere

mikejholly:
Note: `EARTHLY_GLOBAL_WAIT_END=true` also fixes the issue without this hacky patch. Should we just wait until global wait block is enabled?


builder/builder.go
```diff
@@ -217,11 +217,19 @@ func (b *Builder) convertAndBuild(ctx context.Context, target domain.Target, opt
 		// to accomodate parallelism in the WAIT/END PopWaitBlock handling
 		dirIDs = map[int]string{}
 	)
+
```
alexcb, line 220: nit: why not just combine the two `var` sections instead?

builder/builder.go
```diff
@@ -217,11 +217,19 @@ func (b *Builder) convertAndBuild(ctx context.Context, target domain.Target, opt
...
 		depIndex   = 0
 		imageIndex = 0
 		dirIndex   = 0
 	)
+
+	// Delay closing the registry proxy server until this method
+	// returns. Previously, it was being closed when the buildFunc returned,
+	// which led to a problem with fetching some images.
+	stopRegistryProxyFunc := func() {}
+	defer stopRegistryProxyFunc()
```
alexcb, line 231: I don't think this works, see the following for an example:

```
alex@surfperch:~$ cat main.go 
package main

import "fmt"

func main() {

	stopRegistryProxyFunc := func() {
		fmt.Println("first func")
	}
	defer stopRegistryProxyFunc()

	stopRegistryProxyFunc = func() {
		fmt.Println("is this actually called?")
	}
}
alex@surfperch:~$ go run main.go 
first func
```

mikejholly: Good point. I was originally using the below approach but wanted to condense the code. It works, but this is because the registry never closes.

```golang
	var stopRegistryProxyFunc func()
	defer func() {
		if stopRegistryProxyFunc != nil {
			stopRegistryProxyFunc()
		}
	}()
```

alexcb: `defer func() { stopRegistryProxyFunc() }` would probably work.

mikejholly: Sure, that's fine too.

alexcb:
we should validate that this fixes the issues that #3624 surfaced prior to merging this.

builder/builder.go
mikejholly: I think it's ok to group var & const declarations by purpose or subject.

mikejholly:
> we should validate that this fixes the issues that #3624 surfaced prior to merging this.

Does it make more sense to merge this PR and then address any other issues in a PR to enable remote registry for version 0.8?

alexcb:
> > we should validate that this fixes the issues that #3624 surfaced prior to merging this.
> 
> Does it make more sense to merge this PR and then address any other issues in a PR to enable remote registry for version 0.8?

We should try to add tests to every PR when possible -- that could be done by cherry-picking #3624 and combining them into a single PR, or possibly by creating a new test that enables both `--use-inline-cache --use-remote-registry` features.

alexcb:
Linking #3628 to this PR

cmd/earthly/flag/global.go
```diff
@@ -517,12 +516,5 @@ func (global *Global) RootFlags(installName string, bkImage string) []cli.Flag {
 			Usage:       "A remote docker image tag use as explicit cache",
 			Destination: &global.RemoteCache,
 		},
-		&cli.BoolFlag{
-			Name:        "use-remote-registry",
-			EnvVars:     []string{"EARTHLY_USE_REMOTE_REGISTRY"},
-			Usage:       "Use remote registry with Docker pull",
-			Destination: &global.UseRemoteRegistry,
-			Hidden:      true, // Internal.
-		},
```
vladaionescu, line 526: I wonder if we should keep this flag around (and just flipping the default value), just in case a user needs to turn it off for some reason (e.g. running into a bug)?

builder/builder.go
```diff
@@ -217,11 +211,21 @@ func (b *Builder) convertAndBuild(ctx context.Context, target domain.Target, opt
...
 	)
+
 	var (
 		depIndex   = 0
 		imageIndex = 0
 		dirIndex   = 0
 	)
+
+	// Delay closing the registry proxy server until this method
+	// returns. Previously, it was being closed when the buildFunc returned,
```
alexcb, line 222: nit: no need to mention a past bug that was fixed in this comment.

I would reword this as simply: `Delay closing the registry proxy server until after the buildFunc returns`

cmd/earthly/flag/global.go
alexcb: I'm on the fence regarding this, yes it's a nice escape hatch; however keeping around all these old features adds to maintenance cost.

We could leave it as-is for a few releases, and then fully remove it down the road; alternatively if a user runs into the need for this, we could re-introduce it, and the user could pin to an older version until it's done.

vladaionescu: Yeah, that's reasonable. Let's nuke it then

mikejholly: @alexcb and I were talking about this as well. On the one hand, it could be useful to retain the option to disable it. On the other, there's a small, accumulating cost with keeping these around (code, tests, etc.).

I'd vote for removing it and fixing bugs as they come up. All the tests are green :green_apple:

builder/builder.go
mikejholly: Sure. I wanted to allude to the fact that this code can be reverted once global wait block support is enabled. Updated.

alexcb (approved):

Issue #3632, opened 2023-12-20
document how args are passed implicitly to targets in the same Earthfile but not to other Earthfiles


alexcb:
Consider an Earthfile:

```
VERSION --pass-args 0.7
FROM alpine

a:
  BUILD +b
  BUILD ./subdir+sub # this does NOT pass args implicitly since it's an external target

b:
  ARG foo=beta
  RUN --no-cache echo in b with $foo

test:
  BUILD +a --foo=fromtest
```

and a second one under `subdir`:

```
VERSION 0.7

sub:
  FROM alpine
  ARG foo=subdirdefault
  RUN --no-cache echo in sub with $foo
```

when running `earthly +test`, one gets:

```
                  +b | in b with fromtest
        ./subdir+sub | in sub with subdirdefault
```

Note that `fromtest` is implicitly passed from the initial `test` target into `b` since it's in the _same_ Earthfile; however this is not the case when calling an external Earthfile.

In order to "implicitly" pass the args to the sub target, one must use the `--pass-args` flag, for example:

```
  BUILD --pass-args ./subdir+sub
```

which will then produce:

```
                  +b | in b with fromtest
        ./subdir+sub | in sub with fromtest
```

### Documentation request

I can't find a reference to this in our docs; I checked:
- https://docs.earthly.dev/docs/guides/build-args
- https://docs.earthly.dev/basics/part-4-args
- https://docs.earthly.dev/docs/earthfile#arg
- https://docs.earthly.dev/docs/earthfile#build

PR #3636, opened 2023-12-20
docs: --pass-args

alexcb:


mikejholly (approved):

alexcb:
This somewhat touches on #3632 -- it would be nice to link this to a higher level document, but I couldn't find anything.

docs/earthfile/earthfile.md
```diff
@@ -140,6 +140,11 @@ then one can build `my-target` by invoking earthly with the `--allow-privileged`
 earthly --allow-privileged +my-target
 ```
 
+##### `--pass-args`
+
+Earthly will automatically pass all existing arguments to referenced targets that exist in the _same_ Earthfile.
+When the `--pass-args` flag is set, Earthly will also propagate all arguments to the externally referenced target.
```
vladaionescu, line 146: Improve clarity a bit.

```suggestion
Earthly automatically passes all current arguments to referenced targets in the _same_ Earthfile. However, when the `--pass-args` flag is set, Earthly will also propagate all arguments to an externally referenced target.
```

vladaionescu (commented):
Don't forget about `DO`

vladaionescu (approved):

PR #3637, opened 2023-12-20
Fix flaky Podman test

mikejholly:
The test was checking for the chan to be closed before waiting on the goroutine in `BeforeEach` to close the chan.

Before:

```bash
go test -count 2000 -run TestPodmanProvider 
--- FAIL: TestPodmanProvider (0.00s)
    --- FAIL: TestPodmanProvider/it_falls_back_to_XDG_RUNTIME_DIR/containers/auth.json (0.00s)
        podman_test.go:63: channel open
            podman_test.go:63
--- FAIL: TestPodmanProvider (0.00s)
    --- FAIL: TestPodmanProvider/it_prefers_REGISTRY_AUTH_FILE (0.00s)
        podman_test.go:63: channel open
            podman_test.go:63
--- FAIL: TestPodmanProvider (0.00s)
    --- FAIL: TestPodmanProvider/it_prefers_REGISTRY_AUTH_FILE (0.00s)
        podman_test.go:63: channel open
            podman_test.go:63
--- FAIL: TestPodmanProvider (0.00s)
    --- FAIL: TestPodmanProvider/it_falls_back_to_XDG_RUNTIME_DIR/containers/auth.json (0.00s)
        podman_test.go:63: channel open
            podman_test.go:63
--- FAIL: TestPodmanProvider (0.00s)
    --- FAIL: TestPodmanProvider/it_falls_back_to_XDG_RUNTIME_DIR/containers/auth.json (0.00s)
        podman_test.go:63: channel open
            podman_test.go:63
FAIL
exit status 1
FAIL    github.com/earthly/earthly/util/llbutil/authprovider    4.130s
```
After:
```bash
$ go test -count 2000 -run TestPodmanProvider
PASS
ok      github.com/earthly/earthly/util/llbutil/authprovider    4.243s

### DIFF
--- a/util/llbutil/authprovider/podman_test.go
+++ b/util/llbutil/authprovider/podman_test.go
@@ -60,8 +60,8 @@ func TestPodmanProvider(topT *testing.T) {
 	defer o.Run()
 
 	o.AfterEach(func(tt testCtx) {
-		<-tt.result
-		tt.expect(tt.result).To(beClosed()) // Ensure tests don't leak goroutines.
+		_, ok := <-tt.result
+		tt.expect(ok).To(beFalse()) // Ensure that the channel was closed
 	})
 
 	type authFile struct {
--- a/util/llbutil/authprovider/podman_test.go
+++ b/util/llbutil/authprovider/podman_test.go
@@ -60,6 +60,7 @@ func TestPodmanProvider(topT *testing.T) {
 	defer o.Run()
 
 	o.AfterEach(func(tt testCtx) {
+		<-tt.result
 		tt.expect(tt.result).To(beClosed()) // Ensure tests don't leak goroutines.
 	})
 
--- a/util/llbutil/authprovider/imports_test.go
+++ b/util/llbutil/authprovider/imports_test.go
@@ -14,6 +14,7 @@ const (
 
 var (
 	beTrue       = matchers.BeTrue
+	beFalse      = matchers.BeFalse
 	not          = matchers.Not
 	haveOccurred = matchers.HaveOccurred
 	equal        = matchers.Equal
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: util/llbutil/authprovider/podman_test.go [hunk 1]
```diff
@@ -60,6 +60,7 @@ func TestPodmanProvider(topT *testing.T) {
 	defer o.Run()
 
 	o.AfterEach(func(tt testCtx) {
+		<-tt.result
```
alexcb, line 63: I don't understand this -- this seems to be discarding the results of a test without verifying them. Other parts in this test call things like:

```
			select {
			case res := <-tt.result:
				_, ok := res.(credentials)
				tt.expect(ok).To(beTrue())
```

mikejholly: `AfterEach` runs after each test in `TableSpec`. In those tests, the `result` chan is indeed read from. The new addition then waits for `result` to be closed with `<-tt.result`. The latter read is meant to verify the `close(tt.result)` occurred.

alexcb: I had to look over this test code a bit more closely; it feels more complicated than it needs to be (I'm not suggesting we change it).

I would feel a bit more comfortable with:

```
_, ok := <-tt.result
tt.expect(ok).To(beFalse()) // ensure the channel was closed
```

to prevent cases where we accidentally leave (unchecked) results on the channel.

mikejholly: Yes, I agree that it's overly complicated. Sure, I'm fine with that as well.

mikejholly: Updated.

alexcb (approved):
good find.

PR #3639, opened 2023-12-21
docs: CACHE --persist

alexcb:


docs/earthfile/earthfile.md
```diff
@@ -1422,14 +1422,23 @@ The sharing mode for the cache mount, from one of the following:
...
 The cache ID for a global cache volume to be used across other targets or Earthfiles.
 
+##### `--persist`
+
+Make a copy of the cache available to any children that inherit from this target, by copying the contents of the cache to the child image.
+
+{% hint style='warning' %}
+Caches were persisted by default in version 0.7, which led to bloated images being pushed to registries. Version 0.8 changed the default behavior
+to prevent copying the contents to children targets unless explicitly enabled by the newly added `--persist` flag.
+{% endhint %}
```
vladaionescu, line 1440: Can you remove this paragraph from above?

```
At the end of the target, the contents of the cache mountpoint are persisted as an additional layer in the image. This means that the contents are available to subsequent targets in the same build using `FROM`, or to any saved images `SAVE IMAGE`.
```

alexcb: I removed the warning entirely -- not sure if that's what you meant?

vladaionescu: No... grep for the words I quoted. Remove those :-)

alexcb: Thanks for pointing that out -- it was initially lost on me :laughing:

vladaionescu (approved):

PR #3642, opened 2023-12-21
Remove second var declaration in +smoke-test

mikejholly:
Declaring an ARG that's meant to depend on a global doesn't seem to be supported. Perhaps this should be supported by Earthly, but for now, here's a fix for the CircleCI test.


alexcb (commented):
can you also change the `all` target to be:

```
all:
    BUILD +alpine-kind
    BUILD +ubuntu-kind
```

alexcb (approved):

mikejholly:
Thanks. I missed that. Updated.

PR #3643, opened 2023-12-21
Changelog for 0.8

vladaionescu:


mikejholly (approved):

idodod (commented):
I think there are a few things that we updated unrelated to 0.8.
Should we include them as well in rc1?

PR #3647, opened 2023-12-21
Auto-skip: retain quotes when expanding args

mikejholly:
This fixes an issue with stripping quotes when expanding arguments.

The new code retains quotes and ensures the argument values are maintained after expansion.

Previous:

```
Original: COPY (+my-target --my-arg="hello world" --other="$FOO")
Expanded: COPY (+my-target --my-arg=hello world --other=value)
```

Note that the above `--my-arg` value will lead to an invalid argument error.

New:

```
Original: COPY (+my-target --my-arg="hello world" --other="$FOO")
Expanded: COPY (+my-target --my-arg="hello world" --other="value")
```

### DIFF
--- a/inputgraph/loader.go
+++ b/inputgraph/loader.go
@@ -894,8 +894,10 @@ func (l *loader) load(ctx context.Context) error {
 
 	// Ensure that args passed to this target are always hashed. Globals,
 	// built-ins, and ARG values are hashed elsewhere.
-	for _, val := range l.overridingVars.BuildArgs() {
-		l.hasher.HashString(fmt.Sprintf("VAR %s", val))
+	if l.overridingVars != nil {
+		for _, val := range l.overridingVars.BuildArgs() {
+			l.hasher.HashString(fmt.Sprintf("VAR %s", val))
+		}
 	}
 
 	ef := buildCtx.Earthfile
--- a/inputgraph/loader.go
+++ b/inputgraph/loader.go
@@ -21,7 +21,6 @@ import (
 	"github.com/earthly/earthly/features"
 	"github.com/earthly/earthly/util/buildkitskipper/hasher"
 	"github.com/earthly/earthly/util/flagutil"
-	"github.com/earthly/earthly/util/stringutil"
 	"github.com/earthly/earthly/variables"
 	"github.com/pkg/errors"
 )
@@ -48,12 +47,14 @@ type loader struct {
 }
 
 func newLoader(ctx context.Context, opt HashOpt) *loader {
+	h := hasher.New()
+	h.HashJSONMarshalled(opt.BuiltinArgs)
 	// Other important values are set by load().
 	return &loader{
 		conslog:         opt.Console,
 		target:          opt.Target,
 		visited:         map[string]struct{}{},
-		hasher:          hasher.New(),
+		hasher:          h,
 		isBaseTarget:    opt.Target.Target == "base",
 		ci:              opt.CI,
 		builtinArgs:     opt.BuiltinArgs,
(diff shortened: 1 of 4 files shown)


alexcb:
I'm trying to wrap my head around the change to expandArgs is needed?

this seems to work under v0.7.23 (without auto-skip)

```
VERSION 0.7

echofoo:
  FROM alpine
  ARG foo
  ARG bar
  RUN --no-cache env | grep foo
  RUN --no-cache env | grep bar
  RUN env | grep foo > data
  SAVE ARTIFACT data

test:
  FROM alpine
  ARG value="hello world"
  COPY (+echofoo/data --foo="$value" --bar=baz) .
```

`earthly-v0.7.23 +test` gives me:

```
            +echofoo | --> RUN --no-cache env | grep foo
            +echofoo | foo=hello world
            +echofoo | bar=baz foo=hello world
            +echofoo | --> RUN --no-cache env | grep bar
            +echofoo | bar=baz
```

I'm a bit worried about changing `variables/collection.go`  since it could introduce differences between how the inputgraph and converters parse commands, which could lead to the input graph producing different hashes, when the converter still produces the same llb.

alexcb:
I think part of the difference is under https://github.com/earthly/earthly/blob/0fd4ae2cc257b78cf228bcb808c797ea6b2eea73/inputgraph/loader.go#L199

the args have already been replaced -- for example:


```
flagutil.ParseParams (["+echofoo/data", "--foo=hello", "world", "--bar=baz"])
```

is being called when instead I think it should mimic what the converter passes, which would be closer to:

```
flagutil.ParseParams (["+echofoo/data", "--foo=\"$value\"", "--bar=baz"])
```


I think it's worthwhile changing the inputgraph code to be similar to the parser's code.

mikejholly:
@alexcb The main issue is that the `spec.Command` args for this `COPY` are provided as follows.  

```
([]string) (len=3 cap=4) {
  (string) (len=15) "(+copy-target/x",
  (string) (len=20) "--foo=\"hello world\")",
  (string) (len=1) "."
}
```

These need to be joined such that the first and second args are combined. `ProcessParamsAndQuotes` seemed to suit that need. 

```
([]string) (len=2 cap=2) {
  (string) (len=36) "(+copy-target/x --foo=\"hello world\")",
  (string) (len=1) "."
}
```

Previously, the quotes were being stripped around `hello world`, but the quotes are required for subsequent target parsing. 

The main difference with the interpreter code is that there is individual arg parsing and expanding code for each command & statement. In the loader code, I tried to reduce the boilerplate with a single expand step that also formats the args correctly for later parsing. The loader code needs to expand & hash the commands, but also keep them intact for later parsing. 

There may be a good reason to keep the code consistent, but this approach keeps the code concise and ultimately restructures the args correctly for auto-skip. The new `Expand` options should be safe and shouldn't affect other code.

mikejholly:
It looks like this change is leading to other some test failures. Perhaps it's a good time to bite the bullet and split the handlers. Expanding large strings doesn't seem to be working well for all cases. I'll take a stab.

mikejholly:
I've updated the PR with a new approach:

Instead of parsing, expanding, & hashing arg strings upfront in hopes of preserving them for multiple purposes, the new code leaves these strings intact and only expands values when further target exploration is required. Rather than expanding the args and hashing the resultant strings, both raw strings and arg values are hashed independently. This allows for more efficient auto-skip checks (less 10s on `+test`) and helps keep the code concise by removing the need to write custom handlers & differing treatment of args for each command type. Arg changes still bust the cache as expected.

vladaionescu:
> less 10s on +test

Yayy!

vladaionescu (approved):

PR #3649, opened 2023-12-22
Retry auth provider token fetch requests

mikejholly:
I see the below error from time to time but I assumed that we had retries in place for these requests. That's not the case, so I've added a drop-in retryable client that should cover these requests. I've extended the base retry policy to support `io.EOF`. 

```
failed to login to registry-1.docker.io with username griswoldthecat (using credentials from /earthly-technologies/cloud/std/registry/registry-1.docker.io/): Post "https://auth.docker.io/token": EOF
```
image
https://github.com/earthly/cloud/actions/runs/7303264176/job/19903432258


vladaionescu (approved):

vladaionescu:
Very nice fix, BTW - I had seen this kind of error before, and I didn't know that we were missing retries here.

Issue #3158, opened 2023-08-02
`earthly secrets ls /user` no longer works if the org is not specified/selected


idodod:
**What went wrong?**

Since user secrets do not belong to any (regular) org, an org need not be selected in order to list user secrets.
However, since user secrets can be viewed as the personal org secrets, (selecting the personal org + removing the `/user` prefix), the command now expects for the personal org to be selected which is a regression.

  
`earthly secrets ls /user`

**What should have happened?**

The user/personal org secrets should be listed

Specify the org, e.g : `earthly secrets --org <my-personal-org> ls /user`

**Other Helpful Information**

Relevant PR that may have introduced the regression: https://github.com/earthly/earthly/pull/3112

idodod:
FYI @brandonSc

PR #3651, opened 2023-12-29
Improve user secret ls ergonomics

mikejholly:
Addresses https://github.com/earthly/earthly/issues/3158

This change ensures the following user secret listing styles all behave as expected:

```bash
$ EARTHLY_ORG="" EARTHLY_PROJECT="" earthly secrets ls /user
$ EARTHLY_ORG="my-org" EARTHLY_PROJECT="my-project" earthly secrets ls /user
$ EARTHLY_ORG="my-org" EARTHLY_PROJECT="" earthly secrets ls /user
$ EARTHLY_ORG="my-personal-org" EARTHLY_PROJECT="" earthly secrets ls
```

However, because of how user secret names are stripped of the `/user/` prefix on display, users may be confused by the following scenario:

```bash
# Using the /user prefix here will essentially ignore the organization & project envs.
$ EARTHLY_ORG="my-org" EARTHLY_PROJECT="my-project" earthly secrets ls /user
/my/secret
# The above secret is stored as a user secret, but the /user prefix has been removed.
$ EARTHLY_ORG="my-org" EARTHLY_PROJECT="my-project" earthly secrets set /my/secret foo
# The new /my/secret secret will be stored in the specified organization & project.
```

We can either leave this as is or disallow non-personal organizations when interacting with `/user` secrets. The latter may break existing workflows as organizations & projects are often set as environmental variables.

### DIFF
--- a/cmd/earthly/subcmd/common.go
+++ b/cmd/earthly/subcmd/common.go
@@ -2,10 +2,16 @@ package subcmd
 
 import (
 	"context"
+	"strings"
+
 	"github.com/earthly/earthly/cloud"
 	"github.com/pkg/errors"
 	"github.com/urfave/cli/v2"
 )
+
+type orgLister interface {
+	ListOrgs(ctx context.Context) ([]*cloud.OrgDetail, error)
+}
 
 // projectOrgName returns the specified org or retrieves the default org from the API.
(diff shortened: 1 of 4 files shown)


alexcb (approved):
nice test coverage :)

Issue #3147, opened 2023-07-26
Crashes if global ARG variable missing (when `--if-exists` flag is used)


9876691:
**What went wrong?**

If you run `earthly +build` on this Earthfile it crashes. If you comment the ARG back in it's fine.

```
VERSION 0.7
FROM ubuntu

build:
    COPY --if-exists $FOLDER $FOLDER
```

Part of the stack trace.

```
panic: runtime error: slice bounds out of range [1:0]

goroutine 169 [running]:
github.com/earthly/earthly/util/llbutil.CopyOp({_, _}, {{{0x175a848, 0xc00035f440}, 0xc0009382d0, {0x12edf40, 0xc00090c6e0}, 0xc0003763d8, {0x0, 0x0, ...}, ...}}, ...)
        /earthly/util/llbutil/copyop.go:44 +0xb73
github.com/earthly/earthly/earthfile2llb.(*Converter).CopyClassical(0xc0005d6000, {0x1761dd0, 0xc00035eb70}, {0xc00035ee10, 0x1, 0x3}, {0x0, 0x0}, 0x0, 0x0, ...)
        /earthly/earthfile2llb/converter.go:558 +0x613
github.com/earthly/earthly/earthfile2llb.(*Interpreter).handleCopy(0xc000910000, {0x1761dd0, 0xc00035eb70}, {{0x155cf23, 0x4}, {0x0, 0x0}, {0xc000333dc0, 0x3, 0x4}, ...})
        /earthly/earthfile2llb/interpreter.go:1008 +0x139e
github.com/earthly/earthly/earthfile2llb.(*Interpreter).handleCommand(0xc000910000, {0x1761dd0, 0xc00035eb10}, {{0x155cf23, 0x4}, {0x0, 0x0}, {0xc000333dc0, 0x3, 0x4}, ...})
```

note this depends on `--if-exists` being set.

PR #3652, opened 2023-12-30
Guard against empty COPY sources

mikejholly:
Addresses https://github.com/earthly/earthly/issues/3147

This PR adds a check against empty COPY source arguments. This fixes the above `--if-exists` panic and validates all other sources that expand to empty strings. Another option would be to fail during arg expansion if the arg doesn't exist.

### DIFF
--- a/earthfile2llb/interpreter.go
+++ b/earthfile2llb/interpreter.go
@@ -971,11 +971,6 @@ func (i *Interpreter) handleCopy(ctx context.Context, cmd spec.Command) error {
 	if !allClassical && !allArtifacts {
 		return i.errorf(cmd.SourceLocation, "combining artifacts and build context arguments in a single COPY command is not allowed: %v", srcs)
 	}
-	for _, src := range srcs {
-		if src == "" {
-			return i.wrapError(err, cmd.SourceLocation, "empty COPY source encountered")
-		}
-	}
 	if slices.ContainsFunc(strings.Split(dest, "/"), func(s string) bool {
 		return s == "~" || strings.HasPrefix(s, "~")
 	}) {
--- a/earthfile2llb/interpreter.go
+++ b/earthfile2llb/interpreter.go
@@ -922,7 +922,6 @@ func (i *Interpreter) handleCopy(ctx context.Context, cmd spec.Command) error {
 	if err != nil {
 		return i.wrapError(err, cmd.SourceLocation, "parse platform %s", expandedPlatform)
 	}
-
 	allClassical := true
 	allArtifacts := true
 	for index, src := range srcs {
@@ -972,7 +971,11 @@ func (i *Interpreter) handleCopy(ctx context.Context, cmd spec.Command) error {
 	if !allClassical && !allArtifacts {
 		return i.errorf(cmd.SourceLocation, "combining artifacts and build context arguments in a single COPY command is not allowed: %v", srcs)
 	}
-
+	for _, src := range srcs {
+		if src == "" {
+			return i.wrapError(err, cmd.SourceLocation, "empty COPY source encountered")
+		}
+	}
 	if slices.ContainsFunc(strings.Split(dest, "/"), func(s string) bool {
 		return s == "~" || strings.HasPrefix(s, "~")
 	}) {
--- a/tests/Earthfile
+++ b/tests/Earthfile
@@ -232,8 +232,9 @@ copy-test:
     DO +RUN_EARTHLY --earthfile=copy.earth --output_does_not_contain="which does not expand to a home directory"
 
 copy-test-empty-src:
-    DO +RUN_EARTHLY --earthfile=copy.earth --target=+copy-empty-src --should_fail=true --output_contains="empty COPY source encountered"
-    DO +RUN_EARTHLY --earthfile=copy.earth --target=+copy-empty-src-if-exists --should_fail=true --output_contains="empty COPY source encountered"
+    # Regression tests for 'COPY --if-exists "" <dst>' (empty string)
+    DO +RUN_EARTHLY --earthfile=copy.earth --target=+copy-empty-src
+    DO +RUN_EARTHLY --earthfile=copy.earth --target=+copy-empty-src-if-exists
 
 copy-test-verbose-output:
     RUN mkdir -p subdir/a.txt && \
--- a/tests/Earthfile
+++ b/tests/Earthfile
@@ -30,6 +30,7 @@ ga-no-qemu-group1:
     BUILD +copy-test-verbose-output
     BUILD +copy-tilde-test
     BUILD +copy-keep-own-test
+    BUILD +copy-test-empty-src
     BUILD +git-clone-test
     BUILD +builtin-args-invalid-default-test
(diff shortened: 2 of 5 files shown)


earthfile2llb/interpreter.go
```diff
@@ -972,7 +971,11 @@ func (i *Interpreter) handleCopy(ctx context.Context, cmd spec.Command) error {
 	if !allClassical && !allArtifacts {
 		return i.errorf(cmd.SourceLocation, "combining artifacts and build context arguments in a single COPY command is not allowed: %v", srcs)
 	}
-
+	for _, src := range srcs {
+		if src == "" {
+			return i.wrapError(err, cmd.SourceLocation, "empty COPY source encountered")
```
alexcb, line 976:
```suggestion
			return i.wrapError(err, cmd.SourceLocation, "empty COPY source encountered")
			return i.errorf(cmd.SourceLocation, "COPY command is referencing an empty src path (maybe an $ARG is missing or set to \"\")")
```

idodod: I'm actually not sure we should necessarily error in this case (I originally worked on this ticket and meant to get back to it) - I _think_ technically (not sure of the use case though) the arg could be intentionally left empty.
Perhaps printing a warning instead? thoughts?

mikejholly: For `COPY`, specifically, I think we'll need a non-empty file path.

alexcb: it's not possible to have an empty file name under linux:

```
alex@surfperch:~$ touch ""
touch: cannot touch '': No such file or directory
```

however, for the case of a command like `COPY --if-exists "" dst-file` , maybe this should pass, and essentially becomes a non-operation?

idodod: right, but the srcs slice can contain multiple args, so I can see something like this:
```
my-target:
   ARG file1=foo
   ARG file2
   COPY --if-exists $file1 $file2 some/dest/.
```

`earthly +my-target --file2=bar` => copies foo & bar into some/dest/.

`earthly +my-target` => copies foo (only) into some/dest/.

if we return an error when file2 is empty it would be a breaking change (which I agree can be worked around with an IF statement but still a breaking change).

Perhaps warning when any `src` is empty and error is all `srcs` are empty?

tests/Earthfile
```diff
@@ -230,6 +231,10 @@ copy-test:
         echo "sub" > in/sub/file
     DO +RUN_EARTHLY --earthfile=copy.earth --output_does_not_contain="which does not expand to a home directory"
 
+copy-test-empty-src:
+    DO +RUN_EARTHLY --earthfile=copy.earth --target=+copy-empty-src --should_fail=true --output_contains="empty COPY source encountered"
+    DO +RUN_EARTHLY --earthfile=copy.earth --target=+copy-empty-src-if-exists --should_fail=true --output_contains="empty COPY source encountered"
```
alexcb, line 236: @idodod got me thinking, I think this should pass when the `--if-exists` flag is set.
```suggestion
    DO +RUN_EARTHLY --earthfile=copy.earth --target=+copy-empty-src-if-exists
```

alexcb (changes requested):
I jumped the gun on approving this PR; we need to figure out how `--if-exists` works when attempting to reference an invalid empty string file.

earthfile2llb/interpreter.go
alexcb: great point regarding multiple files!

> Perhaps warning when any src is empty and error is all srcs are empty?

I don't think we need to display a warning, or error in this case -- the user explicitly told us to ignore missing files by setting the `--if-exists` flag.

idodod: in that case perhaps the simplest way to address this is to skip empty src  in https://github.com/earthly/earthly/blob/96676e71fb7700801d5847f860679bec378d0468/util/llbutil/copyop.go#L32

idodod: I don't feel strongly about the warning - my only thinking was that there could be a difference between 
 - the arg set but the file does not exist
 - the arg is not set (by mistake).
 and that the user did mean to copy the file (if it exists) but forgot to set the arg for it (so essnetially the file might exist but the arg doesn't point to it).
 
 But like I said, no strong feelings about it.

mikejholly: That makes sense. The second point was basically what I was thinking when I wrote this code. An empty source seems erroneous. We can leave that point for later and just use the change you suggested, though.

idodod: > I see the point, but in what case would '' ever be a valid COPY arg? Is it a missing file, or rather an invocation error? Does it make sense to address both scenarios (empty sources & current panic bug)? Is this a breaking change if we're fixing a preexisting panic? I'm not bound to either approach, just looking to clarify :)

you're right, it would still panic, but this wouldn't:
```
my-target:
   ARG file1=foo //foo exists
   ARG file2
   COPY $file1 $file2 some/dest/.
```
 so maybe skipping/erroring in the copyop.go instead of interperter.go would be better.

mikejholly: Good point. I'll stick with ignoring the empty file for now. Thanks for the discussion.

idodod: side note: what would be great is to be able to know if the used ARG is even defined, but I think that's harder to do (it would require parsing src and finding all the used args in the src expression, then checking if they are all in declared). A problem for another issue I guess.

mikejholly: I considered that as well. It does seem odd to expand an undefined arg to `""`. Perhaps this ought to lead to an error instead.

alexcb: > It does seem odd to expand an undefined arg to `""`. Perhaps this ought to lead to an error instead.

I _think_ it's done to mimic the default behaviour of bash, but I agree it can be error-prone.

It would be neat if we could have a `set -u` inspired keyword that could be defined once per Earthfile.

mikejholly:
@idodod @alexcb Updated. The new code simply covers the `src` modifications to prevent the index panic.

alexcb (commented):
I think it's worthwhile to raise an error when the `--if-exists` flag is not present.

for example,

this should pass:
```
COPY --if-exists "" .
```

but this should fail:
```
COPY "" .
```

mikejholly:
That would introduce a behavior change (see thread above) as the below example currently completes without error. 

```
VERSION 0.7

FROM alpine

my-target:
   ARG file1
   ARG file2
   COPY $file1 $file2 some/dest/.
```

alexcb:
> That would introduce a behavior change (see thread above) as the below example currently completes without error.
> 
> ```
> VERSION 0.7
> 
> FROM alpine
> 
> my-target:
>    ARG file1
>    ARG file2
>    COPY $file1 $file2 some/dest/.
> ```

this feels like a bug; but I understand that some users might be relying on this behaviour, so perhaps we need to introduce a feature flag for this -- I'll open a separate issue for tracking it.

alexcb (approved):

Issue #3284, opened 2023-09-22
`earthly account login --token <token>` should ignore SSH_AUTH_SOCK (but doesnt)


alexcb:
If I try to run `earthly account login --token <token>`, 1Password prompts me to allow use of my SSH key.

After I approve it, the Earthly command exits with `Error: failed to login: failed executing ping request: failed refreshing expired auth token: agent: failed to sign challenge`.

_Originally posted by @Nezteb in https://github.com/earthly/earthly/issues/3179#issuecomment-1731861993_

alexcb:
@nezteb can you try running `SSH_AUTH_SOCK="" earthly account login --token <token>` -- does that prevent the key from being used?

(technically earthly will never read your private key, when I say "used", i'm referring to earthly requesting the ssh-agent to sign a challenge..... this is irrelevant to this ticket, but just wanted to mention our stance on never loading your private key into earthly's memory space)

Nezteb:
> can you try running `SSH_AUTH_SOCK="" earthly account login --token <token>`

That works!

```
❯ SSH_AUTH_SOCK="" earthly account login --token <token> && earthly org select <org>
Logged in as "<email>" using token auth
Log sharing is enabled by default. If you would like to disable it, run:

        earthly config global.disable_log_sharing true
Updated selected org in /Users/noah/.earthly/config.yml
```

alexcb:
we'll need to get this fixed -- the `<token>` should take precedence over ssh-based login.

extra note: When we fix this, we should also **test that once the JWT expires, the <token> still takes precedence**.

PR #3653, opened 2023-12-30
Prefer auth token to SSH login when provided

mikejholly:
Addresses #3284

This change allows for the Cloud client's auth token to be set when using the below command to log in. Previously, the command was performing SSH-based auth despite the `--token` parameter.

```
earthly account login --token <token>
```

Setting the `--token` parameter will also ignore any cached JWT. This seems more intuitive as the token is explicitly provided.

### DIFF
--- a/cloud/client.go
+++ b/cloud/client.go
@@ -76,11 +76,20 @@ type Client struct {
 	lastAuthMethodExpiry     time.Time
 }
 
+// ClientOpt is used to customize the Cloud client.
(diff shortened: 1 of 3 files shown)


mikejholly:
@vladaionescu would you mind confirming that this is the desired behavior?

alexcb (approved):

vladaionescu:
This is a good change - thanks!

PR #3655, opened 2024-01-01
Ensure satellite is ready before running test

mikejholly:
The `earthly-image-sat.sh` test has been failing periodically due to hitting unexpected states. This change will handle all states appropriately & wait for the satellite to become available. 

image

### DIFF
--- a/scripts/tests/earthly-image-sat.sh
+++ b/scripts/tests/earthly-image-sat.sh
@@ -35,6 +35,24 @@ function finish {
 }
 trap finish EXIT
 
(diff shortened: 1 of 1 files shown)


idodod (approved):
Nice!

alexcb (approved):
nice improvement.

I wonder if we should change the `wake` command to also wait on the shutdown stage -- but that's best left for a different PR.

mikejholly:
@alexcb AFAIK the relevant states are `Stopping` & `Sleeping`. This code will wait for `Stopping` and then call `wake` for both cases.

Issue #3582, opened 2023-12-09
Target auto-discovery: `BUILD ./*+test`


vladaionescu:
**Use case**

As a developer, I would like to be able to setup a mono-repository in such a way that sub-projects have standardized build targets. E.g. `+build`, `+test`, `+lint` etc. Defining such a target would automatically run as part of the CI/CD pipeline.

Combining this with `BUILD --auto-skip` (https://github.com/earthly/earthly/issues/3581) allows for a powerful monorepo setup.

**Expected Behavior**

To accomplish the above, the root Earthfile needs to execute targets in certain sub-directories, without the user having to add each target manually.

Currently, there is no way to accomplish this without significant downsides (using `LOCALLY`, which is not designed for CI use, or `COPY`-ing the entire repository into the build environment, which can be very slow).

The proposal is to support globbing in the path of local `BUILD` references. For example,

```Earthfile
# Supported
BUILD ./*+test
BUILD ./js-*+test
BUILD ./*/*+lint
BUILD ../*+build
BUILD ./projects/*+lint
BUILD ./src/*+compile
BUILD ./src{abc,def}+code

# Maybe? These would be nice because devs coming from bash expect them when you say that "globbing is supported".
BUILD ./foo?+test
BUILD ./foo{bar,buz}+test
BUILD ./[a-c]foo+test
# ... and other stuff listed here https://tldp.org/LDP/abs/html/globbingref.html

# Maybe? (any deep dir)
BUILD ./**+test

# Not allowed
BUILD ./foo+*                    # Glob can only be in the path
COPY ./*+build/artifact ./       # Cannot use with anything other than BUILD
FROM ./*+test                    # Cannot use with anything other than BUILD
BUILD github.com/foo/*+test      # Remote targets cannot use globbing
BUILD github.com/foo/bar/*+test  # Not even when the repo is clear
```

weaversam8:
Would this work on the command line as well?

PR #3661, opened 2024-01-03
Wildcard BUILDs

mikejholly:
This PR is my first pass on wildcard BUILD expansion. That is, builds in the form of `BUILD ./*+target`. This code will attempt to expand the BUILD into multiple commands if both the Earthfile & the target exist. The code will only explore directories 1 level beneath the referenced directory and globstar (`**`) is not yet supported.

Example: 

```
$ tree tests/wildcard-build 
tests/wildcard-build
├── bar
│   └── Earthfile
├── baz
│   └── Earthfile
├── foo
│   └── Earthfile
├── no-earthfile
└── no-target
    └── Earthfile
```

```
VERSION 0.7

test:
    FROM alpine
    BUILD ./wildcard-build/*+test
```

https://cloud.earthly.dev/builds/ec88fff1-d353-4045-a345-e3ce10cadff0

image

### DIFF
--- a/buildcontext/git.go
+++ b/buildcontext/git.go
@@ -66,7 +66,7 @@ type resolvedGitProject struct {
 	state pllb.State
 }
 
-func (gr *gitResolver) expandWildcardPath(ctx context.Context, gwClient gwclient.Client, platr *platutil.Resolver, target domain.Target, pattern string) ([]string, error) {
+func (gr *gitResolver) expandWildcard(ctx context.Context, gwClient gwclient.Client, platr *platutil.Resolver, target domain.Target, pattern string) ([]string, error) {
 	if !target.IsRemote() {
 		return nil, errors.Errorf("unexpected local reference %s", target.String())
 	}
--- a/buildcontext/git.go
+++ b/buildcontext/git.go
@@ -77,7 +77,10 @@ func (gr *gitResolver) expandWildcard(ctx context.Context, gwClient gwclient.Cli
 		return nil, err
 	}
 
-	fullPattern := filepath.Clean("./" + filepath.Join(subDir, pattern))
+	fullPattern := filepath.Join(subDir, pattern)
+	if !strings.HasPrefix(fullPattern, ".") {
+		fullPattern = "./" + fullPattern
+	}
 
 	var matches []string
 
--- a/buildcontext/git.go
+++ b/buildcontext/git.go
@@ -63,7 +63,8 @@ type resolvedGitProject struct {
 	// refs is the git refs
 	refs []string
 	// state is the state holding the git files.
-	state pllb.State
+	state          pllb.State
+	earthfilePaths []string
 }
 
 func (gr *gitResolver) expandWildcard(ctx context.Context, gwClient gwclient.Client, platr *platutil.Resolver, target domain.Target, pattern string) ([]string, error) {
@@ -76,23 +77,20 @@ func (gr *gitResolver) expandWildcard(ctx context.Context, gwClient gwclient.Cli
 		return nil, err
 	}
 
-	gitRef, err := llbutil.StateToRef(ctx, gwClient, rgp.state, false, platr.SubResolver(platutil.NativePlatform), nil)
-	if err != nil {
-		return nil, errors.Wrap(err, "state to ref git meta")
-	}
-
-	path, _ := filepath.Split(pattern)
-
-	stats, err := gitRef.ReadDir(ctx, gwclient.ReadDirRequest{Path: filepath.Join(subDir, path)})
-	if err != nil {
-		return nil, errors.Wrap(err, "failed to read dir")
-	}
+	fullPattern := "./" + filepath.Join(subDir, pattern)
 
 	var matches []string
 
-	for _, st := range stats {
-		if st.IsDir() {
-			matches = append(matches, filepath.Join(path, st.GetPath()))
+	for _, path := range rgp.earthfilePaths {
+		path = strings.TrimSuffix(path, "/Earthfile")
+		ok, err := filepath.Match(fullPattern, path)
+		if err != nil {
+			return nil, err
+		}
+		if ok {
+			// Ensure we return paths that are relative to the sub-directory.
+			path = strings.TrimPrefix(path, fmt.Sprintf("./%s/", subDir))
+			matches = append(matches, path)
 		}
 	}
 
@@ -268,6 +266,7 @@ func (gr *gitResolver) resolveGitProject(ctx context.Context, gwClient gwclient.
 					"git log -1 --format=%ae >/dest/git-author || touch /dest/git-author ; " +
 					"git log -1 --format=%b >/dest/git-body || touch /dest/git-body ; " +
 					"git for-each-ref --contains HEAD --format '%(refname:lstrip=-1)' >/dest/git-refs || touch /dest/git-refs ; " +
+					"find -name Earthfile > /dest/Earthfile-paths || touch /dest/Earthfile-paths ; " +
 					"",
 			}),
 			llb.Dir("/git-src"),
@@ -343,6 +342,12 @@ func (gr *gitResolver) resolveGitProject(ctx context.Context, gwClient gwclient.
 		if err != nil {
 			return nil, errors.Wrap(err, "read git-refs")
 		}
+		earthfilePathsRaw, err := gitMetaRef.ReadFile(ctx, gwclient.ReadRequest{
+			Filename: "Earthfile-paths",
+		})
+		if err != nil {
+			return nil, errors.Wrap(err, "read Earthfile-paths")
+		}
 
 		gitHash := strings.SplitN(string(gitHashBytes), "\n", 2)[0]
 		gitShortHash := strings.SplitN(string(gitShortHashBytes), "\n", 2)[0]
@@ -397,15 +402,16 @@ func (gr *gitResolver) resolveGitProject(ctx context.Context, gwClient gwclient.
 		}
 
 		rgp := &resolvedGitProject{
-			hash:        gitHash,
-			shortHash:   gitShortHash,
-			branches:    gitBranches2,
-			tags:        gitTags2,
-			committerTs: gitCommiterTs,
-			authorTs:    gitAuthorTs,
-			author:      gitAuthor,
-			coAuthors:   gitCoAuthors,
-			refs:        gitRefs2,
+			hash:           gitHash,
+			shortHash:      gitShortHash,
+			branches:       gitBranches2,
+			tags:           gitTags2,
+			committerTs:    gitCommiterTs,
+			authorTs:       gitAuthorTs,
+			author:         gitAuthor,
+			coAuthors:      gitCoAuthors,
+			refs:           gitRefs2,
+			earthfilePaths: strings.Split(strings.TrimSpace(string(earthfilePathsRaw)), "\n"),
 			state: pllb.Git(
 				gitURL,
 				gitHash,
--- a/buildcontext/git.go
+++ b/buildcontext/git.go
@@ -66,6 +66,39 @@ type resolvedGitProject struct {
 	state pllb.State
 }
 
+func (gr *gitResolver) expandWildcardPath(ctx context.Context, gwClient gwclient.Client, platr *platutil.Resolver, target domain.Target, pattern string) ([]string, error) {
+	if !target.IsRemote() {
+		return nil, errors.Errorf("unexpected local reference %s", target.String())
+	}
+
+	rgp, _, subDir, err := gr.resolveGitProject(ctx, gwClient, platr, target)
+	if err != nil {
+		return nil, err
+	}
+
+	gitRef, err := llbutil.StateToRef(ctx, gwClient, rgp.state, false, platr.SubResolver(platutil.NativePlatform), nil)
+	if err != nil {
+		return nil, errors.Wrap(err, "state to ref git meta")
+	}
+
+	path, _ := filepath.Split(pattern)
+
+	stats, err := gitRef.ReadDir(ctx, gwclient.ReadDirRequest{Path: filepath.Join(subDir, path)})
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to read dir")
+	}
+
+	var matches []string
+
+	for _, st := range stats {
+		if st.IsDir() {
+			matches = append(matches, filepath.Join(path, st.GetPath()))
+		}
+	}
+
+	return matches, nil
+}
+
 func (gr *gitResolver) resolveEarthProject(ctx context.Context, gwClient gwclient.Client, platr *platutil.Resolver, ref domain.Reference, featureFlagOverrides string) (*Data, error) {
 	if !ref.IsRemote() {
 		return nil, errors.Errorf("unexpected local reference %s", ref.String())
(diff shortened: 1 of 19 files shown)


tests/wildcard-build.earth
```diff
@@ -0,0 +1,9 @@
+VERSION 0.7
```
alexcb, line 1: main now supports v0.8
```suggestion
VERSION 0.8
```

tests/Earthfile
```diff
@@ -565,6 +566,11 @@ fail-invalid-artifact-test:
     DO +RUN_EARTHLY --earthfile=fail-invalid-artifact.earth --should_fail=true --target="--artifact +test/foo /tmp/stuff" \
         --output_contains="cannot save artifact +test/foo, since it does not exist"
 
+wildcard-build:
+    COPY --dir wildcard-build .
+    DO +RUN_EARTHLY --earthfile=wildcard-build.earth --target=+wildcard-build --output_does_not_contain="not-test run"
```
alexcb, line 571: can you add some extra tests here to verify `earthly.output` (which is created by `DO +RUN_EARTHLY`) contains the expected "hello world" strings.

alexcb (commented):
looking good! any thoughts on how hard it will be to support remote targets?

tests/Earthfile
mikejholly: Thanks. I forgot about that output file.

vladaionescu:
> any thoughts on how hard it will be to support remote targets?

We are assuming that we don't support remote targets for this. (This is where I documented it: https://github.com/earthly/earthly/issues/3582)

But now that I'm thinking about this a bit more... I wonder what happens if you're invoking a remote target that in turn invokes a `BUILD ./*+something` within it. Will this work?

tests/Earthfile
```diff
@@ -565,6 +566,15 @@ fail-invalid-artifact-test:
     DO +RUN_EARTHLY --earthfile=fail-invalid-artifact.earth --should_fail=true --target="--artifact +test/foo /tmp/stuff" \
         --output_contains="cannot save artifact +test/foo, since it does not exist"
 
+wildcard-build:
+    COPY --dir wildcard-build .
+    DO +RUN_EARTHLY --earthfile=wildcard-build.earth --target=+wildcard-build
+    RUN cat earthly.output | acbgrep 'hello world 1'
+    RUN cat earthly.output | acbgrep 'hello world 2'
+    RUN cat earthly.output | acbgrep 'hello world 3'
+    RUN cat earthly.output | (! acbgrep 'not-test run')
```
alexcb, line 575:
```suggestion
    RUN cat earthly.output | acbgrep -v 'not-test run'
```

this will cause it to print an error stating the command failed, e.g.

```
alex@surfperch:~/gh/earthly/earthly$ echo "not-test run" | ./scripts/acbtest/acbgrep -v 'not-test run'
grep '-v' 'not-test run' failed
alex@surfperch:~/gh/earthly/earthly$ echo $?
1
```

where as using `(! ...` defeats the purpose of acbgrep (which is to print out a helpful message when it fails), 

for example, the output of a pass is confusing:

```
alex@surfperch:~/gh/earthly/earthly$ echo "blah" | (! ./scripts/acbtest/acbgrep 'not-test run')
blah
grep 'not-test run' failed
alex@surfperch:~/gh/earthly/earthly$ echo $?
0
```

tests/wildcard-build/foo/Earthfile
```diff
@@ -0,0 +1,5 @@
+VERSION 0.7
```
alexcb, line 1:
```suggestion
VERSION 0.8
```

alexcb:
we should also add a feature flag for this, to make it more clear to users who are using an older version of earthly with this new feature.

mikejholly:
Updated with feature flag & support for remote targets which use `BUILD` with wildcards.

Example:

```
earthly-dev github.com/earthly/test-remote/wildcard-build:882190030762e620e932e3fab1091bdd50c93131+wildcard-build
```

https://cloud.earthly.dev/builds/90ae164f-3bdf-4c25-a06b-b871b9ce805b

buildcontext/git.go
```diff
@@ -66,6 +66,39 @@ type resolvedGitProject struct {
...
+	if !target.IsRemote() {
+		return nil, errors.Errorf("unexpected local reference %s", target.String())
+	}
+
+	rgp, _, subDir, err := gr.resolveGitProject(ctx, gwClient, platr, target)
+	if err != nil {
+		return nil, err
+	}
+
+	gitRef, err := llbutil.StateToRef(ctx, gwClient, rgp.state, false, platr.SubResolver(platutil.NativePlatform), nil)
```
mikejholly, line 79: I also tried to add `llb.Run(llb.Shlex("ls <path"))` here but I couldn't get it to work. The code here works well, but is limited to simple wildcard paths that end with `*` like `my/path/*`.

mikejholly: Here's the error. I found this surprising because the base image is `alpine/git:v2.43.0` and appears to include `ls`.

```
            internal *failed* | Repeating the failure error...
            internal *failed* | runc run failed: unable to start container process: exec: "/bin/sh": stat /bin/sh: no such file or directory
            internal *failed* | ERROR
            internal *failed* |       The internal command
            internal *failed* |           /bin/sh -c ls ./subdir/*
            internal *failed* |       did not complete successfully. Exit code 1
```

buildcontext/resolver.go
```diff
@@ -83,6 +83,17 @@ func NewResolver(cleanCollection *cleanup.Collection, gitLookup *GitLookup, cons
 	}
 }
 
+// ExpandWildcard will expand a wildcard BUILD target in a remote Git
+// repository. The pattern is the path relative to the target path and should be
+// in the form 'my/path/*'
+func (r *Resolver) ExpandWildcard(ctx context.Context, gwClient gwclient.Client, platr *platutil.Resolver, target domain.Target, pattern string) ([]string, error) {
+	if !target.IsRemote() {
+		return nil, errors.Errorf("unexpected local reference %s", target.String())
```
vladaionescu, line 91: Should we just move the implementation for the local case over here too, for symmetry?

earthfile2llb/interpreter.go
```diff
@@ -1239,6 +1248,79 @@ func (i *Interpreter) handleBuild(ctx context.Context, cmd spec.Command, async b
...
+		childTarget, err := domain.ParseTarget(childTargetName)
+		if err != nil {
+			return i.wrapError(err, cmd.SourceLocation, "failed to parse target %q", childTargetName)
+		}
+
+		data, _, _, err := i.converter.ResolveReference(ctx, childTarget)
+		if err != nil {
+			notExist := buildcontext.ErrEarthfileNotExist{}
+			if errors.As(err, &notExist) {
+				continue
```
vladaionescu, line 1289: So if an Earthfile does not exist in a directory, it's not an error, right? Can you confirm that this works for both local and remote cases?

util/fileutil/fileutil.go
```diff
@@ -65,3 +65,24 @@ func EnsureUserOwned(dir string, owner *user.User) error {
 		return os.Chown(path, uid, gid)
 	})
 }
+
+// GlobDirs will return any sub-directories which match the provided glob
+// pattern. Example: "/tmp/*" will return all sub-directories under "/tmp/".
+func GlobDirs(pattern string) ([]string, error) {
+	matches, err := filepath.Glob(pattern)
```
vladaionescu, line 72: The implementation for the local case seems much more rich than the remote version, since it permits more than just `*`. This can result in inconsistent behavior where Earthly builds something differently depending on how the target is referenced (can cause "works on my machine" type of issues).

Could we perhaps detect any unpermitted patterns here before passing to `filepath.Glob`? For example we could search for `[` or `?` and return an error in those cases.

(Maybe you're already checking for this somewhere else and I didn't see it)

buildcontext/git.go
vladaionescu: Hmm.. not sure why that happens.

It may be worth trying to get that to work, since supporting patterns like `BUILD ./cmd-*+binary` would be useful.

buildcontext/resolver.go
mikejholly: Sure, sounds good. I wasn't certain about where the new code belongs.

earthfile2llb/interpreter.go
mikejholly: That's right. The path is ignored in that case. I'll add some more tests. There was one at one point, but I think I dropped it by mistake while refactoring.

util/fileutil/fileutil.go
mikejholly: Yep, the code in `interpreter.go` currently limits support to the most basic case (e.g., `my/path/*`).

I can try to get `ls` working in the Git resolver to improve globbing support.

earthfile2llb/interpreter.go
mikejholly: Oh, the directory was missing a `.gitkeep`. Thanks.

buildcontext/git.go
alexcb: it might be a bit dirty (since it's no longer a git command), but maybe we could just add another line to https://github.com/earthly/earthly/blob/f51a0c1cbbbb44520d5343b373e66d8004f5ee75/buildcontext/git.go#L237

which would run a `find -name Earthfile > /dest/Earthfile-paths`, then a corresponding `gitMetaRef.ReadFile` which would return the list of Earthfiles in a new field under the `resolvedGitProject` struct.

vladaionescu: That's actually pretty reasonable IMO

mikejholly: Good idea. This would help remove the new code path in `gitResolver` & remove the need to run a custom `ls` or `find` command.

earthfile2llb/interpreter.go
```diff
@@ -1239,6 +1248,79 @@ func (i *Interpreter) handleBuild(ctx context.Context, cmd spec.Command, async b
...
+
+	var matches []string
+
+	if isRemoteTarget {
+		matches, err = i.converter.ExpandRemoteWildcard(ctx, i.target, parsedTarget.GetLocalPath())
+	} else {
+		matches, err = fileutil.GlobDirs(parsedTarget.GetLocalPath())
+	}
+	if err != nil {
+		return i.wrapError(err, cmd.SourceLocation, "invalid BUILD wildcard pattern")
```
alexcb, line 1273: the `"invalid BUILD wildcard pattern"` error message is confusing for cases where `ExpandWildcard` returns an error that's not related to the pattern (e.g. a network-related error)

mikejholly: Updated in upcoming commit. The language is now more generic.

mikejholly:
Updated with improved Git target globbing. `filepath.Match` is useful here.

buildcontext/git.go
```diff
@@ -310,6 +342,12 @@ func (gr *gitResolver) resolveGitProject(ctx context.Context, gwClient gwclient.
 		if err != nil {
 			return nil, errors.Wrap(err, "read git-refs")
 		}
+		earthfilePathsRaw, err := gitMetaRef.ReadFile(ctx, gwclient.ReadRequest{
```
vladaionescu, line 345: Just realizing that at some point it might be better to put all this metadata in a single file, to minimize the back-and-forth between earthly and buildkit with all these `ReadFile` requests. In any case, that can be a change that we make in a different PR.

buildcontext/git.go
```diff
@@ -235,6 +266,7 @@ func (gr *gitResolver) resolveGitProject(ctx context.Context, gwClient gwclient.
 					"git log -1 --format=%ae >/dest/git-author || touch /dest/git-author ; " +
 					"git log -1 --format=%b >/dest/git-body || touch /dest/git-body ; " +
 					"git for-each-ref --contains HEAD --format '%(refname:lstrip=-1)' >/dest/git-refs || touch /dest/git-refs ; " +
+					"find -name Earthfile > /dest/Earthfile-paths || touch /dest/Earthfile-paths ; " +
```
vladaionescu, line 269: Maybe add `-type f`, to ensure that they're files and not dirs?

buildcontext/git.go
```diff
@@ -63,7 +63,38 @@ type resolvedGitProject struct {
...
+		return nil, err
+	}
+
+	fullPattern := "./" + filepath.Join(subDir, pattern)
+
+	var matches []string
+
+	for _, path := range rgp.earthfilePaths {
+		path = strings.TrimSuffix(path, "/Earthfile")
+		ok, err := filepath.Match(fullPattern, path)
```
vladaionescu, line 86: Neat!

buildcontext/git.go
```diff
@@ -63,7 +63,38 @@ type resolvedGitProject struct {
...
+	if !target.IsRemote() {
+		return nil, errors.Errorf("unexpected local reference %s", target.String())
+	}
+
+	rgp, _, subDir, err := gr.resolveGitProject(ctx, gwClient, platr, target)
+	if err != nil {
+		return nil, err
+	}
+
+	fullPattern := "./" + filepath.Join(subDir, pattern)
```
vladaionescu, line 80: Nit: Not sure if necessary, but should this path be cleaned? Like if it contains things like `././foo/bar/../../buz`?

earthfile2llb/converter.go
```diff
@@ -1594,6 +1595,71 @@ func (c *Converter) Pipeline(ctx context.Context) error {
...
+		childTargetName := fmt.Sprintf("./%s+%s", match, parsedTarget.GetName())
+
+		childTarget, err := domain.ParseTarget(childTargetName)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to parse target %q", childTargetName)
+		}
+
+		data, _, _, err := c.ResolveReference(ctx, childTarget)
+		if err != nil {
+			notExist := buildcontext.ErrEarthfileNotExist{}
```
vladaionescu, line 1632: Just realizing another edge case. What if the Earthfile does exist, but it does not contain the target we're looking for? Should we also skip that error?

vladaionescu (commented):
Looks great! Some minor remaining remarks...

earthfile2llb/converter.go
mikejholly: IIRC `ExpandWildcard` & `ResolveReference` do not verify that the target exists, but there's a check on L1639 which does. There are also tests for this scenario.

buildcontext/git.go
mikejholly: Sure.

earthfile2llb/converter.go
vladaionescu: Oh yeah - makes sense!

buildcontext/git.go
mikejholly: `Join` should do most of the work but `Clean` will strip the `./` prefix. I'll add. 

https://go.dev/play/p/Q2GwzWroqVI

vladaionescu: Oh yeah, `Join` seems to be doing a clean of its own 👍

vladaionescu (approved):
🎉

buildcontext/git.go
mikejholly: `Clean` actually strips off the leading `./`, which the following code expects.

vladaionescu: SG!

alexcb (approved):
nice work!

PR #3671, opened 2024-01-04
pass default args when --pass-args is set

### LINKED ISSUE #3449: `--pass-args` should pass default `ARG`s too
consider

```
VERSION --pass-args 0.7
FROM alpine

a:
  ARG foo=alpha
  BUILD --pass-args +b

b:
  ARG foo=beta
  RUN --no-cache echo in b with $foo

test:
  ARG foo=hello
  BUILD --pass-args +a
```

When one runs `earthly +test`, it ultimately prints out `in b with beta`; which feels wrong.

Instead we expect the default value `ARG foo=hello` to be set, then passed via `BUILD --pass-args +a`, which ultimately should print out `in b with hello`.

alexcb:
fixes #3449

### DIFF
--- a/earthfile2llb/converter.go
+++ b/earthfile2llb/converter.go
@@ -1704,7 +1704,7 @@ func (c *Converter) EnterScopeDo(ctx context.Context, command domain.Command, ba
 		return err
 	}
 	if passArgs {
-		overriding = variables.CombineScopes(overriding, c.varCollection.Overriding())
+		overriding = variables.CombineScopesInactive(overriding, c.varCollection.Overriding(), c.varCollection.Args(), c.varCollection.Globals())
 	}
 	c.varCollection.EnterFrame(
 		scopeName, command, overriding, baseMts.Final.VarCollection.Globals(),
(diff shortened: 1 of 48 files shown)


earthfile2llb/converter.go
```diff
@@ -1828,7 +1828,9 @@ func (c *Converter) prepBuildTarget(ctx context.Context, fullTargetName string,
 	}
 	// Don't allow transitive overriding variables to cross project boundaries (unless --pass-args is used).
 	propagateBuildArgs := !relTarget.IsExternal()
-	if passArgs || propagateBuildArgs {
+	if passArgs {
+		overriding = variables.CombineScopes(overriding, c.varCollection.Overriding(), c.varCollection.Args(), c.varCollection.Globals())
+	} else if propagateBuildArgs || passArgs {
```
mikejholly, line 1833: This condition seems a bit confusing as it won't be executed if `passArgs` is true given L1831.

variables/scope.go
```diff
@@ -21,6 +22,22 @@ func NewScope() *Scope {
 	}
 }
 
+// DebugString returns a string that can be printed while debugging
+func (s *Scope) DebugString() string {
```
mikejholly, line 26: :+1:

mikejholly (approved):
LGTM.

earthfile2llb/converter.go
alexcb: good catch! that slipped back in when I was rebasing.

mikejholly (approved):

PR #3673, opened 2024-01-06
Clean up the builtin args, including platform

vladaionescu:


alexcb (approved):
we have a lot of args now.... this is much cleaner :)

PR #3677, opened 2024-01-08
Add support for additional sizes to launch and update satellite commands

dchw:
Adds support for our upcoming (2,3,4)xlarge instances

### DIFF
--- a/cloud/satellite.go
+++ b/cloud/satellite.go
@@ -39,11 +39,14 @@ const (
 )
 
 const (
-	SatelliteSizeXSmall = "xsmall"
-	SatelliteSizeSmall  = "small"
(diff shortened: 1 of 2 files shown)


vladaionescu (approved):

alexcb (approved):

mikejholly (approved):
LGTM. It might be worth looking at fetching this list from the server at some point.

dchw:
> LGTM. It might be worth looking at fetching this list from the server at some point.

Absolutely. To me, it seems like we shouldn't be validating this at all on the client; but instead prettify a server response code? I'm sure theres _reasons_ its like this, and fixing it would take longer than implementing the sizes themselves :(

PR #3680, opened 2024-01-09
Logstream: ensure no deltas are dropped on failed resume

mikejholly:
This change fixes a stream resumption issue where a delta may be dropped if the resume attempt immediately fails.  On L150, the code attempts to send the resume delta and any delta that previously failed to send. The `last` variable was not updated in the earlier code, so the `last` delta would not be properly passed to the subsequent retry. The new code ensures `last` is always set appropriately.

### DIFF
--- a/cloud/logstream_test.go
+++ b/cloud/logstream_test.go
@@ -35,59 +35,56 @@ func (s *streamNops) SendMsg(m interface{}) error  { return nil }
 func (s *streamNops) RecvMsg(m interface{}) error  { return nil }
 
 type flakyStream struct {
-	mu        sync.Mutex
-	fail      chan struct{}
-	err       error
-	calls     map[string]int
-	sent      []*pb.Delta
-	attempted []*pb.Delta
-	recv      chan *pb.StreamLogResponse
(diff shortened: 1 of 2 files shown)


cloud/logstream_test.go
```diff
@@ -123,18 +120,19 @@ func TestStreamLogs(t *testing.T) {
...
-	require.Equal(t, 1, stream.calls["Recv"], "expected 1 Recv")
+	require.Equal(t, stream.sendCalls, 12, "expected 10 Sends plus first manifest & EOF (12)")
 	require.Len(t, stream.sent, 11, "expected 10 deltas sent and 1 manifest (11)")
 }
 
 func TestStreamLogsResume(t *testing.T) {
-	stream := newFlakyStream()
-	streams := []*flakyStream{stream}
-
+	streams := []*flakyStream{newFlakyStream(4), newFlakyStream(0), newFlakyStream(-1)}
```
mikejholly, line 128: Each stream in this slice will fail after N send attempts.

alexcb (approved):

PR #3682, opened 2024-01-10
Add Logbus/Logstream target link to allow for association

mikejholly:
This is rather simple but it seems to work. Perhaps I'm missing something. 

Example debug output:
```
Parent target 523272c3-de17-45db-af41-372af13e1b0b +for-linux
Parent target dd0122f0-fcb3-4901-b512-f9fe4a470ccf ./ast/parser+parser
Parent target 523272c3-de17-45db-af41-372af13e1b0b +for-linux
Parent target 901cf3be-0061-49f3-99e6-1c98d9ed72af ./buildkitd+buildkitd
```

### DIFF
--- a/earthfile2llb/earthfile2llb.go
+++ b/earthfile2llb/earthfile2llb.go
@@ -276,9 +276,6 @@ func Earthfile2LLB(ctx context.Context, target domain.Target, opt ConvertOpt, in
 		}
 	}
 
-	// Update the target ID that child converters will reference as the parent.
-	opt.ParentTargetID = sts.ID
-
 	tiHash, err := sts.TargetInput().Hash()
 	if err != nil {
 		return nil, err
--- a/earthfile2llb/earthfile2llb.go
+++ b/earthfile2llb/earthfile2llb.go
@@ -183,6 +183,10 @@ type ConvertOpt struct {
 	// FilesWithCommandRenameWarning keeps track of the files for which the COMMAND => FUNCTION warning was displayed
 	// this can be removed in VERSION 0.8
 	FilesWithCommandRenameWarning map[string]bool
+
+	// ParentTargetID is the Logbus target ID of the parent target, if any. It
+	// is used to link together targets.
+	ParentTargetID string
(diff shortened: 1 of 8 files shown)


util/deltautil/apply.go
```diff
@@ -124,6 +124,9 @@ func setManifestFields(dm *pb.DeltaManifest, ret *pb.RunManifest) {
 		if t2.GetLocalPath() != "" {
 			t.LocalPath = t2.GetLocalPath()
 		}
+		if len(t2.GetDependsOn()) > 0 {
+			t.DependsOn = append(t.DependsOn, t2.GetDependsOn()...)
+		}
```
vladaionescu, line 129: Remember to update the serve-side with this part.

vladaionescu (commented):
It's a reasonable attempt :-)

There's one gotcha, though... the visited collection might de-duplicate the target, if it's referenced from multiple places, causing the converter to only run once. In such cases, the `dependsOn` might only mention a single parent.

For instance, does this get represented correctly?

```Earthly
all:
  BUILD +a
  BUILD +b

a:
  BUILD +final

b:
  BUILD +final

final:
  RUN echo foo
```

earthfile2llb/earthfile2llb.go
```diff
@@ -265,6 +269,16 @@ func Earthfile2LLB(ctx context.Context, target domain.Target, opt ConvertOpt, in
...
 	}
+
+	if opt.ParentTargetID != "" {
+		if parentTarget, ok := opt.Logbus.Run().Target(opt.ParentTargetID); ok {
+			parentTarget.AddDependsOn(sts.ID)
+		}
+	}
+
+	// Update the target ID that child converters will reference as the parent.
+	opt.ParentTargetID = sts.ID
```
vladaionescu, line 280: We generally prepare the updated `opt` object under the converter `prepBuildTarget` function. Can you move this line in there for consistency?

vladaionescu (approved):

PR #3701, opened 2024-01-12
test overridden args are used by expanded arg

alexcb:
This tests that a global arg that performs a shell-out correctly uses an overridden arg while expanding the argument value.

This additionally reverts a decision with the `--arg-scope-and-set` feature, which prevented overriding global args -- without the ability to override the global arg, it's not possible to expand an arg to the correct value.


tests/Earthfile
```diff
@@ -1075,17 +1075,11 @@ command:
...
-command-nested-global:
-    DO +RUN_EARTHLY --earthfile=command-nested-global.earth --target="+test-overriding --foo=expected"
-
 function:
     RUN echo "hello function" >./message.txt
     DO +RUN_EARTHLY --earthfile=function.earth --target=+all-positive
     DO +RUN_EARTHLY --earthfile=function.earth --should_fail=true --target=+test-command-fails
 
-function-explicit-global:
-    DO +RUN_EARTHLY --earthfile=function-explicit-global.earth
```
alexcb, line 1087: Note this test is no longer needed as it's covered by https://github.com/earthly/earthly/commit/97e8876af2bfe5464d66e79c5defce2bdb7a7ece

alexcb, line 1087: note that this test is no longer needed as it's covered by the new test in this PR

tests/Earthfile
```diff
@@ -171,6 +169,7 @@ ga-no-qemu-group4:
     BUILD --pass-args ./autoskip+test-all
     BUILD +if-exists
     BUILD +let-set
+    BUILD +let-scope
```
alexcb, line 172: this BUILD was accidentally missed when the target was added, fortunately it's still passing :sweat_smile:

mikejholly (approved):

PR #3704, opened 2024-01-13
test that functions do not propagate args

alexcb:
Unlike targets, calling a function does not cause args to be automatically propagated.

### DIFF
--- a/tests/functions-do-not-propagate-args/Earthfile
+++ b/tests/functions-do-not-propagate-args/Earthfile
@@ -1,0 +1,20 @@
+VERSION 0.8
+FROM --pass-args ..+base
+
+test-all:
+    BUILD +test-v07
+    BUILD +test-v08
+
+test-v07:
+    COPY propagate.earth Earthfile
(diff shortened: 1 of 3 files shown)


tests/functions-do-not-propagate-args/Earthfile
```diff
@@ -0,0 +1,20 @@
...
+
+test-v07:
+    COPY propagate.earth Earthfile
+    RUN sed -i "s/VERSION_STRING_TO_REPLACE/VERSION --use-function-keyword 0.7/g" Earthfile
+    DO --pass-args +RUN_EARTHLY --target=+test
+
+test-v08:
+    COPY propagate.earth Earthfile
+    RUN sed -i "s/VERSION_STRING_TO_REPLACE/VERSION 0.8/g" Earthfile
+    DO --pass-args +RUN_EARTHLY --target=+test
```
mikejholly, line 16: Can the output of `F` be checked here?

alexcb: might as well be extra safe and do that, good call.

mikejholly (approved):

PR #3707, opened 2024-01-15
split up tests-that-require-earthly-technologies-account-access

alexcb:
split up the the
`tests-that-require-earthly-technologies-account-access` target into the 4 seperate targets it builds.


alexcb:
If approved, I'll change the required tests to match the new split up ones.

mikejholly (approved):

PR #3712, opened 2024-01-16
do-not-track test: display stub server logs

alexcb:
If the do-not-track test fails, display the api.earthly.dev stub server logs.

### DIFF
--- a/tests/do-not-track/test-earthly-script.sh
+++ b/tests/do-not-track/test-earthly-script.sh
@@ -1,0 +1,30 @@
+#!/bin/sh
+set -e
+test -n "$earthly_config"
+
+# Start the stub server; it will create a file under /server-got-a-connection if any connections are received.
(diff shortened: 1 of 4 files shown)


tests/do-not-track/test-earthly-script.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/sh
+set -e
+test -n "$earthly_config"
+
+# Star the stub server; it will create a file under /server-got-a-connection if any connections are received
```
mikejholly, line 5: Super nit: "_Start_ the stub..."

mikejholly (approved):

tests/do-not-track/test-earthly-script.sh
alexcb: good catch :)

PR #3715, opened 2024-01-17
Satellites best practices guide

brandonSc:
A guide within our cloud/satellites doc section to help users get the most from their satellite usage.

Open to adding more best practices here if anyone has ideas.

### DIFF
--- a/docs/cloud/satellites/best-practices.md
+++ b/docs/cloud/satellites/best-practices.md
@@ -30,7 +30,7 @@ build-all:
   BUILD +library-2
 ```
 
-If the `+build-all` target is too intensive for a single satellite, we could instead create dedicated satellites for each library, and run their builds separately:
+If the `+build-all` target is too intensive for a single satellite, dedicated satellites can be created for each library:
 
 ```
 earthly --satellite sat-1 +library-1
--- a/docs/cloud/satellites/best-practices.md
+++ b/docs/cloud/satellites/best-practices.md
@@ -80,7 +80,7 @@ In this output, there are a couple of metrics which are worth keeping an eye on:
 
 On top of the metrics mentioned above, these other symptoms may indicate an overloaded satellite:
 * "No space left on device" errors
-* “Out of memory” errors
+* "Out of memory" errors
 * Op load metrics printing during the build
 * Noticeably long delays between steps in the build
 * New builds failing to connect to the satellite
--- a/docs/cloud/satellites/best-practices.md
+++ b/docs/cloud/satellites/best-practices.md
@@ -1,6 +1,6 @@
 # Satellites Best Practices
 
-Earthly Satellites are remote runner instances provided by the Earthly team, offering the ability to perform builds in the cloud while retaining cache between runs. This best practices guide aims to help users optimize their workflow and leverage the full potential of Earthly Satellites.
+Earthly Satellites are remote build runners hosted by Earthly, which enables building in the cloud with instantly available cache. This best practices guide aims to help users optimize their workflow and leverage the full potential of Earthly Satellites.
 
 ## Size Satellites for Builds
 
--- a/docs/cloud/satellites/best-practices.md
+++ b/docs/cloud/satellites/best-practices.md
@@ -1,6 +1,6 @@
 # Satellites Best Practices
 
-Earthly Satellites are remote build runners hosted by Earthly, which enables building in the cloud with instantly available cache. This best practices guide aims to help users optimize their workflow and leverage the full potential of Earthly Satellites.
+Earthly Satellites are [remote build runners](../../remote-runners.md) hosted by Earthly, which enables building in the cloud with instantly available cache. This best practices guide aims to help users optimize their workflow and leverage the full potential of Earthly Satellites.
 
 ## Size Satellites for Builds
 
@@ -87,4 +87,3 @@ On top of the metrics mentioned above, these other symptoms may indicate an over
 * Unexpected build failures or crashes during a build
 
 If you are unsure that your satellite is overloaded, please reach out via [email](mailto:support@earthly.dev) or our [Community Slack channel](https://earthly.dev/slack). An Earthly team member will be happy to investigate your instance and provide advice.
-
--- a/docs/cloud/satellites/best-practices.md
+++ b/docs/cloud/satellites/best-practices.md
@@ -1,0 +1,88 @@
+# Satellites Best Practices
+
+Earthly Satellites are remote runner instances provided by the Earthly team, offering the ability to perform builds in the cloud while retaining cache between runs. This best practices guide aims to help users optimize their workflow and leverage the full potential of Earthly Satellites.
+
+## Size Satellites for Builds
+
+Properly sizing Earthly Satellites is crucial to achieving optimal performance and resource utilization. Consider the following when determining which Satellite size is appropriate for your use case.
+
+### 1. Resource Requirements of a Single Build
+
+Understand the resource demands of your build, including memory usage, CPU requirements, and required cache size on disk. 
+
+For example, builds utilizing computationally intensive compilation steps, or large integration tests may require additional CPU and memory. Builds that download many gigabytes of dependencies may require a large amount of cache on disk.
+
+### 2. Concurrent Build Capacity
+
+Take into account the number of simultaneous builds that are likely to be run by your development team. The Satellite size should be chosen to handle the expected concurrency level effectively. 
+
+One metric to consider is the number of commits made per hour that will trigger builds on the satellite. For example, an xsmall or small satellites might handle a couple of concurrent builds, however, larger teams contributing more frequently may require a larger instance to accommodate the extra load.
+
+## Splitting Builds and Using More Satellites
+
+Large projects will often benefit from using multiple satellites in a single CI run, balancing the load across different machines in parallel. With this strategy, it’s best to optimize for the reuse of cache. Running the same target on the same satellite is more effective than alternating runs of the same target across different satellites.
+
+As an example, consider this Earthfile for a monorepo that contains a couple of libraries:
+
+build-all:
+  BUILD +library-1
+  BUILD +library-2
+
+If the +build-all target is too intensive for a single satellite, we could instead create dedicated satellites for each library, and run their builds separately:
+
+```
+earthly --satellite sat-1 +library-1
+earthly --satellite sat-2 +library-2
+```
+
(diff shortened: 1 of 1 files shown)


docs/cloud/satellites/best-practices.md
```diff
@@ -0,0 +1,90 @@
+# Satellites Best Practices
+
+Earthly Satellites are remote runner instances provided by the Earthly team, offering the ability to perform builds in the cloud while retaining cache between runs. This best practices guide aims to help users optimize their workflow and leverage the full potential of Earthly Satellites.
```
alexcb, line 3: reworded the first sentence to align with the text from https://earthly.dev/earthly-satellites
```suggestion
Earthly Satellites are remote build runners hosted by Earthly, which enables building in the cloud with instantly available cache. This best practices guide aims to help users optimize their workflow and leverage the full potential of Earthly Satellites.
```

Maybe we should even link to that page from the docs?

docs/cloud/satellites/best-practices.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Size Satellites for Builds
+
+Properly sizing Earthly Satellites is crucial to achieving optimal performance and resource utilization. Consider the following when determining which Satellite size is appropriate for your use case.
```
alexcb, line 7: "crucial to achieving optimal performance" is very wordy, can we condense this sentence?

alexcb: maybe even replace this entirely with:

```suggestion
Consider the following when picking a satellite size:
```

docs/cloud/satellites/best-practices.md
```diff
@@ -0,0 +1,90 @@
...
+
+### 1. Resource Requirements of a Single Build
+
+Understand the resource demands of your build, including memory usage, CPU requirements, and required cache size on disk. 
+
+For example, builds utilizing computationally intensive compilation steps, or large integration tests may require additional CPU and memory. Builds that download many gigabytes of dependencies may require a large amount of cache on disk.
+
+### 2. Concurrent Build Capacity
+
+Take into account the number of simultaneous builds that are likely to be run by your development team. The Satellite size should be chosen to handle the expected concurrency level effectively.
```
alexcb, line 17:
```suggestion
Take into account the number of simultaneous builds that are run by your development team. The Satellite size should be chosen to handle the expected concurrency level effectively. 
```

docs/cloud/satellites/best-practices.md
```diff
@@ -0,0 +1,90 @@
...
+
+Understand the resource demands of your build, including memory usage, CPU requirements, and required cache size on disk. 
+
+For example, builds utilizing computationally intensive compilation steps, or large integration tests may require additional CPU and memory. Builds that download many gigabytes of dependencies may require a large amount of cache on disk.
+
+### 2. Concurrent Build Capacity
+
+Take into account the number of simultaneous builds that are likely to be run by your development team. The Satellite size should be chosen to handle the expected concurrency level effectively. 
+
+One metric to consider is the number of commits made per hour that will trigger builds on the satellite. For example, an xsmall or small satellites might handle a couple of concurrent builds, however, larger teams contributing more frequently may require a larger instance to accommodate the extra load.
```
alexcb, line 19:
```suggestion
One metric to consider is the number of commits made per hour that will trigger builds on the satellite. For example, an xsmall or small satellites might handle a couple of concurrent builds; however, larger teams contributing more frequently may require a larger instance to accommodate the extra load.
```

docs/cloud/satellites/best-practices.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Splitting Builds and Using More Satellites
+
+Large projects will often benefit from using multiple satellites in a single CI run, balancing the load across different machines in parallel. With this strategy, it’s best to optimize for the reuse of cache. Running the same target on the same satellite is more effective than alternating runs of the same target across different satellites.
```
alexcb, line 23:
```suggestion
Large projects will often benefit from using multiple satellites in a single CI run, balancing the load across different machines in parallel. With this strategy, it’s best to optimize for cache reuse, by running the same target on the same satellite rather than alternating runs of the target across different satellites.
```

docs/cloud/satellites/best-practices.md
brandonSc: Yeah I'll add that too in a follow up commit

docs/cloud/satellites/best-practices.md
```diff
@@ -0,0 +1,90 @@
...
+
+As an example, consider this Earthfile for a monorepo that contains a couple of libraries:
```
alexcb, line 25:
```suggestion
As an example, consider an Earthfile that builds multiple libraries:
```

brandonSc:
Great revisions, thanks @alexcb !

docs/cloud/satellites/best-practices.md
```diff
@@ -0,0 +1,90 @@
...
+earthly --satellite sat-2 +library-2
+```
+
+This strategy is easiest to implement when there are distinct components that can be split apart. This is natural in monorepos, but the same considerations on cache reuse should be made in any case.
+
+## Cancelling Builds from Stale Commits
+
+When running Earthly on pull requests, it is common for new commits to be pushed before a previous build has finished. Since the previous results may no longer be useful, it’s good practice to cancel the previous build – freeing up resources on the satellite for the current build.
+
+Many CI systems allow this behavior to be configured. For example, on Github Actions the concurrency block can be used:
```
alexcb, line 46:
```suggestion
Many CI systems allow this behavior to be configured. Github Actions, for example, can be configured with:
```

docs/cloud/satellites/best-practices.md
```diff
@@ -0,0 +1,90 @@
...
+           satellite | Utilization: 0 other builds, 0/12 op load
+           satellite | GC stats: 2.1 MB cache, avg GC duration 0s, all-time GC duration 23ms, last GC duration 0s, last cleared 0 B
+```
+
+In this output, there are a couple of metrics which are worth keeping an eye on:
+* `Utilization` – The number of other concurrent builds and op load usage. The op load indicates the number of concurrent operations being executed, compared to the max allowed by the satellite, where an operation is a step within an Earthfile. When the op load is at the max, operations may begin to queue, resulting in slower performance.
+* `GC Stats` – The amount of cache on disk, and time spent doing garbage collection. Compare this to the amount of [disk available per the satellite’s size](https://earthly.dev/pricing). The cache typically hovers around 50% of the total disk size, however, if the cache size is close to the max, then aggressive garbage collection can cause degraded performance.
+
+On top of the metrics mentioned above, these other symptoms may indicate an overloaded satellite:
+* “No space left on device” errors
```
alexcb, line 82: stick with the ascii double-quote
```suggestion
* "No space left on device" errors
```

docs/cloud/satellites/best-practices.md
```diff
@@ -0,0 +1,90 @@
...
+
+jobs:
+  ... 
+```
+
+## Determining when Satellites are Overloaded
+
+A satellite may become sluggish, or even fail a build (in extreme cases) if it has taken on more jobs than it has the resources to handle. If you are experiencing a drop in performance, consider using a larger instance, or splitting up your build across multiple satellites. Note that in some cases, performance issues can be the result of a problem in the build script or external factors such as network operations.
+
+There are some metrics printed by the Earthly CLI that can be used to gauge the performance and overall health of a satellite. These metrics are printed at the start of the build during the “Init” phase, when running the satellite inspect command, or sometimes printed midway through the build if Earthly detects a performance problem. Here is an example of the metrics printed during Init:
```
alexcb, line 63:
```suggestion
There are some metrics printed by the Earthly CLI that can be used to gauge the performance and overall health of a satellite. These metrics are printed at the start of the build during the "Init" phase, when running the satellite inspect command, or sometimes printed midway through the build if Earthly detects a performance problem. Here is an example of the metrics printed during Init:
```

docs/cloud/satellites/best-practices.md
brandonSc: This slipped through a copy-and-paste from google doc

docs/cloud/satellites/best-practices.md
```diff
@@ -0,0 +1,90 @@
...
+On top of the metrics mentioned above, these other symptoms may indicate an overloaded satellite:
+* “No space left on device” errors
+* “Out of memory” errors
+* Op load metrics printing during the build
+* Noticeably long delays between steps in the build
+* New builds failing to connect to the satellite
+* Unexpected build failures or crashes during a build
+
+If you are unsure that your satellite is overloaded, please reach out via [email](mailto:support@earthly.dev) or our [Community Slack channel](https://earthly.dev/slack). An Earthly team member will be happy to investigate your instance and provide advice.
+
```
alexcb, line 90: nit: trailing newline

docs/cloud/satellites/best-practices.md
```diff
@@ -0,0 +1,90 @@
...
+## Size Satellites for Builds
+
+Properly sizing Earthly Satellites is crucial to achieving optimal performance and resource utilization. Consider the following when determining which Satellite size is appropriate for your use case.
+
+### 1. Resource Requirements of a Single Build
+
+Understand the resource demands of your build, including memory usage, CPU requirements, and required cache size on disk. 
+
+For example, builds utilizing computationally intensive compilation steps, or large integration tests may require additional CPU and memory. Builds that download many gigabytes of dependencies may require a large amount of cache on disk.
```
alexcb, line 13: can we point out `earthly --exec-stats +target` in some way here (or elsewhere in this doc?)

alexcb (commented):
looking good! might be good to get @vladaionescu to have a quick read over too.

docs/cloud/satellites/best-practices.md
alexcb: figured as much -- nice to have google docs take care of spell checking though :)

docs/cloud/satellites/best-practices.md
brandonSc: Good idea. Added this as a tip below. 

P.S. I was going to link it to a doc page on the `--exec-stats` page but noticed we didn't have one.

alexcb (approved):

vladaionescu (approved):

vladaionescu:
> looking good! might be good to get @vladaionescu to have a quick read over too.

:+1:

Issue #3475, opened 2023-11-08
ARG shellout doesn't correctly handle escaped `#`


alexcb:
**What went wrong?**
```Earthfile
thebug:
    FROM alpine
    ARG myarg=$(echo "a#b#c" | cut -f2 -d\#)
    RUN touch /some-file
    RUN echo "myarg is \"$myarg\""
    RUN test -f /some-file
```

fails with:

```
              alpine | --> Load metadata alpine linux/amd64
             +thebug | --> FROM alpine
             +thebug | [----------] 100% FROM alpine
             +thebug | *cached* --> mkdir /run/buildargs
             +thebug | *cached* --> expandargs echo "a#b#c" | cut -f2 -d#
             +thebug | --> RUN echo "myarg is \"$myarg\""
             +thebug | myarg is "b
             +thebug |     RUN touch /some-file"
             +thebug | --> RUN test -f /some-file
             +thebug | ERROR Earthfile line 27:4
             +thebug |       The command
             +thebug |           RUN test -f /some-file
             +thebug |       did not complete successfully. Exit code 1
```

Note that `ARG myarg ...` is incorrectly including the following `RUN touch /some-file` command:

```
             +thebug | myarg is "b
             +thebug |     RUN touch /some-file"
```

which ultimately means the `RUN touch /some-file` doesn't get executed, which causes `RUN test -f /some-file` to fail.

alexcb:
work-around: use `"#"` instead of `\#`, e.g.

```
    ARG myarg=$(echo "a#b#c" | cut -f2 -d"#")
```

Issue #3131, opened 2023-07-21
Quoting issues with `RUN` and "shell-out" executions


rrjjvv:
**What went wrong?**

Attempting to pass a single double-quote to commands in some shell scenarios fail in various ways.  For example, trying to remove quotes from strings via `tr -d '"'` or `tr -d "\""`, which should be functionally equivalent, fail in their own ways.  I don't know if this is restricted to passing a single double-quote; I did not explicitly try scenarios with single single-quotes.

Below are the different variations.  Since the first two are lexer issues, they had to be broken into separate files to illustrate.  (When they are in the same file as the others, depending where they are physically located, they sometimes execute, but with very wrong results.)

Summary:

|  | `'"'` | `"\""` |
| - | --- | ------ |
| `RUN` (earthly) | shell error | works |
| `ARG` shell-out | lex error | lex error |
| `RUN` (Dockerfile) | works | works |

```Earthfile
VERSION 0.7
FROM alpine

run-plain:
    RUN echo run | tr -d '"'
```
```
$ earthly +run-plain
Error: build target: build main: failed to solve: resolve build context for target +run-plain: lexer error: Earthfile
syntax error: line 5:26: token recognition error at: '"'

'
```

```Earthfile
VERSION 0.7
FROM alpine

arg-plain:
    ARG val=$(echo run | tr -d '"')
    RUN echo $val
```
```
$ earthly +arg-plain
Error: build target: build main: failed to solve: resolve build context for target +arg-plain: lexer error: Earthfile
syntax error: line 5:32: token recognition error at: '"')
    RUN echo $val
'
```

```Earthfile
VERSION 0.7
FROM alpine

# This variant works correctly
run-esc:
    RUN echo single | tr -d "\""

# The shell-out form should also work, but fails at runtime
arg-esc:
    ARG val=$(echo single | tr -d "\"")
    RUN echo $val

# The same constructs work fine when not parsed by earthly itself
doc-plain:
    FROM DOCKERFILE -f +dockerfiles/Dockerfile.plain .

doc-esc:
    FROM DOCKERFILE -f +dockerfiles/Dockerfile.esc .


dockerfiles:
    WORKDIR work
    RUN echo "
FROM alpine
RUN echo run | tr -d '\"'
" > Dockerfile.plain

    RUN echo "
FROM alpine
RUN echo run | tr -d \"\\\"\"
" > Dockerfile.esc
    SAVE ARTIFACT *
```
```
$ earthly +arg-esc
[...]

Issue #1280, opened 2021-10-08
Lexer error: unable to handle `#` in single quoted string


zaufi:
The following sample `Earthfile` can't be parsed properly:

```
test:
    FROM debian:9
    RUN set -x \
     && sed -i \
            -e 's, universe multiverse, universe # multiverse,' \
            /etc/apt/sources.list
    SAVE IMAGE --push blah
```

Trying to build fails with the error:

```console
$ earthly +test
           buildkitd | Found buildkit daemon as docker container (earthly-buildkitd)
Error: build target: build main: bkClient.Build: failed to solve: resolve build context for target +test: lexer error: Earthfile
syntax error: line 6:12 token recognition error at: '/'
syntax error: line 6:13 token recognition error at: 'etc/'
syntax error: line 6:17 token recognition error at: 'apt/'
syntax error: line 6:21 token recognition error at: 'sources.list
'
```

Rewriting the command one line (in my real-life case the line is too long, that is why I split it (and it was Ok for `Dockerfile`) do not fail with lexer, but fall to execute:
```
test:
    FROM debian:9
    RUN sed -i -e 's, universe multiverse, universe # multiverse,' /etc/apt/sources.list
    SAVE IMAGE --push blah
```

with the following:
``` console
$ earthly +test
           buildkitd | Found buildkit daemon as docker container (earthly-buildkitd)
            debian:9 | --> Load metadata linux/amd64
               +test | --> FROM debian:9
               +test | [██████████] resolve docker.io/library/debian:9@sha256:c1274c24446b98394e1594ceefac1012e1b39b752586ca993fe3c3bea6f6d20f ... 100%
               +test | --> RUN sed -i -e 's, universe multiverse, universe
               +test | /bin/sh: 1: Syntax error: Unterminated quoted string
               +test | Command /bin/sh -c 'sed -i -e '"'"'s, universe multiverse, universe' failed with exit code 2
               +test |
               +test | ERROR: Command exited with non-zero code: RUN sed -i -e 's, universe multiverse, universe
Repeating the output of the command that caused the failure
================================ FAILURE [main] ================================
               +test *failed* | --> RUN sed -i -e 's, universe multiverse, universe
               +test *failed* | /bin/sh: 1: Syntax error: Unterminated quoted string
               +test *failed* | Command /bin/sh -c 'sed -i -e '"'"'s, universe multiverse, universe' failed with exit code 2
[...]

PR #3716, opened 2024-01-17
Fix single quoted # lexing

mikejholly:
Addresses: https://github.com/earthly/earthly/issues/1280, https://github.com/earthly/earthly/issues/3475, & some of https://github.com/earthly/earthly/issues/3131.

The lexer rules were not correctly handling single-quoted strings & escaped characters in certain contexts.


Earthfile
```diff
@@ -155,7 +155,7 @@ lint-newline-ending:
     # test that line endings are unix-style
     RUN set -e; \
         code=0; \
-        for f in $(find . -not -path "./.git/*" -type f \( -iname '*.go' -o -iname 'Earthfile' -o -iname '*.earth' -o -iname '*.md' -o -iname '*.json'\) | grep -v "ast/tests/empty-targets.earth" ); do \
+        for f in $(find . -not -path "./.git/*" -type f \( -iname '*.go' -o -iname 'Earthfile' -o -iname '*.earth' -o -iname '*.md' -o -iname '*.json' \) | grep -v "ast/tests/empty-targets.earth" ); do \
```
mikejholly, line 158: Once I fixed the lexer rules for `$()`, the shell command was correctly parsed as a single token, rather than a series of values that were later joined.

The missing space before `\)` is invalid when parsed correctly.
```
$ find . -not -path "./.git/*" -type f \( -iname '*.go' -o -iname 'Earthfile' -o -iname '*.earth' -o -iname '*.md' -o -iname '*.json'\) 
find: invalid expression; I was expecting to find a ')' somewhere but did not see one.
```

ast/ast_test.go
```diff
@@ -1,11 +1,10 @@
-package ast_test
+package ast
```
mikejholly, line 1: I'm not sold on the merits of using separate testing packages. The advertised benefit is that one should only test exposed interfaces. I think it's ok to write tests for unexported code as well. For example, sometimes there's a complicated helper function that warrants more attention.

vladaionescu (commented):
Looks good to me!

Can you also update the changelog? It's important to articulate well what has changed here, and how people may need to migrate Earthfiles accordingly.

@alexcb - FYI, I told @mikejholly that this is a rare opportunity to change the lexer without a feature flag if we can make changes as part of 0.8. It's normally really difficult to change the lexer, since it's hard to get a feature flag in there.

ast/ast_test.go
alexcb: > sometimes there's a complicated helper function that warrants more attention.

Have you added such a test in this PR? (I quickly scanned over the test but it wasn't clear).

If this change isn't needed for this PR -- I would rather undo this change and leave it for a separate discussion.

alexcb: it's also possible to have multiple test files, some with `package ast_test` and others with `package ast`, which can be used to communicate the goal of the file -- testing exposed vs internal functions.

mikejholly: > Have you added such a test in this PR?

No, that was just a general comment. It didn't strike me as necessary to add tests for unexported code in this PR.

I intended to include the package rename change in https://github.com/earthly/earthly/pull/3717, but left it out by accident.

Do you have concerns about the change? It seems like this pattern is only used with the `onpar` tests. The differing package names also confuse my editor.

alexcb: The "_test" package idiom is nice because it allows us to communicate that this test is only for external methods. My concern is that we'll loose the ability to have go enforce this convention.

mikejholly: Right, but it's not like it was used extensively here or that removing it introduces further risk. If anything removing it reduces complexity a bit. I just considered this part of the test cleanup & refactoring.

alexcb: my general preference is if a change isn't needed, and the existing code is working, we just leave it as-is, but if you strongly feel that the "_test" package idiom should be avoided then I don't have a strong opinion against the change.

alexcb: We do use the "_test" package idiom in other places too, e.g. https://github.com/earthly/earthly/blob/587f9f694fc093d0c640a6024c93442422d18fa2/util/buildkitskipper/hasher/hasher_test.go#L1

we should come to a decision on if it's something we want to remove everywhere.

mikejholly: > change isn't needed, and the existing code is working, we just leave it as-is

Many changes aren't strictly needed but may simplify or improve. 

> remove everywhere

I don't agree that the same approach needs to be used everywhere. The context is important. 

I'll remove this change for now as it's now a bit of a distraction. My original intention was to tidy up and simplify the tests.

mikejholly:
@vladaionescu @alexcb updated with changelog commands & package name revert.

CHANGELOG.md
```diff
@@ -70,6 +70,10 @@ It should be noted that some of these features break backwards compatibility. Se
 
 - Parallelism is improved when running the same target with different arguments in certain cases (e.g. the target uses `WITH DOCKER`).
 - Fixed a log sharing upload-resumption bug
+- Fixed multiple issues with the lexer failing to parse certain characters in shell command substitution (`$()`) and single quoted strings.
```
alexcb, line 73: this should go under `unreleased` since it wasn't fixed in v0.8.0-rc2

mikejholly: Updated.

CHANGELOG.md
```diff
@@ -70,6 +70,10 @@ It should be noted that some of these features break backwards compatibility. Se
...
+  - Some escaped characters, like `\#`, were failing to parse when used inside shell expressions. Example: `$(echo "a#b#c" | cut -f2 -d\#)`
+  - Some characters, like `#`, were failing to parse when used inside single-quoted strings: Example: `'this is a # string'`
+  - Some invalid shell commands may have been parsed without issue. Example: `$(find . \( -name 'Earthfile'\))` is missing a space before `\)`
```
alexcb: I don't think the example is accurate, because if I create an Earthfile as:

```
VERSION 0.7

test:
  FROM alpine
  RUN mkdir sub
  RUN touch sub/Earthfile
  RUN echo $(find . \( -name 'Earthfile'\))
```

when I run `earthly-v0.7.23 +test` it fails with:

```
               +test | --> RUN echo $(find . \( -name 'Earthfile'\))
               +test | find: unpaired '('
```

however I thought the case you found would cause `find` to work correctly under v0.7.23 but not with this fix.

alexcb: I just tested our root Earthfile with:

```
earthly-v0.7.23 --no-cache +lint-newline-ending
```

and it contains the same `find: unpaired '('` error:

```
+lint-newline-ending | --> RUN export f=ast/tests/empty-targets.earth && if [ "$(tail -c 1 $f)" = "$(printf '\n')" ]; then echo "$f is a special-case test which must not end with a newline."; exit 1; fi
+lint-newline-ending | DOCKERHUB_MIRROR_AUTH_FROM_CLOUD_SECRETS=true
+lint-newline-ending | --> RUN set -e; code=0; for f in $(find . -not -path "./.git/*" -type f \( -iname '*.go' -o -iname 'Earthfile' -o -iname '*.earth' -o -iname '*.md' -o -iname '*.json'\) | grep -v "ast/tests/empty-targets.earth" | grep -v "ast/parser/earth_parser.go" | grep -v "ast/parser/earth_lexer.go" ); do if [ "$(tail -c 2 $f)" == "$(printf '\n\n')" ]; then echo "$f has trailing newlines"; code=1; fi; done; exit $code
+lint-newline-ending | find: unpaired '('
```

could it be that we've just always had this error and never noticed it? If that's the case we could just delete this line.

CHANGELOG.md
mikejholly: Oh, interesting. It turns out this hasn't been working in `main`. I noticed this while testing and thought it must have been due to the lexer changes. I'll remove it. 

https://github.com/earthly/earthly/actions/runs/7561776859/job/20590798774#step:4:35

CHANGELOG.md
mikejholly: Oops, I added this comment to the wrong thread. Just for completeness:

--

Oh, interesting. It turns out this hasn't been working in main. I noticed this while testing and thought it must have been due to the lexer changes. I'll remove it.

https://github.com/earthly/earthly/actions/runs/7561776859/job/20590798774#step:4:35

alexcb (approved):
nice work getting these errors fixed, they've been here for quite some time now.

PR #3721, opened 2024-01-19
More quote parsing fixes

mikejholly:
Fixes remaining issues described in https://github.com/earthly/earthly/issues/3131

Escaped double-quotes like `"\""` were incorrectly translated into `"""` before being passed to BuildKit.

Also includes a few extra lexer tests to ensure the other issues in #3131 are working.


vladaionescu (commented):
I'll leave the main review to someone more knowledgeable. Just came here to ask for a changelog entry too :-)

alexcb:
can we add a test under `./test/...` too, maybe along the lines of:

```
# quote.earth
VERSION 0.8
FROM alpine

test:
    RUN echo '"' > data
    RUN echo "'" >> data
    RUN echo $(echo '"') >> data
    RUN echo $(echo "'") >> data
    RUN echo $(echo "\"") >> data
    ARG a=$( echo $(echo '"'))
    ARG b=$( echo $(echo "'"))
    ARG c=$( echo $(echo "\""))
    RUN env | grep ^a=.*$ >> data
    RUN env | grep ^b=.*$ >> data
    RUN env | grep ^c=.*$ >> data
    SAVE ARTIFACT data AS LOCAL actual
```

then create an `expected` file such as:

```
"
'
"
'
"
a="
b='
c="
```

then ultimately we can do a diff on the expected vs actual:

```
test-quote:
    COPY expected .
    DO +RUN_EARTHLY --earthfile=quotes.earth --target=+test
    RUN diff actual expected
```

this will allow us to be certain that the argument substitution/shell outs are expected and the expected results are not modified by the ast code in any way (since it's just a file we copy in)

mikejholly:
@alexcb @vladaionescu Updated with new tests & changelog entry.

alexcb (approved):
looks good, I even checked this out and tried hard at breaking it (but couldn't).

PR #3723, opened 2024-01-19
Put satellites to sleep before starting updates

brandonSc:
Satellites must be asleep to start a manual update. We would previously produce an error if the satellite was not asleep, but this has been confusing to users.

This PR introduces ~~a prompt which will ask the user permission~~ flag which will need to provided before automatically sleeping the satellite and starting the update.

Screen Shot 2024-01-19 at 14 49 28

### DIFF
--- a/cmd/earthly/subcmd/satellite_cmds.go
+++ b/cmd/earthly/subcmd/satellite_cmds.go
@@ -18,6 +18,7 @@ import (
 	"github.com/earthly/earthly/buildkitd"
 	"github.com/earthly/earthly/cloud"
 	"github.com/earthly/earthly/cmd/earthly/base"
+	"github.com/earthly/earthly/cmd/earthly/common"
 	"github.com/earthly/earthly/cmd/earthly/helper"
 	"github.com/earthly/earthly/config"
 	"github.com/earthly/earthly/conslogging"
@@ -965,9 +966,9 @@ func (a *Satellite) actionUpdate(cliCtx *cli.Context) error {
 		return err
 	}
 
-	satName, err := base.GetSatelliteName(cliCtx.Context, orgName, a.cli.Flags().SatelliteName, cloudClient)
+	sat, err := cloudClient.GetSatellite(cliCtx.Context, a.cli.Flags().SatelliteName, orgName)
 	if err != nil {
-		return err
+		return errors.Wrap(err, "failed getting satellite")
 	}
 
 	if window != "" {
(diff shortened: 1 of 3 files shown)


cmd/earthly/subcmd/satellite_cmds.go
```diff
@@ -987,8 +988,30 @@ func (a *Satellite) actionUpdate(cliCtx *cli.Context) error {
 		return errors.Errorf("not a valid platform: %q", platform)
 	}
 
+	if sat.State != cloud.SatelliteStatusSleep {
+		a.cli.Console().Printf("")
+		a.cli.Console().Printf("The satellite must be asleep to start the update.")
+		a.cli.Console().Printf("Putting the satellite to sleep will interrupt any running builds.")
+		a.cli.Console().Printf("")
+		answer, err := common.PromptInput(cliCtx.Context, "Would you like to put it to sleep now? [y/N]: ")
```
alexcb, line 996: can we have a `--force` option to override this

cmd/earthly/subcmd/satellite_cmds.go
```diff
@@ -987,8 +988,30 @@ func (a *Satellite) actionUpdate(cliCtx *cli.Context) error {
...
+		if err != nil {
+			return errors.Wrap(err, "failed to read permission")
+		}
+		answer = strings.ToLower(answer)
```
alexcb, line 1000: nit: I think we should refactor this into a common `IsResponseYes(s string) bool` function, then make use of it here and https://github.com/earthly/earthly/blob/b337b88e4e78827c0566d009161ed9dee95b0a87/cmd/earthly/subcmd/project_cmds.go#L183

alexcb (commented):
thanks for tackling this, just a couple small suggestions.

brandonSc:
Thanks for the quick review @alexcb - I've added your suggestions

cmd/earthly/subcmd/satellite_cmds.go
vladaionescu: I would go even further and never make anything interactive, because it makes scripting more difficult. It's probably best to just never have this kind of option and it's always forced. It's not like someone will be like "oh I didn't know that it was sleeping... maybe I won't bother it then" :-)

cmd/earthly/subcmd/satellite_cmds.go
```diff
@@ -989,20 +997,21 @@ func (a *Satellite) actionUpdate(cliCtx *cli.Context) error {
...
-			a.cli.Console().Printf("Update aborted.")
-			return nil
+		if !a.forceUpdate {
+			a.cli.Console().Printf("")
+			a.cli.Console().Printf("The satellite must be asleep to start the update.")
+			a.cli.Console().Printf("Putting the satellite to sleep will interrupt any running builds.")
+			a.cli.Console().Printf("")
+			answer, err := common.PromptInput(cliCtx.Context, "Would you like to put it to sleep now? [y/N]: ")
+			if err != nil {
+				return errors.Wrap(err, "failed to read permission")
```
alexcb, line 1007:
```suggestion
				return errors.Wrap(err, "failed to prompt input")
```

cmd/earthly/subcmd/satellite_cmds.go
alexcb: we can alternatively call https://github.com/earthly/earthly/blob/761740f72931261bf6f7a10168d006e398388057/util/termutil/is_tty.go#L6

vladaionescu: I think Alex's original suggestion to have a `--force` option makes sense. And perhaps it just fails without `--force`, if it's not already asleep. The failure message should perhaps include the suggestion to use `--force`.

BTW, is there a good way to tell whether a build is actually ongoing, and only fail in that case (but not if the satellite is awake and idle)? This might need some sort of support end-to-end to make it atomic perhaps.

brandonSc: We can detect if there are no builds ongoing, but to make it atomic, we'd need backend support (as you mentioned). I think the backend support could be bit for work than it's worth (e.g. different API that supports streaming, or some way to lock the satellite momentarily blocking new builds).

I think I would vote for the error message just telling the user to re-run with `--force` , but personally I do think the interactive mode will be more user-friendly in most cases (with `--force` still being script-friendly)

vladaionescu: Makes sense. BTW we made the decision to avoid interactive for many other commands. We only kept it for things that are never going to be used for scripting for sure (like account registration). I guess in this specific case, it's a consistency argument more than anything.

brandonSc: Fair enough - I wasn't sure if this would be a common command to script, but I see the value in consistency. 

Shall we proceed with showing an error and prompting to run with `--force`  if the satellite is awake instead?

vladaionescu: SGTM!

alexcb (approved):

vladaionescu:
Hey @brandonSc - can you also add a changelog entry?

PR #3733, opened 2024-01-22
Display warning message when configured cache size is too small

mikejholly:
It looks like the computed cache values are already exposed by the `ListWorkers` endpoint via the `GCPolicy` attribute. 

The [`CACHE_ALL_KEEP_BYTES`](https://github.com/earthly/earthly/blob/main/buildkitd/entrypoint.sh#L186) variable is set to the byte value of `CACHE_SIZE_MB`, so it should be ok to rely on this value.

It should be possible to add the `gckeepstorage` value to the endpoint response, but that will require a few code changes in BuildKit. 

image

### DIFF
--- a/buildkitd/buildkitd.go
+++ b/buildkitd/buildkitd.go
@@ -938,19 +938,20 @@ func printBuildkitInfo(bkCons conslogging.ConsoleLogger, info *client.Info, work
 		}
 	}
 
-	var cacheInfo *client.PruneInfo
-	for _, p := range workerInfo.GCPolicy {
-		if p.All {
-			cacheInfo = &p
-			break
-		}
-	}
-
-	if cacheInfo != nil && cacheInfo.KeepBytes < minRecommendedCacheSize {
+	if size, ok := getGCPolicySize(*workerInfo); ok && size < minRecommendedCacheSize {
 		bkCons.Warnf("Configured cache size of %s is smaller than the minimum recommended size of %s",
-			units.HumanSize(float64(cacheInfo.KeepBytes)), units.HumanSize(minRecommendedCacheSize))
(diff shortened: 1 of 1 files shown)


buildkitd/buildkitd.go
```diff
@@ -934,6 +937,20 @@ func printBuildkitInfo(bkCons conslogging.ConsoleLogger, info *client.Info, work
 		default:
 		}
 	}
+
+	var cacheInfo *client.PruneInfo
+	for _, p := range workerInfo.GCPolicy {
```
alexcb, line 942: nit: move this logic into a helper func so we could do `cacheSize, ok := getGCPolicySize(workerInfo.GCPolicy)` rather than have to worry about `cacheInfo` being nil.

alexcb (approved):
cool to see it can be done without any buildkit code changes!

buildkitd/buildkitd.go
mikejholly: Updated.

It's interesting how verbosity or DRY differ in Go versus other languages :slightly_smiling_face:

alexcb (approved):

buildkitd/buildkitd.go
```diff
@@ -30,6 +31,8 @@ import (
 	"github.com/earthly/earthly/util/semverutil"
 )
 
+const minRecommendedCacheSize = 10 << 30 // 10 GiB
```
vladaionescu, line 34: `10.74 GB` looks a bit strange. Shall we use simply `10,000,000,000`?

mikejholly: That's fine by me. Some of the other code was using GiB so I assumed that was the preference.

Issue #3740, opened 2024-01-23
Mac registry proxy support container fails to start


mikejholly:
**What went wrong?**

Mac registry proxy container seems to produce this error without ending the build. 

```
registry-proxy | Failed to start registry proxy: failed to start Darwin support container: context deadline exceeded
```

**What should have happened?**

No error here & continue as normal.

PR #3741, opened 2024-01-23
Fix issue with Darwin proxy container wait time

mikejholly:
Fixes: https://github.com/earthly/earthly/issues/3740

The Darwin proxy healthcheck was timing out immediately. This wasn't obvious on faster machines where the proxy came up immediately. 

I also renamed a context value to disambiguate it from the one in the `defer`. 

Thanks to @brandonSc for the testing and confirmation help!

### DIFF
--- a/regproxy/controller.go
+++ b/regproxy/controller.go
@@ -45,6 +45,7 @@ func NewController(
 		containerFrontend: containerFrontend,
 		darwinProxy:       darwinProxy,
 		darwinProxyImage:  darwinProxyImage,
+		darwinProxyWait:   darwinProxyWait,
 		cons:              cons,
 	}
 }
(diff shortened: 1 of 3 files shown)


vladaionescu (approved):
🐛 🦶

regproxy/controller.go
```diff
@@ -45,6 +45,7 @@ func NewController(
 		containerFrontend: containerFrontend,
 		darwinProxy:       darwinProxy,
 		darwinProxyImage:  darwinProxyImage,
+		darwinProxyWait:   darwinProxyWait,
```
alexcb, line 48: :detective: :heavy_check_mark:

alexcb (approved):
good find!

PR #3742, opened 2024-01-23
override buildkit's automatic cache size

alexcb:
automatically set cache size to min(55%, max(10%, 20GB)), rather than using buildkit's 10% default.

### DIFF
--- a/buildkitd/entrypoint.sh
+++ b/buildkitd/entrypoint.sh
@@ -177,16 +177,11 @@ if [ "$CACHE_SIZE_PCT" -gt "0" ]; then
     fi
 fi
 
-# Remains unset if neither percent nor size were specified.  It would be simpler to just process whether it was
+# EFFECTIVE_CACHE_SIZE_MB remains unset if neither percent nor size were specified.  It would be simpler to just process whether it was
 # set (or not), but we'll continue setting to "0" in case anyone has become dependent on that behavior.
 CACHE_SIZE_MB="${EFFECTIVE_CACHE_SIZE_MB:-0}"
-if [ "$CACHE_SIZE_MB" -gt "0" ]; then
-    SOURCE_FILE_KEEP_BYTES="$(echo "($CACHE_SIZE_MB * 1024 * 1024 * 0.5) / 1" | bc)" # Note /1 division truncates to int
-    export SOURCE_FILE_KEEP_BYTES
-    CATCH_ALL_KEEP_BYTES="$(echo "$CACHE_SIZE_MB * 1024 * 1024" | bc)"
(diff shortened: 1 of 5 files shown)


buildkitd/entrypoint.sh
```diff
@@ -186,6 +186,19 @@ if [ "$CACHE_SIZE_MB" -gt "0" ]; then
...
+    # no config value was set by the user; buildkit would set this to 10% by default:
+    #   https://github.com/moby/buildkit/blob/54b8ff2fc8648c86b1b8c35e5cd07517b56ac2d5/cmd/buildkitd/config/gcpolicy_unix.go#L16
+    # however, we will be aggresive and set it to min(55%, max(10%, 20GB))
+    CACHE_MB_10PCT="$(stat -c "10 * %b * %S / 100 / 1024 / 1024" -f "$EARTHLY_TMP_DIR" | bc)"
+    CACHE_MB_55PCT="$(stat -c "55 * %b * %S / 100 / 1024 / 1024" -f "$EARTHLY_TMP_DIR" | bc)"
+    CACHE_SIZE_MB="20480" # first start with 20GB
+    if [ "$CACHE_MB_10PCT" -gt "$CACHE_SIZE_MB" ]; then
+        CACHE_SIZE_MB="$CACHE_MB_10PCT" # increase it to 10% of the disk if bigger
+    elif [ "$CACHE_MB_55PCT" -lt "$CACHE_SIZE_MB" ]; then
+        CACHE_SIZE_MB="$CACHE_MB_55PCT" # otherwise, prevent it from being bigger than 55% of the disk
```
mikejholly, line 199: The `envsubst` command on L188 will still have to be run, no? I don't think that the new `CACHE_SIZE_MB` in the else branch will be used as it stands.

vladaionescu (commented):
Should we also update minRecommendedSize that was introduced in https://github.com/earthly/earthly/pull/3733?

buildkitd/entrypoint.sh
mikejholly: Also, my earlier PR uses the value written out from `CATCH_ALL_KEEP_BYTES` to determine whether we display the warning message. 

https://github.com/earthly/earthly/pull/3733

alexcb: very good catch :)

alexcb:
> Should we also update minRecommendedSize that was introduced in https://github.com/earthly/earthly/pull/3733?

The warning could be annoying to users on smaller disks that hit the 55% limit.

maybe we should only display it when the user hasn't configured any cache_size_mb or cache_size_pct config options?

mikejholly (approved):
LGTM

Issue #3736, opened 2024-01-23
SAVE IMAGE connection refused


mikejholly:
**What went wrong?**

A user reported the following error when he attempted to run a build with a `SAVE IMAGE` on Jenkins CI with Linux.

```
[2024-01-22T23:20:34.964Z]               output | --> exporting outputs
[2024-01-22T23:20:47.159Z]               output | [----------] 100% exporting outputs[K
[2024-01-22T23:20:55.244Z] Error: pull ping error: pull ping response: rpc error: code = Unknown desc = image pull: 1 error occurred:
[2024-01-22T23:20:55.244Z] 	* command failed: docker pull 127.0.0.1:35137/sess-jwd728vcbn4ni32eu360wnxmi/pullping:img-0: exit status 1: Error response from daemon: Get "http://127.0.0.1:35137/v2/": dial tcp 127.0.0.1:35137: connect: connection refused: exit status 1
```

**What should have happened?**

The build should have been completed as usual.

**Other Helpful Information**

https://earthlycommunity.slack.com/archives/C01DL2928RM/p1705973717313219

mikejholly:
Workaround:

mikejholly:
It looks as though the registry proxy is failing after startup. If it fails on start-up, the build should fall back to tar-based image transfer.

https://github.com/earthly/earthly/blob/main/builder/builder.go#L198

mikejholly:
This message in a user's logs indicates that the proxy started correctly. 

```
registry-proxy | Starting registry proxy on 127.0.0.1:33849
```

mikejholly:
It turns out the user was running Earthly in a Docker container. 

Here's how to reproduce:

```
docker run --privileged -e NO_BUILDKIT=1 -v /var/run/docker.sock:/var/run/docker.sock -v $(pwd):/workspace -v earthly-tmp:/tmp/earthly:rw  -it docker.io/earthly/earthly -V +for-linux
```

Excerpt:
```
Error: pull ping error: pull ping response: rpc error: code = Unknown desc = image pull: 1 error occurred:
	* command failed: docker pull 127.0.0.1:34133/sess-kx4b26mhnc5s3ewlbj57ozoah/pullping:img-0: exit status 1: Error response from daemon: Get "http://127.0.0.1:34133/v2/": dial tcp 127.0.0.1:34133: connect: connection refused: exit status 1

```

mikejholly:
Another workaround would be to use `--network=host` when running the container.

[...]

PR #3753, opened 2024-01-25
Disable registry proxy when running in container

mikejholly:
Addresses: https://github.com/earthly/earthly/issues/3736

See comment in Earthfile :slightly_smiling_face:


alexcb (approved):

mikejholly:
> I wonder if we could somehow detect if `--network=host` is specified, and if so maybe not set the env variable -- but this feels like a bigger (lower priority) item which we should create a separate issue for.

Yeah, it would be nice to make this a little smarter. I'll create another issue.

mikejholly:
https://github.com/earthly/earthly/issues/3754

Issue #3761, opened 2024-01-29
Token-based login is leaving Eearthly in a partially logged-in state


mikejholly:
**What went wrong?**

`earthly account login --token=<token>` login, when SSH is disabled, doesn't appear to work correctly.

**What should have happened?**

Earthly should remain logged in after a successful `--token` login.

**Other Helpful Information**

How to reproduce: 

```
➜  earthly git:(main) export SSH_AUTH_SOCK=""
➜  earthly git:(main) rm ~/.earthly/auth.*                                                                          
➜  earthly git:(main) earthly account login                                                                         
➜  earthly git:(main) earthly account login --token=xxxxx
Logged in as "mike@earthly.dev" using token auth
➜  earthly git:(main) earthly org select earthly-technologies
Updated selected org in /home/mike/.earthly/config.yml
➜  earthly git:(main) earthly sat ls
Error: unable to authenticate: unauthorized
➜  earthly git:(main) ls ~/.earthly/
auth.jwt  autocomplete.log  buildkitd-startup.lock  certs  config.yml  earthly-prerelease  install_id
```

PR #3762, opened 2024-01-29
Ensure auth.credentials file is created with --token

mikejholly:
Fixes: https://github.com/earthly/earthly/issues/3761

Using `--token` was working correctly, but did not ultimately create the `auth.credentials` file as the code assumed the user was already logged in. This led to some issues with Satellite commands. This change ensures that using `--token` will always create `auth.credentials`.

### DIFF
--- a/cmd/earthly/subcmd/account_cmds.go
+++ b/cmd/earthly/subcmd/account_cmds.go
@@ -609,7 +609,7 @@ func (a *Account) actionLogin(cliCtx *cli.Context) error {
 		return nil
 	}
 
-	if c := errors.Cause(err); c != nil && c != cloud.ErrUnauthorized && c != cloud.ErrAuthTokenExpired {
+	if err != nil && !errors.Is(err, cloud.ErrUnauthorized) && !errors.Is(err, cloud.ErrAuthTokenExpired) {
 		return errors.Wrap(err, "failed to login")
 	}
 
--- a/cmd/earthly/subcmd/account_cmds.go
+++ b/cmd/earthly/subcmd/account_cmds.go
@@ -602,12 +602,14 @@ func (a *Account) actionLogin(cliCtx *cli.Context) error {
 	}
 
 	loggedInEmail, authType, _, err := cloudClient.WhoAmI(cliCtx.Context)
-	if err == nil {
+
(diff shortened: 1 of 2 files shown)


cmd/earthly/subcmd/account_cmds.go
```diff
@@ -602,12 +602,14 @@ func (a *Account) actionLogin(cliCtx *cli.Context) error {
...
-	if err == nil {
+
+	if err == nil && authType == cloud.AuthMethodCachedJWT {
 		// already logged in, don't re-attempt a login
 		a.cli.Console().Printf("Logged in as %q using %s auth\n", loggedInEmail, authType)
 		return nil
 	}
-	if errors.Cause(err) != cloud.ErrUnauthorized && errors.Cause(err) != cloud.ErrAuthTokenExpired {
+
+	if c := errors.Cause(err); c != nil && c != cloud.ErrUnauthorized && c != cloud.ErrAuthTokenExpired {
```
brandonSc, line 612: Could this use `errors.Is`?
```suggestion
	if !errors.Is(err, cloud.ErrUnauthorized) && !errors.Is(err, cloud.ErrAuthTokenExpired) {
```

brandonSc (approved):

vladaionescu (approved):

cmd/earthly/subcmd/account_cmds.go
```diff
@@ -602,12 +602,14 @@ func (a *Account) actionLogin(cliCtx *cli.Context) error {
 	}
 
 	loggedInEmail, authType, _, err := cloudClient.WhoAmI(cliCtx.Context)
-	if err == nil {
+
+	if err == nil && authType == cloud.AuthMethodCachedJWT {
```
idodod, line 606: I wonder if this also fixes the bug/behavior change where in order to login to user Y, while already being logged into user X, forces an explicit `earthly account logout`

idodod (approved):

cmd/earthly/subcmd/account_cmds.go
mikejholly: Good question. I can look into that. We should also verify whether we want to keep this skipping logic.

Issue #3581, opened 2023-12-09
`BUILD --auto-skip`


vladaionescu:
**Use case**

As a developer working with a monorepo, I would like to be able to skip executing targets in CI that are related to sub-projects unaffected by the changes that triggered the CI. E.g. if a monorepo has 7 projects, you would normally change only one of those. Ideally, you don't re-run the builds of the other 6 (unless they depend on the sub-project that was changed).

**Expected Behavior**

Currently, auto-skip is an all-or-nothing operation: either the entire build is skipped or it runs entirely (falling back to the buildkit-based caches).

The ability to specify in an Earthfile sub-targets that may be skipped allows us to extend the behavior to include skipping sub-graphs. Which sub-graphs can be auto-skipped are entirely under the user's control, via adding the `--auto-skip` flag to `BUILD` targets.

PR #3767, opened 2024-01-31
Allow for --auto-skip on BUILD commands

mikejholly:
Implements: https://github.com/earthly/earthly/issues/3581

Adds support for `BUILD --auto-skip <target-name>` where individual targets may be auto-skipped. 

Example: 

```
VERSION 0.8

FROM alpine

a:
    RUN echo "hello from a"

b:
    RUN echo "hello from b"

basic:
    BUILD --auto-skip +a --foo=1
    BUILD --auto-skip +b
```

Output for second run (after auto-skip results captured): 

```
 Init 🚀

Streaming logs to https://cloud.earthly.dev/builds/75a1763e-9995-4494-b115-ba6bbf361621

 Build 🔧

           /tmp+base | DOCKERHUB_AUTH=true
           /tmp+base | --> FROM alpine
           /tmp+base | [----------] 100% FROM alpine
           auto-skip | Target /tmp+a (hash 0f81324c8af0001d1b53451da4fcb3a1eada3e00) has already been run. Skipping.
           auto-skip | Target /tmp+b (hash 2f3459fae0f3b5bb071e869fde11360f8856675b) has already been run. Skipping.

 Push Summary ⏫ (disabled)

 Local Output Summary 🎁



View logs at https://cloud.earthly.dev/builds/75a1763e-9995-4494-b115-ba6bbf361621
```

### DIFF
--- a/ast/commandflag/flags.go
+++ b/ast/commandflag/flags.go
@@ -89,6 +89,7 @@ type BuildOpts struct {
 	BuildArgs       []string `long:"build-arg" description:"A build arg override passed on to a referenced Earthly target"`
 	AllowPrivileged bool     `long:"allow-privileged" description:"Allow targets to assume privileged mode"`
 	PassArgs        bool     `long:"pass-args" description:"Pass arguments to external targets"`
+	AutoSkip        bool     `long:"auto-skip" description:"Use auto-skip to bypass the target if nothing has changed"`
 }
 
 type GitCloneOpts struct {
--- a/earthfile2llb/converter.go
+++ b/earthfile2llb/converter.go
@@ -26,7 +26,6 @@ import (
 	"github.com/earthly/earthly/ast/commandflag"
 	"github.com/earthly/earthly/ast/spec"
 	"github.com/earthly/earthly/buildcontext"
-	"github.com/earthly/earthly/cmd/earthly/bk"
 	debuggercommon "github.com/earthly/earthly/debugger/common"
 	"github.com/earthly/earthly/domain"
 	"github.com/earthly/earthly/features"
@@ -1825,6 +1824,14 @@ func (c *Converter) absolutizeTarget(fullTargetName string, allowPrivileged bool
 }
 
 func (c *Converter) checkAutoSkip(ctx context.Context, fullTargetName string, allowPrivileged, passArgs bool, buildArgs []string) (bool, func(), error) {
+	console := c.opt.Console.WithPrefix("auto-skip")
+
+	nopFn := func() {}
+
+	if c.opt.BuildkitSkipper == nil {
+		console.Warnf("--auto-skip flags are disabled due to client initialization failure")
+		return false, nopFn, nil
+	}
 
 	target, relTarget, _, err := c.absolutizeTarget(fullTargetName, allowPrivileged)
 	if err != nil {
@@ -1847,17 +1854,12 @@ func (c *Converter) checkAutoSkip(ctx context.Context, fullTargetName string, al
 		return false, nil, errors.Wrapf(err, "auto-skip is unable to calculate hash for %s", target)
 	}
 
-	skipDB, err := bk.NewBuildkitSkipper(c.opt.AutoSkipLocalDB, c.varCollection.Org(), target, c.opt.AutoSkipClient)
-	if err != nil {
-		return false, nil, err
-	}
+	orgName := c.varCollection.Org()
 
-	console := c.opt.Console.WithPrefix("auto-skip")
-
-	exists, err := skipDB.Exists(ctx, targetHash)
+	exists, err := c.opt.BuildkitSkipper.Exists(ctx, orgName, targetHash)
 	if err != nil {
 		console.Warnf("Unable to check if target %s (hash %x) has already been run: %s", target.String(), targetHash, err.Error())
-		return false, func() {}, nil
+		return false, nopFn, nil
 	}
 
 	if exists {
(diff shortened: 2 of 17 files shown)


idodod (approved):

util/flagutil/parse.go
```diff
@@ -196,3 +196,36 @@ func ParseParams(str string) (string, []string, error) {
...
+		}
+	}
+	if IsInParamsForm(target) {
+		target, extraArgs, err = ParseParams(target)
+		if err != nil {
+			return "", "", nil, err
+		}
+	}
+	return image, target, extraArgs, nil
+}
```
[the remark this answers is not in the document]
mikejholly: Yes, to resolve an import cycle.

mikejholly:
@vladaionescu 

> Should there be a way to ignore auto-skipping from the terminal? Maybe earthly --no-auto-skip?

Sounds good. Is the idea to allow users to disable all `BUILD --auto-skip` instances? I suppose this should be incompatible with `--auto-skip`, correct? We can display an error message when both are set. 

> Should checkAutoSkip happen one level lower, in earthfile2llb.go? Pro: you only have to reference checkAutoSkip once in the code, which simplifies the logic a bit. Con: you have to pass an auto-skip flag in to the converter.

I think `checkAutoSkip` is currently called once if I'm not mistaken. I thought the interpreter was a good spot as the check here allows us to completely bypass the `Build` & `BuildAsync` calls. I looked at integrating the auto-skip check into the `opt.Visited` cache in `Earthfile2LLB`, but that didn't seem to work as auto-skip doesn't create a `SingleTarget`.

> I believe that we need to add a VERSION --use-auto-skip flag too. We've generally done this when introducing new flags or syntax, in order to guarantee that every 0.8 binary can run any VERSION 0.8 earthfile.

Good point. I'll add a new feature flag.

earthfile2llb/converter.go
```diff
@@ -1799,33 +1801,89 @@ func (c *Converter) ExpandArgs(ctx context.Context, runOpts ConvertRunOpts, word
...
+		Console:        c.opt.Console,
+		CI:             c.opt.IsCI,
+		BuiltinArgs:    c.opt.BuiltinArgs,
+		OverridingVars: overriding,
+	})
+	if err != nil {
+		return false, nil, errors.Wrapf(err, "auto-skip is unable to calculate hash for %s", target)
+	}
+
+	skipDB, err := bk.NewBuildkitSkipper(c.opt.AutoSkipLocalDB, c.varCollection.Org(), target, c.opt.AutoSkipClient)
```
alexcb, line 1850: calling this in the converter doesn't seem great, but I understand it's needed because of the org name or target might be different.

What if instead we introduce a factory, e.g.

change `NewBuildkitSkipper` to 

```
func NewBuildkitSkipperFactory(localSkipDB, cloudClient buildkitskipper.ASKVClient) (BuildkitSkipperCreator, error) {

    return func(orgName string, target domain.Target) (BuildkitSkipper, error) {

	if localSkipDB != "" {
		skipDB, err := buildkitskipper.NewLocal(localSkipDB)
		if err != nil {
			return nil, errors.Wrapf(err, "failed to open buildkit skipper database %s", localSkipDB)
		}
		return skipDB, nil
	}

	skipDB, err := buildkitskipper.NewCloud(orgName, target, cloudClient)
	if err != nil {
		return nil, errors.Wrapf(err, "failed to create cloud-based buildkit skipper database")
	}
	return skipDB, nil
    }
}

```

it has the benefit that we don't have to propagate localSkipDB and cloudClient through to the converter.

alexcb (changes requested):
we need to add a new entry to `features.go` because this introduces a new flag which could break old versions of earthly.

vladaionescu:
> > Should there be a way to ignore auto-skipping from the terminal? Maybe earthly --no-auto-skip?
> 
> Sounds good. Is the idea to allow users to disable all BUILD --auto-skip instances? I suppose this should be incompatible with --auto-skip, correct? We can display an error message when both are set.

Agreed!

> > Should checkAutoSkip happen one level lower, in earthfile2llb.go? Pro: you only have to reference checkAutoSkip once in the code, which simplifies the logic a bit. Con: you have to pass an auto-skip flag in to the converter.
> 
> I think checkAutoSkip is currently called once if I'm not mistaken. I thought the interpreter was a good spot as the check here allows us to completely bypass the Build & BuildAsync calls. I looked at integrating the auto-skip check into the opt.Visited cache in Earthfile2LLB, but that didn't seem to work as auto-skip doesn't create a SingleTarget.

There are two spots that I saw (the other one is needed for the matrix BUILD). Agreed that integrating with the visited cache doesn't make sense. To me conceptually it's a little simpler for each target to decide on its own whether to auto-skip or not (rather than the parent deciding). But all in all, this is not a strong opinion. Both versions are fine.

vladaionescu:
> There are two spots that I saw
NVM there's only one spot. Feel free to ignore my comment.

earthfile2llb/converter.go
mikejholly: I think we can keep the existing structure by moving the org & target names into the methods. `NewBuildkitSkipper` is already a factory of sorts. 

As I explored this further, I realized that we can only initialize a local Bolt DB instance once. Secondary calls to `bolt.Open` block. Hence, I moved the DB initialization up & now supply it to `initAutoSkip` & as a build opt.

alexcb: > moving the org & target names into the methods

good plan!

> I realized that we can only initialize a local Bolt DB instance once

good observation!

ast/commandflag/flags.go
```diff
@@ -89,6 +89,7 @@ type BuildOpts struct {
 	BuildArgs       []string `long:"build-arg" description:"A build arg override passed on to a referenced Earthly target"`
 	AllowPrivileged bool     `long:"allow-privileged" description:"Allow targets to assume privileged mode"`
 	PassArgs        bool     `long:"pass-args" description:"Pass arguments to external targets"`
+	AutoSkip        bool     `long:"auto-skip" description:"Use auto-skip to bypass the target if nothing has changed"`
```
alexcb, line 92:
```suggestion
	AutoSkip        bool     `long:"auto-skip" description:"Enables the auto-skip BUILD flag, which will bypass the target if nothing has changed"`
```

alexcb (approved):
TIL: "absolutize" is a word, nice work.

PR #3770, opened 2024-01-31
Satellite `rm` requires a `--force` if it's running

brandonSc:
If a satellite is not in a sleep or "offline" state, then the command will ask the user to re-run using a `--force` flag instead. This prevents a user from mistakenly killing builds that may be running.

Screenshot 2024-01-31 151626

### DIFF
--- a/cmd/earthly/subcmd/satellite_cmds.go
+++ b/cmd/earthly/subcmd/satellite_cmds.go
@@ -33,7 +33,7 @@ type Satellite struct {
 	maintenanceWindow      string
 	maintenaceWeekendsOnly bool
 	version                string
-	forceUpdate            bool
+	force                  bool
 	printJSON              bool
 	listAll                bool
 	dropCache              bool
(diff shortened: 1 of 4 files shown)


cmd/earthly/subcmd/satellite_cmds.go
```diff
@@ -612,21 +622,31 @@ func (a *Satellite) actionRemove(cliCtx *cli.Context) error {
...
 				return errors.New("cannot delete hidden satellites")
 			}
 		}
 	}
-	if !found {
+	if sat == nil {
 		return fmt.Errorf("could not find %q for deletion", a.cli.Flags().SatelliteName)
 	}
 
+	isOffline := sat.State == cloud.SatelliteStatusOperational || sat.State == cloud.SatelliteStatusOffline
```
mikejholly, line 638: Does `sat.State == cloud.SatelliteStatusOperational` denote offline?

cmd/earthly/subcmd/satellite_cmds.go
```diff
@@ -612,21 +622,31 @@ func (a *Satellite) actionRemove(cliCtx *cli.Context) error {
...
+	if !a.force && !isOffline {
```
mikejholly, line 639: Would `isOnline` be more clear?

brandonSc: Yes, thanks, for noticing. I refactored and got the name backwards.

brandonSc: There was more wrong than just a name. Good thing you noticed.

mikejholly (approved):

PR #3791, opened 2024-02-07
Docs: Self-Hosted Satellites

brandonSc:
New doc page for self-hosted satellites.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -50,7 +50,7 @@
         '(:|\\s+)earthly/earthly:(?<currentValue>.+?)[\\s\\n/]+',
         'the `latest` tag is `(?<currentValue>.+?)`',
         'earthly/compare/(?<currentValue>.+?)\\.\\.\\.main',
-        'earthly/satellite:(?<currentValue>.+?)',
+        'earthly/satellite:(?<currentValue>.+?)\\n',
       ],
       depNameTemplate: 'earthly/earthly',
       datasourceTemplate: 'github-releases',
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -50,6 +50,7 @@
         '(:|\\s+)earthly/earthly:(?<currentValue>.+?)[\\s\\n/]+',
         'the `latest` tag is `(?<currentValue>.+?)`',
         'earthly/compare/(?<currentValue>.+?)\\.\\.\\.main',
+        'earthly/satellite:(?<currentValue>.+?)',
       ],
       depNameTemplate: 'earthly/earthly',
       datasourceTemplate: 'github-releases',
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -84,5 +84,5 @@
 * [Earthly Satellites](cloud/satellites.md)
     * [Managing Satellites](cloud/satellites/managing.md)
     * [Using Satellites](cloud/satellites/using.md)
-    * [Best Practices](cloud/satellites/best-practices.md)
     * [Self-Hosted Satellites](cloud/satellites/self-hosted.md)
+    * [Best Practices](cloud/satellites/best-practices.md)
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -85,3 +85,4 @@
     * [Managing Satellites](cloud/satellites/managing.md)
     * [Using Satellites](cloud/satellites/using.md)
     * [Best Practices](cloud/satellites/best-practices.md)
+    * [Self-Hosted Satellites](cloud/satellites/self-hosted.md)
--- a/docs/ci-integration/guides/kubernetes.md
+++ b/docs/ci-integration/guides/kubernetes.md
@@ -2,7 +2,7 @@
 
 {% hint style='info' %}
 ##### Note
-This guide is related to self-hosting Buildkit, however, Self-Hosted Satellites _*beta*_ are now available with more features and better security. Check out the [Self-Hosted Satellites Guide](../../cloud/satellites/self-hosted.md) for more details and instructions to deploy in Kubernetes.
+This guide is related to self-hosting Buildkit, however, Self-Hosted Satellites _\*beta\*_ are now available with more features and better security. Check out the [Self-Hosted Satellites Guide](../../cloud/satellites/self-hosted.md) for more details and instructions to deploy in Kubernetes.
 {% endhint %}
 
 
--- a/docs/ci-integration/guides/kubernetes.md
+++ b/docs/ci-integration/guides/kubernetes.md
@@ -1,5 +1,11 @@
 # Kubernetes
 
+{% hint style='info' %}
+##### Note
+This guide is related to self-hosting Buildkit, however, Self-Hosted Satellites _*beta*_ are now available with more features and better security. Check out the [Self-Hosted Satellites Guide](../../cloud/satellites/self-hosted.md) for more details and instructions to deploy in Kubernetes.
+{% endhint %}
+
+
 ## Overview
 
 Kubernetes isn't a CI per-se, but it _can_ serve as the underpinning for many modern CI systems. As such, this example serves as a bare-bones example to base your implementations on.
--- a/docs/ci-integration/guides/kubernetes.md
+++ b/docs/ci-integration/guides/kubernetes.md
@@ -2,7 +2,7 @@
 
 {% hint style='info' %}
 ##### Note
-This guide is related to self-hosting Buildkit, however, Self-Hosted Satellites **beta** are now available with more features and better security. Check out the [Self-Hosted Satellites Guide](../../cloud/satellites/self-hosted.md) for more details and instructions to deploy in Kubernetes or in AWS EC2.
+This guide is related to self-hosting a remote Buildkit, however, Self-Hosted Satellites **beta** are now available. Self-Hosted Satellites provide more features, have better security, and are easier to deploy than remote Buildkit. Check out the [Self-Hosted Satellites Guide](../../cloud/satellites/self-hosted.md) for more details and instructions to deploy in Kubernetes or AWS EC2.
 {% endhint %}
 
 
--- a/docs/ci-integration/guides/kubernetes.md
+++ b/docs/ci-integration/guides/kubernetes.md
@@ -2,7 +2,7 @@
 
 {% hint style='info' %}
 ##### Note
-This guide is related to self-hosting Buildkit, however, Self-Hosted Satellites _\*beta\*_ are now available with more features and better security. Check out the [Self-Hosted Satellites Guide](../../cloud/satellites/self-hosted.md) for more details and instructions to deploy in Kubernetes.
+This guide is related to self-hosting Buildkit, however, Self-Hosted Satellites **beta** are now available with more features and better security. Check out the [Self-Hosted Satellites Guide](../../cloud/satellites/self-hosted.md) for more details and instructions to deploy in Kubernetes or in AWS EC2.
 {% endhint %}
 
 
--- a/docs/ci-integration/remote-buildkit.md
+++ b/docs/ci-integration/remote-buildkit.md
@@ -2,7 +2,7 @@
 
 {% hint style='info' %}
 ##### Note
-[Self-Hosted Satellites _*beta*_](../cloud/satellites/self-hosted.md) are now available. Satellites provide more features, better security, and are easier to deploy. If your use case cannot tolerate any cloud integration, however, then running Buildkit directly is still supported.
+[Self-Hosted Satellites _\*beta\*_](../cloud/satellites/self-hosted.md) are now available. Satellites provide more features, better security, and are easier to deploy. If your use case cannot tolerate any cloud integration, however, then running Buildkit directly is still supported.
 {% endhint %}
 
 ## Introduction
--- a/docs/ci-integration/remote-buildkit.md
+++ b/docs/ci-integration/remote-buildkit.md
@@ -1,5 +1,10 @@
 # Remote BuildKit 
 
+{% hint style='info' %}
+##### Note
+[Self-Hosted Satellites _*beta*_](../cloud/satellites/self-hosted.md) are now available. Satellites provide more features, better security, and are easier to deploy. If your use case cannot tolerate any cloud integration, however, then running Buildkit directly is still supported.
+{% endhint %}
+
 ## Introduction
 
 In some cases, you may want to run a remote instance of [`earthly/buildkitd`](https://hub.docker.com/r/earthly/buildkitd). This guide is intended to help you identify if you might benefit from this configuration, and to help you set it up correctly.
--- a/docs/ci-integration/remote-buildkit.md
+++ b/docs/ci-integration/remote-buildkit.md
@@ -2,7 +2,7 @@
 
 {% hint style='info' %}
 ##### Note
-This guide is related to self-hosting a remote Buildkit, however, Self-Hosted Satellites **beta** are now available. Self-Hosted Satellites provide more features, have better security, and are easier to deploy than remote Buildkit. Check out the [Self-Hosted Satellites Guide](../cloud/satellites/self-hosted.md) for more details. If your use case cannot tolerate any cloud integration, however, then self-hosting a remote Buildkit is still supported.
+This guide is related to self-hosting a remote Buildkit, however, Self-Hosted Satellites **beta** are now available. Self-Hosted Satellites provide more features, have better security, and are easier to deploy than remote Buildkit. Check out the [Self-Hosted Satellites Guide](../cloud/satellites/self-hosted.md) for more details. If your use case cannot tolerate a cloud-based control plane, however, then self-hosting a remote Buildkit is still supported.
 {% endhint %}
 
 ## Introduction
--- a/docs/ci-integration/remote-buildkit.md
+++ b/docs/ci-integration/remote-buildkit.md
@@ -2,7 +2,7 @@
 
 {% hint style='info' %}
 ##### Note
-[Self-Hosted Satellites _\*beta\*_](../cloud/satellites/self-hosted.md) are now available. Satellites provide more features, better security, and are easier to deploy. If your use case cannot tolerate any cloud integration, however, then running Buildkit directly is still supported.
+This guide is related to self-hosting a remote Buildkit, however, Self-Hosted Satellites **beta** are now available. Self-Hosted Satellites provide more features, have better security, and are easier to deploy than remote Buildkit. Check out the [Self-Hosted Satellites Guide](../cloud/satellites/self-hosted.md) for more details. If your use case cannot tolerate any cloud integration, however, then self-hosting a remote Buildkit is still supported.
 {% endhint %}
 
 ## Introduction
--- a/docs/cloud/satellites/self-hosted.md
+++ b/docs/cloud/satellites/self-hosted.md
@@ -198,7 +198,7 @@ If you are having problems using or deploying your self-hosted satellite, please
 
 **Problem:** Satellite is not listed in the output of earthly satellite ls
 
-**Resolution:** Check the logs from the satellite’s Docker container. There may be a message containing the phrase "SATELLITE IS NOT REGISTERED". This usually means there was a problem with the values supplied to the satellite’s run command. Check the error for additional context. Ensure the values provided for account token, earthly org, etc are correct.
+**Resolution:** Check the logs from the satellite’s Docker container. There may be a message containing the phrase `SATELLITE IS NOT REGISTERED`. This usually means there was a problem with the values supplied to the satellite’s run command. Check the error for additional context. Ensure the values provided for account token, earthly org, etc are correct.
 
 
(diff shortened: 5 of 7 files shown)


docs/ci-integration/guides/kubernetes.md
```diff
@@ -1,5 +1,11 @@
 # Kubernetes
 
+{% hint style='info' %}
+##### Note
+This guide is related to self-hosting Buildkit, however, Self-Hosted Satellites _\*beta\*_ are now available with more features and better security. Check out the [Self-Hosted Satellites Guide](../../cloud/satellites/self-hosted.md) for more details and instructions to deploy in Kubernetes.
```
vladaionescu, line 5:
```suggestion
This guide is related to self-hosting Buildkit, however, Self-Hosted Satellites **beta** are now available with more features and better security. Check out the [Self-Hosted Satellites Guide](../../cloud/satellites/self-hosted.md) for more details and instructions to deploy in Kubernetes or in AWS EC2.
```

docs/cloud/satellites/self-hosted.md
```diff
@@ -0,0 +1,215 @@
+# Self-Hosted Satellites
+
+Self-hosted satellites _\*Beta\*_ are a good alternative to Earthly Cloud’s [managed satellites](../satellites.md) when CI/CD is required to run in your own environment. Self-hosted satellites provide most of the benefits of Earthly Cloud while ensuring that build-related data never leaves your network.
```
vladaionescu, line 3:
```suggestion
Self-hosted satellites **Beta** are a good alternative to Earthly Cloud’s [managed satellites](../satellites.md) when CI/CD is required to run in your own environment. Self-hosted satellites provide most of the benefits of Earthly Cloud while ensuring that your code and build-related data never leaves your network.
```

docs/cloud/satellites/self-hosted.md
```diff
@@ -0,0 +1,215 @@
...
+
+In addition, Self-Hosted Satellites have the following features:
+* Lightning-fast caching
+* Cloud-enabled control plane
+* Encrypted connections by default using mTLS
+* Ready to run builds with just a single command
```
vladaionescu, line 9: Perhaps we should mention only the benefits specific to self-hosted satellites, otherwise we're repeating content from https://docs.earthly.dev/earthly-cloud/satellites.

Looping in @thtmnisamnstr in case he has thoughts on messaging here.

vladaionescu: Side note: We should also amend step 2 from the page I linked above to say something along the lines of "if you want to launch a self-hosted satellite, go to this other page". Otherwise all the steps on that page are the same.

docs/cloud/satellites/self-hosted.md
```diff
@@ -0,0 +1,215 @@
...
+
+On the other hand, they may have the following drawbacks when compared to managed satellites offered in Earthly Cloud:
```
vladaionescu, line 11: @thtmnisamnstr - thoughts on what to call the regular satellites? I think we use the term "Cloud" on our pricing. Do we then call these "Cloud Satellites" instead of managed satellites?

docs/cloud/satellites/self-hosted.md
```diff
@@ -0,0 +1,215 @@
...
+* No automatic updates
+* Does not automatically sleep to save costs while idle
+* Requires more knowledge and tuning to achieve good performance
+
+## Getting Started
+Launching a self-hosted satellite is as easy as running the public Docker image. A few environment variables are needed to link the satellite with your Earthly account. The satellite will use these values to automatically register with Earthly Cloud servers once it starts. Note that Earthly Cloud will never connect to your satellite directly.
```
vladaionescu, line 17:
```suggestion
Launching a self-hosted satellite is as easy as running the public Docker image. A few environment variables are needed to link the satellite with your Earthly account. The satellite will use these values to automatically register with Earthly Cloud servers once it starts. Note that Earthly Cloud will never connect to your satellite directly. The host and port of your satellite can thus be private to your internal corporate network or VPC, if you wish.
```

docs/cloud/satellites/self-hosted.md
```diff
@@ -0,0 +1,215 @@
...
+Here is a minimal command to start a self-managed satellite using Docker:
+
+```
+docker run --privileged \
+    -p 8372:8372 \
+    -e EARTHLY_TOKEN=GuFna*****nve7e \ 
+    -e EARTHLY_ORG=my-org \
+    -e SATELLITE_NAME=my-satellite \
+    -e SATELLITE_HOST=153.65.8.0 \
+  earthly/satellite:v0.8.3
```
vladaionescu, line 28: Can we get this line to auto-update via renovate?

docs/cloud/satellites/self-hosted.md
```diff
@@ -0,0 +1,215 @@
...
+* `SATELLITE_HOST` - Hostname or IP address for the Earthly CLI to connect. Note that users will select the satellite by name when running builds.
+
+Using a volume for cache storage is recommended for preserving cache after the container is destroyed. See [Advanced Configuration](#advanced-configuration) below for more details and a list of optional environment variables that can be used to finetune your deployment. See also [Platform Specific Guides](#platform-specific-guides) for examples of deploying to platforms like Kubernetes or EC2.
+
+The self-hosted satellite will not accept localhost as an address. For instructions on testing locally, please refer to [Running on Localhost](#running-on-localhost) below.
+
+
+{% hint style='info' %}
+##### Security Advisory
+The `--privileged` flag is currently required. Privileged mode is used by some features of earthly, including the `WITH DOCKER` command in Earthfiles. If using privileged mode is a concern, running satellites in a dedicated VM or Kubernetes cluster can provide better isolation compared to running the container directly in your production environment.
```
vladaionescu, line 44:
```suggestion
The `--privileged` flag is currently required. Privileged mode is used by some features of earthly, including the `WITH DOCKER` command in Earthfiles. If using privileged mode is a concern, running satellites in a dedicated VM or separate Kubernetes cluster can provide better isolation compared to running the container directly in your production environment.
```

vladaionescu: I would perhaps add that a rootless version will be available in the future.

docs/cloud/satellites/self-hosted.md
```diff
@@ -0,0 +1,215 @@
...
+{% endhint %}
+
+### Auto-discovery of Host _\*experimental\*_
+
+When `SATELLITE_HOST` is left unset, the satellite will attempt to auto-discover its host address. This feature may not always work, and currently only supports environments using cloud-init (such as EC2).
```
vladaionescu, line 49:
```suggestion
When `SATELLITE_HOST` is left unset, the satellite will attempt to auto-discover its host address. Auto-discovery may not always work, and currently only supports environments using cloud-init (such as EC2).
```

docs/cloud/satellites/self-hosted.md
```diff
@@ -0,0 +1,215 @@
...
+earthly sat select my-satellite
+```
+
+Or, to invoke a build once on the satellite, without persistently selecting it:
+
+```
+earthly --sat my-satellite +my-target
+```
+
+Note that `earthly-v0.8.0` or later is required to connect to a self-hosted satellite.
```
vladaionescu, line 66:
```suggestion
Note that `earthly` `v0.8.0` or later is required to connect to a self-hosted satellite.
```

docs/cloud/satellites/self-hosted.md
```diff
@@ -0,0 +1,215 @@
...
+NAME          PLATFORM     SIZE         VERSION  STATE
+my-satellite  linux/amd64  self-hosted  v0.8.3   operational
+...
+```
+
+The size field of a self-hosted satellite will be listed as `self-hosted`, as opposed to one of the sizes of Earthly Cloud’s managed satellites.
+
+The state field will be either `operational` or `offline`. A self-hosted satellite automatically transitions to an `offline` state when it is gracefully terminated, or if no heartbeat has been received by Earthly’s servers for a while.
+
+You can run `satellite rm` on a self-hosted satellite when it is in an offline state to remove it from your account permenantly.
```
vladaionescu, line 83: Add mention of `--force`?

docs/cloud/satellites/self-hosted.md
```diff
@@ -0,0 +1,215 @@
...
+            fieldRef:
+              fieldPath: status.podIP
+```
+
+Note that this example uses the Pod’s IP address (via [Downward API](https://kubernetes.io/docs/concepts/workloads/pods/downward-api/)) as the `SATELLITE_HOST` value. The Earthly CLI must be able to reach the IP on its network.
+
+If the Earthly CLI should connect via a different address, such as DNS, then this value should be provided as the `SATELLITE_HOST` instead.
+
+
+### AWS EC2
```
vladaionescu, line 128: I wonder if the EC2 version should be more encouraged in this guide since VMs provide a better security sandbox for builds. Just a thought.

docs/cloud/satellites/self-hosted.md
```diff
@@ -0,0 +1,215 @@
...
+        earthly/satellite:v0.8.3
+```
+
+Note that the `SATELLITE_HOST` variable is unset in this example so that the host is auto-discovered by the satellite when it starts. This should result in the instance’s private DNS being used as the host. 
+
+If the Earthly CLI will be unable to connect to the satellite via the EC2’s private DNS, then `SATELLITE_HOST` should be provided in the `docker run` command with an alternate value.
+
+## Advanced Configuration
+
+### Cache Disk
```
vladaionescu, line 156: I think that our position should be stronger here. Not using a volume causes extremely bad `WITH DOCKER` performance (due to overlayfs). We should warn users that they MUST use a volume for this, and that volume should be backed by an ext4 filesystem (Docker volumes work great).

All our examples above should include the volume mount (the minimal docker run command and k8s example).

docs/cloud/satellites/self-hosted.md
```diff
@@ -0,0 +1,215 @@
...
+```
+# Local test for Earthly
+127.0.0.1	earthly.local
+```
+
+Starting your satellite with `SATELLITE_HOST` set to `earthly.local` should allow for a localhost connection using your Earthly CLI.
+
+## Debugging
+
+If you are having problems using or deploying your self-hosted satellite, please refer to the following tips or reach out to us through our community Slack channel.
```
vladaionescu, line 197:
```suggestion
If you are having problems using or deploying your self-hosted satellite, please refer to the following tips or reach out to us through our community [Slack channel](https://earthly.dev/slack).
```

docs/cloud/satellites/self-hosted.md
```diff
@@ -0,0 +1,215 @@
...
+
+**Problem:** Satellite is not listed in the output of earthly satellite ls
```
vladaionescu, line 199:
```suggestion
### Problem: Satellite is not listed in the output of `earthly satellite ls`
```

Ditto for all the others...

docs/cloud/satellites/self-hosted.md
```diff
@@ -0,0 +1,215 @@
...
+
+**Resolution:** Check the logs from the satellite’s Docker container. There may be a message containing the phrase "SATELLITE IS NOT REGISTERED". This usually means there was a problem with the values supplied to the satellite’s run command. Check the error for additional context. Ensure the values provided for account token, earthly org, etc are correct.
```
vladaionescu, line 201:
```suggestion
**Resolution:** Check the logs from the satellite’s Docker container. There may be a message containing the phrase `SATELLITE IS NOT REGISTERED`. This usually means there was a problem with the values supplied to the satellite’s run command. Check the error for additional context. Ensure the values provided for account token, earthly org, etc are correct.
```

docs/remote-runners.md
```diff
@@ -26,6 +26,4 @@ When using remote runners, even though the build executes remotely, the followin
 
 To get started with free remote runners managed by Earthly, check out [Earthly Satellites](cloud/satellites.md).
 
-To get started with self-managed remote runners, see
-* The [remote buildkit page](ci-integration/remote-buildkit.md)
-* The [Kubernetes setup page](ci-integration/guides/kubernetes.md)
+To get started with self-hosted runners, see the [Self-Hosted Satellites Guide](cloud/satellites/self-hosted.md).
```
vladaionescu, line 29: I wouldn't remove our open-source offering here. Perhaps this is where we have a little comparison between self-hosted satellites vs remote Buildkit. See a comparison example here: https://docs.earthly.dev/docs/caching#sharing-cache. Perhaps we can reuse the format / style of it.

docs/SUMMARY.md
```diff
@@ -85,3 +85,4 @@
     * [Managing Satellites](cloud/satellites/managing.md)
     * [Using Satellites](cloud/satellites/using.md)
     * [Best Practices](cloud/satellites/best-practices.md)
+    * [Self-Hosted Satellites](cloud/satellites/self-hosted.md)
```
vladaionescu, line 88: I would move this a big higher, between Using Satellites and Best Practices.

vladaionescu (commented):
Thanks for putting this together! The important stuff is there! Sorry for the long review...

* Can you also update the [Managing Satellites](https://docs.earthly.dev/earthly-cloud/satellites/managing#launching-and-removing-satellites) page?

docs/cloud/satellites/self-hosted.md
brandonSc: Not sure if we should promote that for self-hosted. I think of it as an escape hatch just in case something gets stuck in a bad state, as opposed to normal operation.

docs/cloud/satellites/self-hosted.md
brandonSc: People seem very inclined to use K8S, but what if we put a "recommended" label on AWS , and/or bump it higher in the doc?

docs/cloud/satellites/self-hosted.md
vladaionescu: It's probably a very useful escape hatch - sometimes you don't have control over your infrastructure and you really need to disconnect it. Anyway... it's a small feature, we can also just leave it out and let them find it via `rm` attempt - presumably the error says that you can use `--force`.

docs/cloud/satellites/self-hosted.md
brandonSc: I didn't realize there were legitimate issues with not using a volume. I'll put this in the other examples as well then.

docs/cloud/satellites/self-hosted.md
vladaionescu: SGTM

docs/cloud/satellites/self-hosted.md
thtmnisamnstr: That makes sense. Cloud Satellites, Self-Hosted Satellites, and Dedicated Satellites.

docs/cloud/satellites/self-hosted.md
thtmnisamnstr: Replying to this comment: "Perhaps we should mention only the benefits specific to self-hosted satellites, otherwise we're repeating content from https://docs.earthly.dev/earthly-cloud/satellites."

I removed the "In addition to, ..." wording and updated copy a little, and I like it better. I don't think reiterating the benefits here is a bad thing.

docs/cloud/satellites/self-hosted.md
thtmnisamnstr: That makes sense to me too. EC2 should be above k8s and encouraged.

docs/remote-runners.md
thtmnisamnstr: I added the paragraph below to set context and link to the remote Buildkit page. It mentions the comparison  table you recommended. So that needs to be added if we want to keep the below paragraph.

docs/cloud/satellites/self-hosted.md
brandonSc: Yes, it will say that they can re-run with force if they try it. 

Also (in case it wasn't obvious) this wouldn't trigger the actual self-hosted satellite to terminate. Just remove it from the list on earthly side.

docs/ci-integration/remote-buildkit.md
```diff
@@ -1,5 +1,10 @@
 # Remote BuildKit 
 
+{% hint style='info' %}
+##### Note
+This guide is related to self-hosting a remote Buildkit, however, Self-Hosted Satellites **beta** are now available. Self-Hosted Satellites provide more features, have better security, and are easier to deploy than remote Buildkit. Check out the [Self-Hosted Satellites Guide](../cloud/satellites/self-hosted.md) for more details. If your use case cannot tolerate any cloud integration, however, then self-hosting a remote Buildkit is still supported.
```
thtmnisamnstr, line 5: @vladaionescu What do you think about changing "any cloud integration" to "a cloud control plane"? I feel like cloud control plane sounds less invasive than could integration. Or maybe it doesn't matter.

docs/remote-runners.md
```diff
@@ -26,6 +26,6 @@ When using remote runners, even though the build executes remotely, the followin
...
+
+If your use case cannot tolerate any cloud integration, then self-hosting a remote Buildkit is the best approach. Remote Buildkit has less features, is less secure, and is more difficult to deploy than Self-Hosted Satellites (see diagram below for comparison). To get started self-hosting Buildkit, see the [remote buildkit page](ci-integration/remote-buildkit.md).
```
thtmnisamnstr, line 31: @brandonSc Were you planning on adding a comparison diagram below this line? If not, can you update the line as needed?

docs/ci-integration/remote-buildkit.md
vladaionescu: Yeah, that can work. Maybe `a cloud-based control plane`? (minor clarity edit)

docs/remote-runners.md
brandonSc: @thtmnisamnstr I'll see what I can do for that diagram.

docs/remote-runners.md
```diff
@@ -26,6 +26,14 @@ When using remote runners, even though the build executes remotely, the followin
...
+
+Below is a comparison of the different features available with each kind of remote runner.
+
+| Type | Runner Location | Cloud Control-Plane | TLS Encryption By Default | Auto-Sleep | Auto-Updates |
+| --- | --- | --- | --- | --- | --- |
+| Cloud Satellites | Earthly Cloud | ✅ Yes | ✅ Yes | ✅ Yes | ✅ Yes |
+| Self-Hosted Satellites | Your Infrastructure | ✅ Yes | ❌ No | ❌ No | ❌ No |
+| Remote Buildkit | Your Infrastructure | ❌ No Control Plane | ❌ Requires manual configuration | ❌ No | ❌ No |
```
brandonSc, line 39: @thtmnisamnstr @vladaionescu what do you think of the comparison diagram? Any columns we should add/remove?

.github/renovate.json5
```diff
@@ -50,6 +50,7 @@
         '(:|\\s+)earthly/earthly:(?<currentValue>.+?)[\\s\\n/]+',
         'the `latest` tag is `(?<currentValue>.+?)`',
         'earthly/compare/(?<currentValue>.+?)\\.\\.\\.main',
+        'earthly/satellite:(?<currentValue>.+?)',
```
brandonSc, line 53: @Ido did I set this up correctly? The intent is to update lines that look like `earthly/satellite:v0.8.2`

docs/remote-runners.md
brandonSc: I was also picturing how `Earthly Dedicated` could be slotted in here with checkmarks in the auto-sleep and auto-update columns.

thtmnisamnstr: I think this is good. We can add more columns with more features in the future if we want.

brandonSc: Screen Shot 2024-02-12 at 15 18 29
The text wrapping doesn't look great

vladaionescu: A few thoughts...

1. We should flip rows and columns. Then we can add as many features as we want without wrapping issues. It's more common anyway to have the products as columns in such tables (and it would maintain consistency with our other comparison table).

2. I would rename the `TLS` feature to `Managed TLS Certificates`, perhaps. Then we can say just "No" for Remote Buildkit. I think a stronger message (i.e. a clear "No") here is important, because it's pretty difficult to do this yourself, and it's not immediately obvious why.

3. I wonder if we should add the following additional differences:
  * `Managed By`. It would be "You" for Self-hosted and remote buildkit and Earthly for Cloud. For Dedicated, in the future it will also be "Earthly".
  * `Cache persistence`. For Cloud - it's marked as "✅ Yes", for the others, I would put "🟡 Needs configuration". The reason to show this is that sometimes people don't realize that these are not trivial to manage because they tend to be more like pets rather than cattle.

4. Possible option: Add the Earthly Dedicated column now - sales cycles are anyway pretty long for those.

brandonSc: Good suggestions, thanks. I reworked the chart but did not add dedicated, as it seemed confusing without having the corresponding docs to explain what it is.

thtmnisamnstr (approved):

Issue #3786, opened 2024-02-06
earthly/dind release process needs improvement


alexcb:
currently there's no easy way to only release a single dind image (e.g. only alpine), to do so, I had to manually comment out code:

```
@@ -664,16 +664,16 @@ all-dind:
         --platform=linux/amd64 \
         --platform=linux/arm64 \
         +dind-alpine --DATETIME=$DATETIME
-    BUILD \
-        --pass-args \
-        --platform=linux/amd64 \
-        --platform=linux/arm64 \
-        +dind-ubuntu-20.04 --DATETIME=$DATETIME
-    BUILD \
-        --pass-args \
-        --platform=linux/amd64 \
-        --platform=linux/arm64 \
-        +dind-ubuntu-23.04 --DATETIME=$DATETIME
+    #BUILD \
+    #    --pass-args \
+    #    --platform=linux/amd64 \
+    #    --platform=linux/arm64 \
+    #    +dind-ubuntu-20.04 --DATETIME=$DATETIME
+    #BUILD \
+    #    --pass-args \
+    #    --platform=linux/amd64 \
+    #    --platform=linux/arm64 \
+    #    +dind-ubuntu-23.04 --DATETIME=$DATETIME
 
 # all builds all of the following:
 # - Buildkitd for both linux amd64 and linux arm64
```

furthermore, the `alpine-3.19` tag was never created (but out tutorials reference them), to get unstuck, I had to manually add it in:

```
@@ -574,7 +574,7 @@ dind:
       SAVE IMAGE --push --cache-from=earthly/dind:$OS_IMAGE-main $DOCKERHUB_USER/dind:$OS_IMAGE
     END
     ARG DATETIME="$(date --utc +%Y%m%d%H%M%S)" # note this must be overriden when building a multi-platform image (otherwise the values wont match)
-    SAVE IMAGE --push --cache-from=earthly/dind:$OS_IMAGE-main $DOCKERHUB_USER/dind:$TAG "$DOCKERHUB_USER/dind:$TAG-$DATETIME"
+    SAVE IMAGE --push --cache-from=earthly/dind:$OS_IMAGE-main $DOCKERHUB_USER/dind:alpine-3.19 $DOCKERHUB_USER/dind:$TAG "$DOCKERHUB_USER/dind:$TAG-$DATETIME"
```

## What would be nicer ##

It would be nice to be able to first build and push a tag such as `alpine-3.19-docker-25.0.2-r0-20240206171222`, then we could do some testing with it.

[...]

PR #3798, opened 2024-02-08
dind:alpine-3.19 install iptables-legacy

alexcb:
earthly attempts to create a symbolic link to /sbin/iptables-legacy whenever wsl is detected.

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -561,6 +561,8 @@ dind:
         # in order to display the upstream-version, e.g. "24.0.5-1".
         SET DOCKER_VERSION_TAG="$(echo $DOCKER_VERSION | sed 's/^[0-9]*:\([^~]*\).*$/\1/')"
         RUN if echo $DOCKER_VERSION_TAG | grep "[^0-9.-]"; then echo "DOCKER_VERSION_TAG looks bad; got $DOCKER_VERSION_TAG" && exit 1; fi
+    ELSE IF [ "$OS_IMAGE" = "alpine" ]
+        RUN apk add iptables-legacy # required for older kernels
     END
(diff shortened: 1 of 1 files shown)


mikejholly (approved):

Earthfile
```diff
@@ -561,6 +561,8 @@ dind:
         # in order to display the upstream-version, e.g. "24.0.5-1".
         SET DOCKER_VERSION_TAG="$(echo $DOCKER_VERSION | sed 's/^[0-9]*:\([^~]*\).*$/\1/')"
         RUN if echo $DOCKER_VERSION_TAG | grep "[^0-9.-]"; then echo "DOCKER_VERSION_TAG looks bad; got $DOCKER_VERSION_TAG" && exit 1; fi
+    ELSE IF [ "$OS_IMAGE" = "alpine" ]
+        RUN apk add iptables-legacy # required for older kernels
```
idodod, line 565: I know this would probably be harder to maintain but I wonder if we should try to pin to a specific version to ensure to avoid surprises (unless the fact it's a legacy version means it's not going to get anymore updates?)

idodod (approved):

Earthfile
alexcb: there might still be updates to the iptables-legacy package, but they should be backwards compatable, pinning to a specific version here seems overkill -- once this image is built and pushed, we shouldn't need to continually rebuild it.

when we do rebuild it (e.g. a new docker version, which is pinned), we'll need to do more testing.

overall the dind process needs some improvements, which I quickly highlighted in #3786

I'm inclined to leave this as-is while we sort out the immediate wsl issue.

PR #3805, opened 2024-02-12
Append platform to git project errors

brandonSc:
We've had reports of internal errors resolving git projects, which could be related to using an image for the wrong architecture.  Appending the intended platform to the error could shed some light on the issue.

Example error: 

```
 exec /bin/sh: exec format error
 ERROR
       The internal command
           GET GIT META github.com/org/repo:refs/main
    did not complete successfully. Exit code 1 
```


alexcb (approved):

brandonSc:
This error may happen when the git image is [overridden in the Earthly config](https://docs.earthly.dev/docs/earthly-config#git_image).

PR #3810, opened 2024-02-14
Expand args in CACHE --id

brandonSc:
Users reported that that ARGs were not expanded in the `--id` of `CACHE` commands. For example, the following would not expand the arg, and result in a cache id of the literal string `$something`:
```
ARG something
CACHE --id $something
```


idodod (approved):

alexcb (approved):
nice test :)

Issue #3807, opened 2024-02-13
Capture more target data in Logstream deltas


mikejholly:
From Slack discussion:

1. For each command, we need references to other targets mentioned (e.g. `COPY +foo/bar.txt ./` would have to point to the target ID of `+foo`)
1. We noticed that BUILD commands don't show up in Logstream. I guess it makes sense why we never needed them before - we never display those in the terminal output. In any case - it would be useful to have an entry for them, such that Miron can properly represent targets that have nothing in them other than a series of BUILD instructions (pretty common). Note that the duration of the BUILD command should match the duration of the target being referenced.

PR #3815, opened 2024-02-16
Logstream: link nested targets to parent commands & BUILD command

mikejholly:
Addresses: https://github.com/earthly/earthly/issues/3807

This PR adds new Logbus commands for pseudo-commands like `BUILD` & `FROM` that don't necessarily lead to a command being run in BuildKit. It also links commands to any referenced targets. For example: `COPY +foo/file .` will ensure the Logstream data reflects the fact that the `COPY` command refers to the `+foo` target. This will allow for some more interesting UI features.

The main changes are:
* Some Logbus commands are created ahead of time rather than while listening for solve updates
* Some commands are created and updated in place for pseudo-commands
* If set, the command ID will be used in Earthfile2LLB to link a command to a target

Please have a look at the implementation. I'll take a look at adding tests where possible if this approach makes sense.

### DIFF
--- a/earthfile2llb/converter.go
+++ b/earthfile2llb/converter.go
@@ -115,7 +115,6 @@ type Converter struct {
 	waitBlockStack      []*waitBlock
 	isPipeline          bool
 	logbusTarget        *logbus.Target
-	logbusCommand       *logbus.Command
 	nextCmdID           int
 }
 
@@ -162,35 +161,6 @@ func NewConverter(ctx context.Context, target domain.Target, bc *buildcontext.Da
 
 	logbusTarget.SetStart(time.Now())
 
-	var logbusCommand *logbus.Command
-
-	if opt.BuildCommand {
-		cmdID := fmt.Sprintf("target-command-%s", sts.ID)
-		var gitURL, gitHash, fileRelToRepo string
-		if gitMeta := bc.GitMetadata; gitMeta != nil {
-			gitURL = gitMeta.RemoteURL
-			gitHash = gitMeta.Hash
-			fileRelToRepo = path.Join(gitMeta.RelDir, "Earthfile")
-		}
-		logbusCommand, err = run.NewCommand(
-			cmdID,
-			"build",
-			sts.ID,
-			target.String(),
-			opt.PlatformResolver.Materialize(opt.PlatformResolver.Current()).String(),
-			false, // cached
-			false, // local
-			false, // interactive
-			SourceLocationFromContext(ctx),
-			gitURL,
-			gitHash,
-			fileRelToRepo,
-		)
-		if err != nil {
-			return nil, errors.Wrap(err, "new logbus command")
-		}
-	}
-
 	c := &Converter{
 		target:              target,
 		gitMeta:             bc.GitMetadata,
@@ -206,7 +176,6 @@ func NewConverter(ctx context.Context, target domain.Target, bc *buildcontext.Da
 		containerFrontend:   opt.ContainerFrontend,
 		waitBlockStack:      []*waitBlock{opt.waitBlock},
 		logbusTarget:        logbusTarget,
-		logbusCommand:       logbusCommand,
 	}
 
 	if c.opt.GlobalWaitBlockFtr {
@@ -254,7 +223,7 @@ func (c *Converter) fromClassical(ctx context.Context, imageName string, platfor
 		internal = false
 	}
 	vm := c.vertexMeta(ctx, c.newCmdID(), local, false, internal, nil)
-	err := c.newLogbusCommand(ctx, vm)
+	err := c.newLogbusCommandFromVertex(ctx, vm)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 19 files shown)


earthfile2llb/earthfile2llb.go
```diff
@@ -277,12 +281,18 @@ func Earthfile2LLB(ctx context.Context, target domain.Target, opt ConvertOpt, in
...
+		if parentTarget, ok := opt.Logbus.Run().Target(opt.parentTargetID); ok {
 			parentTarget.AddDependsOn(sts.ID)
 		}
 	}
 
+	if opt.parentCommandID != "" {
+		if parentCmd, ok := opt.Logbus.Run().Command(opt.parentCommandID); ok {
+			parentCmd.AddDependsOn(sts.ID, target.GetName())
+		}
+	}
```
vladaionescu, line 294: This happens before the visited `found` short-circuit. Good.

logbus/command.go
```diff
@@ -83,6 +85,28 @@ func (c *Command) SetStart(start time.Time) {
 	})
 }
 
+// AddDependsOn creates a delta that will be used to merge the specified target
+// ID & name into the command's list of targets on which it depends.
+func (t *Command) AddDependsOn(targetID, refName string) {
+	// Only add the dependency link once to avoid sending duplicates to Logstream.
+	t.mu.Lock()
+	if _, ok := t.dependsOn[targetID]; ok {
+		defer t.mu.Unlock()
```
vladaionescu, line 94: Nit: This is slightly clearer.

```suggestion
		t.mu.Unlock()
```

logbus/solvermon/solvermon.go
```diff
@@ -74,29 +74,48 @@ func (sm *SolverMonitor) handleBuildkitStatus(ctx context.Context, status *clien
...
-				meta.RepoGitURL, meta.RepoGitHash, meta.RepoFileRelToRepo)
-			if err != nil {
-				return err
+			var cp *logbus.Command
+			// Operations initiated from Earthly have created Logbus commands
+			// ahead-of-time. Others may originate from BuildKit, so we'll have
+			// to create a command at this point.
+			if createCmd {
+				var err error
+				cp, err = bp.NewCommand(
```
vladaionescu, line 101: Oh.. so this is still necessary for things that don't map directly to Earthfile commands, right? Stuff like "exporting outputs", "context", and other internal operations, right?

earthfile2llb/converter.go
```diff
@@ -2535,29 +2659,86 @@ func (c *Converter) processNonConstantBuildArgFunc(ctx context.Context) variable
...
+		vm.Interactive,
+		SourceLocationFromContext(ctx),
+		vm.RepoGitURL,
+		vm.RepoGitHash,
+		vm.RepoFileRelToRepo,
+	)
+	return err
+}
+
+func (c *Converter) vertexMeta(ctx context.Context, cmdID int, local, interactive, internal bool, secrets []string) *vertexmeta.VertexMeta {
```
vladaionescu, line 2716: What if we change this to `newVertexMeta` and instead generate the new `cmdID` within? It would simplify the API / call sequence of this a little.

vladaionescu (commented):
This makes sense!

I'm pretty worried that the output of Earthly might be affected by these changes in some non-obvious way, since it's really hard to test for these things automatically.

We should do some pretty intense manual testing for this.

logbus/solvermon/solvermon.go
mikejholly: Yep, you got it. I wasn't able to remove this completely as some statuses don't map to commands on the other end.

vladaionescu: Yeah, makes sense...

earthfile2llb/converter.go
mikejholly: I started with that but decided to separate the two for reasons I can't quite recall... :sweat_smile: 

I'll take a look at combining them. Perhaps it was just that they seem like disparate operations.

vladaionescu: All good if you find that they need to be called independently too - I didn't look carefully.

mikejholly:
> I'm pretty worried that the output of Earthly might be affected by these changes in some non-obvious way, since it's really hard to test for these things automatically.

I haven't noticed any differences so far. The same commands are created upfront instead of in `SolverMonitor`. 

I'll run some large targets using `main` and this branch and `diff` they output to be sure.

earthfile2llb/converter.go
mikejholly: I think that was my original thinking, but as it happens they are always called together. I'll merge them.

vladaionescu:
> I'll run some large targets using main and this branch and diff they output to be sure.

Good idea!

Maybe also check `GIT CLONE` output too via another test. I think there were changes to some of that code too, IIUC.

earthfile2llb/converter.go
mikejholly: Updated.

vladaionescu (approved):

Issue #3706, opened 2024-01-14
Unable to `--allow-privileged` when using artifact reference syntax with `FROM DOCKERFILE`


dustyhorizon:
**What went wrong?**
When the build chain requires `RUN --privileged ...`, `FROM DOCKERFILE` is not able to propagate the necessary flags to allow the artifacts to be properly copied. 

A minimal Earthfile to reproduce this is as follows:

```
VERSION 0.7

source:
    RUN \
        --privileged \
        \
        set -ex \
        && touch a

    SAVE ARTIFACT a

build:
    FROM DOCKERFILE \
        -f +source/a \
        +source
```

It is also noted that there are no flags to specify `--allow-privileged` within a `FROM DOCKERFILE` command.

**What should have happened?**
By specifying `earthly --allow-privileged +build` I should be able to perform the build. 

**Other Helpful Information**
Output from the minimal reproduction

```
❯ earthly -i -P +build
 Init 🚀
————————————————————————————————————————————————————————————————————————————————

           buildkitd | Found buildkit daemon as docker container (earthly-buildkitd)

 Build 🔧
————————————————————————————————————————————————————————————————————————————————

Error: Earthfile line 13:4 from dockerfile: earthfile2llb for +source: Earthfile line 4:4 Permission denied: unwilling to run privileged command; did you reference a remote Earthfile without the --allow-privileged flag?
in              x:develop+source
in              x:develop+build
🛰️ Reuse cache between CI runs with Earthly Satellites! 2-20X faster than without cache. Generous free tier https://cloud.earthly.dev
```

dustyhorizon:
Still an issue on Earthly v0.8.3 with `VERSION 0.8`

PR #3818, opened 2024-02-19
add allow-privileged flag for FROM DOCKERFILE command

dustyhorizon:
This PR introduces the `--allow-privileged` flag to `FROM DOCKERFILE` command.

Fixes https://github.com/earthly/earthly/issues/3706 although not extensively tested yet but seems to be working thus far.


alexcb:
Thanks for the PR, I kicked off a test via #3818 (unfortunately the Approve and run github feature doesn't work with our setup).

For this to be merged in, we'll need to create a new feature flag for the `VERSION` keyword -- here's an example of a similar change when the `--network=none` feature was added: https://github.com/earthly/earthly/commit/2721c63ee6d923d6d801606913de47a90117c697#diff-acbd3cf883cc5793dc1ab8d931dd1c5805e6c80785fe4b9979f415087b04e0d9R35

additionally, can we get a test written for this? You could create a new target under https://github.com/earthly/earthly/blob/34e5a37c0f2ed6f7f21ba9374077d43aa630acd7/tests/dockerfile/Earthfile and under the Dockerfile you could run some commands like we do under https://github.com/earthly/earthly/blob/34e5a37c0f2ed6f7f21ba9374077d43aa630acd7/tests/privileged.earth

alexcb (commented):
These changes look good, see my previous comment on what's needed to get this PR merged.

dustyhorizon:
@alexcb Thanks for the review and inputs, will work out something!

dustyhorizon:
@alexcb I went ahead and added the feature flag and tests per your comment, do let me know if there are any further issues!

tests/Earthfile
```diff
@@ -513,6 +513,8 @@ dockerfile-test:
         --mount=type=tmpfs,target=/tmp/earthly \
         -- --no-output +test
 
+    DO +RUN_EARTHLY --earthfile=dockerfile/privileged.earth --extra_args="--allow-privileged --no-output" --target=+test-allow-privileged-with-privileged-run-command
```
alexcb, line 516: I think all that's needed is to change the `+test` on line 514 to `+test-all`, then delete this line.

earthfile2llb/interpreter.go
```diff
@@ -806,6 +806,12 @@ func (i *Interpreter) handleFromDockerfile(ctx context.Context, cmd spec.Command
 	if len(args) < 1 {
 		return i.errorf(cmd.SourceLocation, "invalid number of arguments for FROM DOCKERFILE")
 	}
+
+	if !i.converter.ftrs.AllowPrivilegedFromDockerfile && opts.AllowPrivileged {
+		return i.errorf(cmd.SourceLocation, "the FROM DOCKERFILE --allow-privileged flag must be enabled with the VERSION --allow-privileged-from-dockerfile feature flag.")
+	}
+	allowPrivileged := opts.AllowPrivileged && i.allowPrivileged
```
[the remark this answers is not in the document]
alexcb: sorry, I take this back. this appears to be consistent with the way `FROM --allow-privileged` works.

alexcb (commented):
This is looking great.

once the minor comment re tests/Earthfile is addressed I can kick of a test.

tests/Earthfile
alexcb: can this line be deleted?

dustyhorizon: Done. Just realized that with the latest fix I don't have to abstract the test anymore :sweat_smile:

alexcb (approved):

alexcb:
Thanks for the contribution!

PR #3833, opened 2024-02-22
Restore missed commits 

mikejholly:
Restores commits from `77eba05`, which was the tip of `docs-0.8` before the updates yesterday.


alexcb:
can we just set `docs-0.8` to be `77eba05` rather than create a new commit for this.

mikejholly:
> can we just set `docs-0.8` to be `77eba05` rather than create a new commit for this.

Sure, I can force push it. I figured I'd get confirmation in a PR.

vladaionescu (approved):

mikejholly:
Force pushed instead.

PR #3836, opened 2024-02-23
Bring back support for org env/flag for log streaming

mikejholly:
We need to support `EARTHLY_ORG` & `--org` to get org ID data to Logstream in all cases. It looks like we inadvertently lost support for these settings with this change: https://github.com/earthly/earthly/commit/cbcad005b087f2177fabbbb8dee908de371eb69d#diff-1ca730ef8be44ea6574fb0c5e06a96627cea73b86f46ba51ba09c5807418527aL158

I'll look into adding a test or two if this looks good.

### DIFF
--- a/cmd/earthly/subcmd/build_cmd.go
+++ b/cmd/earthly/subcmd/build_cmd.go
@@ -625,8 +625,8 @@ func (a *Build) ActionBuildImp(cliCtx *cli.Context, flagArgs, nonFlagArgs []stri
 		// Use the flag/env values for org & project if specified, but fallback to
 		// the PROJECT command if provided.
 		orgName := d.EarthlyOrgName
-		if a.cli.Flags().OrgName != "" {
-			orgName = a.cli.Flags().OrgName
+		if a.cli.OrgName() != "" {
+			orgName = a.cli.OrgName()
 		}
 		projectName := d.EarthlyProjectName
 		if a.cli.Flags().ProjectName != "" {
--- a/cmd/earthly/subcmd/build_cmd.go
+++ b/cmd/earthly/subcmd/build_cmd.go
@@ -9,7 +9,6 @@ import (
 	"path"
 	"path/filepath"
 	"strings"
-	"time"
 
 	"github.com/containerd/containerd/platforms"
 	"github.com/docker/cli/cli/config"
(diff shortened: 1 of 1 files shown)


alexcb (approved):

cmd/earthly/subcmd/build_cmd.go
```diff
@@ -619,17 +618,27 @@ func (a *Build) ActionBuildImp(cliCtx *cli.Context, flagArgs, nonFlagArgs []stri
...
+	// Kick off log streaming upload only when we've passed the necessary
 	// information to logbusSetup. This function will be called right at the
 	// beginning of the build within earthfile2llb.
 	buildOpts.MainTargetDetailsFunc = func(d earthfile2llb.TargetDetails) error {
-		a.cli.Console().VerbosePrintf("Logbus: setting organization %q and project %q at %s", d.EarthlyOrgName, d.EarthlyProjectName, time.Now().Format(time.RFC3339Nano))
-		analytics.AddEarthfileProject(d.EarthlyOrgName, d.EarthlyProjectName)
+		// Use the flag/env values for org & project if specified, but fallback to
+		// the PROJECT command if provided.
+		orgName := d.EarthlyOrgName
+		if a.cli.Flags().OrgName != "" {
```
idodod, line 628: should we fallback to the org in the config?
That is what `a.cli.OrgName()` does

idodod:
> We need to support `EARTHLY_ORG` & `--org` to get org ID data to Logstream in all cases. It looks like we inadvertently lost support for these settings with this change: [cbcad00#diff-1ca730ef8be44ea6574fb0c5e06a96627cea73b86f46ba51ba09c5807418527aL158](https://github.com/earthly/earthly/commit/cbcad005b087f2177fabbbb8dee908de371eb69d#diff-1ca730ef8be44ea6574fb0c5e06a96627cea73b86f46ba51ba09c5807418527aL158)
> 
> I'll look into adding a test or two if this looks good.

For context: I could be wrong but I think we identified that the block of code you linked above was never executed.

cmd/earthly/subcmd/build_cmd.go
mikejholly: Thanks! I'll check it out.

Issue #3840, opened 2024-02-27
Wildcard BUILDs not taking target path into account.


mikejholly:
Should the expansion of wildcards take the path of the target into account?
For context, I ran into the following scenario:

directory:
```
~/projects/lib/utils (main*) » ls -1R Earthfile */Earthfile                                                                                      
Earthfile
dind/Earthfile
git/Earthfile
github/Earthfile
ssh/Earthfile
```
~/projects/lib/utils/Earthfile:

```
VERSION --wildcard-builds 0.8

test:
    BUILD ./*+test
```

earthly command:
```
~/projects/lib (main*) » earthly ./utils+test                                                                                                1 ↵ idodavid@Idos-MacBook-Air
 Init 🚀
————————————————————————————————————————————————————————————————————————————————

           buildkitd | Found buildkit daemon as docker container (earthly-buildkitd)

Streaming logs to https://cloud.earthly.dev/builds/79f27edd-b7d2-4026-be65-7be9d63799d4

 Build 🔧
————————————————————————————————————————————————————————————————————————————————

        ./utils+test | --> FROM +base
              output | --> exporting outputs

 Push Summary ⏫ (disabled)
————————————————————————————————————————————————————————————————————————————————

To enable pushing use earthly --push

 Local Output Summary 🎁
————————————————————————————————————————————————————————————————————————————————



========================== 🌍 Earthly Build  ✅ SUCCESS ==========================

View logs at https://cloud.earthly.dev/builds/79f27edd-b7d2-4026-be65-7be9d63799d4
```
This does work if I cd utils and then:
```
~/projects/lib/utils (main*) » earthly +test
```

PR #3841, opened 2024-02-27
Wildcard BUILD: Account for build command target path

mikejholly:
Addresses: https://github.com/earthly/earthly/issues/3840

Wildcard `BUILD`'s will now account for the correct file path when building using a relative target path (e.g., `earthly ./some/dir/*+test`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to [Earthly](https://github.com/earthly/earthly) will be doc
 
 ## Unreleased
 
-## Fixed
+### Fixed
 
 - Fixes an issue where wildcard `BUILD`'s are invoked from a relative directory (e.g., `earthly ./my-dir/*+target`). [#3840](https://github.com/earthly/earthly/issues/3840)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to [Earthly](https://github.com/earthly/earthly) will be doc
 
 ## Unreleased
 
+## Fixed
+
+- Fixes an issue where wildcard `BUILD`'s are invoked from a relative directory (e.g., `earthly ./my-dir/*+target`). [#3840](https://github.com/earthly/earthly/issues/3840)
+
 ## v0.8.4 - 2024-02-21
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ All notable changes to [Earthly](https://github.com/earthly/earthly) will be doc
 
 ### Fixed
 
-- Fixes an issue where wildcard `BUILD`'s are invoked from a relative directory (e.g., `earthly ./my-dir/*+target`). [#3840](https://github.com/earthly/earthly/issues/3840)
+- Fixes an issue where wildcard `BUILD`'s are invoked from a relative directory (e.g., an `Earthfile` containing `BUILD ./*+test` invoked with `earthly ./rel-dir+target`). [#3840](https://github.com/earthly/earthly/issues/3840)
 
 ## v0.8.4 - 2024-02-21
 
--- a/earthfile2llb/converter.go
+++ b/earthfile2llb/converter.go
@@ -1721,14 +1721,14 @@ func (c *Converter) ExpandWildcard(ctx context.Context, fullTargetName string, c
 		target = parsedTarget
 	}
 
-	matches, err := c.opt.Resolver.ExpandWildcard(ctx, c.opt.GwClient, c.platr, target, parsedTarget.GetLocalPath())
+	matches, err := c.opt.Resolver.ExpandWildcard(ctx, c.opt.GwClient, c.platr, c.opt.RootTarget, target, parsedTarget.GetLocalPath())
 	if err != nil {
 		return nil, err
 	}
 
 	children := []spec.Command{}
 	for _, match := range matches {
-		childTargetName := fmt.Sprintf("./%s+%s", match, parsedTarget.GetName())
+		childTargetName := fmt.Sprintf("%s+%s", match, parsedTarget.GetName())
 
 		childTarget, err := domain.ParseTarget(childTargetName)
 		if err != nil {
--- a/earthfile2llb/converter.go
+++ b/earthfile2llb/converter.go
@@ -1704,7 +1704,6 @@ func (c *Converter) Pipeline(ctx context.Context) error {
 }
 
 func (c *Converter) ExpandWildcard(ctx context.Context, fullTargetName string, cmd spec.Command) ([]spec.Command, error) {
-
 	parsedTarget, err := domain.ParseTarget(fullTargetName)
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to [Earthly](https://github.com/earthly/earthly) will be doc
 
 ## Unreleased
 
+### Fixed
+
+- Fixes an issue where wildcard `BUILD`'s are invoked from a relative directory (e.g., `earthly ./my-dir/*+target`). [#3840](https://github.com/earthly/earthly/issues/3840)
```
vladaionescu, line 9: Isn't this related to `BUILD ./my-dir/*+target` (in the context of `earthly ./rel-path+something`), as opposed to `earthly ./my-dir/*+target`?

earthfile2llb/earthfile2llb.go
```diff
@@ -198,6 +198,9 @@ type ConvertOpt struct {
 
 	// NoAutoSkip disables auto-skip usages.
 	NoAutoSkip bool
+
+	// RootTarget tracks the topmost target used to invoke the build.
+	RootTarget domain.Target
```
vladaionescu, line 203: I'm trying to understand if this is really needed, since the non-glob resolving doesn't have something equivalent. Also, what happens if there's a chain of more than 2 relative target references? Would the logic yield the correct reference after going through the chain?

See the logic here for the non-glob case: https://github.com/earthly/earthly/blob/main/domain/reference.go#L57

Could we adapt that logic to make it work for the glob case too instead?

CHANGELOG.md
mikejholly: Yes, thanks. My example shouldn't contain the `*`.

earthfile2llb/earthfile2llb.go
mikejholly: It's probably not needed. I wasn't sure of the best way to derive the topmost target's path.

What do you mean by "chain" in this context? Can you give me an example?

vladaionescu: Yeah, like `BUILD ./something+foo` then calling `BUILD ../something-else+bar` then calling `BUILD ./my-dir/*+buz`. Maintaining the correct relative path shouldn't depend on the path of the root target, but rather you need to keep track of it every step of the way. This can get a bit complex when remote references are thrown in the mix. That's why I would reuse the code for JoinReferences, because it's been tried and tested to keep track of this stuff for the non-glob case.

tests/Earthfile
```diff
@@ -610,6 +610,26 @@ wildcard-build-pwd:
     RUN cat earthly.output | acbgrep 'hello world 3'
     RUN cat earthly.output | acbgrep -v 'not-test run'
 
+wildcard-build-pwd-rel:
+    COPY wildcard-build subdir
+    COPY wildcard-build.earth subdir/Earthfile
+    DO +RUN_EARTHLY --earthfile=wildcard-build.earth --target=./subdir+wildcard-build-pwd
```
alexcb, line 616: the `--earthfile=wildcard-build.earth` will copy that file into `./Earthfile` which is wrong, since you've already copied it into `./subdir`.
```suggestion
    DO +RUN_EARTHLY --target=./subdir+wildcard-build-pwd
```

tests/Earthfile
```diff
@@ -610,6 +610,26 @@ wildcard-build-pwd:
...
+    RUN cat earthly.output | acbgrep 'hello world 1'
+    RUN cat earthly.output | acbgrep 'hello world 2'
+    RUN cat earthly.output | acbgrep 'hello world 3'
+    RUN cat earthly.output | acbgrep -v 'not-test run'
+
+wildcard-build-rel-dir:
+    WORKDIR /usr/local
```
alexcb, line 623: is this intentional? can we pick a different directory instead, like `/root/special/test/directory`? Using `/usr/local` makes this test feel like this test is installing something on the container's system.

mikejholly: I chose an arbitrary system directory to help test the relative path resolution. I'd rather not create a long-winded custom directory. What about `/home/root`?

alexcb: `/home/root` seems a bit odd too, since typically root is located under `/root`, `/root/some-test-dir` feels more in line with linux file paths.

mikejholly: Yes, thanks. I meant the home dir for root. The difference slipped my mind.

earthfile2llb/earthfile2llb.go
mikejholly: Should be good for another look.

vladaionescu (commented):
Looks good, I think!

Do we have a test for the remote case (the current target is remote, and within this remote target we encounter `BUILD ./*+something`)?

mikejholly:
> Looks good, I think!
> 
> Do we have a test for the remote case (the current target is remote, and within this remote target we encounter `BUILD ./*+something`)?

Good question. I'll look into it.

mikejholly:
I'm not certain there is a remote analog. The original problem was that, when passed a non-current-directory target, the globbing code failed to reconstruct the full path relative to the current directory. With remote targets, the target "path" is always absolute. 

Existing test:

```
earthly github.com/earthly/test-remote/wildcard-build:fb0ebe1c6181bdfc5a7d92165f3317e85e199794+wildcard-glob
```

```
wildcard-build:
    FROM alpine
    BUILD ./subdir/*+test
```

earthfile2llb/converter.go
```diff
@@ -1718,7 +1718,13 @@ func (c *Converter) ExpandWildcard(ctx context.Context, fullTargetName string, c
...
+		// For local targets, we need to determine the full path relative to the
+		// working directory of Earthly in order to glob for matching paths.
+		ref, err := domain.JoinReferences(c.target, parsedTarget)
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to join references")
+		}
+		target = ref.(domain.Target)
 	}
 
 	matches, err := c.opt.Resolver.ExpandWildcard(ctx, c.opt.GwClient, c.platr, target, parsedTarget.GetLocalPath())
```
alexcb, line 1730: I was wondering if `parsedTarget.GetLocalPath()` should be changed to `target.GetLocalPath()`?

Additionally, when I looked up the function which is being called which looks like:

```
// ExpandWildcard will expand a wildcard BUILD target in a local path or remote
// Git repository. The pattern is the path relative to the target path and
// should be in the form 'my/path/*'
func (r *Resolver) ExpandWildcard(ctx context.Context, gwClient gwclient.Client, platr *platutil.Resolver, target domain.Target, pattern string) ([]string, error) {
```

and it's expecting a `pattern` rather than a path here; the difference appears that is it's missing a `/*` suffix?

mikejholly: The arguments have changed a bit. Let me see if I can clean this up a bit by moving the `IsRemote` checks into `ExpandWildcard`. It's a little confusing at the moment.

mikejholly: Updated with some improvements. The remote/local handling is now fully abstracted, which is nice.

vladaionescu (approved):

alexcb (approved):

earthfile2llb/converter.go
alexcb: nice change, definitely easier to follow.

PR #3842, opened 2024-02-28
RUN --aws

mikejholly:
This PR contains an initial draft for the new `RUN --aws` (flag name not finalized) feature.

Give this Earthfile.

```
VERSION --run-with-aws 0.8

FROM alpine

test:
  RUN apk add aws-cli jq
  RUN --no-cache --aws aws ec2 describe-instances | jq -r '.Reservations[].Instances[].InstanceId' | head -10
```

We can inspect the environment & standard config files for AWS credentials and provide them to the build using a new secret provider. 

```
 Init 🚀

Streaming logs to https://cloud.earthly.dev/builds/ea63a53d-a8f0-4cc5-ba73-3ac4e2c3501e

 Build 🔧

              logbus | Setting organization "earthly-technologies" and project "cloud"
           /tmp+test | DOCKERHUB_AUTH=true
           /tmp+test | --> FROM +base
           /tmp+base | DOCKERHUB_AUTH=true
           /tmp+base | --> FROM alpine
           /tmp+base | [----------] 100% FROM alpine
           /tmp+test | DOCKERHUB_AUTH=true
           /tmp+test | *cached* --> RUN apk add aws-cli jq
           /tmp+test | DOCKERHUB_AUTH=true
           /tmp+test | --> RUN --no-cache aws ec2 describe-instances | jq -r '.Reservations[].Instances[].InstanceId' | head -10
           /tmp+test | i-026975d**********
           /tmp+test | i-0ea508e**********
           /tmp+test | i-0951d61**********
           /tmp+test | i-0341024**********
           /tmp+test | i-09b2e95**********
           /tmp+test | i-0e6b0ab**********
           /tmp+test | i-0090ab7**********
           /tmp+test | i-076f49e**********
           /tmp+test | i-0a906b0**********
           /tmp+test | i-017ecedb8fa4f76b7

 Push Summary ⏫ (disabled)

 Local Output Summary 🎁



View logs at https://cloud.earthly.dev/builds/ea63a53d-a8f0-4cc5-ba73-3ac4e2c3501e

```


earthfile2llb/converter.go
```diff
@@ -2236,6 +2237,18 @@ func (c *Converter) internalRun(ctx context.Context, opts ConvertRunOpts) (pllb.
 			extraEnvVars = append(extraEnvVars, fmt.Sprintf("%s=\"$(cat %s)\"", envVar, secretPath))
 		}
 	}
+	// AWS credential import.
+	if opts.WithAWSCredentials {
+		if !c.ftrs.RunWithAWS {
+			return pllb.State{}, errors.New("RUN --aws requires the --run-with-aws feature flag")
```
alexcb, line 2243: These checks tend to be done in the interrupter, which would then allow you to use `i.errorf` which would include an error message containing the Earthfile source location.

alexcb, line 2243: These checks tend to be done in the interrupter, which would then allow you to use `i.errorf` which would include an error message containing the Earthfile source location.

earthfile2llb/converter.go
```diff
@@ -2416,6 +2429,65 @@ func (c *Converter) internalRun(ctx context.Context, opts ConvertRunOpts) (pllb.
 	}
 }
 
+// detectAWSCredentials determines whether AWS envs are set and/or ~/.aws
+// exists. These are then used when fetching the secrets during a build.
+func (c *Converter) detectAWSCredentials(ctx context.Context) ([]llb.RunOption, []string, error) {
+
+	homeDir, err := os.UserHomeDir()
```
alexcb, line 2436: I wonder if we should use https://github.com/earthly/earthly/blob/34320d5111d7488bee0ecbc961b4e9d51d289e06/util/fileutil/homedir.go#L53 instead?

earthfile2llb/converter.go
```diff
@@ -2416,6 +2429,65 @@ func (c *Converter) internalRun(ctx context.Context, opts ConvertRunOpts) (pllb.
...
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "failed to determine user home directory")
+	}
+
+	credsPath := fmt.Sprintf("%s/.aws/credentials", homeDir)
```
alexcb, line 2441: maybe it's because it's still a work in progress, but I don't see where this file gets used.

earthfile2llb/converter.go
```diff
@@ -2416,6 +2429,65 @@ func (c *Converter) internalRun(ctx context.Context, opts ConvertRunOpts) (pllb.
 	}
 }
 
+// detectAWSCredentials determines whether AWS envs are set and/or ~/.aws
+// exists. These are then used when fetching the secrets during a build.
+func (c *Converter) detectAWSCredentials(ctx context.Context) ([]llb.RunOption, []string, error) {
```
alexcb, line 2434: I think the logic related to fetching credentials could be moved somewhere under `util/...`, in order to limit the converter code to llb-specific functionality.

earthfile2llb/converter.go
alexcb: nevermind.... I scrolled down and saw the secret provider below -- this is just to check that these exist before hand.

in this case I think the majority of this error checking could be moved into a `util/...` package, then call it from the interrupter since it's related to the secondary-level parsing of flags, then only keep the llb-specific code in the converter source code.

util/llbutil/secretprovider/aws_creds.go
```diff
@@ -0,0 +1,116 @@
...
+
+// NewAWSCredentialProvider creates and returns a credential provider for AWS.
+func NewAWSCredentialProvider() *AWSCredentialProvider {
+	return &AWSCredentialProvider{}
+}
+
+// GetSecret attempts to find an AWS secret from either the environment or a local config file.
+func (c *AWSCredentialProvider) GetSecret(ctx context.Context, name string) ([]byte, error) {
+
+	names := map[string]int{
```
alexcb, line 30: nit: use a `map[string]struct{}{` to make it more clear this is being used as a set rather than a map.

util/llbutil/secretprovider/aws_creds.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	q, err := url.ParseQuery(name)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to parse secret info")
+	}
+
+	secretName := q.Get("name")
+
+	if _, ok := names[secretName]; ok {
```
alexcb, line 44: only using the name might be too fragile, as some users might have a `RUN --secret AWS_ACCESS_KEY_ID`, which would trigger this code if I'm not mistaken?

util/llbutil/secretprovider/aws_creds.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	return nil, secrets.ErrNotFound
+}
+
+func (c *AWSCredentialProvider) loadFromEnv(ctx context.Context, name string) (string, bool) {
+	return os.LookupEnv(name)
+}
+
+func (c *AWSCredentialProvider) loadFromConfig(ctx context.Context, name string) (string, bool, error) {
```
alexcb, line 66: Have we discussed if anything else is needed to support assuming different roles, or is that outside of this use-case?

If we need to support that, I wonder if there's an aws-specific library that we should be using instead of reading directly from the ini.

util/llbutil/secretprovider/aws_creds.go
mikejholly: I often use `struct{}` for larger maps considering the byte savings. But in this case, I think `int` is a little prettier than this:

```
	names := map[string]struct{}{
		"AWS_ACCESS_KEY_ID":     struct{}{},
		"AWS_SECRET_ACCESS_KEY": struct{}{},
		"AWS_SESSION_TOKEN":     struct{}{},
		"AWS_DEFAULT_REGION":    struct{}{},
	}
```

Update: my editor reminded me that the `struct{}` is redundant here.

```
	names := map[string]struct{}{
		"AWS_ACCESS_KEY_ID":     {},
		"AWS_SECRET_ACCESS_KEY": {},
		"AWS_SESSION_TOKEN":     {},
		"AWS_DEFAULT_REGION":    {},
	}
```

That's nicer. 

image

earthfile2llb/converter.go
mikejholly: Sounds good. I assume you mean "interpreter", right?

util/llbutil/secretprovider/aws_creds.go
mikejholly: IIRC it's not easy to access the credentials using the Go lib once they've been parsed. I'll take another look. I think the goal for the first version was to support the envs & `~/.aws` in a limited fashion so perhaps we can iterate on other profiles, etc.

earthfile2llb/converter.go
alexcb: hah, yeah. spell check fail :facepalm:

mikejholly:
Updated with the following:
* The code now uses "internal" secrets names with an `aws:` prefix to identify these special AWS secrets
* Right now the code can load credentials from the environment or standard config files
* For config files, only the default profile is currently supported. Perhaps a `--aws-profile` would be worth adding next?

earthfile2llb/converter.go
```diff
@@ -2416,6 +2428,41 @@ func (c *Converter) internalRun(ctx context.Context, opts ConvertRunOpts) (pllb.
...
+			llb.SecretFileOpt(0, 0, 0444),
+		}
+		runOpts = append(runOpts, llb.AddSecret(secretPath, secretOpts...))
+		if envName, ok := secretprovider.AWSEnvName(secretName); ok {
+			extraEnvs = append(extraEnvs, fmt.Sprintf("%s=\"$(cat %s)\"", envName, secretPath))
+		}
+	}
+
+	return runOpts, extraEnvs, nil
+}
```
vladaionescu, line 2464: Normally the LLB should not be influenced by the local system, because it can cause the same operation to have different cache hashes, causing unexpected cache misses on different systems. LLB should only be influenced by the syntax (e.g. the existence of the `--aws` flag under `RUN`).

earthfile2llb/converter.go
```diff
@@ -2416,6 +2428,41 @@ func (c *Converter) internalRun(ctx context.Context, opts ConvertRunOpts) (pllb.
...
+	// sourced from the environment during lookup.
+	for _, secretName := range secretprovider.AWSCredentials {
+		secretPath := path.Join("/run/secrets", secretName)
+		secretOpts := []llb.SecretOption{
+			llb.SecretID(c.secretID(secretName)),
+			llb.SecretFileOpt(0, 0, 0444),
+		}
+		runOpts = append(runOpts, llb.AddSecret(secretPath, secretOpts...))
+		if envName, ok := secretprovider.AWSEnvName(secretName); ok {
+			extraEnvs = append(extraEnvs, fmt.Sprintf("%s=\"$(cat %s)\"", envName, secretPath))
```
vladaionescu, line 2459: Is there a good way to make this a bit more uniform? Like if one developer uses `~/.aws`, but another might use the env vars, could we normalize into one of the two representations every time, so that it always shows up the same way within Earthly? It'll create better consistency.

earthfile2llb/converter.go
```diff
@@ -2416,6 +2428,41 @@ func (c *Converter) internalRun(ctx context.Context, opts ConvertRunOpts) (pllb.
 	}
 }
 
+// detectAWSCredentials determines whether AWS envs are set and/or ~/.aws
+// exists. These are then used when fetching the secrets during a build.
+func (c *Converter) detectAWSCredentials(ctx context.Context) ([]llb.RunOption, []string, error) {
+
+	credsAvail, err := awsutil.CredsAvailable()
```
alexcb, line 2435: can this be moved into the into the interpreter, so we could make use of `i.errorf` in order to provide the user with the source code line that is associated with this failure?

mikejholly: Updated.

earthfile2llb/converter.go
mikejholly: Can you elaborate on what you mean by "uniform"?

The same special secret names (below) are added here as env vars whether they're present in the host's env or config file. They are already normalized, in a sense. 

https://github.com/earthly/earthly/blob/mh%2Faws-creds/util/llbutil/secretprovider/aws_creds.go#L17

vladaionescu: Ah ok you're already doing this - awesome! I missed the normalization part.

earthfile2llb/converter.go
mikejholly: The local system is only used at this point to determine if there is some sort of support for AWS creds. The code then uses special, internal secret names to help the AWS secret provider expand them based on the local system's env vars or config files. I don't think the LLB will differ depending on the cred storage type.

vladaionescu: I don't fully understand the logic, but my gut is telling me that something isn't right. LLB should almost never depend on the local system.

Could the logic be made like this:

* Add the LLB only when `--aws` is passed to a `RUN` command.
* If the local system does not have AWS creds to provide, the build results in an error when Buildkit attempts to use the secret provider.

mikejholly: Here's how it works at the moment:

1. Check for `--aws` flag on `RUN`
2. If flag exists and creds are available in environment, add internal secret names as LLB
3. Secret values are then loaded into env vars using the new AWS secret provider
4. Secret provider errors when it can't find one of the expected secrets

The LLB only depends on the environment to an extent--in that it checks if the envs or config files exist beforehand. 

Is there a way to provide these secrets without the LLB definitions?

util/llbutil/secretprovider/aws_creds.go
idodod: Can we use [LoadDefaultConfig](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/config#LoadDefaultConfig) or do we need to load the values in a different way?

mikejholly: I took another look at this. I was initially wary of using the config loader because it hits several different sources and it's not as configurable as it should be IMO.

Unfortunately, it doesn't seem possible to limit the sources to envs and standard config files only. It may also hit other AWS APIs to retrieve some settings.

However, I think it's alright to use this approach as it will likely improve the overall UX (as in it will discover settings from the environment without much user consideration).

The one feature I did disable manually is the EC2 metadata remote source for config. This seems rather peripheral in terms of features we need to support. Plus it generates some ugly errors.  

Example:
```
earthfile2llb group error: async force execution for +basic: unlazy force execution: failed to load AWS credentials: failed to refresh cached credentials, no EC2 IMDS role found, operation error ec2imds: GetMetadata, access disabled to EC2 IMDS via client option, or "AWS_EC2_METADATA_DISABLED" environment variable
```

CC @vladaionescu @alexcb

alexcb: > It may also hit other AWS APIs to retrieve some settings.

I think this is fine? I wouldn't be surprised if their MFA would require something like this?

> Plus it generates some ugly errors.
Maybe we could add to this switch statement under
https://github.com/earthly/earthly/blob/bafb458fda83a990c35d16038bc698497e7bf70e/cmd/earthly/app/run.go#L185 to display these errors in a more user-friendly way?

brandonSc (approved):
Cool stuff! The implementation was pretty straight forward too.

earthfile2llb/converter.go
mikejholly: @vladaionescu and I spoke about this last night. He suggested removing the upfront checks for credentials to remove potential differences in the generated LLB. The new credential provider will return an error if the requested secret isn't available in the environment.

util/llbutil/secretprovider/aws_creds.go
mikejholly: I believe MFA is usually run upfront and will save credentials in the environment or config files.

mikejholly:
@vladaionescu @alexcb @idodod This should be ready to merge. I removed the hack that disables the EC2 metadata check. I figure we can start with the standard `LoadDefaultConfig` heuristic and refine the approach if there are any issues before feature GA.

vladaionescu:
I'll leave the final review to someone else. This is good from my perspective

alexcb (approved):

PR #3847, opened 2024-02-28
Mention DEEP_CLONE can be used in git clone best practices

idodod:


.github/workflows/docs-checks-links.yml
idodod, line 1: the changes here were merged to main already but I had to manually put them here for the workflow to start working (without merging main to docs-0.8 yet)

vladaionescu (approved):

Issue #3868, opened 2024-03-05
Remote cache harbor support


hiddenmarten:
Hello!
_Text mostly copied from a Slack [conversation](https://earthlycommunity.slack.com/archives/C01DL2928RM/p1709565865786759)._

**What existing functionality needs improvement?**
I have a question that I couldn’t find in the documentation. Is there a way to add BuildKit parameters for exporting cache to Earthly’s `--remote-cache`?
In my case, I’m using Harbor and Earthly, and I’m encountering the following [issue](https://github.com/goharbor/harbor/pull/18105):

`Error: error writing manifest blob: failed commit on ref "sha256:e5cabe1a9719d98fd03ef2e67036faeef57795a2d2ba03fd33c68586e7560188": unexpected status from PUT request to https://harbor.company.team/v2/earthly-cache/test/manifests/cache: 404 Not Found`

As a workaround, it might be helpful to try using the flags mentioned in this [comment](https://github.com/goharbor/harbor/pull/18105#issuecomment-1811951274). 
I’ve attempted the following:
`--remote-cache=harbor.company.team/earthly-cache/test:cache,image-manifest=true,oci-mediatypes=true`

However, I encountered an error – invalid reference format Is there another way I can achieve this?

**Expected Behavior**
I would appreciate it if it would be possible to somehow use Harbor as a remote cache ref.

PR #3869, opened 2024-03-05
Parse remote cache attributes

hiddenmarten:
Hi!

Addresses: https://github.com/earthly/earthly/issues/3868
A straightforward solution that works on my side. 

I'm not sure that it's suitable to put these options next to the max mode, if not please lead me to the new flags in earthly that should be created.

### DIFF
--- a/builder/solver.go
+++ b/builder/solver.go
@@ -106,7 +106,7 @@ func (s *solver) newSolveOptMulti(ctx context.Context, eg *errgroup.Group, onIma
 		cacheExports = append(cacheExports, newCacheExportOpt(cacheExportName, attrs, false))
 	}
 	if s.maxCacheExport != "" {
-		maxCacheExportName, attrs, err := parseImageNameAndAttrs(s.cacheExport)
+		maxCacheExportName, attrs, err := parseImageNameAndAttrs(s.maxCacheExport)
 		if err != nil {
 			return nil, errors.Wrapf(err, "parse max cache export %s", s.maxCacheExport)
 		}
--- a/builder/solver.go
+++ b/builder/solver.go
@@ -166,6 +166,8 @@ func newCacheExportOpt(ref string, max bool) client.CacheOptionsEntry {
 	registryCacheOptAttrs["ref"] = ref
 	if max {
 		registryCacheOptAttrs["mode"] = "max"
+		registryCacheOptAttrs["image-manifest"] = "true"
+		registryCacheOptAttrs["oci-mediatypes"] = "true"
 	}
 	return client.CacheOptionsEntry{
 		Type:  "registry",
--- a/builder/solver.go
+++ b/builder/solver.go
@@ -3,7 +3,7 @@ package builder
 import (
 	"context"
 	"io"
-	"strings"
+	"maps"
 
 	"github.com/earthly/earthly/conslogging"
 	"github.com/earthly/earthly/domain"
@@ -32,16 +32,17 @@ type onArtifactFunc func(context.Context, string, domain.Artifact, string, strin
 type onFinalArtifactFunc func(context.Context) (string, error)
 
 type solver struct {
-	sm              *outmon.SolverMonitor
-	logbusSM        *solvermon.SolverMonitor
-	useLogstream    bool
-	bkClient        *client.Client
-	attachables     []session.Attachable
-	enttlmnts       []entitlements.Entitlement
-	cacheImports    *states.CacheImports
-	cacheExport     string
-	maxCacheExport  string
-	saveInlineCache bool
+	sm               *outmon.SolverMonitor
+	logbusSM         *solvermon.SolverMonitor
+	useLogstream     bool
+	bkClient         *client.Client
+	attachables      []session.Attachable
+	enttlmnts        []entitlements.Entitlement
+	cacheImports     *states.CacheImports
+	cacheExport      string
+	maxCacheExport   string
+	cacheExportAttrs map[string]string
+	saveInlineCache  bool
 }
 
 func (s *solver) buildMainMulti(ctx context.Context, bf gwclient.BuildFunc, onImage onImageFunc, onArtifact onArtifactFunc, onFinalArtifact onFinalArtifactFunc, onPullCallback pullping.PullCallback, phaseText string, console conslogging.ConsoleLogger) error {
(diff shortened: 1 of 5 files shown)


alexcb:
That's great to hear it unblocks your issue; however to accept this PR I think we would need to support parsing a config string that's formatted the same way that docker.

I had a look under buildkit and couldn't find any utility function for parsing these attributes -- maybe you could have better luck grepping for something?

Otherwise it feels easy enough for us to parse them via something like:

```
	func parseImageNameAndAttrs(s) (string, map[string]string, error) {
		entries := strings.Split(s, ",")
		imageName := entries[0]
		attrs := map[string][string]
		for _, entry := range entries[1:] {
			pair := strings.Split(strings.TrimSpace(entry), "=")
			//TODO check that len(pair) == 2
			attrs[k] = v
		}
	}
              
```

builder/solver.go
```diff
@@ -166,6 +166,8 @@ func newCacheExportOpt(ref string, max bool) client.CacheOptionsEntry {
 	registryCacheOptAttrs["ref"] = ref
 	if max {
 		registryCacheOptAttrs["mode"] = "max"
+		registryCacheOptAttrs["image-manifest"] = "true"
```
alexcb, line 169: once we handle parsing the image string, I think it's better to assign the attr values outside of this `if` statement.

hiddenmarten: Got it.

hiddenmarten:
I'm not quite sure about error handling, it should be better I suppose.
In general, solution works for me.

alexcb:
> I'm not quite sure about error handling, it should be better I suppose. In general, the solution works for me.
> 
> @alexcb Thanks for the suggestion.

good question!

how about parse them before calling the `newCacheExportOpt` function? something like:

```
$ git diff
diff --git a/builder/solver.go b/builder/solver.go
index be22709a..da7e74a6 100644
--- a/builder/solver.go
+++ b/builder/solver.go
@@ -98,10 +98,18 @@ func (s *solver) newSolveOptMulti(ctx context.Context, eg *errgroup.Group, onIma
        }
        var cacheExports []client.CacheOptionsEntry
        if s.cacheExport != "" {
-               cacheExports = append(cacheExports, newCacheExportOpt(s.cacheExport, false))
+               cacheExportName, attrs, err := parseImageNameAndAttrs(s.cacheExport)
+               if err != nil {
+                       return nil, errors.Wrapf(err, "parse cache export %s", s.cacheExport)
+               }
+               cacheExports = append(cacheExports, newCacheExportOpt(cacheExportName, attrs, false))
        }
        if s.maxCacheExport != "" {
-               cacheExports = append(cacheExports, newCacheExportOpt(s.maxCacheExport, true))
+               maxCacheExportName, attrs, err := parseImageNameAndAttrs(s.cacheExport)
+               if err != nil {
+                       return nil, errors.Wrapf(err, "parse max cache export %s", s.maxCacheExport)
+               }
+               cacheExports = append(cacheExports, newCacheExportOpt(maxCacheExportName, attrs, true))
        }
        if s.saveInlineCache {
                cacheExports = append(cacheExports, newInlineCacheOpt())
```

builder/solver.go
```diff
@@ -178,3 +182,18 @@ func newInlineCacheOpt() client.CacheOptionsEntry {
...
+
+func parseImageNameAndAttrs(s string) (string, map[string]string, error) {
+	entries := strings.Split(s, ",")
+	imageName := entries[0]
+	attrs := make(map[string]string)
+	var err error
+	for _, entry := range entries[1:] {
+		pair := strings.Split(strings.TrimSpace(entry), "=")
+		if len(pair) != 2 {
+			err = errors.New("Export attributes couldn't be parsed")
```
alexcb, line 194: we need to return here, otherwise we will keep executing below and will try to read from `pair[1]` which might not exist.
```suggestion
			return "", "", errors.Errorf("failed to parse export attribute: expected a key=value pair while parsing %q", entry)
```

builder/solver.go
```diff
@@ -178,3 +182,18 @@ func newInlineCacheOpt() client.CacheOptionsEntry {
...
+		}
+		attrs[pair[0]] = pair[1]
```
alexcb, line 196: should we also trim the key/value variables here too?

hiddenmarten: Sure. :)

hiddenmarten:
Okay, it was a bit challenging for me :D

I wasn't able to implement all the modifications in `builder/solver.go` because it caused issues with the imports. 

Consequently, I opted to relocate the parsing to `build_cmd.go`. I hope this won't cause any inconvenience.

cmd/earthly/subcmd/build_cmd.go
```diff
@@ -500,21 +500,28 @@ func (a *Build) ActionBuildImp(cliCtx *cli.Context, flagArgs, nonFlagArgs []stri
 	}
 
 	cacheImports := make([]string, 0)
+	var cacheImportImageName string
 	if a.cli.Flags().RemoteCache != "" {
-		cacheImports = append(cacheImports, a.cli.Flags().RemoteCache)
+		cacheImportImageName, _, err = a.parseImageNameAndAttrs(a.cli.Flags().RemoteCache)
```
alexcb, line 505: What's the reason for discarding the attributes here? It would be better to raise an error saying attributes are not supported by 

What about error handling?

cmd/earthly/subcmd/build_cmd.go
```diff
@@ -500,21 +500,28 @@ func (a *Build) ActionBuildImp(cliCtx *cli.Context, flagArgs, nonFlagArgs []stri
...
 	if len(a.cacheFrom.Value()) > 0 {
 		cacheImports = append(cacheImports, a.cacheFrom.Value()...)
 	}
 	var cacheExport string
 	var maxCacheExport string
+	var cacheExportAttrs map[string]string
 	if a.cli.Flags().RemoteCache != "" && a.cli.Flags().Push {
 		if a.cli.Flags().MaxRemoteCache {
-			maxCacheExport = a.cli.Flags().RemoteCache
+			maxCacheExport, cacheExportAttrs, err = a.parseImageNameAndAttrs(a.cli.Flags().RemoteCache)
```
alexcb, line 516: since the max cache setting ends up in a different variable name, I would create a similar one for attributes
```suggestion
			maxCacheExport, maxCacheExportAttrs, err = a.parseImageNameAndAttrs(a.cli.Flags().RemoteCache)
```

(alternatively I would remove the `MaxCacheExport` field entirely, and just set the `max` attribute to true, but please don't do this in this PR, since it might be a larger task)

cmd/earthly/subcmd/build_cmd.go
```diff
@@ -500,21 +500,28 @@ func (a *Build) ActionBuildImp(cliCtx *cli.Context, flagArgs, nonFlagArgs []stri
...
 		} else {
-			cacheExport = a.cli.Flags().RemoteCache
+			cacheExport, cacheExportAttrs, err = a.parseImageNameAndAttrs(a.cli.Flags().RemoteCache)
 		}
 	}
+	if err != nil {
```
alexcb, line 521: can you do this immediately after each call

and please use a clearer error message, like the one I suggested earlier:

> return nil, errors.Wrapf(err, "parse max cache export %s", s.maxCacheExport)

cmd/earthly/subcmd/build_cmd.go
hiddenmarten: From what I understand, in [buildkit](https://github.com/moby/buildkit?tab=readme-ov-file#output), there are only output keys and no input keys, which in Earthly terminology translates to export attributes without any corresponding import attributes. Consequently, these cannot be utilized in the import process without causing issues.

The critical aspect here is the --remote-cache flag, which facilitates the provision of a ref that is applicable to both exporting and importing. To address my concern, I should be able to specify attributes exclusively for the export process.

Additionally, I'm unclear about the import mechanism in Earthly. Specifically, I attempted to eliminate input attributes in the `solver.go` file, but this effort was unsuccessful. Earthly defaulted to using export attributes for the import process, which resulted in the malfunction of the import function.

hiddenmarten:
I'll update the error handling soon.

cmd/earthly/subcmd/build_cmd.go
alexcb: I'm just unsure if there's a chance the code was designed to support both a min and max export mode; that's why I'm wary of making an even bigger change in this PR.

hiddenmarten: I agree with you on this point.

P.S. Honestly, I would recommend considering deprecating the `--max-remote-cache` option and just adding a flag like: `--export-attributes`, it would be easier from my point of view.

hiddenmarten: About your commit suggestion, it will lead to creating additional fields 
```
	maxCacheExportAttrs map[string]string
```
in `builder` and `solver` structs, is that okay for you?

alexcb: yes that is what I'm asking for, thanks.

cmd/earthly/subcmd/build_cmd.go
alexcb: this hasn't been fully addressed -- can you do this immediately after each call

cmd/earthly/subcmd/build_cmd.go
hiddenmarten: Okay, I'll do it now :)

hiddenmarten: @alexcb done

cmd/earthly/subcmd/build_cmd.go
hiddenmarten: I've added it in one more place

cmd/earthly/subcmd/build_cmd.go
alexcb: Sorry I misunderstood this change earlier, I didn't appreciate the fact that the import is _also_ set to the same `RemoteCache` value that's used for the export.

can you move the error handling to into this if statement -- currently the error handling is being run even when this code block is skipped.

In the case of imports, it might have been easier to just move `parseImageNameAndAttrs` into a utility package (and rename to `ParseImageNameAndAttrs`), which could be called both here and in the solver.go code.

hiddenmarten: Sure, done.
@alexcb

alexcb (approved):

alexcb:
Thanks for the help with this issue, and getting it across the finish line :)

hiddenmarten:
@alexcb 
Thank you for the perfect lead :)

Issue #3875, opened 2024-03-07
Regression: COPY --if-exists + wildcard is broken in v0.8


idelvall:
The following command fails in `v.0.8.x` but works in `v0.7.23`

```
earthly --no-cache +build
```
```
VERSION 0.7

build:
 FROM alpine
 COPY --if-exists non_existing_folder/* /
```

mikejholly:
[This PR](https://github.com/earthly/earthly/pull/3652) may have introduced a regression with the wildcard case.

idelvall:
Thanks Mike!

PR #3877, opened 2024-03-07
Fix COPY --if-exists with wildcard

mikejholly:
Fixes: https://github.com/earthly/earthly/issues/3875

This PR fixes the issue, but I'm curious why the `!strings.Contains(src, "*")` was added in the first place. It seems like all of the relevant tests pass without this extra condition.

### DIFF
--- a/tests/Earthfile
+++ b/tests/Earthfile
@@ -261,7 +261,7 @@ copy-test-wildcard:
     DO +RUN_EARTHLY --earthfile=copy.earth --target=+copy-wildcard
 
 copy-test-if-exists-wildcard:
-    #DO +RUN_EARTHLY --earthfile=copy.earth --target=+copy-if-exists-wildcard
+    DO +RUN_EARTHLY --earthfile=copy.earth --target=+copy-if-exists-wildcard
 
     RUN mkdir existing_folder
     DO +RUN_EARTHLY --earthfile=copy.earth --target="+copy-if-exists-wildcard-exists-empty"
--- a/tests/Earthfile
+++ b/tests/Earthfile
@@ -33,6 +33,7 @@ ga-no-qemu-group1:
     BUILD +copy-keep-own-test
     BUILD +copy-test-empty-src
     BUILD +copy-test-if-exists-wildcard
+    BUILD +copy-test-wildcard
     BUILD +git-clone-test
     BUILD +builtin-args-invalid-default-test
     BUILD +builtin-args-invalid-pass-test
@@ -253,7 +254,6 @@ copy-test:
     DO +RUN_EARTHLY --earthfile=copy.earth --output_does_not_contain="which does not expand to a home directory"
 
 copy-test-empty-src:
-    # Regression tests for 'COPY --if-exists "" <dst>' (empty string)
     DO +RUN_EARTHLY --earthfile=copy.earth --target=+copy-empty-src
     DO +RUN_EARTHLY --earthfile=copy.earth --target=+copy-empty-src-if-exists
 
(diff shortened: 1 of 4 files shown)


tests/Earthfile
```diff
@@ -256,6 +257,19 @@ copy-test-empty-src:
     DO +RUN_EARTHLY --earthfile=copy.earth --target=+copy-empty-src
     DO +RUN_EARTHLY --earthfile=copy.earth --target=+copy-empty-src-if-exists
 
+copy-test-wildcard:
+    DO +RUN_EARTHLY --earthfile=copy.earth --target=+copy-wildcard
+
+copy-test-if-exists-wildcard:
+    # Regression test for 'COPY --if-exists non_existent_folder/* /'
```
alexcb, line 264: Nit: I would drop the "Regression test" from this comment, maybe just:
```suggestion
    # test 'COPY --if-exists non_existent_folder/* /' works when "non_existent_folder" doesn't exist.
```

tests/Earthfile
```diff
@@ -256,6 +257,19 @@ copy-test-empty-src:
...
+    DO +RUN_EARTHLY --earthfile=copy.earth --target=+copy-if-exists-wildcard
+
+    RUN mkdir non_existent_folder
```
alexcb, line 267: hah.... maybe "non_existent_folder" should be named to something else, since it exists now.

tests/copy.earth
```diff
@@ -222,3 +222,9 @@ copy-empty-src:
 copy-empty-src-if-exists:
     ARG SRC=""
     COPY --if-exists $SRC $SRC
+
+copy-if-exists-wildcard:
+    COPY --if-exists non_existing_folder/* /
```
alexcb, line 227: what about adding:

```
ARG test_foo_exists=false
IF [ "$test_foo_exists" ]
    RUN test -f touch non_existent_folder/foo
END
```

so we can then call:

```
DO +RUN_EARTHLY --earthfile=copy.earth --target="+copy-if-exists-wildcard --test_foo_exists=true"
```

and perhaps another similar arg for testing the expected outcome when `non_existent_folder` exists but is empty?

tests/Earthfile
mikejholly: I removed it as it's somewhat redundant anyhow.

util/llbutil/copyop.go
```diff
@@ -30,17 +30,12 @@ func CopyOp(ctx context.Context, srcState pllb.State, srcs []string, destState p
...
 	for _, src := range srcs {
-		// If the copy came in as optional (ifExists), then we need to trigger
-		// the underlying wildcard matching and allow empty wildcards. The
-		// matching uses the filepath match syntax. So by simply creating a
-		// wildcard where the first letter needs to match the current first
-		// letter gets us the single match; and no error if it is missing. Paths
-		// with wildcards will be automatically ignored when no match is present.
-		if ifExists && len(src) != 0 && !strings.Contains(src, "*") {
-			// Normalize path by dropping './'
+		// HACK: For COPY --if-exists, we can use a glob expression (e.g., '[f]oo') to
```
alexcb, line 33: I'm confused by the comment -- where are we using a glob expression?

mikejholly:
```
src = fmt.Sprintf("[%s]%s", string(src[0]), string(src[1:]))
```
This code uses a bit of a hack to have BK treat the file(s) as optional. `foo` becomes `[f]oo`, which has BK interpret the source as an expression to expand.

tests/copy.earth
mikejholly: Updated with `test -f`. I used separate targets as the parameters were more complicated than the extra 3 lines.

tests/Earthfile
mikejholly: Updated.

util/llbutil/copyop.go
alexcb: oh wow.... yeah I didn't understand what that particular line was doing. This is really hacky.

A comment directly above this line would make it easier, maybe something like:

```
// HACK: a regular expression is constructed, e.g. "foo" becomes "[f]oo"
src = fmt.Sprintf("[%s]%s", string(src[0]), string(src[1:]))
```

mikejholly: Sure, I moved the comment.

alexcb (approved):

util/llbutil/copyop.go
alexcb: thanks, it feels easier to follow now.

Issue #3641, opened 2023-12-21
Emit notice when unnecessary feature flags are used


eliottwiener-gridunity:
**What existing functionality needs improvement?**

I've found Earthly's [feature flags](https://docs.earthly.dev/docs/earthfile/features) to be a useful way to try out experimental features. However, over time my Earthfile's `VERSION` line became cluttered with feature flags. At some point, I looked at the documentation again, and realized that most of the flags I had been using were no longer necessary, as the features had been accepted and integrated into the release as non-experimental features. I was thinking that it would be nice if Earthly notified the user if their Earthfile has feature flags that are no longer necessary.

EXAMPLE:

Let's say I have an Earthfile like:

```
VERSION --use-copy-link 0.7
FROM scratch
```

The feature corresponding to `--use-copy-link` was accepted for 0.7, and so the `--use-copy-link` flag is unnecessary in this example Earthfile. 

**Expected Behavior**

Upon invocation, Earthly would emit a notice to the user, informing them that the feature flag is unnecessary. This message would be informational only, and would not affect Earthly's execution otherwise.

Using the note about `COMMAND` being renamed to `FUNCTION` in 0.8 as an example, this message might appear in the output like so:

```
$ earthly +base
 Init 🚀
——————————————————————————————————————————————

buildkitd | Found buildkit daemon as docker container (earthly-buildkitd)

 Build 🔧
——————————————————————————————————————————————
Note that the feature flag --use-copy-link is not necessary for version 0.7, since the feature was enabled by default. You may remove this flag from your Earthfile.
```

alexcb:
I like the idea; it doesn't seem that difficult to implement either.

we have https://github.com/earthly/earthly/blob/c48f19feca0d4742990b5ed48150df93eb48e58f/features/features.go#L193-L196

which parses the initial `VERSION --flag --otherflag ... X.Y` string,

then the corresponding parts that sets the bools to true: https://github.com/earthly/earthly/blob/c48f19feca0d4742990b5ed48150df93eb48e58f/features/features.go#L256-L266

all we would have to do is check to see if these bools are already true and raise a warning if they are.

the naive approach would be to do, for example:

```
                if ftrs.NoNetwork {
                    issue warning
                }
		ftrs.NoNetwork = true
```

but that seems rather tedious (and error-prone) and should be avoided if possible -- maybe we can use `go generate` to come up with a different approach for this. The good news is this is all contained in this single `Get` method, which limits scope.

[...]

PR #3896, opened 2024-03-12
Adjust feature flag handling in parseFeatures.go and features.go

danqixu:
Issue: #3641


buildcontext/parsefeatures.go
```diff
@@ -27,6 +28,13 @@ func parseFeatures(buildFilePath string, featureFlagOverrides string, projectRef
 		return nil, fmt.Errorf("No version specified in %s/Earthfile", projectRef)
 	}
 
+	warningStrs, err := ftrs.Adjust()
+	if err != nil {
+		return nil, err
+	}
+
+	console.Printf("Note that the feature flag %s is not necessary for version %s, since the feature was enabled by default. You may remove this flag from your Earthfile.", strings.Join(warningStrs, ", "), ftrs.Version())
```
alexcb, line 36: this should only be printed when `warningStrs` has len > 0

features/features.go
```diff
@@ -309,3 +255,68 @@ func FromContext(ctx context.Context) *Features {
...
+
+	if f.ArgScopeSet && !f.ShellOutAnywhere {
+		// ArgScopeSet uses new ARG declaration logic that requires
+		// ShellOutAnywhere. We're erroring here to ensure that users get that
+		// feedback as early as possible.
+		return nil, errors.New("--arg-scope-and-set requires --shell-out-anywhere")
+	}
+
+	return warningStrs, nil
+}
```
alexcb, line 322: nit: this is not a POSIX file; we need a final newline.

alexcb (commented):
I like the approach, but is there a way we can avoid duplicating the `--flag-names`? maybe we could use reflection to pull them out of the `Features` struct?

danqixu:
> I like the approach, but is there a way we can avoid duplicating the `--flag-names`? maybe we could use reflection to pull them out of the `Features` struct?


There is no good way. If reflection is used, we still need to fill in the serial numbers of all fields. Once the order in the structure changes, we still need to remember to modify this code, which is not intuitive and easy to maintain. I thought about using reflection but gave up

alexcb:
> > I like the approach, but is there a way we can avoid duplicating the `--flag-names`? maybe we could use reflection to pull them out of the `Features` struct?
> 
> There is no good way. If reflection is used, we still need to fill in the serial numbers of all fields. Once the order in the structure changes, we still need to remember to modify this code, which is not intuitive and easy to maintain. I thought about using reflection but gave up

I wonder if we could use the Features struct's tags to also include the earthly version where the values are enabled by default?

If you haven't already, maybe look at github.com/jessevdk/go-flags for ideas of how it uses reflection to read the tags as well as set the fields in the struct.

danqixu:
If you don't mind the problem I asked, I can use reflection

like:
```golang
// Enable version-specific features.
    if versionAtLeast(*f, 0, 5) {
        enableFunc(&f.ExecAfterParallel, 3)
        enableFunc(&f.ParallelLoad, 4)
        enableFunc(&f.UseRegistryForWithDocker, 5)
    }
    if versionAtLeast(*f, 0, 6) {
        enableFunc(&f.ForIn, 6)
        enableFunc(&f.NoImplicitIgnore, 7)
        enableFunc(&f.ReferencedSaveOnly, 8)
        enableFunc(&f.RequireForceForUnsafeSaves, 9)
        enableFunc(&f.UseCopyIncludePatterns, 10)
    }
...
```

features/features.go
danqixu: amended

alexcb:
> If you don't mind the problem I asked, I can use reflection
> 
> like:
> 
> ```go
> // Enable version-specific features.
>     if versionAtLeast(*f, 0, 5) {
>         enableFunc(&f.ExecAfterParallel, 3)
>         enableFunc(&f.ParallelLoad, 4)
>         enableFunc(&f.UseRegistryForWithDocker, 5)
>     }
>     if versionAtLeast(*f, 0, 6) {
>         enableFunc(&f.ForIn, 6)
>         enableFunc(&f.NoImplicitIgnore, 7)
>         enableFunc(&f.ReferencedSaveOnly, 8)
>         enableFunc(&f.RequireForceForUnsafeSaves, 9)
>         enableFunc(&f.UseCopyIncludePatterns, 10)
>     }
> ...
> ```

this is really hard to follow.

I think looking into using tags, as I previously suggested, would be cleaner.

danqixu:
> > If you don't mind the problem I asked, I can use reflection
> > like:
> > ```go
> > // Enable version-specific features.
> >     if versionAtLeast(*f, 0, 5) {
> >         enableFunc(&f.ExecAfterParallel, 3)
> >         enableFunc(&f.ParallelLoad, 4)
> >         enableFunc(&f.UseRegistryForWithDocker, 5)
> >     }
> >     if versionAtLeast(*f, 0, 6) {
> >         enableFunc(&f.ForIn, 6)
> >         enableFunc(&f.NoImplicitIgnore, 7)
> >         enableFunc(&f.ReferencedSaveOnly, 8)
> >         enableFunc(&f.RequireForceForUnsafeSaves, 9)
> >         enableFunc(&f.UseCopyIncludePatterns, 10)
> >     }
> > ...
> > ```
> 
> this is really hard to follow.
> 
> I think looking into using tags, as I previously suggested, would be cleaner.

I've updated my previous comment to clarify my concerns about using reflection to get the tag.

alexcb:
I think there's a disconnect in communicate here; I was suggesting we add a new filed, such as:

```
enabled_in_version: "0.7"
```

to the `Features` struct,

So for example, this field

```
CheckDuplicateImages     bool `long:"check-duplicate-images" description:"check for duplicate images during output"`
```

would be updated to

```
CheckDuplicateImages     bool `long:"check-duplicate-images" description:"check for duplicate images during output" enabled_in_version: "0.7"`
```

Then we would use reflect (in the same way that github.com/jessevdk/go-flags works).

Using serial numbers of fields is too difficult to maintain.

danqixu:
> I think there's a disconnect in communication here; I was suggesting we add a new filed, such as:
> 
> ```
> enabled_in_version: "0.7"
> ```
> 
> to the `Features` struct,
> 
> So for example, this field
> 
> ```
> CheckDuplicateImages     bool `long:"check-duplicate-images" description:"check for duplicate images during output"`
> ```
> 
> would be updated to
> 
> ```
> CheckDuplicateImages     bool `long:"check-duplicate-images" description:"check for duplicate images during output" enabled_in_version: "0.7"`
> ```
> 
> Then we would use reflect (in the same way that github.com/jessevdk/go-flags works).
> 
> Using serial numbers of fields is too difficult to maintain.

Oh I got it, I will try your suggestion.

danqixu:
I have made the changes as requested, could you please check if there are any issues?

danqixu:
@alexcb

features/features.go
```diff
@@ -309,3 +252,57 @@ func FromContext(ctx context.Context) *Features {
...
+
+func ParseVersion(version string) (int, int) {
```
alexcb, line 291: since this function panics, we should rename it to `must....` to make it more clear.

```suggestion
func MustParseVersion(version string) (int, int) {
```

buildcontext/parsefeatures.go
```diff
@@ -27,6 +28,15 @@ func parseFeatures(buildFilePath string, featureFlagOverrides string, projectRef
 		return nil, fmt.Errorf("No version specified in %s/Earthfile", projectRef)
 	}
 
+	warningStrs, err := ftrs.Adjust()
+	if err != nil {
+		return nil, err
+	}
+
+	if len(warningStrs) > 0 {
+		console.Printf("Note that the feature flag %s is not necessary for version %s, since the feature was enabled by default. You may remove this flag from your Earthfile.", strings.Join(warningStrs, ", "), ftrs.Version())
```
alexcb, line 37: Let's reword this to be a bit shorter:
```suggestion
		console.Printf("NOTE: The %s feature is enabled by default under VERSION %s, and can be safely removed from the VERSION command", strings.Join(warningStrs, ", "), ftrs.Version())
```

features/features.go
```diff
@@ -309,3 +252,57 @@ func FromContext(ctx context.Context) *Features {
 	}
 	return nil
 }
+
+func (f *Features) Adjust() ([]string, error) {
```
alexcb, line 256: I think having "flags" appear in this method name would make it a bit more readable

```suggestion
func (f *Features) processFlags() ([]string, error) {
```

alexcb (commented):
This is looking really good! just a few minor suggestions, thanks for sticking with it!

features/features.go
```diff
@@ -309,3 +252,57 @@ func FromContext(ctx context.Context) *Features {
...
+		if len(version) == 0 {
+			continue
+		}
+		majorVersion, minorVersion := ParseVersion(field.Tag.Get("enabled_in_version"))
+		if versionAtLeast(*f, majorVersion, minorVersion) && value.Kind() == reflect.Bool {
+			if value.Bool() {
+				tagName := field.Tag.Get("long")
+				warningStrs = append(warningStrs, fmt.Sprintf("--%s", strings.ToLower(tagName)))
+			}
+		value.SetBool(true)
```
alexcb, line 275: indentation issue (although it didn't cause `+lint` to fail which is odd)
```suggestion
			value.SetBool(true)
```

alexcb (approved):

Issue #3895, opened 2024-03-12
autoskip - targets which should be rerun are skipped as cached


idodod:
**What went wrong?**

While using the go docker image as the base image to build and run tests for go applications, I noticed that after upgrading the go version, the targets that build & test are still skipped even though I expected those targets to run with the new go version.

common/Earthfile
```Earthfile
VERSION 0.8

ARG GOLANG_VERSION=1.21-alpine
FROM golang:$GOLANG_VERSION
```

some-service/Earthfile
```Earthfile
FROM ../common+base
RUN apk add --update --no-cache build-base ca-certificates curl
WORKDIR /app

deps:
    #COPY go files
    COPY go.mod go.sum ./
    RUN go mod download
```

Earthfile:
```Earthfile
build-services:
    BUILD --autoskip ./some-service+deps
```

`earthly --autoskip +build-services`

**What should have happened?**

Autoskip should have detected that the base image changed and rerun the target

**Other Helpful Information**

mikejholly:
I was able to reproduce this with:

```
VERSION --build-auto-skip 0.8

ARG VERSION=3.18.6
FROM alpine:$VERSION

parent:
    RUN cat /etc/os-release

expand-args-from:
    BUILD --auto-skip +parent
```

PR #3903, opened 2024-03-14
Auto-skip: ensure we account for +base

mikejholly:
The auto-skip code did not account for the "base" target. Hence, the change to root-level `ARG`s were not breaking the cache. 

https://github.com/earthly/earthly/issues/3895

### DIFF
--- a/tests/autoskip/Earthfile
+++ b/tests/autoskip/Earthfile
@@ -33,6 +33,7 @@ test-all:
     BUILD +test-build-flag
     BUILD +test-flag-conflict
     BUILD +test-init-failure
+    BUILD +test-arg-change
 
 test-files:
     RUN echo hello > my-file
@@ -276,6 +277,14 @@ test-init-failure:
     DO --pass-args tests+RUN_EARTHLY --extra_args="--auto-skip-db-path=/tmp --auto-skip" \
                                      --earthfile=build-flag.earth --target=+basic \
                                      --output_contains="Failed to initialize auto-skip database"
+
+test-arg-change:
+    COPY arg-change.earth Earthfile
+    RUN cat Earthfile
+    DO --pass-args tests+RUN_EARTHLY --extra_args="--auto-skip-db-path=test.db" --target=+expand-args-from --output_contains="VERSION_ID=3.18.6"
+    RUN sed -i "s/3.18.6/3.19.1/g" Earthfile
+    RUN cat Earthfile
+    DO --pass-args tests+RUN_EARTHLY --extra_args="--auto-skip-db-path=test.db" --target=+expand-args-from --output_contains="VERSION_ID=3.19.1"
 
 RUN_EARTHLY_ARGS:
     FUNCTION
--- a/tests/autoskip/Earthfile
+++ b/tests/autoskip/Earthfile
@@ -279,10 +279,15 @@ test-init-failure:
                                      --output_contains="Failed to initialize auto-skip database"
 
 test-arg-change:
+    # Please don't update. These versions are pinned to ensure a version
+    # argument change triggers auto-skip cache-busting.
+    ARG V1="3.18.6"
+    ARG V2="3.19.1"
     COPY arg-change.earth Earthfile
-    DO --pass-args tests+RUN_EARTHLY --extra_args="--auto-skip-db-path=test.db" --target=+expand-args-from --output_contains="VERSION_ID=3.18.6"
-    RUN sed -i "s/3.18.6/3.19.1/g" Earthfile
-    DO --pass-args tests+RUN_EARTHLY --extra_args="--auto-skip-db-path=test.db" --target=+expand-args-from --output_contains="VERSION_ID=3.19.1"
+    RUN sed -i "s/<VERSION>/$V1/g" Earthfile
+    DO --pass-args tests+RUN_EARTHLY --extra_args="--auto-skip-db-path=test.db" --target=+expand-args-from --output_contains="VERSION_ID=$V1"
+    RUN sed -i "s/$V1/$V2/g" Earthfile
+    DO --pass-args tests+RUN_EARTHLY --extra_args="--auto-skip-db-path=test.db" --target=+expand-args-from --output_contains="VERSION_ID=$V2"
 
 RUN_EARTHLY_ARGS:
     FUNCTION
--- a/tests/autoskip/Earthfile
+++ b/tests/autoskip/Earthfile
@@ -280,10 +280,8 @@ test-init-failure:
 
 test-arg-change:
     COPY arg-change.earth Earthfile
-    RUN cat Earthfile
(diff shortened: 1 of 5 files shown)


tests/autoskip/arg-change.earth
```diff
@@ -0,0 +1,12 @@
+VERSION --build-auto-skip 0.8
+
+ARG VERSION=3.18.6
+FROM alpine:$VERSION
+
+PROJECT earthly-technologies/cloud
```
idodod, line 6: nit: 
```suggestion
PROJECT earthly-technologies/core
```

or remove completely?

idodod (approved):
awesome! thanks for the quick turnaround

tests/autoskip/arg-change.earth
```diff
@@ -0,0 +1,10 @@
+VERSION --build-auto-skip 0.8
+
+ARG VERSION=3.18.6
```
alexcb, line 3: I worry that someone might be tempted to bump this version (thinking it's good to keep up to date); however it won't be obvious that it needs to stay in sync with the corresponding test in the Earthfile.

I have two ideas:

1. change this to `ARG VERSION=PLACEHOLDER_FOR_DEFAULT_VALUE`, and then in the `test-arg-change` target, do a `sed -i "s/PLACEHOLDER_FOR_DEFAULT_VALUE/3.18.6"`.

or

2. instead of changing the versions, just change the image, e.g.

```
ARG IMG=alpine
FROM $IMG
```

then `sed -i "s/alpine/busybox/g" Earthfile`, and then they will just always use the latest.

util/deltautil/Earthfile
```diff
@@ -1,6 +1,7 @@
 VERSION 0.8
 
-FROM github.com/earthly/earthly+base
+# Note: the tag below is needed for auto-skip.
+FROM github.com/earthly/earthly:tags/v0.8.5+base
```
alexcb, line 4: could we just go back to using `../..+base` instead?

tests/autoskip/arg-change.earth
mikejholly: Good point. Does a comment or special Renovate ignore flag sound fair? @idodod do you know how to achieve the latter?

util/deltautil/Earthfile
mikejholly: I think these full target names were added because `deltautil` & `ast` are separate packages. I'm concerned that using relative paths may break something.

alexcb: how about using `ARG EARTHLY_GIT_BRANCH` instead?

e.g. 

```
ARG EARTHLY_GIT_BRANCH
FROM github.com/earthly/earthly:$EARTHLY_GIT_BRANCH+base
```

alexcb: or `EARTHLY_GIT_HASH`

mikejholly: Good idea. Branches don't work with auto-skip, but hashes will.

tests/autoskip/arg-change.earth
mikejholly: To elaborate, it would be useful to have a way to pin versions without the extra effort.

util/deltautil/Earthfile
mikejholly: Actually, the hash will break local runs for unpushed commits. :thinking:

tests/autoskip/arg-change.earth
idodod: We don't use Renovate (yet) to upgrade the images in Earthfiles but once we do it shouldn't be a problem for it to ignore this specific image version. 

To Alex's point though, Someone (like me:)) might still feel the urge to upgrade it manually. Maybe using a digest might make it clearer that it shouldn't be upgraded?

util/deltautil/Earthfile
idodod: When running locally wouldn't EARTHLY_GIT_HASH point to the local hash that might not exist in the remote repo yet?

idodod: If we use the tag/version, we can configure Renovate to upgrade the version together with all the other places in the repo.

mikejholly: 3-4 poor options :cry: 
* Use relative paths and possibly break other code that depends on this
* Use hashes and break builds locally
* Use version tag and have to update (unexpected results for local runs)
* Copy and paste the 3-4 root Earthfile `+base` commands

I think I prefer the last option.

alexcb: I don't see any discussion about why this was done under #2579 and don't see any references to it under our other repos either.

it was initially added as `FROM github.com/earthly/earthly+base`; if instead it was changed to it from a relative import there would be more git history for us to dig into to understand why.

I think we should try using `../..+base` and then if we discover anything breaks then we come back and revert this, and add a comment explaining why.

tests/autoskip/arg-change.earth
alexcb: > it would be useful to have a way to pin versions

my second idea of using alpine vs busybox would allow this.

```
ARG IMG=alpine:3.19
FROM $IMG
```

then update this:

```
DO --pass-args tests+RUN_EARTHLY --extra_args="--auto-skip-db-path=test.db" --target=+expand-args-from --output_contains="IMG=alpine"
RUN sed -i "s/alpine:.*$/busybox:1.36.1/g" Earthfile
DO --pass-args tests+RUN_EARTHLY --extra_args="--auto-skip-db-path=test.db" --target=+expand-args-from --output_contains="IMG=busybox"
```

alexcb: update: my incorrectly understood the `output_contains="VERSION_ID=3.18.6"` part of the test, this requires `cat /etc/os-release`.

Mike and I chatted about this, and instead we can go with the first idea where the version is only specified in `tests/autoskip/Earthfile`, and will use sed to update the value in the `tests/autoskip/arg-change.earth` file

tests/autoskip/Earthfile
```diff
@@ -279,10 +279,15 @@ test-init-failure:
                                      --output_contains="Failed to initialize auto-skip database"
 
 test-arg-change:
+    # Please don't update. These versions are pinned to ensure a version
```
alexcb, line 282: I would reword this to be:

```
# this test requires two different versions of alpine
```

then add a following

```
RUN acbtest "$V1" != "$V2"
```

alexcb: the reasoning is to prevent cases where it's somehow auto-updated by an automated process, or user running a `sed` against all Earthfiles, which would miss such a comment.

mikejholly: @alexcb Updated.

util/deltautil/Earthfile
mikejholly: We decided to use relative paths as the full canonical target names did not seem necessary.

alexcb (approved):

PR #3936, opened 2024-03-25
Logstream: add command ID to formatted logs

mikejholly:
Used to filter log entries by command ID in the UI.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,7 +20,7 @@ require (
 	github.com/docker/go-connections v0.4.0
 	github.com/docker/go-units v0.5.0
 	github.com/dustin/go-humanize v1.0.1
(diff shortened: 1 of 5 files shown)


vladaionescu (approved):

vladaionescu:
Side note - if we don't need the raw logs anymore, the architecture of logbus could be simplified a little. In any case - that can be something that we do in the future.

mikejholly:
> Side note - if we don't need the raw logs anymore, the architecture of logbus could be simplified a little. In any case - that can be something that we do in the future.

Good point. It may be worth keeping the current structure until we're sure.

alexcb (approved):

PR #3950, opened 2024-03-27
exec-stats-summary: output summary of stats at end

alexcb:
Outputs a list of targets which include the max memory and cpu time that was consumed from a single RUN command.

### DIFF
--- a/util/execstatssummary/tracker.go
+++ b/util/execstatssummary/tracker.go
@@ -1,0 +1,91 @@
+package execstatssummary
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"os"
+	"sort"
+	"sync"
+	"text/tabwriter"
+	"time"
+
+	"github.com/dustin/go-humanize"
+)
+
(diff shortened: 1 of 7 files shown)


alexcb:
example output:
```
target                           memory     cpu
+earthly-integration-test-base   172417024  1.386318s
./tests+test-platform-expansion  99618816   4.629893s
+debugger                        193654784  1.007853s
+earthly                         924483584  7.801467s
+earthly-docker                  58712064   381.747ms
```

util/execstatssummary/tracker.go
```diff
@@ -0,0 +1,86 @@
...
+	"os"
+	"sort"
+	"sync"
+	"text/tabwriter"
+	"time"
+
+	"github.com/dustin/go-humanize"
+)
+
+type Tracker struct {
```
mikejholly, line 16: I think we should be consistent about adding Godoc comments to exported values.

mikejholly (approved):
LGTM. Nice addition!

PR #3951, opened 2024-03-27
Logstream: ensure all failed targets are marked as such

mikejholly:
A missing func call was leading to some target statuses not being updated.

### DIFF
--- a/earthfile2llb/converter.go
+++ b/earthfile2llb/converter.go
@@ -1862,6 +1862,7 @@ func (c *Converter) FinalizeStates(ctx context.Context) (*states.MultiTarget, er
 		if c.ftrs.ExecAfterParallel {
 			err = c.forceExecution(ctx, c.mts.Final.MainState, c.mts.Final.PlatformResolver)
 			if err != nil {
+				c.RecordTargetFailure(ctx, err)
 				return errors.Wrapf(err, "async force execution for %s", c.mts.FinalTarget().String())
 			}
 		}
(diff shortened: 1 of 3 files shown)


earthfile2llb/converter.go
```diff
@@ -1862,6 +1862,7 @@ func (c *Converter) FinalizeStates(ctx context.Context) (*states.MultiTarget, er
 		if c.ftrs.ExecAfterParallel {
 			err = c.forceExecution(ctx, c.mts.Final.MainState, c.mts.Final.PlatformResolver)
 			if err != nil {
+				c.RecordTargetFailure(ctx, err)
```
vladaionescu, line 1865: Will this still work even if the ctx is canceled?

vladaionescu (approved):

earthfile2llb/converter.go
mikejholly: AFAICT, most cancellations are being caught [here](https://github.com/earthly/earthly/blob/main/earthfile2llb/interpreter.go#L74). But some weren't as there is a change the cancellation may trigger a failure in [`FinalizeStates`](https://github.com/earthly/earthly/blob/main/earthfile2llb/earthfile2llb.go#L351-L359) as well. Maybe there's a better spot to handle all cases.

vladaionescu: Perhaps logstream should generally work despite any regular cancelation, with the exception being if the cancelation was user-issued.

mikejholly: That's the idea. With this change and the corresponding backend change, all messages should make their way to the server except after the 2nd control-c. This particular change is to fix incorrect "in progress" statuses in the database &, subsequently, the UI.

vladaionescu: SGTM!

mikejholly: I played around with a few different locations for `RecordTargetFailure` but they didn't seem to catch all cancelations. It seems like the code in this PR is working well.

PR #3952, opened 2024-03-28
revert some changes from 9822b7321

alexcb:
9822b7321dc1202209b503f3e06a51c4d9ac6214 accidentally included some changes to ast which should not have been merged (possibly a bad search-and-replace?)

This additionally un-exports mustParseVersion, which is only used within the features package.


features/features_integration_test.go
alexcb, line 1: nothing in this file changed, except for moving `TestMustParseVersion` to a seperate file that uses `package features` instead, to allow testing unexported funcs.

mikejholly (approved):

idodod (approved):

PR #3956, opened 2024-03-29
Handle source location containing SSH URL

mikejholly:
SSH URLs were breaking links in the new UI.

### DIFF
--- a/logbus/run.go
+++ b/logbus/run.go
@@ -2,12 +2,16 @@ package logbus
 
 import (
 	"errors"
+	"fmt"
+	"regexp"
+	"strings"
 	"sync"
 	"time"
 
 	"github.com/earthly/cloud-api/logstream"
 	"github.com/earthly/earthly/ast/spec"
 	"github.com/earthly/earthly/domain"
+	"github.com/moby/buildkit/util/sshutil"
 )
 
 // GenericDefault is the internal name used to identify messages unrelated to a specific target or command.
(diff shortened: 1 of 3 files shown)


logbus/run.go
```diff
@@ -200,6 +202,14 @@ func sourceLocationToProto(repoURL, repoHash, fileRelToRepo string, sl *spec.Sou
 	if fileRelToRepo == "" && repoURL == "" {
 		file = sl.File
 	}
+	// Some repository URLs are being provided in Git SSH form. Let's convert
+	// these to HTTP. Example: git@github.com:earthly/earthly.git
+	if strings.HasPrefix(repoURL, "git@") {
```
alexcb, line 207: Some ssh-based git repos use other usernames than `git`; people who have esoteric self-hosted git repos won't trigger this codepath (e.g. using a plain old vanilla remote server and hosting the git repo under their home directory).

alexcb: maybe https://github.com/earthly/buildkit/blob/50fffdb32b13748ba42e45c5097a10dade2e5ff7/source/git/identifier.go#L87C164-L87C186

or

https://github.com/earthly/buildkit/blob/earthly-main/util/sshutil/scpurl.go#L12

could be useful?

mikejholly: Yep, I'm familiar with the fact that the usernames can differ. My thinking was to target `github.com` URLs initially, as not all SSH URLs will have HTTP equivalents. Or perhaps it's safe to assume they do. What do you think?

alexcb: I think it's worthwhile targeting all of them since we can use code (from buildkit) that has already been written. Not all of our users use github.

mikejholly: Sounds good. It should work in most cases.

alexcb: i searched for sshutil and discovered we already use it under https://github.com/earthly/earthly/blob/e1761bdefb23fa33b33c12af886a2dc39370bf89/buildcontext/gitlookup.go#L580

seems like this function could be moved to an earthly util and we could call that?

mikejholly: `parseGitProtocol` doesn't really do what I'm looking for. `IsImplicitSSHTransport` should be fine.

alexcb: what about the case of explicit git repos? e.g. `ssh://...`

alexcb: thinking a bit more about this, some of these self-hosted repos won't even have a http equivalent, so it's probably fine.

mikejholly: > what about the case of explicit git repos? e.g. ssh://...

I haven't yet seen these in the source location data. I'll revisit this code if we encounter them. AFAIK, these are extracted from 

> some of these self-hosted repos won't even have a http equivalent

That's true.

Issue #1743, opened 2022-03-11
`HOST` bug or expected behavior?


wataruian:
I've been following the experimental feature to add host in /etc/hosts for the duration of the build (`HOST`) and found something (bug or expected behavior?):

1. I can't seem to pass dynamic value to `HOST`, if I tried passing the ARG it really tries to add `${ARG_NAME}` in /etc/hosts, I tried even setting as ENV, still the same. But hardcoding the value is working
2. `SAVE IMAGE` does not seem to respect what is in /etc/hosts. So I am unable to push to the local registry, which when you try to do `curl` in `+success-access-host` is successful.

```
VERSION --use-host-command 0.6

FROM alpine:3.14

ARG IP="192.168.1.9"
ARG DOMAIN="mylocalhost.com"

success-set-host:
  HOST mylocalhost.com 192.168.1.9

failed-set-host:
  RUN echo ${DOMAIN}
  RUN echo ${IP}
  HOST ${DOMAIN} ${IP}

success-access-host:
  HOST mylocalhost.com 192.168.1.9
  RUN apk update && apk add curl && curl -GET mylocalhost.com:5000/v2/_catalog

failed-push:
  HOST mylocalhost.com 192.168.1.9
  SAVE IMAGE --insecure --push mylocalhost.com:5000/some-image:1.0
```

Any , idea how to achieve this?

Thanks!

dchw:
Right now, the `HOST` keyword only accepts static values. It seems reasonable to incorporate this.

wataruian:
Looks like this is the one I am needing for the long time (did not even know this existed): 
`earthly global.buildkit_additional_args ["--add-host", "yourdomain.com:127.0.0.1"]`

I am not sure if `HOST` interfering with buildkit hosts is a wanted/good thing though, if needed it can be easily done with (`--add-host` - Thanks @dchw). 

But `HOST` being able to accept variables for DOMAIN and IP makes sense.

PR #3957, opened 2024-03-30
Expand args on the HOST command so you can have dynamic hostnames

pbecotte:
This should fix [#1743](https://github.com/earthly/earthly/issues/1743)

### DIFF
--- a/tests/host.earth
+++ b/tests/host.earth
@@ -20,7 +20,7 @@ expand-args:
     ARG fake_host=example.com
     ARG target_ip=1.2
     HOST $fake_host $target_ip.3.4
-    RUN cat /etc/hosts | grep "1.2.3.4\W*example.com"
+    RUN cat /etc/hosts | acbgrep "1.2.3.4\W*example.com"
 
 invalid-ip:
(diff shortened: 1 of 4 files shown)


vladaionescu (approved):

tests/host.earth
```diff
@@ -16,6 +16,12 @@ add-ipv6:
     HOST cool.website dead:beef::cafe
     RUN cat /etc/hosts | grep "dead:beef::cafe\W*cool.website"
 
+expand-args:
+    ARG fake_host=example.com
+    ARG target_ip=1.2
+    HOST $fake_host $target_ip.3.4
+    RUN cat /etc/hosts | grep "1.2.3.4\W*example.com"
```
alexcb, line 23: we have a grep wrapper called [`acbgrep`](https://github.com/alexcb/acbtest?tab=readme-ov-file#acbgrep) which prints out failures messages.
```suggestion
    RUN cat /etc/hosts | acbgrep "1.2.3.4\W*example.com"
```

idodod:
I think you may also need to run `earthly ./ast/tests+update-all` to update some of the expected ast.json files

PR #3960, opened 2024-04-01
Print HTTP retries at debug level instead of warn

brandonSc:
Several users (myself included) have noticed HTTP retries printing in red text. These errors are often due to a network blip that is resolved naturally by the retry mechanism. The red text, however, makes it look like a concerning error.

This PR removes the red text and only logs those errors on retry when debug mode is used. 

In any case, the underlying error is always printed if the retry limit is reached. 

Before:
Screenshot 2024-04-01 at 08 14 20

After:
Screenshot 2024-04-01 at 08 14 05


idodod (approved):

alexcb (approved):

idodod:
@brandonSc Thanks for the PR, mind adding a changelog entry?

PR #3963, opened 2024-04-01
update go to 1.22

alexcb:


Earthfile
```diff
@@ -305,7 +305,7 @@ lint-changelog:
 # debugger builds the earthly debugger and saves the artifact in build/earth_debugger
 debugger:
     FROM +code
-    ENV CGO_ENABLED=0
+    ENV CGO_ENABLED=1
```
alexcb, line 308: from the 1.22 Release notes:

> go build commands that invoke the linker now error out if an external (C) linker will be used but cgo is not enabled. (The Go runtime requires cgo support to ensure that it is compatible with any additional libraries added by the C linker.)

alexcb: because we make use of `-linkmode external`

PR #3985, opened 2024-04-03
Earthly can connect to satellites in BYOC plan

brandonSc:


brandonSc:
Merging in with https://github.com/earthly/earthly/pull/4024

PR #3986, opened 2024-04-03
Fix host test use of grep

mikejholly:
`acbgrep` isn't available here. Plus, the other tests in this file just use `grep`.

### DIFF
--- a/ast/tests/host.ast.json
+++ b/ast/tests/host.ast.json
@@ -147,7 +147,7 @@
               "cat",
               "/etc/hosts",
               "|",
-              "acbgrep",
+              "grep",
               "\"1.2.3.4\\W*example.com\""
             ],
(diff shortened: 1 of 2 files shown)


alexcb (approved):

alexcb:
sorry about that :(

alexcb:
looks like you need a `earthly ./ast/tests+update-all` too

alexcb:
another (longer term) idea: should we stop using the default `alpine` image and instead have an `earthly/alpine-base` image that we include other things we sometimes need in our tests?

mikejholly:
> another (longer term) idea: should we stop using the default `alpine` image and instead have an `earthly/alpine-base` image that we include other things we sometimes need in our tests?

Sounds reasonable.

PR #3989, opened 2024-04-04
Allow for local --BUILDKIT_PROJECT when running RUN_EARTHLY tests

mikejholly:
The BuildKit specified when running `+for-linux` and friends is not also being used by `RUN_EARTHLY` tests which use the `earthly-integration-test-base` image. This PR adds a similar argument for tests based on the latter image. 

```
earthly -P +test-no-qemu-group8 --BUILDKIT_PROJECT="../buildkit"
```

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -475,7 +475,8 @@ earthly-docker:
 # if no dockerhub mirror is not set it will attempt to login to dockerhub using the provided docker hub username and token.
 # Otherwise, it will attempt to login to the docker hub mirror using the provided username and password
 earthly-integration-test-base:
-    FROM +earthly-docker
+    ARG BUILDKIT_PROJECT
+    FROM --pass-args +earthly-docker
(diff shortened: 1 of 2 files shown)


Earthfile
```diff
@@ -475,7 +475,8 @@ earthly-docker:
 # if no dockerhub mirror is not set it will attempt to login to dockerhub using the provided docker hub username and token.
 # Otherwise, it will attempt to login to the docker hub mirror using the provided username and password
 earthly-integration-test-base:
-    FROM +earthly-docker
+    ARG BUILDKIT_PROJECT
```
idodod, line 478: nit: I'm not even sure the arg definition is necessary because I think --pass-args would just pass the value from the caller but maybe it's good to be explicit

idodod (approved):

Earthfile
mikejholly: That's true. I removed it.

Issue #3966, opened 2024-04-02
Support Wildcard targets for COPY


idodod:
**Use case**

We recently introduced support of wildcards in a target path when using `BUILD` on a target to ease execution of targets in a monorepo (See https://github.com/earthly/earthly/pull/3661).
It can be useful to do the same for `COPY`, e.g. :
```Earthfile
VERSION 0.8

test:
    FROM alpine
    COPY --dir ./wildcard-build/*+get-files/* .
```

**Expected Behavior**

The behavior should be the same as it was implemented for `BUILD`

Note that if the different targets save an artifact to the same file/directory, the artifact may overwrite each other.
For this command to. be effective, the artifact needs to be saved in a different path.

idodod:
will be included in next release (v0.8.8)

PR #3995, opened 2024-04-08
Support wildcard/glob in COPY

idodod:
Address #3966


domain/artifact.go
```diff
@@ -14,6 +14,12 @@ type Artifact struct {
 	Artifact string
 }
 
+// Clone returns a copy of the Artifact
+func (a Artifact) Clone() Artifact {
+	newArtifact := a
```
mikejholly, line 19: Nit: I think `return a` here is equivalent as no pointers involved.

earthfile2llb/converter.go
```diff
@@ -2906,6 +2876,74 @@ func (c *Converter) newCmdID() int {
...
+	}
+
+	if len(targets) == 0 {
+		return nil, errors.Errorf("no matching targets found for pattern %q", parsedTarget.GetLocalPath())
+	}
+
+	return targets, nil
+}
+
+func clonesWithExpandedTargets[T cloneable[T]](expandedTargets []string, c T, setTarget func(t *T, expandedTarget string) error) ([]T, error) {
```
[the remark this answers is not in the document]
mikejholly: Up to you, but I wonder if the use of generics here is more complicated than the simple loop plus `Clone` calls they are meant to abstract.

mikejholly (approved):
Nice job! Cool new addition. Thanks for adding the docs as well.

idodod:
> Nice job! Cool new addition. Thanks for adding the docs as well.

Thanks, I actually meant and forgot to thank you for paving the way. After your implementation this feature was easy to implement. So thank you for making it easy :)

domain/artifact.go
idodod: good point, it's a leftover from previous commit where I thought I might need to do more to fully clone.
I'll update.

earthfile2llb/converter.go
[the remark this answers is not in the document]
idodod: yeah, I was wondering if the indirection is worth it and I'm open to removing it.
I tended to go with generics after all because I liked how it forces the user to clone the object, which I think it might be something that can be easily forgotten. 
I also considered whether this function can be used again, and figured it _might_ in something like `WITH DOCKER --load ./services/*+image`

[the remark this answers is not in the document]
mikejholly: That sounds fair.

mikejholly:
> > Nice job! Cool new addition. Thanks for adding the docs as well.
> 
> Thanks, I actually meant and forgot to thank you for paving the way. After your implementation this feature was easy to implement. So thank you for making it easy :)

Thanks for saying so :smiling_face_with_tear:

docs/earthfile/earthfile.md
```diff
@@ -1100,7 +1163,7 @@ The `WITH DOCKER` clause only supports the command [`RUN`](#run). Other commands
 A typical example of a `WITH DOCKER` clause might be:
 
 ```Dockerfile
-FROM earthly/dind:alpine-3.19-docker-25.0.2-r0
+FROM earthly/dind:alpine-3.19-docker-25.0.3-r2
```
alexcb, line 1166: nit: this seems irrelevant to this PR.

alexcb (approved):
nice tests too!

docs/earthfile/earthfile.md
alexcb: ah ha... I see the reason now, nevermind me here :)

Issue #3997, opened 2024-04-09
`LET`/`SET` values cannot be used in parallel commands such as `BUILD`


idodod:
**What went wrong?**

Trying to defined the name of the target in a `LET` statement, the name of the target is not properly expanded and the target cannot be built.

```Earthfile
VERSION 0.8

FROM alpine

foo:
    LET target=+bar
    BUILD $target

bar:
    RUN echo hello from bar
```

`earthly +foo`

**What should have happened?**

The command should succeed in eventually building `+bar`.
chan

**Other Helpful Information**

Changing the `LET` command to an `ARG` allows the target to be built.

It appears LET/SET do not block parallel commands, thus BUILD is executed before LET is evaluated.

idodod:
This also fixes a bug where a BUILD might get executed twice with different values of an arg.
Consider the folliwing:
```Earthfile

FROM alpine

foo:
   LET my_val=abc
   BUILD +bar --my_arg=$my_val
   
bar:
  ARG my_arg="123"
  RUN echo $my_arg   
```

`earthly +foo` would result in +bar getting executed twice - once with my_arg="" and once with my_arg="abc", where the latter is the intended behavior.

idodod:
will be included in next releases (v0.8.8)

PR #3998, opened 2024-04-09
LET/SET should block parallel commands that come after it

idodod:
Address #3997


earthfile2llb/interpreter.go
```diff
@@ -125,7 +125,8 @@ func (i *Interpreter) handleBlock(ctx context.Context, b spec.Block) error {
 		if err != nil {
 			return err
 		}
-		prevWasArg = (stmt.Command != nil && stmt.Command.Name == "ARG")
+		prevWasArg = stmt.Command != nil && (stmt.Command.Name == "ARG" || (i.converter.ftrs.LetSetBlockParallel && (stmt.Command.Name == "LET" || stmt.Command.Name == "SET")))
```
mikejholly, line 128: I find these long conditional expressions difficult to mentally parse. How about a function with guards instead?

```golang
func (i *interpreter) prevWasArg(stmt *Statement) bool {
	if (stmt == nil || stmt.Command == nil) {
		return false
	}

	if i.converter.ftrs.LetSetBlockParallel && (stmt.Command.Name == "LET" || stmt.Command.Name == "SET") {
		return true
	}

	return stmt.Command.Name == "ARG"
}
```

mikejholly (approved):
LGTM with one suggestion.

earthfile2llb/converter.go
```diff
@@ -1467,7 +1467,7 @@ func (c *Converter) UpdateArg(ctx context.Context, argKey string, argValue strin
 		pncvf = c.processNonConstantBuildArgFunc(ctx)
 	}
 
-	err := c.varCollection.UpdateVar(argKey, argValue, pncvf, isBase)
+	err := c.varCollection.UpdateVar(argKey, argValue, pncvf)
```
idodod, line 1470: unrelated, but noticed this var is not used

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to [Earthly](https://github.com/earthly/earthly) will be doc
 
 ## Unreleased
 
+### Changed
```
alexcb, line 7: should this be fixed instead? the old behaviour seems rather buggy.
```suggestion
### Fixed
```

I wonder if there's even a need for the feature flagging? I can't think up a case where people wouldn't want the correct value passed through.

idodod: > should this be fixed instead? the old behaviour seems rather buggy.
I agree, it seems like more a bug fix than a behavior change, I'm good with changing it.
I guess I decided on calling it a change because I wasn't sure if for whatever reason this fix would somehow change expectations for our users.

> I wonder if there's even a need for the feature flagging? I can't think up a case where people wouldn't want the correct value passed through.

Related to the previous point, I just didn't know if there could be some builds that "work" despite this bug.
For example:

```Earthfile

FROM alpine

foo:
   LET my_val=abc
   BUILD +bar --my_arg=$my_val
   
bar:
  ARG my_arg="123"
  RUN echo $my_arg   
```

https://cloud.earthly.dev/builds/76aeae46-969d-40eb-a8f6-b58e9b2ccbad
It results in +foo getting executed twice (my_arg="", my_arg="abc")

alexcb: twice? wow that seems incredibly buggy.

I think we should just fix it (without a feature flag), I can't imagine people needing the old buggy behaviour.

earthfile2llb/interpreter.go
```diff
@@ -141,7 +142,7 @@ func (i *Interpreter) handleBlockParallel(ctx context.Context, b spec.Block, sta
 		stmt := b[index]
 		if stmt.Command != nil {
 			switch stmt.Command.Name {
-			case command.Arg, command.Locally, command.From, command.FromDockerfile:
+			case command.Arg, command.Locally, command.From, command.FromDockerfile, command.Let, command.Set:
```
idodod, line 145: the actual fix

alexcb (approved):

PR #4002, opened 2024-04-09
doc: ENV doesn't support setting multiple key value pairs

Nasfame:


alexcb:
I don't understand the intention of this PR -- it seems to be introducing a test to validate that we don't have dockerfile parity?

I think it would be better to just work on the `handleEnv` function under https://github.com/earthly/earthly/blob/0531f282391292d477aa4eceaae1521848f56f3f/earthfile2llb/interpreter.go#L1430

to support multiple envs.

Nasfame:
Well @alexcb , i understand your concerns on parity. Parsing the multiple ENV with multiple ENV instruction simplifies our solution early own. I do have concerns on handling whitespaces later down the lane

Nasfame (commented):
Its ready now. Supports env like `export`

tests/env.earth
```diff
@@ -4,3 +4,14 @@ FROM alpine:3.18
...
+
+test-multiple-on-single-env:
+    ENV GOLANG=1.22.2 \
+        GO_VERSION=1.22.2 \
+    	GOOS=linux \
+    	GOARCH=amd64 \
+        GO_DOWNLOAD_SHA256=5901c52b7a78002aeff14a21f93e0f064f74ce1360fce51c6ee68cd471216a17
+
+    RUN env
+    RUN echo GOLANG=$GOLANG
```
alexcb, line 16: these tests are not adequate; instead you'll need to do something like
```suggestion
    RUN test "$GOLANG" = "1.22.2"
```

can you additionally rename these values to be something more generic -- using go-specific env names doesn't seem as readable, since at quick glance one might think this test is related to compiling go.

maybe

```
ENV A=alpha \
  B=bravo
```

alexcb (commented):
thanks for the submission; we'll need the following changes:
- make sure you can run the top-level `+test` target (which should include the `+lint` target, which I suspect will fail due to newline issues)
- errors must be returned rather than using logrus

tests/env.earth
Nasfame: upto u, i m open to it. Just felt it was more relatable to real life

Lemniscite:
- [  ] make sure you can run the top-level +test target (which should include the +lint target, which I suspect will fail due to newline issues)
@alexcb i m unable to run +lint, as i don't have access to `PROJECT earthly-technologies/core`
- [x] errors must be returned rather than using logrus

Nasfame:
@alexcb 

```
 earthly +go
            frontend | Unable to detect Docker or Podman. Use --verbose to see details (or errors)
 Init 🚀

           satellite | my-satellite is waking up. Please wait...
           satellite | ...modifying downlink frequency...
           satellite | ...calibrating transponders...
           satellite | ...reporting fuel levels...
           satellite | ...System online.
           satellite | Connecting to my-satellite...
           satellite | ...Done
           satellite | Version github.com/earthly/buildkit v0.8.7 422af4977207d5a65ea59c25526a7a7782655df4
           satellite | Platforms: linux/amd64 (native) linux/amd64/v2 linux/amd64/v3 linux/amd64/v4 linux/386 linux/arm64 linux/riscv64 linux/ppc64 linux/ppc64le linux/s390x linux/mips64le linux/mips64
           satellite | Utilization: 0 other builds, 0/12 op load
           satellite | GC stats: 17 GB cache, avg GC duration 0s, all-time GC duration 774ms, last GC duration 0s, last cleared 0 B

Streaming logs to https://cloud.earthly.dev/builds/e5b703da-73a7-4dc6-a121-9d9a4be44513

Build Minutes: 64 out of 6000 used

 Build 🔧

      registry-proxy | Failed to start registry proxy: failed to start Darwin support container: failed to start support container: frontend (e.g. docker/podman) not initialized
              logbus | Setting organization "Nasfame" and project "core"
Error: Earthfile line 0:0 target go not found
in              github.com/Nasfame/earthly/tests:bugfix/env-multi-line-eval+go
View logs at https://cloud.earthly.dev/builds/e5b703da-73a7-4dc6-a121-9d9a4be44513
        * log streamer: 1 error occurred:
        * failed waiting for log stream server: rpc error: code = NotFound desc = project "core" not found



```

tests/env.earth
Nasfame: `    RUN test "$GOLANG" = "1.22.2"` i didn't realize this is possible . I m curious why it wasn't done for L5

Nasfame: done, complete with generic vars

Issue #3996, opened 2024-04-08
Autoskip - Autoskip fails target depends on a LET/SET argument


idodod:
**What went wrong?**

When referencing a target with path that should be expanded from a LET statement, the path is not properly expanded.

```Earthfile
VERSION 0.8

test:
    FROM alpine
    LET path_to_target="subdir"
    BUILD ./$path_to_target+some-target
```

Run `earthly --auto-skip +test`

Get
```bash
Error: auto-skip is unable to calculate hash for +test: target "test" not found
```

**What should have happened?**

**Other Helpful Information**

The command should have succeeded.
Note, if the `LET` statement is changed to `ARG` , the example works.

Additionally, it also seems the error message should state `some-target` was not found instead of test

idodod:
will be included in next release (v0.8.8)

PR #4007, opened 2024-04-10
Support Let/Set in autoskip

idodod:
Address https://github.com/earthly/earthly/issues/3996

### DIFF
--- a/inputgraph/loader.go
+++ b/inputgraph/loader.go
@@ -463,10 +463,11 @@ func (l *loader) handleLet(ctx context.Context, cmd spec.Command) error {
 	argsCpy := flagutil.GetArgsCopy(cmd)
 	args, err := flagutil.ParseArgsCleaned("LET", &opts, argsCpy)
 	if err != nil {
-		return errors.Wrap(err, "failed to parse LET args")
+		return wrapError(err, cmd.SourceLocation, "failed to parse LET args")
 	}
+	// performing minimal validation to avoid index out of bound (full validation occurs in the interpreter)
 	if len(args) != 3 {
-		return errors.New("failed to parse LET args")
+		return newError(cmd.SourceLocation, "failed to parse LET args")
 	}
 	key := args[0]
 	baseVal := args[2]
@@ -479,10 +480,7 @@ func (l *loader) handleLet(ctx context.Context, cmd spec.Command) error {
 
 	_, _, err = l.varCollection.DeclareVar(key, variables.WithValue(val))
 	if err != nil {
-
-		if err != nil {
-			return wrapError(err, cmd.SourceLocation, "failed to declare variable")
-		}
+		return wrapError(err, cmd.SourceLocation, "failed to declare variable")
 	}
 	return nil
 }
@@ -490,19 +488,20 @@ func (l *loader) handleLet(ctx context.Context, cmd spec.Command) error {
 func (l *loader) handleSet(ctx context.Context, cmd spec.Command) error {
 	var opts commandflag.SetOpts
 	argsCpy := flagutil.GetArgsCopy(cmd)
-	args, err := flagutil.ParseArgsCleaned("LET", &opts, argsCpy)
+	args, err := flagutil.ParseArgsCleaned("SET", &opts, argsCpy)
 	if err != nil {
-		return errors.Wrap(err, "failed to parse LET args")
+		return wrapError(err, cmd.SourceLocation, "failed to parse SET args")
 	}
+	// performing minimal validation to avoid index out of bound (full validation occurs in the interpreter)
 	if len(args) != 3 {
-		return errors.New("failed to parse LET args")
+		return newError(cmd.SourceLocation, "failed to parse SET args")
 	}
 
 	key := args[0]
 	baseVal := args[2]
 	val, err := l.expandArgs(ctx, baseVal)
 	if err != nil {
-		return wrapError(err, cmd.SourceLocation, "failed to expand LET value %q", baseVal)
+		return wrapError(err, cmd.SourceLocation, "failed to expand SET value %q", baseVal)
 	}
 
 	l.hasher.HashString(fmt.Sprintf("SET %s=%s", key, val))
--- a/inputgraph/loader.go
+++ b/inputgraph/loader.go
@@ -374,6 +374,10 @@ func (l *loader) handleCommand(ctx context.Context, cmd spec.Command) error {
 		return l.handleCopy(ctx, cmd)
 	case command.Arg:
 		return l.handleArg(ctx, cmd, false)
+	case command.Let:
+		return l.handleLet(ctx, cmd)
+	case command.Set:
+		return l.handleSet(ctx, cmd)
 	case command.FromDockerfile:
 		return l.handleFromDockerfile(ctx, cmd)
 	case command.Import:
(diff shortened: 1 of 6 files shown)


tests/wildcard-copy.earth
```diff
@@ -45,9 +45,10 @@ wildcard-multi-with-flags:
     RUN test -f somedir1/somefile1
 
 wildcard-if-exists:
-    #COPY --if-exists ./wildcard/*+test/non-existing .
-    #DO +TEST --EXPECTED_COUNT=0
-    #RUN ! test -e non-existing
+    COPY --if-exists ./wildcard/*+test/non-existing .
```
idodod, line 48: forgot to uncomment this test in previous PR

mikejholly (approved):
Nice work!

inputgraph/loader.go
```diff
@@ -454,6 +458,64 @@ func (l *loader) handleArg(ctx context.Context, cmd spec.Command, isBase bool) e
...
 }
 
+func (l *loader) handleLet(ctx context.Context, cmd spec.Command) error {
+	var opts commandflag.LetOpts
+	argsCpy := flagutil.GetArgsCopy(cmd)
+	args, err := flagutil.ParseArgsCleaned("LET", &opts, argsCpy)
+	if err != nil {
+		return errors.Wrap(err, "failed to parse LET args")
+	}
+	if len(args) != 3 {
```
alexcb, line 468: should we also check for a `=` somewhere? currently we'll accept any of `LET foo <- bar`, `LET foo gets bar`, or `LET foo = bar`
```suggestion
	if len(args) != 3 || args[1] != '=' {
```

inputgraph/loader.go
```diff
@@ -454,6 +458,64 @@ func (l *loader) handleArg(ctx context.Context, cmd spec.Command, isBase bool) e
...
+	if err != nil {
+		return wrapError(err, cmd.SourceLocation, "failed to expand LET value %q", baseVal)
+	}
+
+	l.hasher.HashString(fmt.Sprintf("LET %s=%s", key, val))
+
+	_, _, err = l.varCollection.DeclareVar(key, variables.WithValue(val))
+	if err != nil {
+
+		if err != nil {
```
alexcb, line 483: you must really want to check err isn't null here :)

(there's a duplicate if above)

inputgraph/loader.go
```diff
@@ -454,6 +458,64 @@ func (l *loader) handleArg(ctx context.Context, cmd spec.Command, isBase bool) e
...
+			return wrapError(err, cmd.SourceLocation, "failed to declare variable")
+		}
+	}
+	return nil
+}
+
+func (l *loader) handleSet(ctx context.Context, cmd spec.Command) error {
+	var opts commandflag.SetOpts
+	argsCpy := flagutil.GetArgsCopy(cmd)
+	args, err := flagutil.ParseArgsCleaned("LET", &opts, argsCpy)
```
alexcb, line 493:
```suggestion
	args, err := flagutil.ParseArgsCleaned("SET", &opts, argsCpy)
```

inputgraph/loader.go
```diff
@@ -454,6 +458,64 @@ func (l *loader) handleArg(ctx context.Context, cmd spec.Command, isBase bool) e
...
 
+func (l *loader) handleLet(ctx context.Context, cmd spec.Command) error {
+	var opts commandflag.LetOpts
+	argsCpy := flagutil.GetArgsCopy(cmd)
+	args, err := flagutil.ParseArgsCleaned("LET", &opts, argsCpy)
+	if err != nil {
+		return errors.Wrap(err, "failed to parse LET args")
+	}
+	if len(args) != 3 {
+		return errors.New("failed to parse LET args")
```
alexcb, line 469: not sure if this is actually the syntax that we need here, but maybe there's a way to include the source location for this error too?
```suggestion
		return wrapError(errors.New("failed to parse LET args"), cmd.SourceLocation)
```

inputgraph/loader.go
idodod: The reason for only checking length is that I didn't want to potentially get an index out of bound error. 
I think the rest of the validation should still occur later on when LET/SET are handled by the interperter.
Thoughts?

inputgraph/loader.go
alexcb: actually.....

```
return newError(with.Command.SourceLocation,"failed to parse LET args")
```

alexcb: there's a number of places in this PR where we could do this.

inputgraph/loader.go
```diff
@@ -454,6 +458,64 @@ func (l *loader) handleArg(ctx context.Context, cmd spec.Command, isBase bool) e
...
+	}
+	return nil
+}
+
+func (l *loader) handleSet(ctx context.Context, cmd spec.Command) error {
+	var opts commandflag.SetOpts
+	argsCpy := flagutil.GetArgsCopy(cmd)
+	args, err := flagutil.ParseArgsCleaned("LET", &opts, argsCpy)
+	if err != nil {
+		return errors.Wrap(err, "failed to parse LET args")
```
alexcb, line 495:
```suggestion
		return newError(with.Command.SourceLocation,"failed to parse SET args")
```

inputgraph/loader.go
```diff
@@ -454,6 +458,64 @@ func (l *loader) handleArg(ctx context.Context, cmd spec.Command, isBase bool) e
...
+	}
+	if len(args) != 3 {
+		return errors.New("failed to parse LET args")
```
alexcb, line 498: let -> set

inputgraph/loader.go
```diff
@@ -454,6 +458,64 @@ func (l *loader) handleArg(ctx context.Context, cmd spec.Command, isBase bool) e
...
+	}
+
+	key := args[0]
+	baseVal := args[2]
+	val, err := l.expandArgs(ctx, baseVal)
+	if err != nil {
+		return wrapError(err, cmd.SourceLocation, "failed to expand LET value %q", baseVal)
```
alexcb, line 505:
```suggestion
		return wrapError(err, cmd.SourceLocation, "failed to expand SET value %q", baseVal)
```

alexcb (commented):
looking good; just a number of small suggestions.

inputgraph/loader.go
idodod: added a comment instead

inputgraph/loader.go
```diff
@@ -454,6 +458,64 @@ func (l *loader) handleArg(ctx context.Context, cmd spec.Command, isBase bool) e
...
+}
+
+func (l *loader) handleSet(ctx context.Context, cmd spec.Command) error {
+	var opts commandflag.SetOpts
+	argsCpy := flagutil.GetArgsCopy(cmd)
+	args, err := flagutil.ParseArgsCleaned("LET", &opts, argsCpy)
+	if err != nil {
+		return errors.Wrap(err, "failed to parse LET args")
+	}
+	if len(args) != 3 {
```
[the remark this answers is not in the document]
idodod: see discussion above

idodod:
> looking good; just a number of small suggestions.

Thanks for the review, I think I addressed all comments.

alexcb (approved):

PR #4013, opened 2024-04-11
init: go.work

Nasfame:
- Recommended over replace directives pointing to local modules
- Will allow earthly to be installed via go toolchain with `go install github.com/earthly/earthly@latest` 
- renamed incorrect go modules in examples with proper path
- upgraded direct deps of go modules


Nasfame:
Remark: go builds of all go modules successful so far. Compiled binary functioning as expected with args `--no-cache -i github.com/earthly/hello-world:+hello`

alexcb:
We do not want to save the `go.work` file in git; it's meant as a way for developers to override packages on their own computer that's specific to their own workspace setup.

We even have a blog post: https://earthly.dev/blog/go-workspaces/ that includes this as a warning at the bottom of the post.

Nasfame:
Oh I didn't realize that.

Nasfame:
FWIW: If earthly were a go library, then committing it might be a bad idea. Although there is hack to make it work even then.

But in your case its an application, its highly beneficial for you to commit `go.work` . Given its benefits overall. Citing my favorite example : https://github.com/kubernetes/kubernetes/blob/master/go.work  and rel issue: https://github.com/golang/go/issues/53502

I hope you realize the benefits you get with `go.work` committed versus the so many `replace` directives all over the repo.

Nasfame:
Hey @alexcb,

Good point about go.work. The concern about not including it in Git mainly applies when referenced modules aren't committed. Since we have a monorepo with all modules and likely won't separate them, using go workspaces like Kubernetes (k8s) seems perfectly reasonable for Earthly.

PR #4024, opened 2024-04-15
Cloud Installation Commands

brandonSc:
Added `earthly cloud` commands for managing BYOC configurations. Also tweaked some connection logic so that custom TLS certs can be used for any satellites instead of just Self-Hosted.


cmd/earthly/subcmd/cloud_installation_cmds.go
```diff
@@ -0,0 +1,251 @@
...
+	}
+
+	c.cli.Console().Printf("Configuring new Cloud Installation: %s. Please wait...", cloudName)
+
+	install, err := cloudClient.ConfigureCloud(ctx, orgID, cloudName, false)
+	if err != nil {
+		return errors.Wrap(err, "could not install cloud")
+	}
+
+	if err != nil {
```
alexcb, line 107: duplicate err check

brandonSc: Oops. Thanks for catching that!

PR #4042, opened 2024-04-19
Warn if anonymous token is fetched

### LINKED ISSUE #3009: disallow anonymous registry access when using a satellite
Since satellite users share the same IP, that IP will be limited very quickly when using anonymous access -- we should return an error rather than allow anonymous access.

The current authprovider code needs to be refactored to support this -- it's more complicated than I initially thought.

alexcb:
Using a registry without logging in may lead to rate limiting; display a warning if that's the case.

Note that this will be done in all instances even when using a local buildkit instance.

fixes #3009


alexcb:
depends on https://github.com/earthly/buildkit/pull/57

alexcb:
example output:

```
$ ./build/linux/amd64/earthly --no-cache ~/test/simple+test 
 Init 🚀

           buildkitd | Starting buildkit daemon as a docker container (earthly-dev-buildkitd)...
           buildkitd | ...Done

 Build 🔧

              logbus | Setting organization "" and project ""
/h/a/test/simple+test | DOCKERHUB_MIRROR_AUTH_FROM_CLOUD_SECRETS=true
/h/a/test/simple+test | --> FROM +base
Warning: you are not logged into registry-1.docker.io, you may experience rate-limitting when pulling images
/h/a/test/simple+test | DOCKERHUB_MIRROR_AUTH_FROM_CLOUD_SECRETS=true
/h/a/test/simple+test | --> BUILD +echo
/h/a/test/simple+echo | DOCKERHUB_MIRROR_AUTH_FROM_CLOUD_SECRETS=true name=ab
/h/a/test/simple+echo | --> FROM +base
/h/a/test/simple+base | DOCKERHUB_MIRROR_AUTH_FROM_CLOUD_SECRETS=true
/h/a/test/simple+base | --> FROM alpine
/h/a/test/simple+base | [----------] 100% FROM alpine
/h/a/test/simple+echo | DOCKERHUB_MIRROR_AUTH_FROM_CLOUD_SECRETS=true name=ab
/h/a/test/simple+echo | --> RUN echo hello $name
/h/a/test/simple+echo | hello ab

 Push Summary ⏫ (disabled)

 Local Output Summary 🎁



```

image

tests/Earthfile
```diff
@@ -212,6 +212,9 @@ ga-no-qemu-group10:
 ga-no-qemu-group11:
     BUILD --pass-args ./autoskip+test-group3
 
+ga-no-qemu-group12:
```
alexcb, line 215: this was created because the autoskip groups needed to be split up to avoid OOM issues on the m1 buildkite instance (and `ga-no-qemu-group8` was also getting a bit big)

idodod (approved):

PR #4047, opened 2024-04-22
GitHub Actions Integrations - Groups, Errors, Build Summary

adamgordonbell:
Add group headings for earthly phases when in github actions.
```
::group X
::endgroup
```
Add GHA error message on build failure. 
```
::error file=Earthfile,line=15,col=4,title=Error::The command RUN "something" did not complete successfully. Exit code 127
```
Which shows up like this:
https://github.com/adamgordonbell/gha_earthly_test/commit/429a135cfbc59368ad307b4515710f1cb0610f35
Screenshot 2024-04-24 at 8 54 24 AM

Adding `build summary` markdown message with build failure and run output:
https://github.com/adamgordonbell/gha_earthly_test/actions/runs/8816835334
Screenshot 2024-04-24 at 8 55 38 AM

It feels like there is lots that can be done to enrich the build summary message. We could in the future summarize build timing per target in a table there and link to our logs on a successful run. 

See https://github.com/earthly/earthly/issues/2189


conslogging/conslogging.go
```diff
@@ -91,35 +93,37 @@ func New(w io.Writer, mu *sync.Mutex, colorMode ColorMode, prefixPadding int, lo
...
-		logLevel:       logLevel,
+		consoleErrW:     w,
+		errW:            w,
+		colorMode:       colorMode,
+		saltColors:      make(map[string]*color.Color),
+		nextColorIndex:  new(int),
+		prefixPadding:   prefixPadding,
+		mu:              mu,
+		logLevel:        logLevel,
+		isGitHubActions: os.Getenv("GITHUB_ACTIONS") == "true",
```
vladaionescu, line 104: Wondering if we should pull this detection in the main package, as a flag that the user could override.

vladaionescu (commented):
Looks great!

conslogging/conslogging.go
alexcb: I think that would be nice -- maybe a user wants to see the output occur when they are running it locally?

vladaionescu:
Oh don't forget to add a changelog entry

conslogging/conslogging.go
```diff
@@ -266,6 +272,59 @@ func (cl ConsoleLogger) PrefixColor() *color.Color {
...
+	}
+
+	path := os.Getenv("GITHUB_STEP_SUMMARY")
+	if path == ""{
+		return
+	}
+	file, err := os.OpenFile(path, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
+	if err != nil {
+		return
+	}
```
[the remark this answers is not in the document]
adamgordonbell: Oh shoot 👍

conslogging/conslogging.go
```diff
@@ -266,6 +272,59 @@ func (cl ConsoleLogger) PrefixColor() *color.Color {
...
+	if file != "" && line != "" && col != "" {
+		formattedMsg := fmt.Sprintf("::error file=%s,line=%s,col=%s,title=Error::%s", file, line, col, message)
+		cl.printGithubActionsControl(formattedMsg)
+	} else {
+		cl.printGithubActionsControl(fmt.Sprintf("::error title=Error::%s", message))
+	}
+}
+
+// Print GHA control messages like ::group and ::error
+func (cl ConsoleLogger) printGithubActionsControl(msg string) {
```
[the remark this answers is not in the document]
adamgordonbell: I like it. Done

conslogging/conslogging.go
adamgordonbell: Added as `--github-actions`. Not sure if there's a better name.

alexcb: maybe `--github-actions-annotations` or simply `--github-annotations`? I think it gives a better hint that this is specifically for producing annotations.

conslogging/conslogging.go
```diff
@@ -266,6 +272,67 @@ func (cl ConsoleLogger) PrefixColor() *color.Color {
...
+	}
+}
+
+type GHHeader string
+
+const (
+	Error    GHHeader = "::error"
+	Group    GHHeader = "::group::"
+	EndGroup GHHeader = "::endgroup::"
+)
```
idodod, line 320: nit (I didn't notice before it actually doesn't have to be public):
```suggestion
type ghHeader string

const (
	error    ghHeader = "::error"
	group    ghHeader = "::group::"
	endGroup ghHeader = "::endgroup::"
)
```

cmd/earthly/app/before.go
```diff
@@ -58,10 +58,11 @@ func (app *EarthlyApp) before(cliCtx *cli.Context) error {
 		disableOngoingUpdates := !app.BaseCLI.Flags().Logstream || app.BaseCLI.Flags().InteractiveDebugging
 		forceColor := envutil.IsTrue("FORCE_COLOR")
 		noColor := envutil.IsTrue("NO_COLOR")
+		isGitHubActions := envutil.IsTrue("GITHUB_ACTIONS") || app.BaseCLI.Flags().IsGitHubActions
```
idodod, line 61: I _think_ the flag also handles the case that it's set via the env var (could be wrong though,   I don't recall if the flag is already set at the time of calling `before()`), perhaps worth checking

cmd/earthly/flag/global.go
```diff
@@ -539,5 +540,12 @@ func (global *Global) RootFlags(installName string, bkImage string) []cli.Flag {
 			Destination: &global.NoAutoSkip,
 			Value:       false,
 		},
+		&cli.BoolFlag{
+			Name:        "github-annotations",
+			EnvVars:     []string{"GITHUB_ACTIONS"},
```
idodod, line 545:
```suggestion
			EnvVars:     []string{"GITHUB_ANNOTATIONS"},
```

cmd/earthly/flag/global.go
```diff
@@ -101,6 +101,7 @@ type Global struct {
 	LocalSkipDB                string
 	DisableRemoteRegistryProxy bool
 	NoAutoSkip                 bool
+	IsGitHubActions            bool
```
idodod, line 104:
```suggestion
	IsGitHubAnnotations            bool
```

cmd/earthly/main.go
```diff
@@ -153,8 +153,8 @@ func main() {
 	if envutil.IsTrue("EARTHLY_FULL_TARGET") {
 		padding = conslogging.NoPadding
 	}
-
-	logging := conslogging.Current(colorMode, padding, conslogging.Info)
+	isGitHubActions := envutil.IsTrue("GITHUB_ACTIONS") || cli.Flags().IsGitHubActions
```
idodod, line 156: same as my previous comment, I think the flag should account for the env var being set.
```suggestion
	isGitHubActions := cli.Flags().IsGitHubActions
```

conslogging/conslogging.go
```diff
@@ -81,45 +83,47 @@ type ConsoleLogger struct {
 }
 
 // Current returns the current console.
-func Current(colorMode ColorMode, prefixPadding int, logLevel LogLevel) ConsoleLogger {
-	return New(getCompatibleStderr(), &currentConsoleMutex, colorMode, prefixPadding, logLevel)
+func Current(colorMode ColorMode, prefixPadding int, logLevel LogLevel, isGitHubActions bool) ConsoleLogger {
```
idodod, line 86:
```suggestion
func Current(colorMode ColorMode, prefixPadding int, logLevel LogLevel, isGitHubAnnotations bool) ConsoleLogger {
```

conslogging/conslogging.go
```diff
@@ -266,6 +272,67 @@ func (cl ConsoleLogger) PrefixColor() *color.Color {
 	return cl.color(c)
 }
 
+// Prints a GitHub Actions summary message to GITHUB_STEP_SUMMARY
+func (cl *ConsoleLogger) PrintGHASummary(message string) {
+	if !cl.isGitHubActions {
+		return
+	}
+
+	path := os.Getenv("GITHUB_STEP_SUMMARY")
```
idodod, line 281: where is this being set/by what?

conslogging/conslogging.go
```diff
@@ -266,6 +272,67 @@ func (cl ConsoleLogger) PrefixColor() *color.Color {
...
+	if !cl.isGitHubActions {
+		return
+	}
+	// Assumes mu locked.
+	w := new(bytes.Buffer)
+	defer func() {
+		_, _ = w.WriteTo(cl.errW)
+	}()
+
+	fmt.Fprintf(w, string(header)+" "+format+"\n", a...)
```
idodod, line 333: might be an overkill, but wondering if some callers might already add a newline?
In which case maybe we can conditionally add the newline here (`strings.HasSuffix(s, \"n")`)

idodod (commented):
Looks good, a few more small comments.
Also, should we update GithubActions to GithubAnnotations to be consistent?

conslogging/conslogging.go
adamgordonbell: This part is a bit magic. When you run in GHA it sets that to a file that it uses to populate the summary page. So in GHA runner, GITHUB_STEP_SUMMARY is a file path to some temp uuid file and also GITHUB_ACTIONS is always equal to "true".

idodod: oh gotcha, thanks! please disregard my comments about renaming the `GITHUB_ACTIONS` env var.

Question, do you know if this is set by the GH runner or something that is set to the runner?
Just wondering if it's something we need to set in our own GH runner and for Nacho to be aware of.

cmd/earthly/app/before.go
adamgordonbell: oh nice, will check

adamgordonbell: seems to work

conslogging/conslogging.go
adamgordonbell: done

conslogging/conslogging.go
adamgordonbell: @idelvall are these set in your GHA runner by chance?

idelvall: I will check. I am pretty sure the latter it's set

idelvall: One thing to have in mind is that the GHA might mix earthly commands and others. Wondering if these changes would give too much relevance to the Earthly code

CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ All notable changes to [Earthly](https://github.com/earthly/earthly) will be doc
 
 ## Unreleased
 
+### Added
+- New Github Actions Workflow commands integration `--github-annotations` flag or GITHUB_ACTIONS=true env. See [#4047](https://github.com/earthly/earthly/pull/4047)
```
alexcb, line 8: nit: we tend to only have the issue at the end of changelog entries, the "See" is implied by the link.
```suggestion
- New Github Actions Workflow commands integration `--github-annotations` flag or GITHUB_ACTIONS=true env. [#4047](https://github.com/earthly/earthly/pull/4047)
```

cmd/earthly/flag/global.go
```diff
@@ -101,6 +101,7 @@ type Global struct {
 	LocalSkipDB                string
 	DisableRemoteRegistryProxy bool
 	NoAutoSkip                 bool
+	IsGithubAnnotations        bool
```
alexcb, line 104: nit: drop the `Is`, e.g.
```suggestion
	GithubAnnotations        bool
```

or `EnableGithubAnnotations`

CHANGELOG.md
adamgordonbell: Done

cmd/earthly/flag/global.go
adamgordonbell: Done

conslogging/conslogging.go
```diff
@@ -266,6 +272,72 @@ func (cl ConsoleLogger) PrefixColor() *color.Color {
...
 
+// Prints a GitHub Actions summary message to GITHUB_STEP_SUMMARY
+func (cl *ConsoleLogger) PrintGHASummary(message string) {
+	if !cl.isGitHubActions {
+		return
+	}
+
+	path := os.Getenv("GITHUB_STEP_SUMMARY")
+	if path == "" {
+		return
```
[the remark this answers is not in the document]
idodod: maybe using `--github-annotations` should also set `GITHUB_STEP_SUMMARY` env var?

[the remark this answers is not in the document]
adamgordonbell: > maybe using --github-annotations should also set GITHUB_STEP_SUMMARY env var?

What would we set it to? How about just logging the summary, if its not set?  

I'm not sure there is a reason they would want the summary written to a file if not in GHA where its used for something.

conslogging/conslogging.go
```diff
@@ -266,6 +272,72 @@ func (cl ConsoleLogger) PrefixColor() *color.Color {
...
+// The `file`, `line`, and `col` parameters are optional.
+func (cl *ConsoleLogger) PrintGHAError(message string, details ...string) {
+	cl.mu.Lock()
+	defer cl.mu.Unlock()
+
+	file := ""
+	line := ""
+	col := ""
+
+	if len(details) >= 3 {
```
alexcb, line 303: having a different code path based on the length here is a bit hard to follow; a better go idiom is to define a functional option, see https://golang.cafe/blog/golang-functional-options-pattern.html

This would allow you to call it with:

```
PrintGHAError( "my error message", WithGHASourceLocation("my-file", 10, 20))
```

or

```
PrintGHAError( "my error message")
```

which means the code could look like:

```
type GHAError struct {
  message string
  file string
  line int
  col int
}

type GHAErrorOpt func(*GHAError)

func WithGHASourceLocation(file string, line, col int) GHAErrorOpt {
  return func(cfg *GHAError) {
    cfg.file = file
    cfg.line = line
    cfg.col = col
  }
}

func (cl *ConsoleLogger) PrintGHAError(message string, fns ...GHAErrorOpt) {
    cfg := GHAError{
        message: message,
    }
    for _, fn := range fns {
      fn(&cfg)
    }

  if cfg.file != "" {
    cl.printGithubActionsControl(errorCommand, "file=%s,line=%s,col=%s,title=Error::%s", file, line, col, message)
  } else {
    cl.printGithubActionsControl(errorCommand, "title=Error::%s", message)
  }
}
```

conslogging/conslogging.go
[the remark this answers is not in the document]
alexcb: > What would we set it to? How about just logging the summary, if its not set?

this seems like a better idea -- I think a user would like seeing the contents of it :)

conslogging/conslogging.go
adamgordonbell: done

cmd/earthly/flag/global.go
```diff
@@ -539,5 +540,12 @@ func (global *Global) RootFlags(installName string, bkImage string) []cli.Flag {
 			Destination: &global.NoAutoSkip,
 			Value:       false,
 		},
+		&cli.BoolFlag{
```
idodod, line 543: @alexcb 
Should this be a hidden flag?

alexcb: we were thinking some users might want to run `earthly --github-annotations` directly on their terminal (potentially as a way to experiment with the github actions output).

I think it's fine left visible; see https://github.com/earthly/earthly/pull/4047#discussion_r1578134238

conslogging/conslogging.go
```diff
@@ -266,6 +272,91 @@ func (cl ConsoleLogger) PrefixColor() *color.Color {
...
+	}
+	for _, fn := range fns {
+		fn(&cfg)
+	}
+
+	if cfg.file != "" {
+		cl.printGithubActionsControl(errorCommand, "file=%s,line=%d,col=%s,title=Error::%s", cfg.file, cfg.line, cfg.col, cfg.message)
+	} else {
+		cl.printGithubActionsControl(errorCommand, "title=Error::%s", cfg.message)
+	}
```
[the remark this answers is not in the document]
adamgordonbell: Done

CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ All notable changes to [Earthly](https://github.com/earthly/earthly) will be doc
 
 ## Unreleased
 
+### Added
+- New Github Actions Workflow commands integration `--github-annotations` flag or GITHUB_ACTIONS=true env. See [#2189](https://github.com/earthly/earthly/issues/2189)
```
idodod, line 8:
```suggestion
- New Github Actions Workflow commands integration `--github-annotations` flag or GITHUB_ACTIONS=true env. [#2189](https://github.com/earthly/earthly/issues/2189)
```

idodod: I think this was fixed and then reverted?

bla.md
```diff
@@ -0,0 +1,60 @@
+
```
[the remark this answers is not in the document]
adamgordonbell: Oh, shoot, trying to get integration tests passing and making a mess :) will fix.

idodod (approved):

zaufi:
Nice, but GHA isn't the only CI/CD tool. How about `--teamcity-annotations`, `--gitlab-annotations`, or `--jenkins-annotations`?

I'm trying to say that **hardcoding** support for one particular CI/CD doesn't look like a good decision/implementation IMHO.

adamgordonbell:
> Nice, but GHA isn't the only CI/CD tool. How about `--teamcity-annotations`, `--gitlab-annotations`, or `--jenkins-annotations`?

Happy to take feature requests for those, especially outlining what the integration would look like. Looking quickly, it seems like only TeamCity has a similar system of taking commands via logged output. 

> I'm trying to say that **hardcoding** support for one particular CI/CD doesn't look like a good decision/implementation IMHO.

Each integration will likely be different and so our approach is take feature requests, prioritize them based on user interest and work through them as we can.

PR #4052, opened 2024-04-22
add SatelliteAddress overwrite support for SH satellites too

idelvall:
The GH SH runner needs to call the satellite through `earthly.local` hostname.
This PR makes possible overwriting the satellite address also for SH satellites.


alexcb (changes requested):
It's a bit difficult to follow the logic here, how about rewriting it to be:

```
	satelliteAddress := cli.Flags().SatelliteAddress
	if !sat.IsManaged && satelliteAddress == "" {
		// A self-hosted satellite uses its own address
		satelliteAddress = fmt.Sprintf("tcp://%s", sat.Address)
	}
```

alexcb (approved):

PR #4055, opened 2024-04-23
tests: fix should_fail arg

alexcb:
`should-fail` is not an ARG, but `should_fail` is.


alexcb:
@ido-namely I noticed that `--should-fail` was introduced in e7e02b407b74f6b1a3df75f5286b18422be9a9c5

But that commit _also_ introduced a target called `test-fails` which doesn't seem to be referenced anywhere.

My gut is telling me we should update the referenced target name too?

alexcb:
> My gut is telling me we should update the referenced target name too?

just added this; let's see if this helps.

Issue #3485, opened 2023-11-11
Incorrect image loading time in `WITH DOCKER`


rrjjvv:
**What went wrong?**

Nothing went wrong; it's closer to a cosmetic issue than an true bug.  The *very awesome* change in #3377 artificially inflates its awesomeness due to differences in `date` precision.  I exclusively use the alpine d-in-d images, which will always (falsely) show single- or double-digit millisecond load times.  Only the ubuntu variant is accurate.

Log snippet from a realistic project:
```
   ./functional+test | Loading images from BuildKit via embedded registry...
   ./functional+test | Pulling 172.30.0.1:8371/sess-k7b9l51bl0bes9p0wdgzzakg2:img-4 and retagging as app-image:latest
   ./functional+test | Pulling 172.30.0.1:8371/sess-k7b9l51bl0bes9p0wdgzzakg2:img-8 and retagging as bats:latest
   ./functional+test | Pulling 172.30.0.1:8371/sess-k7b9l51bl0bes9p0wdgzzakg2:img-6 and retagging as consul-server:latest
   ./functional+test | Pulling 172.30.0.1:8371/sess-k7b9l51bl0bes9p0wdgzzakg2:img-3 and retagging as consul-template:latest
   ./functional+test | Pulling 172.30.0.1:8371/sess-k7b9l51bl0bes9p0wdgzzakg2:img-9 and retagging as gcr.io/cadvisor/cadvisor:v0.47.2
   ./functional+test | Pulling 172.30.0.1:8371/sess-k7b9l51bl0bes9p0wdgzzakg2:img-2 and retagging as hashi-vault-agent:latest
   ./functional+test | Pulling 172.30.0.1:8371/sess-k7b9l51bl0bes9p0wdgzzakg2:img-1 and retagging as k8s-vault-agent:latest
   ./functional+test | Pulling 172.30.0.1:8371/sess-k7b9l51bl0bes9p0wdgzzakg2:img-0 and retagging as kind:latest
   ./functional+test | Pulling 172.30.0.1:8371/sess-k7b9l51bl0bes9p0wdgzzakg2:img-10 and retagging as registry.k8s.io/metrics-server/metrics-server:v0.6.4
   ./functional+test | Pulling 172.30.0.1:8371/sess-k7b9l51bl0bes9p0wdgzzakg2:img-7 and retagging as vault-seeder:latest
[...]

PR #4064, opened 2024-04-26
Acb/with docker caching

### LINKED ISSUE #3399: `WITH DOCKER` with cached docker root dir
**Use case**

Faster `WITH DOCKER` load times (at the cost of slightly less repeatability).

**Expected Behavior**

Currently `WITH DOCKER` wipes the Docker root dir on each invocation. This is great for ensuring that builds are repeatable - nothing in the pre-existing Docker state from a previous (like a previously loaded image) can affect the current to cause inconsistency.

That, unfortunately, comes at the cost of having to load all the images into the inner Docker on each run, an operation that takes time, even though it's just a local transfer.

This proposal introduces an option for the Docker root dir to be kept around between runs, in a similar manner to which users can use global cache directories via `--id`.

Possible syntax:

```Earthly
WITH DOCKER --cache-id=my-string --load ... --pull ...
  RUN ...
END
```

The `--cache-id` option does three things:
1. Skips the removal of the dockerd data root at the end of WITH DOCKER.
2. Reuses an existing dockerd data root with the user-provided ID, if it already exists, instead of creating a new one
3. For added repeatability, resets the Docker state between runs as much as possible, without removing image layers. For example, by re-tagging existing images, deleting old stopped containers, old networks, old volumes etc. Keeping the layers helps with caching, while resetting everything else maintains repeatability to some extent.

Possible extensions:

* Having different modes of sharing the Docker instance between targets or parallel runs similar to the way there are different cache locking options for type=cache mounts.

**Challenges**

An implementation challenge is the fact that keeping the Docker root dirs around requires that they are managed and GC'd appropriately, in a similar manner that Buildkit manages the cache. These root dirs cannot be part of Buildkit's cache management, though, because of the inherent limitation that Buildkit operates on overlayfs, and the Docker root dir has to also be overlayfs, and you can't put overlayfs on top of overlayfs.

For this reason, the lifecycle of these Docker root dirs needs to be managed separately within Earthly's buildkit fork.

A v0 / initial experimental version of this feature could perhaps rely on manual-only clearing of Docker root dirs (via some Earthly command) and not require that we implement automatic GCing for it.

A v1 version could maintain a last-used timestamp in these directories and clean old entries on Buildkit startup (e.g. older than 14 days).

alexcb:
Introduces a new `WITH DOCKER --cache-id=<name>` flag, which when
specified will cache the daya layers from the docker data-root
between runs, even when dependencies have changed.

Note that the docker image tags are NOT cached; and a user will still
have to re-load the image (via the WITH DOCKER --load or --pull flags);
however, since the data layers are cached, the re-tagging should be much
faster.

fixes #3399 #3485

### DIFF
--- a/buildkitd/dockerd-wrapper.sh
+++ b/buildkitd/dockerd-wrapper.sh
@@ -10,6 +10,9 @@ fi
 # This host is used to pull images from the embedded BuildKit Docker registry.
 buildkit_docker_registry='172.30.0.1:8371'
 
+# used to prefix images that are persisted to the WITH DOCKER cache
+earthly_cached_docker_image_prefix="earthly_cached_"
+
 detect_docker_compose_cmd() {
     if command -v docker-compose >/dev/null; then
         echo "docker-compose"
@@ -98,9 +101,9 @@ execute() {
     if [ "$EARTHLY_DOCKERD_CACHE_DATA" = "true" ]; then
         # rename existing tags, so we can track which ones get re-tagged
         for img in $(docker images -q); do
-            docker tag "$img" "earthly_cached_$img"
+            docker tag "$img" "${earthly_cached_docker_image_prefix}${img}"
         done
-        docker images -a | grep -v ^REPOSITORY | grep -v ^earthly_cached_ | awk '{print $1":"$2}' | xargs --no-run-if-empty docker rmi
+        docker images -a --format '{{.Repository}}:{{.Tag}}' | grep -v "^$earthly_cached_docker_image_prefix" | xargs --no-run-if-empty docker rmi
     fi
 
     load_file_images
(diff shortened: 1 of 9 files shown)


buildkitd/dockerd-wrapper.sh
```diff
@@ -95,8 +95,23 @@ execute() {
         fi
     done
 
+    if [ "$EARTHLY_DOCKERD_CACHE_DATA" = "true" ]; then
+        # rename existing tags, so we can track which ones get re-tagged
+        for img in $(docker images -q); do
+            docker tag "$img" "earthly_cached_$img"
+        done
+        docker images -a | grep -v ^REPOSITORY | grep -v ^earthly_cached_ | awk '{print $1":"$2}' | xargs --no-run-if-empty docker rmi
```
idodod, line 103:
```suggestion
        docker images -a --format '{{.Repository}}:{{.Tag}}' | grep -v ^earthly_cached_ | xargs --no-run-if-empty docker rmi
```

buildkitd/dockerd-wrapper.sh
```diff
@@ -95,8 +95,23 @@ execute() {
         fi
     done
 
+    if [ "$EARTHLY_DOCKERD_CACHE_DATA" = "true" ]; then
+        # rename existing tags, so we can track which ones get re-tagged
+        for img in $(docker images -q); do
+            docker tag "$img" "earthly_cached_$img"
```
idodod, line 101: I think it would be good to put the prefix in a var/const to avoid possible typos in the future.

idodod (approved):

buildkitd/dockerd-wrapper.sh
alexcb: good plan.

PR #4073, opened 2024-04-30
Requires FROM error message improvement

adamgordonbell:
Old Error:
```
Error: tests/raw-output2/Earthfile Line:7:4 the first command has to be FROM, FROM DOCKERFILE, LOCALLY, ARG, BUILD or IMPORT
```
New Error:
```
Error: tests/raw-output2/Earthfile:7:4 apply FOR ... IN: requires a FROM, FROM DOCKERFILE, or LOCALLY

  Hint: This command needs to run in a shell. You should be able to solve this by adding 'FROM someImage' on the line before this one.

in              github.com/earthly/earthly/tests/raw-output2:agbell/error_message+for
```

### Error to Source link format

The `path:line:col` is source link format that is clickable in VS Code, intellij and other IDEs. Changing errors to use it instead of `Error: file Line:line:col`

### DIFF
--- a/cmd/earthly/app/run.go
+++ b/cmd/earthly/app/run.go
@@ -386,7 +386,7 @@ func (app *EarthlyApp) run(ctx context.Context, args []string, lastSignal *syncu
 					time.Now(),
 					logstream.FailureType_FAILURE_TYPE_OTHER,
 					"",
-					grpcErr.Message(),
+					messageWithSourceLink(grpcErr.Message()),
 				)
 				return 1
 			}
@@ -502,6 +502,12 @@ func (app *EarthlyApp) printCrashLogs(ctx context.Context) {
 
 func errorWithPrefix(err string) string {
 	return fmt.Sprintf("Error: %s", err)
+}
+
+func messageWithSourceLink(message string) string {
+	re := regexp.MustCompile(`(\S+) line (\d+):(\d+)`)
+	updatedMessage := re.ReplaceAllString(message, "$1:$2:$3")
+	return updatedMessage
 }
 
 func getHintErr(err error, grpcError *status.Status) (*hint.Error, bool) {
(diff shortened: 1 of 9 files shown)


cmd/earthly/app/run.go
```diff
@@ -385,7 +385,7 @@ func (app *EarthlyApp) run(ctx context.Context, args []string, lastSignal *syncu
 					time.Now(),
 					logstream.FailureType_FAILURE_TYPE_OTHER,
 					"",
-					grpcErr.Message(),
+					messageWithSourceLink(grpcErr.Message()),
```
idodod, line 388: It might make more sense to change the text in the source instead of having to replace it with regex, sense I'm not sure it would only apply to gRPC error message.

earthfile2llb/converter.go
```diff
@@ -2835,7 +2836,7 @@ func (c *Converter) checkAllowed(command cmdType) error {
 		case fromCmd, fromDockerfileCmd, locallyCmd, buildCmd, argCmd, letCmd, setCmd, importCmd, projectCmd, pipelineCmd:
 			return nil
 		default:
-			return errors.New("the first command has to be FROM, FROM DOCKERFILE, LOCALLY, ARG, BUILD or IMPORT")
+			return hint.Wrap(errors.New("requires a FROM, FROM DOCKERFILE, or LOCALLY"), "This command needs to run in a shell. You should be able to solve this by adding 'FROM someImage' on the line before this one.")
```
mikejholly, line 2839: `FROM <image>` or `FROM [image]` may be a more familiar way to denote an argument. But, this may be overly pedantic. 

https://en.wikipedia.org/wiki/Usage_message#Pattern

mikejholly (approved):

alexcb (approved):
> The path:line:col is source link format that is clickable in VS Code, intellij and other IDEs. Changing errors to use it instead of Error: file Line:line:col

nice improvement!

adamgordonbell:
> > The path:line:col is source link format that is clickable in VS Code, intellij and other IDEs. Changing errors to use it instead of Error: file Line:line:col
> 
> nice improvement!

Yeah, I couldn't find a name for this `format` documented anywhere, but many things support it and I'm calling it `source link format`.

PR #4078, opened 2024-05-01
return exit code if max run time is exceeded

alexcb:


vladaionescu (approved):

cmd/earthly/app/run.go
```diff
@@ -378,6 +378,7 @@ func (app *EarthlyApp) run(ctx context.Context, args []string, lastSignal *syncu
 			helpMsg := "Unverified accounts have a limit on the duration of RUN commands. Verify your account to lift this restriction."
 			app.BaseCLI.Logbus().Run().SetGenericFatalError(time.Now(), logstream.FailureType_FAILURE_TYPE_OTHER, helpMsg, grpcErr.Message())
 			app.BaseCLI.Console().HelpPrintf(helpMsg)
+			return 1
```
mikejholly, line 381: I think this was my mistake. Thanks for fixing.

cmd/earthly/app/run.go
```diff
@@ -469,6 +470,7 @@ func (app *EarthlyApp) run(ctx context.Context, args []string, lastSignal *syncu
 			)
 			return 1
 		}
+		panic("error was not caught by error-handling switch") // if this happens there is a missing "return" in the above switch statement.
```
mikejholly, line 473: Do we want to panic here or default to `return 1`? And should we use `panic` or just output a message? A `panic` seems less ideal for end users.

idodod: maybe another thing we can try is define exiCode at the top (`exitCode := 1`), remove the early returns and/or change them to `return exitCode`
and then set it to zero when there is no error, so by default all cases will return an error?

alexcb: Every single `break` above should have a `return 1` (or other non-zero code).

This discussion feels a bit too nitpicky given this panic will never be hit unless there's a programmer error (vs a runtime error).

mikejholly: `panic` seems fine if we're ok with users seeing stack traces in this instance. I see we use `panic` elsewhere for similar checks. In [some places](https://github.com/earthly/earthly/blob/main/cmd/earthly/app/run.go#L152-L160) we use different message styles. 

I don't think it's a hard and fast rule to return from every `case`, but it's probably not worth changing this now.

mikejholly (approved):

PR #4084, opened 2024-05-02
feat: SAVE IMAGE --without-earthly-labels

3manuek:
This commit adds the `--without-earthly-labels` for the SAVE IMAGE instruction to not modify the digest of the final artifact if none of its contents changes. Enabled `--allow-without-earthly-labels` feature.

Certain custom registries may require strict checking of the image contents, so this option disables build information.


tests/save-images/disable-earthly-labels/Earthfile
```diff
@@ -0,0 +1,12 @@
...
+
+myimage:
+    FROM scratch
+    COPY file /
+    SAVE IMAGE --disable-earthly-labels myimage:test
+
+myimagewithlabels:
+    FROM scratch
+    COPY file /
+    SAVE IMAGE myimagewithlabels:test
```
alexcb, line 12: this will fail the `+lint` target; please make sure all files end with a newline.

ast/commandflag/flags.go
```diff
@@ -80,11 +80,12 @@ type SaveArtifactOpts struct {
...
-	NoManifestList bool     `long:"no-manifest-list" description:"Do not include a manifest list (specifying the platform) in the creation of the image"`
-	CacheFrom      []string `long:"cache-from" description:"Declare additional cache import as a Docker tag"`
+	Push                 bool     `long:"push" description:"Push the image to the remote registry provided that the build succeeds and also that earthly is invoked in push mode"`
+	CacheHint            bool     `long:"cache-hint" description:"Instruct Earthly that the current target should be saved entirely as part of the remote cache"`
+	Insecure             bool     `long:"insecure" description:"Use unencrypted connection for the push"`
+	NoManifestList       bool     `long:"no-manifest-list" description:"Do not include a manifest list (specifying the platform) in the creation of the image"`
+	CacheFrom            []string `long:"cache-from" description:"Declare additional cache import as a Docker tag"`
+	DisableEarthlyLabels bool     `long:"disable-earthly-labels" description:"Disable build information dev.earthly labels to reduce the chance of changing images digsest."`
```
idodod, line 88: a little bit of a nit pick might be more consistent with other flags to call this :
`no-earthly-labels` maybe?
CC @alexcb

tests/save-images/test.sh
```diff
@@ -63,4 +63,27 @@ echo "=== ($LINENO): Test New Behaviour referencing remote via build ==="
 "$frontend" inspect myimage:buildtest >/dev/null
 "$frontend" inspect earthly-test-saveimage:test >/dev/null
 
+echo "=== ($LINENO): Test Disable Earthly Labels ==="
```
alexcb, line 66: Instead of having this test here, can you do it in an Earthfile instead?

Perhaps something along the lines of:

```
$ cat ./tests/with-docker-validate-labels/Earthfile 
VERSION --allow-disable-earthly-labels 0.8 0.8

ARG --global DIND_IMAGE=earthly/dind:alpine-3.19-docker-25.0.5-r0
FROM $DIND_IMAGE

all:
    BUILD +test-with-labels
    BUILD +test-without-labels

myimage-with-labels:
    FROM alpine
    COPY file /
    SAVE IMAGE myimage:test

myimage-without-labels:
    FROM alpine
    COPY file /
    SAVE IMAGE myimage:test

test-with-labels:
    WITH DOCKER --load=+myimage-with-labels
        RUN docker inspect myimage:test | jq -r '.[].Config.Labels' # ....
    END

test-without-labels:
    ...
```

then under `tests/Earthfile`, you can call it by adding it to

```
ga-no-qemu-group12:
    ...
    BUILD --pass-args ./with-docker-validate-labels+all
```

features/features.go
```diff
@@ -74,6 +74,7 @@ type Features struct {
 	WildcardCopy                  bool `long:"wildcard-copy" description:"allow for the expansion of wildcard (glob) paths for COPY commands"`
 	RawOutput                     bool `long:"raw-output" description:"allow for --raw-output on RUN commands"`
 	GitAuthorEmailNameArgs        bool `long:"git-author-email-name-args" description:"includes EARTHLY_GIT_AUTHOR_EMAIL and EARTHLY_GIT_AUTHOR_NAME builtin ARGs"`
+	AllowDisableEarthlyLabels     bool `long:"allow-disable-earthly-labels" description:"Allow the usage of --disable-earthly-labels in SAVE IMAGE"`
```
alexcb, line 77: I like that you didn't call it `--enable-disable-earthly-labels`, the `--allow-*` prefix was a good call :smile:

earthfile2llb/interpreter.go
```diff
@@ -1170,14 +1170,16 @@ func (i *Interpreter) handleSaveImage(ctx context.Context, cmd spec.Command) err
...
 
-	labels := map[string]string{
-		"dev.earthly.version":  version.Version,
-		"dev.earthly.git-sha":  version.GitSha,
-		"dev.earthly.built-by": version.BuiltBy,
-	}
-	err = i.converter.Label(ctx, labels)
-	if err != nil {
-		return i.wrapError(err, cmd.SourceLocation, "failed to create dev.earthly.* labels during SAVE IMAGE")
+	if !opts.DisableEarthlyLabels {
```
alexcb, line 1173: there should be a check to make sure the `AllowDisableEarthlyLabels` feature was enabled.

e.g.

```
if opts.DisableEarthlyLabels {
	if !i.converter.ftrs.AllowDisableEarthlyLabels {
	return i.errorf(cmd.SourceLocation, "the SAVE IMAGE --disable-earthly-labels flag must be enabled with the VERSION --allow-disable-earthly-labels feature flag.")
	}
	// deliberately don't add any labels (i.e. do nothing here)
} else {
...
}
```

alexcb (changes requested):
thanks for the submission, just a few small requests.

ast/commandflag/flags.go
alexcb: That's a good suggestion, another idea is `--without-earthly-labels`.

3manuek: I like both, I would probably go with the shortest.

features/features.go
3manuek: If we change the SAVE IMAGE to `--no-earthly-labels`, the feature attribute might need to be changed to `AllowNoEarthlyLabels`. The "AllowNo" prefix sounds confusing, probably it makes more sense `AllowWithoutEarhtlyLabels`.

ast/commandflag/flags.go
3manuek: The problem with `--no-earthly-labels` is that the feature attribute would be `AllowNoEarthlyLabels`, which is kinda confusing. `without` would probably don't have this issue.

3manuek:
New commit with changes in the review pushed. cc /@alexcb

earthfile2llb/interpreter.go
```diff
@@ -1171,14 +1171,21 @@ func (i *Interpreter) handleSaveImage(ctx context.Context, cmd spec.Command) err
...
-		"dev.earthly.version":  version.Version,
-		"dev.earthly.git-sha":  version.GitSha,
-		"dev.earthly.built-by": version.BuiltBy,
-	}
-	err = i.converter.Label(ctx, labels)
-	if err != nil {
-		return i.wrapError(err, cmd.SourceLocation, "failed to create dev.earthly.* labels during SAVE IMAGE")
+	if opts.WithoutEarthlyLabels {
+		if !i.converter.ftrs.AllowWithoutEarthlyLabels {
+			return i.errorf(cmd.SourceLocation, "the SAVE IMAGE --disable-earthly-labels flag must be enabled with the VERSION --allow-without-earthly-labels feature flag.")
```
alexcb, line 1176:
```suggestion
			return i.errorf(cmd.SourceLocation, "the SAVE IMAGE --without-earthly-labels flag must be enabled with the VERSION --allow-without-earthly-labels feature flag.")
```

alexcb (approved):
Thanks for the submission!

Issue #3804, opened 2024-02-12
Improve AWS authentication experience V2


mikejholly:
Follows on from https://github.com/earthly/earthly/issues/3803

Expand support by supporting OIDC directly in Earthly.  Exactly how needs to be fleshed out but a new command may be a possibility. 

Example GHA code of using OIDC: https://github.com/aws-actions/configure-aws-credentials/blob/main/src/index.t

PR #4096, opened 2024-05-07
Support AWS authentication with OIDC

idodod:
Address #3804 

High level implementation details:
1. Add new RUN flag behind a feature flag that expects multiple key/values (some are mandatory).
2. Parse and pass the keys to the converter, pass the key values as additional params to the aws secrets IDs
3. Modify the aws credentials provider (which was used so far to fetch credentials from local env) to try to get the credentials with oidc:
    1. If the oidc key/values are present in the secret - try to call the cloud to get the credentials, otherwise fallback to local envs.
    2. If oidc, use cache/locks to overcome the fact that secrets are fetched one by one but AWS creds are a group of secrets.


mikejholly (approved):
Nice work! LGTM.

docs/guides/auth.md
```diff
@@ -189,6 +189,12 @@ You can see examples of configuring Docker to use these, and working with Earthl
 * [Pushing and Pulling Images with GCP Artifact Registry](./registries/gcp-artifact-registry.md)
 * [Pushing and Pulling Images with Azure ACR](./registries/azure-acr.md)
 
+## Cloud Providers
+
+Currently Earthly provides a build-in way to easily authenticate to AWS during a build.
```
alexcb, line 194:
```suggestion
Currently Earthly provides a built-in way to easily authenticate to AWS during a build.
```

tests/Earthfile
```diff
@@ -147,6 +147,7 @@ ga-no-qemu-group5:
     BUILD +test-aws-flag-envs
     BUILD +test-aws-flag-configs
     BUILD +test-aws-flag-none
+    BUILD +test-aws-oidc
```
alexcb, line 150: I would put this under the last `group12` instead _if it didn't require any special auth_ (this is to keep the macstadium instances from becoming overloaded)

However, since it looks like it does require special credentials, it should be moved to the `tests-that-require-earthly-technologies-account-access` target

alexcb: i just scrolled down and saw a different test triggered via `require-earthly-technologies-account-access`, so then this might belong under `group12` :laughing:

tests/oidc/Earthfile
```diff
@@ -0,0 +1,39 @@
...
+
+test-aws-base:
+    COPY test-aws.sh .
+    COPY aws.earth Earthfile
+    ENV EARTHLY_EXEC_CMD=/test/test-aws.sh
+
+test-aws-success:
+    FROM +test-aws-base
+    # happy path
+    ARG ROLE_ARN="arn:aws:iam::404851345508:role/ido-oidc-test"
```
alexcb, line 25: `ido-oidc-test` :raised_eyebrow:

I could look the other way (since it might be a pain to change it to something more descriptive like `oidc-integration-test` since it's not testing you)

tests/oidc/aws.earth
```diff
@@ -0,0 +1,15 @@
+VERSION --run-with-aws --run-with-aws-oidc 0.8
+
+PROJECT ido+testemail/my-project1
```
alexcb, line 3: again, I think it's a bad habit to choose project names based on peoples names.

tests/oidc/test-aws.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+set -eo pipefail # DONT add a set -x or you will leak the key
+
+acbtest -n "$OIDC_USER_TOKEN"
```
alexcb, line 4: maybe I should have called this `verbosetest` :rofl:

alexcb (approved):
nice work.

tests/oidc/Earthfile
idodod: This probably wasn't obvious from my comment but these will change once I can set up the environment in production https://github.com/earthly/earthly/pull/4096/files#r1592957105
So once this PR is approved I will:
* merge my BE changes
* set up the prod env (which would not be using my name)
* Update this test (which would not be using my name)

PR #4112, opened 2024-05-14
Update BK for new inline cache (won't merge)

mikejholly:
BuildKit PR: https://github.com/earthly/buildkit/pull/60

### DIFF
--- a/buildkitd/Earthfile
+++ b/buildkitd/Earthfile
@@ -12,7 +12,7 @@ buildkitd:
             ARG BUILDKIT_BASE_IMAGE=$BUILDKIT_PROJECT+build
         END
     ELSE
-        ARG BUILDKIT_BASE_IMAGE=github.com/earthly/buildkit:531b303aa8ec03c29c2ceaa140eb0a6d32e6f6f3+build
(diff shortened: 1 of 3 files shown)


alexcb (commented):
we're going to need to figure out a way to build both `earthly-main` and `earthly-next` versions for use in testing.

#4117 tackles building and pushing both images, but that won't cover being able to execute our tests on both copies, which I think we need to do.

mikejholly:
> we're going to need to figure out a way to build both `earthly-main` and `earthly-next` versions for use in testing.
> 
> #4117 tackles building and pushing both images, but that won't cover being able to execute our tests on both copies, which I think we need to do.

Agreed. We'll need to come up with something here soon.

alexcb:
I wonder if we could just have GHA do a `earthly +update-buildkit --BUILDKIT_GIT_SHA=$(cat earthly-next-sha-to-use)` before running `+for-linux`?

then all we need to do is create a file called `earthly-next-sha-to-use` in the git repo, and update the commit there?

(this is a temporary solution while we have two different branches to maintain)

PR #4115, opened 2024-05-15
Use default gRPC dialer when connecting to satellites

brandonSc:
Uses a new feature in buildkit which forces use of the default grpc dialer in the client-side connection. The default dialer supports HTTP connect proxies, which allows earthly to connect with BYOC satellites that may be behind a VPN like Tailscale.


buildkitd/buildkitd.go
```diff
@@ -622,7 +622,7 @@ func waitForConnection(ctx context.Context, containerName, address string, opTim
 	if !containerutil.IsLocal(address) {
 		retryInterval = 1 * time.Second
 	}
-	ctxTimeout, cancel := context.WithTimeout(ctx, opTimeout)
+	ctxTimeout, cancel := context.WithTimeout(ctx, 5*time.Second)
```
alexcb, line 625: is this intentional?

brandonSc: No sorry, I thought I removed that

alexcb (approved):

PR #4117, opened 2024-05-15
release.sh: build and push new experimental buildkit branch

alexcb:
This builds and pushes the new experimental buildkit branch (earthly-next) to the a new earthly/buildkitd tag with the `-ticktock` codename suffix.


mikejholly (approved):

earthly-next
```diff
@@ -0,0 +1 @@
+d4e630c482bde6e4f6da3d2257ee6e8d9955f180
```
alexcb, line 1: note that we'll have to start updating this file when we do releases for the `-ticktock` version.

mikejholly (approved):
Nice. I like the approach.

PR #4129, opened 2024-05-17
Remove legacy log sharing & associated code & tests

mikejholly:


cmd/earthly/app/before.go
```diff
@@ -23,69 +23,74 @@ import (
 )
 
 func (app *EarthlyApp) before(cliCtx *cli.Context) error {
-	if app.BaseCLI.Flags().EnableProfiler {
+	flags := app.BaseCLI.Flags()
```
alexcb, line 26: nice change :)

cmd/earthly/app/before.go
```diff
@@ -23,69 +23,74 @@ import (
...
-			disableOngoingUpdates, app.BaseCLI.Flags().LogstreamDebugFile, app.BaseCLI.Flags().BuildID, execStatsTracker, app.BaseCLI.Flags().GithubAnnotations)
-		app.BaseCLI.SetLogbusSetup(newSetup)
-		if err != nil {
-			return errors.Wrap(err, "logbus setup")
-		}
+	var execStatsTracker *execstatssummary.Tracker
+	if flags.ExecStatsSummary != "" {
+		execStatsTracker = execstatssummary.NewTracker(flags.ExecStatsSummary)
+	}
+	newSetup, err := logbussetup.New(
```
alexcb, line 57:
```suggestion
	theOnlySetup, err := logbussetup.New(
```

or maybe just `setup` :laughing:

cmd/earthly/app/before.go
mikejholly: Thanks. It's a little easier on the eyes :slightly_smiling_face:

cmd/earthly/app/before.go
mikejholly: Lol. I don't if that name means "the setup using the new approach" or just "the newly created setup variable". I'll clarify it, nonetheless.

alexcb: yeah `setup` seems like a bad name; I too don't really know what it means at first glance.

mikejholly: Sorry, I meant the existing name `newSetup`.

cmd/earthly/app/before.go
```diff
@@ -23,69 +23,74 @@ import (
...
+		flags.DisplayExecStats,
+		envutil.IsTrue("FORCE_COLOR"),
+		envutil.IsTrue("NO_COLOR"),
+		app.BaseCLI.Flags().InteractiveDebugging,
+		flags.LogstreamDebugFile,
+		flags.BuildID,
+		execStatsTracker,
+		flags.GithubAnnotations,
+	)
+	app.BaseCLI.SetLogbusSetup(newSetup)
```
alexcb, line 71: something seems off here:

```
	theOnlySetup, err := logbussetup.New(
	....
	)
	app.BaseCLI.SetLogbusSetup(newSetup)  <---- Shouldn't this happen after the err != nil check?
	if err != nil {
		return errors.Wrap(err, "logbus setup")
	}
```

mikejholly: That does seem odd. The previous commit has the same order, but it seems incorrect.

alexcb: yeah I see that, still seems odd -- I think fixing it is worthwhile.

alexcb (approved):

mikejholly:
@alexcb @vladaionescu Ok to merge?

alexcb (approved):
merge away!

PR #4155, opened 2024-05-23
fix: the --ticktock flag should also be used for bootstrap

alexcb:
This fixes the issue where the bootstrap command was ignoring the --ticktock option.

### DIFF
--- a/cmd/earthly/base/init_frontend.go
+++ b/cmd/earthly/base/init_frontend.go
@@ -1,6 +1,7 @@
 package base
 
 import (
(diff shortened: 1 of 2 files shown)


mikejholly:
I'm not very familiar with the `prune` command, but it looks like [it references the BuildKit image](https://github.com/earthly/earthly/blob/main/cmd/earthly/subcmd/prune_cmds.go#L88), which will be set when present in the config file.

vladaionescu (approved):

PR #4157, opened 2024-05-24
fix: force cache bust when --no-cache used

### LINKED ISSUE #2593: RUN --no-cache can sometimes be cached
Some users have reported that `RUN --no-cache` statements are displayed as `*cached*` , however, these commands should never be cached.

It appears this issue is not just cosmetic, i.e. the statement and it's next statements are really cached, and not just incorrectly displayed as such. 

Some discussions with users here on Slack:
https://earthlycommunity.slack.com/archives/C01DL2928RM/p1673982960942559

Running the `+docker` target on this opensource project does reproduce it: https://github.com/kairos-io/kairos/tree/eb64057da36c6594864e3bcf51aecf07cd4826df

However, we may need to narrow this down in order to identify the issue.

alexcb:
fixes #2593

### DIFF
--- a/earthfile2llb/converter.go
+++ b/earthfile2llb/converter.go
@@ -51,6 +51,7 @@ import (
 	"github.com/earthly/earthly/util/vertexmeta"
 	"github.com/earthly/earthly/variables"
 	"github.com/earthly/earthly/variables/reserved"
+	"github.com/google/uuid"
 	"github.com/moby/buildkit/client/llb"
 	dockerimage "github.com/moby/buildkit/exporter/containerimage/image"
 	"github.com/moby/buildkit/frontend/dockerfile/dockerfile2llb"
@@ -2306,6 +2307,10 @@ func (c *Converter) internalRun(ctx context.Context, opts ConvertRunOpts) (pllb.
 	// Shell and debugger wrap.
 	prependDebugger := !opts.Locally
 	finalArgs = opts.shellWrap(finalArgs, extraEnvVars, opts.WithShell, prependDebugger, isInteractive)
+	if opts.NoCache {
+		// llb.IgnoreCache is not always enough; we will force a different cache key as a work-around
+		finalArgs = append(finalArgs, "#"+uuid.NewString())
+	}
 	if opts.Locally {
 		// buildkit-hack in order to run locally, we prepend the command with a magic UUID.
 		finalArgs = append(
--- a/tests/Earthfile
+++ b/tests/Earthfile
@@ -216,7 +216,7 @@ ga-no-qemu-group12:
     BUILD --pass-args ./warn-if-not-logged-in+test
     BUILD --pass-args ./with-docker-validate-labels+all
     BUILD --pass-args +test-aws-oidc
-
+    BUILD --pass-args +run-no-cache-save-artifact
 
 ga-no-qemu-slow:
     BUILD +server
@@ -1060,6 +1060,26 @@ run-no-cache:
     # Fail if we cached any of the COPY motd2 lines, which are after the --no-cache
     DO +RUN_EARTHLY --earthfile=run-no-cache.earth --use_tmpfs=false --target=+test-from \
         --grep_flags="-v" --output_contains="\\\*cached\\\* --> .*motd2 \\\.\\\//"
+
+run-no-cache-save-artifact:
+    # based on https://github.com/earthly/earthly/issues/2593#issuecomment-1777657111
+    COPY run-no-cache-save-artifact.earth Earthfile
+    RUN touch some-file
+    LET numruns=10
+    RUN echo "#!/bin/sh
+set -ex
+for x in \$(seq 1 $numruns); do
+  earthly --config \$earthly_config --artifact '+build/*' test/\$x/
+done
+" >/tmp/test-earthly-script && chmod +x /tmp/test-earthly-script
+    ENV EARTHLY_EXEC_CMD="/tmp/test-earthly-script"
+    RUN --privileged \
+    --entrypoint \
+    --mount=type=tmpfs,target=/tmp/earthly \
+    -- -P
+    # test that all output files are different
+    RUN cat test/*/dep1
+    RUN acbtest "$(cat test/*/dep1 | sort | uniq | wc -l)" = "$numruns"
 
 save-artifact-after-push:
     # test that save after push only outputs files before the RUN --push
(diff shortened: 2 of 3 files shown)


earthfile2llb/converter.go
```diff
@@ -2306,6 +2307,10 @@ func (c *Converter) internalRun(ctx context.Context, opts ConvertRunOpts) (pllb.
 	// Shell and debugger wrap.
 	prependDebugger := !opts.Locally
 	finalArgs = opts.shellWrap(finalArgs, extraEnvVars, opts.WithShell, prependDebugger, isInteractive)
+	if opts.NoCache {
+		// llb.IgnoreCache is not always enough; we will force a different cache key as a work-around
+		finalArgs = append(finalArgs, "#"+uuid.NewString())
```
vladaionescu, line 2312: Interestingly simple... I hope this works well in all the weird cases. I can't think of a specific situation where it wouldn't.

alexcb:
I decided to generate a new random UUID each time in order to support this (weird) use-case:

```
if:
  FROM python:3
  IF --no-cache python3 -c 'import sys; import random; sys.exit(random.randint(0,10))'
    RUN --no-cache echo hi
  END
  IF --no-cache python3 -c 'import sys; import random; sys.exit(random.randint(0,10))'
    RUN --no-cache echo hi
  END
  IF --no-cache python3 -c 'import sys; import random; sys.exit(random.randint(0,10))'
    RUN --no-cache echo hi
  END
  IF --no-cache python3 -c 'import sys; import random; sys.exit(random.randint(0,10))'
    RUN --no-cache echo hi
  END
  IF --no-cache python3 -c 'import sys; import random; sys.exit(random.randint(0,10))'
    RUN --no-cache echo hi
  END
  IF --no-cache python3 -c 'import sys; import random; sys.exit(random.randint(0,10))'
    RUN --no-cache echo hi
  END
  IF --no-cache python3 -c 'import sys; import random; sys.exit(random.randint(0,10))'
    RUN --no-cache echo hi
  END
  IF --no-cache python3 -c 'import sys; import random; sys.exit(random.randint(0,10))'
    RUN --no-cache echo hi
  END
  IF --no-cache python3 -c 'import sys; import random; sys.exit(random.randint(0,10))'
    RUN --no-cache echo hi
  END
  IF --no-cache python3 -c 'import sys; import random; sys.exit(random.randint(0,10))'
    RUN --no-cache echo hi
  END
  RUN --no-cache echo end
```

each IF needs to be independent -- we don't want any of them to be cached.

earthfile2llb/converter.go
vladaionescu, line 2312: Should this be used in all the `WITH DOCKER` cases too? Or in `LOCALLY` too?

Any other situations where we do a run underneath but it's not part of the RUN command? (IF? FOR? - don't think those have --no-cache variants though)

idodod: I had a similar solution in a wip for a while.
The difference being attaching a uuid (the same uuid actually) for any subsequent RUNs that come after a RUN --no-cache, just because it was unclear if the bug affected those subsequent commands as well.

alexcb: good question -- it does work `WITH DOCKER`, which calls `internalRun`.

e.g. https://github.com/earthly/earthly/blob/7ad30b8d90034448b2796e1328f9fa875dd30375/earthfile2llb/with_docker_run_local_reg.go#L120

it's run in all cases:

```
$ grep internalRun earthfile2llb/with_docker_*.go
earthfile2llb/with_docker_run_local_reg.go:	_, err = w.c.internalRun(ctx, crOpts)
earthfile2llb/with_docker_run_local_tar.go:	_, err = w.c.internalRun(ctx, crOpts)
earthfile2llb/with_docker_run_reg.go:	_, err = w.c.internalRun(ctx, crOpts)
earthfile2llb/with_docker_run_tar.go:	_, err = w.c.internalRun(ctx, crOpts)
```

the `--no-cache` has no effect on `LOCALLY`, but adding it doesn't hurt.

We actually need this for `IF`, `FOR`, which do support the `--no-cache` option.

alexcb: yeah I figured we needed a new one each time to support the weird case of multiple IF statements which had the same llb deps (see my comment under the main PR if you're curious).

alexcb: oh interesting! yeah maybe it'll solve the COPY+LOCALLY issues :crossed_fingers:

idodod: Yeah, the new uuid makes sense(I only saw your example after posting my comment).
My only question is whether the bug in this example
```
RUN --no-cache echo hello
RUN echo goodbye
````
would sometimes cache and not echo "goodbye" And if so, should we make sure to attach a (new) uuid for every RUN that comes after `RUN --no-cache`

Comment (alexcb) [reply earthfile2llb/converter.go:2312]: both will be executed @idodod :

```
/h/a/t/simple-no-cache+base | --> RUN --no-cache echo hello
/h/a/t/simple-no-cache+base | hello
/h/a/t/simple-no-cache+base | DOCKERHUB_MIRROR_AUTH_FROM_CLOUD_SECRETS=true
/h/a/t/simple-no-cache+base | --> RUN echo goodbye
/h/a/t/simple-no-cache+base | goodbye
```

however this is consistent with what v0.8.12 does.

**Review by mikejholly — APPROVED** [review]

File: earthfile2llb/converter.go (hunk 1 above)
Comment (idodod) [reply earthfile2llb/converter.go:2312]: Right, I was under the assumption that that is what is supposed to happen and that usually happens, but wasn't sure if the bug (which is somewhat inconsistent) affects the scenario I described.

Comment (alexcb) [reply earthfile2llb/converter.go:2312]: yeah it sort of feels like it goes against buildkit's slow-cache idea, but I think it's fine.

If you break it up into two separate targets, e.g.

```
VERSION 0.8
FROM alpine
a:
  RUN --no-cache echo hello > msg
  SAVE ARTIFACT msg

b:
  COPY +a/msg .
  RUN echo goodbye
```

it doesn't run the `echo goodbye`:

```
              logbus | Setting organization "" and project ""
                  +b | --> FROM +base
                  +a | --> FROM +base
               +base | --> FROM alpine
               +base | [----------] 100% FROM alpine
                  +a | --> RUN --no-cache echo hello > msg
                  +b | *cached* --> COPY +a/msg ./
                  +b | *cached* --> RUN echo goodbye
```

which makes it look like it does hit the slowcache (which is also consistent with v0.8.12).

PR #4180, opened 2024-05-30
Better error message for WITH DOCKER (unlazy force execution)

adamgordonbell:
If a run fails in WITH DOCKER you get an error message like this:
```
================================== ❌ FAILURE ===================================

./t/with-docker+fail-test1 *failed* | Repeating the failure error...
./t/with-docker+fail-test1 *failed* | --> WITH DOCKER RUN --privileged echo "dummy" && false
./t/with-docker+fail-test1 *failed* | Starting dockerd with data root /var/earthly/dind/7b5da64c458d405fff1f21d3a985f6812f9eaad6af9c6eb0d2b8c1ca7422a2e2/tmp.lbamkM
./t/with-docker+fail-test1 *failed* | dummy
./t/with-docker+fail-test1 *failed* | ERROR tests/with-docker/Earthfile:186:8
./t/with-docker+fail-test1 *failed* |       The command
./t/with-docker+fail-test1 *failed* |           WITH DOCKER RUN --privileged echo "dummy" && false
./t/with-docker+fail-test1 *failed* |       did not complete successfully. Exit code 1
./t/with-docker+fail-test1 | unlazy force execution: process "/bin/sh -c 
EARTHLY_DOCKERD_DATA_ROOT=\"/var/earthly/dind/7b5da64c458d405fff1f21d3a985f6812f9eaad6af9c6eb0d2b8c1
ca7422a2e2\" EARTHLY_DOCKERD_CACHE_DATA=\"false\" EARTHLY_DOCKER_LOAD_FILES=\"\" 
EARTHLY_IMAGES_WITH_DIGESTS=\"\" EARTHLY_START_COMPOSE=\"false\" EARTHLY_COMPOSE_FILES=\"\" 
EARTHLY_COMPOSE_SERVICES=\"\" DIND_IMAGE=earthly/dind:alpine-3.19-docker-25.0.5-r0 
OTEL_TRACES_EXPORTER=none PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin /var/earthly/dockerd-
wrapper.sh execute /usr/bin/earth_debugger /bin/sh -c 'echo \"dummy\" && false'" did not complete successfully: exit 
code: 1
```

The last part is confusing:
```
unlazy force execution: process "/bin/sh -c 
EARTHLY_DOCKERD_DATA_ROOT=\"/var/earthly/dind/7b5da64c458d405fff1f21d3a985f6812f9eaad6af9c6eb0d2b8c1
ca7422a2e2\" EARTHLY_DOCKERD_CACHE_DATA=\"false\" EARTHLY_DOCKER_LOAD_FILES=\"\" 
EARTHLY_IMAGES_WITH_DIGESTS=\"\" EARTHLY_START_COMPOSE=\"false\" EARTHLY_COMPOSE_FILES=\"\" 
EARTHLY_COMPOSE_SERVICES=\"\" DIND_IMAGE=earthly/dind:alpine-3.19-docker-25.0.5-r0 
OTEL_TRACES_EXPORTER=none PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin /var/earthly/dockerd-
wrapper.sh execute /usr/bin/earth_debugger /bin/sh -c 'echo \"dummy\" && false'" did not complete successfully: exit 
code: 1
```
In some conditions, only `unlazy force execution` generates an error, so we need it there, even if duplicated in this case. 

With this change the last error is properly formatted:
```
/U/a/s/gha_earthly_test+fail1 |       The command
/U/a/s/gha_earthly_test+fail1 |           WITH DOCKER RUN
/U/a/s/gha_earthly_test+fail1 |       did not complete successfully. Exit code 1
```

I also changed vertormon.parseError to use templates and added some tests around the error type and exit code parsing.

### DIFF
--- a/earthfile2llb/with_docker_run_local_reg.go
+++ b/earthfile2llb/with_docker_run_local_reg.go
@@ -124,7 +124,10 @@ func (w *withDockerRunLocalReg) Run(ctx context.Context, args []string, opt With
 
 	// Force synchronous command execution if we're using the local registry for
 	// loads and pulls.
-	return w.c.forceExecution(ctx, w.c.mts.Final.MainState, w.c.platr)
+	// The forced error will be returned elsewhere via magic I don't understand
+	// So swallowing the error here keeps error messages consistent.
+	_ = w.c.forceExecution(ctx, w.c.mts.Final.MainState, w.c.platr)
+	return nil
 }
 
 func (w *withDockerRunLocalReg) load(ctx context.Context, cmdID string, opt DockerLoadOpt) (chan *states.ImageDef, error) {
--- a/earthfile2llb/with_docker_run_local_reg.go
+++ b/earthfile2llb/with_docker_run_local_reg.go
@@ -45,8 +45,10 @@ func (w *withDockerRunLocalReg) Run(ctx context.Context, args []string, opt With
 	}
 
 	defer func() {
-		message := solvermon.FormatError("WITH DOCKER RUN", retErr.Error())
-		cmd.SetEnd(time.Now(), logstream.RunStatus_RUN_STATUS_FAILURE, message)
+		if retErr != nil {
+			message := solvermon.FormatError("WITH DOCKER RUN", retErr.Error())
+			cmd.SetEnd(time.Now(), logstream.RunStatus_RUN_STATUS_FAILURE, message)
+		}
 	}()
 
 	var imagesToBuild []*states.ImageDef
--- a/earthfile2llb/with_docker_run_local_reg.go
+++ b/earthfile2llb/with_docker_run_local_reg.go
@@ -41,8 +41,11 @@ func (w *withDockerRunLocalReg) Run(ctx context.Context, args []string, opt With
 		return errors.Wrap(err, "failed to create command")
 	}
 
+	endErrorExecution := true
 	defer func() {
-		cmd.SetEndError(retErr)
+		if endErrorExecution {
+			cmd.SetEndError(retErr)
+		}
 	}()
 
 	var imagesToBuild []*states.ImageDef
@@ -126,9 +129,9 @@ func (w *withDockerRunLocalReg) Run(ctx context.Context, args []string, opt With
 	// loads and pulls.
 	err = w.c.forceExecution(ctx, w.c.mts.Final.MainState, w.c.platr)
 	if err != nil && errors.Is(err, ErrUnlazyForceExecution) {
-		// The forced error will be returned elsewhere via magic I don't understand
-		// So swallowing the error here keeps error messages consistent.
-		return nil
+		// If error is ErrUnlazyForceExecution, Error is logged via SetEnd in
+		// solvermon.handleBuildkitStatus. So we don't setend we don't SetEndError here
+		endErrorExecution = false
 	}
 	return err
 }
--- a/earthfile2llb/with_docker_run_local_reg.go
+++ b/earthfile2llb/with_docker_run_local_reg.go
@@ -124,10 +124,13 @@ func (w *withDockerRunLocalReg) Run(ctx context.Context, args []string, opt With
 
 	// Force synchronous command execution if we're using the local registry for
 	// loads and pulls.
-	// The forced error will be returned elsewhere via magic I don't understand
-	// So swallowing the error here keeps error messages consistent.
-	_ = w.c.forceExecution(ctx, w.c.mts.Final.MainState, w.c.platr)
-	return nil
+	err = w.c.forceExecution(ctx, w.c.mts.Final.MainState, w.c.platr)
+	if err != nil && errors.Is(err, ErrUnlazyForceExecution) {
+		// The forced error will be returned elsewhere via magic I don't understand
+		// So swallowing the error here keeps error messages consistent.
+		return nil
+	}
+	return err
 }
 
 func (w *withDockerRunLocalReg) load(ctx context.Context, cmdID string, opt DockerLoadOpt) (chan *states.ImageDef, error) {
--- a/earthfile2llb/with_docker_run_local_reg.go
+++ b/earthfile2llb/with_docker_run_local_reg.go
@@ -3,8 +3,11 @@ package earthfile2llb
 import (
 	"context"
 	"fmt"
+	"time"
 
+	"github.com/earthly/cloud-api/logstream"
 	"github.com/earthly/earthly/domain"
+	"github.com/earthly/earthly/logbus/solvermon"
 	"github.com/earthly/earthly/states"
 	"github.com/earthly/earthly/util/containerutil"
 	"github.com/earthly/earthly/util/syncutil/semutil"
@@ -41,11 +44,9 @@ func (w *withDockerRunLocalReg) Run(ctx context.Context, args []string, opt With
 		return errors.Wrap(err, "failed to create command")
 	}
 
-	endErrorExecution := true
 	defer func() {
-		if endErrorExecution {
-			cmd.SetEndError(retErr)
-		}
+		message := solvermon.FormatError("WITH DOCKER RUN", retErr.Error())
+		cmd.SetEnd(time.Now(), logstream.RunStatus_RUN_STATUS_FAILURE, message)
 	}()
 
 	var imagesToBuild []*states.ImageDef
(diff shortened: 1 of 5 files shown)


earthfile2llb/with_docker_run_reg.go
```diff
@@ -126,7 +129,8 @@ func (w *withDockerRunRegistry) Run(ctx context.Context, args []string, opt With
 	}
 
 	defer func() {
-		cmd.SetEndError(retErr)
```
alexcb, line 129: I wonder if you could keep this existing function and instead wrap the error? e.g. `errors.Wrap(retErr, "WITH DOCKER RUN")`?

earthfile2llb/with_docker_run_local_reg.go
```diff
@@ -42,7 +45,8 @@ func (w *withDockerRunLocalReg) Run(ctx context.Context, args []string, opt With
 	}
 
 	defer func() {
-		cmd.SetEndError(retErr)
+		message := solvermon.FormatError("WITH DOCKER RUN", retErr.Error())
+		cmd.SetEnd(time.Now(), logstream.RunStatus_RUN_STATUS_FAILURE, message)
```
alexcb, line 49: this feels wrong, this will be called even when `retErr == nil`. Won't this always mark a run as failed?

logbus/solvermon/vertexmon.go
```diff
@@ -37,111 +37,147 @@ type vertexMonitor struct {
...
-var reErrNotFound = regexp.MustCompile(`^failed to calculate checksum of ref ([^ ]*): (.*)$`)
-var reHint = regexp.MustCompile(`^(?P<msg>.+?):Hint: .+`)
+var reErrExitCode = regexp.MustCompile(`(?:process ".*" did not complete successfully|error calling LocalhostExec): exit code: (?P<exit_code>[0-9]+)$`)
 
-func (vm *vertexMonitor) Write(dt []byte, ts time.Time, stream int) (int, error) {
-	if stream == BuildkitStatsStream {
-		stats, err := vm.ssp.Parse(dt)
+func getExitCode(errString string) uint64 {
+	if reErrExitCode.MatchString(errString) {
+		matches, _ := stringutil.NamedGroupMatches(errString, reErrExitCode)
```
alexcb, line 44: this feels a bit odd, it looks like the `reErrExitCode` regex is being called twice (once in `reErrExitCode.MatchString(errString)`, then a second time inside the `NamedGroupMatches` function).

is it possible to only evaluate the regex once?

logbus/solvermon/vertexmon.go
```diff
@@ -37,111 +37,147 @@ type vertexMonitor struct {
...
+		exitCodeMatch := matches["exit_code"][0]
+		exitCode, err := strconv.ParseUint(exitCodeMatch, 10, 32)
 		if err != nil {
-			return 0, errors.Wrap(err, "failed decoding stats stream")
+			return 0
```
alexcb, line 48: this feels wrong: why are we returning an error code of `0`, which indicates things are ok when in fact they aren't.

I would change this function to return (uint64, error), or maybe even simply (int, error) unless there's a specific reason you need 64 bits or unsigned.

logbus/solvermon/vertexmon.go
```diff
@@ -37,111 +37,147 @@ type vertexMonitor struct {
...
-			if err != nil {
-				return 0, errors.Wrap(err, "stats json encode failed")
-			}
-			_, err = vm.cp.Write(statsJSON, ts, int32(stream))
-			if err != nil {
-				return 0, errors.Wrap(err, "write stats")
-			}
+		return exitCode
+	}
+	return 0
```
alexcb, line 52: similar comment, returning `0` seems odd when no exit code was found.

logbus/solvermon/vertexmon.go
```diff
@@ -37,111 +37,147 @@ type vertexMonitor struct {
...
 }
 
-func (vm *vertexMonitor) parseError() {
-	errString := vm.vertex.Error
-	indentOp := strings.Join(strings.Split(vm.operation, "\n"), "\n          ")
-	internalStr := ""
-	if vm.meta.Internal {
-		internalStr = " internal"
-	}
+func formatErrorMessage(errString, operation, internal string, fatalErrorType logstream.FailureType, exitCode uint64) string {
```
alexcb, line 79: can we change `internal` to be a bool? That way it's more clear what values we can expect `internal` to be.

then inside the function we could create a

```
internalStr := ""
if internal {
    internalStr = " internal"
}
```

logbus/solvermon/vertexmon.go
```diff
@@ -37,111 +37,147 @@ type vertexMonitor struct {
...
-			}
+		return exitCode
+	}
+	return 0
+}
+
+var reErrNotFound = regexp.MustCompile(`^failed to calculate checksum of ref ([^ ]*): (.*)$`)
+var reHint = regexp.MustCompile(`^(?P<msg>.+?):Hint: .+`)
+
+func determineFatalErrorType(errString string, exitCode uint64) (logstream.FailureType, bool) {
```
alexcb, line 58: nit: add a comment describing that the bool represents if it's a fatal error.

logbus/solvermon/vertexmon.go
```diff
@@ -37,111 +37,147 @@ type vertexMonitor struct {
...
+
+func (vm *vertexMonitor) parseError() {
+	errString := vm.vertex.Error
+
+	indentOp := strings.Join(strings.Split(vm.operation, "\n"), "\n          ")
+	internalStr := ""
+	if vm.meta.Internal {
+		internalStr = " internal"
+	}
+	errString = fmt.Sprintf("%s%s", internalStr, errString)
```
alexcb, line 135: is this still needed? If yes, can we move this conversion inside the `formatErrorMessage` function?

logbus/solvermon/vertexmon.go
```diff
@@ -37,111 +37,147 @@ type vertexMonitor struct {
...
+		vm.errorStr = fmt.Sprintf("ERROR%s\n%s", slString, formattedError)
 	} else {
-		vm.errorStr = fmt.Sprintf("WARN%s: %s", slString, errString)
+		vm.errorStr = fmt.Sprintf("WARN%s: %s", slString, formattedError)
 	}
+	vm.isFatalError = isFatalError
+	vm.fatalErrorType = fatalErrorType
+}
+
+func (vm *vertexMonitor) Write(dt []byte, ts time.Time, stream int) (int, error) {
```
alexcb, line 160: my guess is this function wasn't changed, just moved?

adamgordonbell: Yeah, my bad. Fixed.

earthfile2llb/with_docker_run_local_reg.go
adamgordonbell: Yep. Fixed

logbus/solvermon/vertexmon.go
adamgordonbell: Yeah, good point. Fixed.

logbus/solvermon/vertexmon.go
adamgordonbell: fixed and moved to int

logbus/solvermon/vertexmon.go
```diff
@@ -37,111 +37,155 @@ type vertexMonitor struct {
...
-			statsJSON, err := json.Marshal(statsSample)
-			if err != nil {
-				return 0, errors.Wrap(err, "stats json encode failed")
-			}
-			_, err = vm.cp.Write(statsJSON, ts, int32(stream))
-			if err != nil {
-				return 0, errors.Wrap(err, "write stats")
-			}
+		// Check if the exit code can fit into an int
+		if exitCode > int64(^uint(0)>>1) {
```
adamgordonbell, line 52: To have getExitCode handle the existing case of math.MaxUint32 being used to indicate OOM and not get caught by the "go/incorrect-integer-conversion" lint rule, need to handle the case that exit code is greater than MaxInt

logbus/solvermon/vertexmon.go
```diff
@@ -37,111 +37,160 @@ type vertexMonitor struct {
...
-	if err != nil {
-		return 0, errors.Wrap(err, "write log line")
+		if exitCode > 255 {
+			return 0, false, fmt.Errorf("exit code %d out of expected range (0-255)", exitCode)
+		}
+		exitCodeByte := exitCode & 0xFF
+		return int(exitCodeByte), true, nil
 	}
-	return len(dt), nil
+	return 0, false, nil
```
alexcb, line 62: ```return 0, errNoExitCodeFound`

where `var errNoExitCodeFound = errors.New("No exit code was found")
```

alexcb (approved):
looks great! thanks for the various changes.

PR #4182, opened 2024-05-31
Several improvements re GHA integrations

idelvall:
- Adds new subcommands: `ls` and `remove`
- Renames subcommand category from `github` to `gha`, since it is GitHub Actions related
- Removes the hidden attribute
- Updates the documentation accordingly


cmd/earthly/subcmd/github_cmds.go
```diff
@@ -92,14 +152,97 @@ func (a *Github) actionAdd(cliCtx *cli.Context) error {
...
 	if err != nil {
 		return fmt.Errorf("error found running github add: %w", err)
 	}
 	a.cli.Console().Printf("GitHub integration successfully created")
 	return nil
 }
 
+func (a *Github) actionRemove(cliCtx *cli.Context) error {
+	a.cli.SetCommandName("githubRemove")
+	if a.Org == "" {
```
alexcb, line 165: this logic appears to be duplicated in multiple locations, can you move it to a function instead? e.g.

```
func (a *Github) getOrgName() { ...
```

or maybe even somehow done under the `before` function: https://github.com/earthly/earthly/blob/1e0e0ef3ea7b9058aaabfab2ad9088b867a22a5b/cmd/earthly/app/before.go#L127-L129

but maybe before won't work since that's used for the top level `--org` flag, and some of our commands have a different `--org` under subcommands, which is confusing and feels like a bigger issue.

cmd/earthly/subcmd/github_cmds.go
```diff
@@ -92,14 +152,97 @@ func (a *Github) actionAdd(cliCtx *cli.Context) error {
...
+		}
+		a.Org = a.cli.OrgName()
+	}
+	cloudClient, err := helper.NewCloudClient(a.cli)
+	if err != nil {
+		return err
+	}
+	integrations, err := cloudClient.ListGHAIntegrations(cliCtx.Context, a.Org)
+	if err != nil {
+		return fmt.Errorf("error found running github list: %w", err)
```
alexcb, line 197:
```suggestion
		return fmt.Errorf("github list failed: %w", err)
```

cmd/earthly/subcmd/github_cmds.go
```diff
@@ -92,14 +152,97 @@ func (a *Github) actionAdd(cliCtx *cli.Context) error {
...
+			GitHubRepoName: integration.GithubRepoName,
+			CreatedBy:      integration.CreatedBy,
+			CreatedAt:      integration.CreatedAt.AsTime().Format(time.DateTime),
+		}
+	}
+	b, err := json.MarshalIndent(jsonInts, "", "  ")
+	if err != nil {
+		return fmt.Errorf("failed to marshal json: %w", err)
+	}
+	a.cli.Console().Printf(string(b))
```
alexcb, line 225: what if `b` contains a `%` character? If you were calling a function like `Print` or `Println` this would be fine, but `Printf` expects a formatting string as the first argument.
```suggestion
	a.cli.Console().Printf("%s", string(b))
```

or simply `a.cli.Console().Printf("%s", b)`

also, do you need a `\n`?

alexcb (approved):

PR #4184, opened 2024-06-03
Remove Pipelines from Satellite commands

mikejholly:
This PR removes Pipelines-related code from the Satellite commands.

It also seems reasonable to remove the `IsHidden` Satellite listing parameter as, IFAIK, this is only used to include Satellites associated with pipelines. 

I'll create another PR to remove the `PIPELINE` keywords.


alexcb (approved):

brandonSc (approved):
Thanks for cleaning that up!

PR #4186, opened 2024-06-05
move exit code parsing to seperate function

alexcb:


alexcb:
This function will be moved into #4180

PR #4222, opened 2024-06-24
fix: Match internalfailed errors in formatErrorMessage() (fixes #4221)

### LINKED ISSUE #4221: Panic when trying to copy file from another target when the file does not exist
### What went wrong?

The following Earthfile will cause earthly to panic:

```Dockerfile
VERSION 0.8

compile:
  FROM alpine:3
  SAVE ARTIFACT /doesnotexist

build:
  FROM alpine:3
  COPY +compile/doesnotexist .
```

Repro with command:

Output:
```
$ earthly +build
 Init 🚀

Streaming logs to https://cloud.earthly.dev/builds/a24dc822-06ba-484c-89bc-e388665ab9b3

 Build 🔧

              logbus | Setting organization "jahands" and project ""
              +build | --> FROM +base
            alpine:3 | --> Load metadata alpine:3 linux/arm64
       registry auth | logged into registry-1.docker.io with username jhands (using credentials from /user/std/registry/registry-1.docker.io/)
            +compile | --> FROM +base
            +compile | --> FROM alpine:3
            +compile | [----------] 100% FROM alpine:3
panic: runtime error: index out of range [2] with length 0

goroutine 133 [running]:
github.com/earthly/earthly/logbus/solvermon.formatErrorMessage({0x140003c0780, 0x74}, {0x14000118420, 0x20}, 0x1, 0x3, 0x0)
        github.com/earthly/earthly/logbus/solvermon/vertexmon.go:120 +0x4f0
github.com/earthly/earthly/logbus/solvermon.(*vertexMonitor).parseError(0x1400091d900)
        github.com/earthly/earthly/logbus/solvermon/vertexmon.go:154 +0xc8
github.com/earthly/earthly/logbus/solvermon.(*SolverMonitor).handleBuildkitStatus(0x1400032d580, {0x14000813eb8?, 0xa?}, 0x140009982a0)
        github.com/earthly/earthly/logbus/solvermon/solvermon.go:140 +0x5c0
github.com/earthly/earthly/logbus/solvermon.(*SolverMonitor).MonitorProgress(0x1400032d580, {0x103696498, 0x140000e6b40}, 0x1400062e2a0)
        github.com/earthly/earthly/logbus/solvermon/solvermon.go:62 +0x154
github.com/earthly/earthly/builder.(*solver).buildMainMulti.func2()
        github.com/earthly/earthly/builder/solver.go:72 +0x30
golang.org/x/sync/errgroup.(*Group).Go.func1()
        golang.org/x/sync@v0.6.0/errgroup/errgroup.go:78 +0x58
created by golang.org/x/sync/errgroup.(*Group).Go in goroutine 1
        golang.org/x/sync@v0.6.0/errgroup/errgroup.go:75 +0x98
```

### What should have happened?

Shown a useful error message about the file not being found (maybe even have SAVE ARTIFACT fail if it's not found?)

### What earthly version?

$ earthly --version
earthly version v0.8.14 6b41f8409d7ffef0d25072c2c04250074b6e3c72 darwin/arm64; macOS 14.5 homebrew

### Buildkit Logs

_No response_

### Other Helpful Information

This issue was introduced in #4180 here: https://github.com/earthly/earthly/blob/main/logbus/solvermon/vertexmon.go#L120
We're trying to apply a regex `^failed to calculate checksum of ref ([^ ]*): (.*)$`
To a string like this:
```
 internalfailed to calculate checksum of ref p4gz72iufvk3t1nsqq07p9sim::m4m7o7gui4zuuoy9vynbrzx8f: "/doesnotexist": not found
```

jahands:
See #4221 for details and how to repro.

Added tests to validate that the regex matches as expected.

Didn't have time to add tests for formatErrorMessage(), but I confirmed that it correctly formats the error now:
CleanShot 2024-06-24 at 08 54 16@2x

Also verified that it doesn't panic when the regex fails to match:
CleanShot 2024-06-24 at 08 54 00@2x

### DIFF
--- a/logbus/solvermon/vertexmon.go
+++ b/logbus/solvermon/vertexmon.go
@@ -63,7 +63,7 @@ func getExitCode(errString string) (int, error) {
 	return 0, errNoExitCode
 }
 
-var reErrNotFound = regexp.MustCompile(`^failed to calculate checksum of ref ([^ ]*): (.*)$`)
+var reErrNotFound = regexp.MustCompile(`^\s*(internal)?failed to calculate checksum of ref ([^ ]::[^ ]*|[^ ]*): (.*)\s*$`)
 var reHint = regexp.MustCompile(`^(?P<msg>.+?):Hint: .+`)
 
 // determineFatalErrorType returns logstream.FailureType
@@ -114,10 +114,14 @@ func formatErrorMessage(errString, operation string, internal bool, fatalErrorTy
 				"      did not complete successfully. Exit code %d", internalStr, operation, exitCode)
 	case logstream.FailureType_FAILURE_TYPE_FILE_NOT_FOUND:
 		m := reErrNotFound.FindStringSubmatch(errString)
+		reason := fmt.Sprintf("unable to parse file_not_found error:%s", errString)
+		if len(m) > 2 {
+			reason = m[3]
+		}
 		return fmt.Sprintf(
 			"      The%s command\n"+
 				"          %s\n"+
-				"      failed: %s", internalStr, operation, m[2])
+				"      failed: %s", internalStr, operation, reason)
 	case logstream.FailureType_FAILURE_TYPE_GIT:
 		gitStdErr, shorterErr, ok := errutil.ExtractEarthlyGitStdErr(errString)
 		if ok {
--- a/logbus/solvermon/vertexmon_test.go
+++ b/logbus/solvermon/vertexmon_test.go
@@ -5,6 +5,7 @@ import (
 	"testing"
 
 	"github.com/earthly/cloud-api/logstream"
+	"github.com/stretchr/testify/assert"
 )
 
 func TestGetExitCode(t *testing.T) {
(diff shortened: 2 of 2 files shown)


Earthfile
```diff
@@ -1,5 +1,5 @@
 VERSION 0.8
-PROJECT earthly-technologies/core
```
brandonSc, line 2: I assume you had to change this to get it working locally? We should probably revert it back

jahands: Oops, yeah will revert later today

jahands: @brandonSc fixed!

jahands:
CI failure looks unrelated

brandonSc:
> CI failure looks unrelated

Yes, I believe so. The same is failing on other PRs. Working on getting this merged (I don't have permission to myself).

vladaionescu (approved):
This looks good - it would be nice for @mikejholly to take a look too though. This is related to recent panic reports.

mikejholly (approved):

PR #4233, opened 2024-06-26
Fix device name parsing for MTU discovery when using link scoped devices

ben-krieger:
This fix is relevant for tun devices, which are common in VPN implementations. For example, while a "normal" default route may look like:

	default via 192.168.1.1 dev enp1s0 proto dhcp src 192.168.1.123 metric 100

A tun device for a VPN may instead look like:

	default dev tun0 scope link

In the current entrypoint, the device name would be incorrectly parsed as "link" instead of "tun0". By using sed to find the word after "dev", the result is more likely to be correct. Sed is used elsewhere in the entrypoint, so it is already an expected utility.

When the device name is incorrectly parsed, the earthly-buildkitd container fails with the error:

	cat: can't open '/sys/class/net/link/mtu': No such file or directory

### DIFF
--- a/buildkitd/entrypoint.sh
+++ b/buildkitd/entrypoint.sh
@@ -138,7 +138,7 @@ echo "$EARTHLY_GIT_CONFIG" | base64 -d >/root/.gitconfig
 
 #Set up CNI
 if [ -z "$CNI_MTU" ]; then
-  device=$(ip route show | grep default | cut -d' ' -f5 | head -n 1)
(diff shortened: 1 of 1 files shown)


ben-krieger:
It looks like the docs here will also need to be updated: https://docs.earthly.dev/docs/guides/using-the-earthly-docker-images/buildkit-standalone

ben-krieger:
The current workaround for this issue is to set in `~/.earthly/config.yml`:

```
global:
    cni_mtu: 1422
```

brandonSc:
Thanks for the submission!

brandonSc (approved):

PR #4253, opened 2024-07-12
Fix issue in test using `--ip` in default docker bridge network

dchw:


vladaionescu (approved):

brandonSc (approved):

brandonSc:
BTW apparently removing the IP also worked? https://github.com/earthly/earthly/pull/4247

PR #4325, opened 2025-07-16
Remove Cloud

dchw:
This PR should remove all cloud functionality. It makes testing hard because many of the configuration options we used to test this also lived in the cloud...


dchw:
Merging despite CI failures - since `secrets` and other components have been removed.

PR #4329, opened 2025-07-18
Use 16MB step log size

kp-mariappan-ramasamy:


kp-mariappan-ramasamy:
We faced an issue where the builds failed with followin error:
```
2025-07-17T05:55:10.4757117Z Error: failed decoding stats stream: invalid character '\x00' looking for beginning of value
2025-07-17T05:55:10.4761415Z ##[error]failed decoding stats stream: invalid character '\x00' looking for beginning of value
```

I tried to increase the buffer size, but it did not help

PR #4330, opened 2025-07-24
Remove satellites

### LINKED ISSUE #14: Allow configuration of cache location
Currently, `/tmp/earthly` is hard-coded.

I suspect workaround via symlink would not work as docker would mount the symlink, not the dir it links to, but not sure.

gilescope:
Leave only references to self hosted buildkits. Fixes: #14

Todo:
  * [ ] Fixup more of docs


gilescope:
wrong repo!

PR #4345, opened 2026-03-11
fix: duplicate OTel schema

janishorsts:
This eliminates all duplicate OTel attributes, including the failing schema one.


vladaionescu:
Did you mean to PR against EarthBuild?

janishorsts:
@vladaionescu, yes 😃. I still create PRs against the @earthly repositories by accident.

PR #4347, opened 2026-03-15
feat: read IdentityAgent from ~/.ssh/config for SSH agent socket

### LINKED ISSUE #3179: Integration between Earthly and 1Password
**Use case**
1Password has a [integrated SSH agent](https://developer.1password.com/docs/ssh/agent/?utm_medium=organic&utm_source=oph&utm_campaign=macos) but Earthly doesn't find it automatically. My environment is correctly configured, but I always need to pass the  SSH socket `earthly --ssh-auth-sock /Users/user/Library/Group Containers/2BUA8C4S2C.com.1password/t/agent.sock +build`.

**Expected Behavior**
I would expect that Earthly could correctly load the configuration from the `~/.ssh/config`:
```
Host *
	IdentityAgent "~/Library/Group Containers/2BUA8C4S2C.com.1password/t/agent.sock"
```

mateenali66:
Fixes #3179

## Summary

Earthly was ignoring the `IdentityAgent` directive in `~/.ssh/config`. This means users with custom SSH agents (like 1Password at `~/Library/Group Containers/2BUA8C4S2C.com.1password/t/agent.sock`) had to pass `--ssh-auth-sock` manually every single time which is quiet annoying.

This PR adds a fallback: when `SSH_AUTH_SOCK` is not set, Earthly will read `~/.ssh/config` and look for `IdentityAgent` directive (from `Host *` or global scope) and uses that as the SSH auth socket. Tilde and env variables in the path are expanded correctly.

## Changes

- `util/sshutil/identity_agent.go` -- New helper `GetSSHAuthSock()` that checks `SSH_AUTH_SOCK` first then falls back to parsing `~/.ssh/config` for `IdentityAgent`
- `cmd/earthly/flag/global.go` -- Use `sshutil.GetSSHAuthSock()` as default for `--ssh-auth-sock` instead of just reading `SSH_AUTH_SOCK` env directly

## Test plan

- [x] Unit tests for path expansion (tilde, env vars)
- [x] Unit tests for SSH config parsing (wildcard host, global scope, no config, SSH_AUTH_SOCK precedence)
- [x] `go build ./cmd/earthly/` passes
- [x] `go test ./util/sshutil/` -- all 6 tests pass

## Summary by CodeRabbit

* **New Features**
  * Improved SSH authentication socket detection: when SSH_AUTH_SOCK is unset, the app now consults SSH config to locate IdentityAgent entries, handles quoted paths, expands ~ and environment variables, and prefers universal wildcard or global directives.

* **Tests**
  * Added extensive tests covering path expansion, config parsing, precedence of env values, wildcard/host matching, quoted paths, and no-config scenarios.

### DIFF
--- a/cmd/earthly/flag/global.go
+++ b/cmd/earthly/flag/global.go
@@ -1,7 +1,6 @@
 package flag
 
 import (
-	"os"
 	"time"
 
 	"github.com/urfave/cli/v2"
@@ -9,6 +8,7 @@ import (
 	"github.com/earthly/earthly/buildkitd"
 	"github.com/earthly/earthly/cmd/earthly/common"
 	"github.com/earthly/earthly/util/containerutil"
+	"github.com/earthly/earthly/util/sshutil"
 )
 
 const (
@@ -108,7 +108,7 @@ func (global *Global) RootFlags(installName string, bkImage string) []cli.Flag {
 		},
 		&cli.StringFlag{
 			Name:        "ssh-auth-sock",
-			Value:       os.Getenv("SSH_AUTH_SOCK"),
+			Value:       sshutil.GetSSHAuthSock(),
 			EnvVars:     []string{"EARTHLY_SSH_AUTH_SOCK"},
 			Usage:       "The SSH auth socket to use for ssh-agent forwarding",
 			Destination: &global.SSHAuthSock,
--- a/util/sshutil/identity_agent.go
+++ b/util/sshutil/identity_agent.go
@@ -1,3 +1,6 @@
+// Package sshutil provides helpers for SSH configuration and authentication.
+// Currently it exposes GetSSHAuthSock to resolve the SSH agent socket path
+// from the environment or from ~/.ssh/config (IdentityAgent directive).
 package sshutil
 
 import (
--- a/util/sshutil/identity_agent.go
+++ b/util/sshutil/identity_agent.go
@@ -51,23 +51,17 @@ func identityAgentFromConfig() string {
 			continue
 		}
 
-		lower := strings.ToLower(line)
-		if strings.HasPrefix(lower, "host ") {
+		// Extract keyword and value, handling both "keyword value" and
+		// "keyword=value" forms (both are valid per ssh_config(5)).
+		keyword, value := splitDirective(line)
+		switch strings.ToLower(keyword) {
+		case "host":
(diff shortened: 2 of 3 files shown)


mateenali66:
Thanks for the review! Pushed a follow-up commit (eb074f7c) that address the feedback:

- Strip surrounding quotes from IdentityAgent values so quoted paths (common with 1Password and paths with spaces) works correctly
- Updated Host pattern matching to handle multi-pattern lines like "Host * !excluded.example.com" by checking if any non-negation token contains a wildcard
- Added docstrings to all functions to bring coverage up
- Added tests for quoted paths, multi-pattern Host, stripQuotes and isWildcardHostPattern

mateenali66:
recheck

mateenali66:
Pushed another commit (15aa4127) to fix the two bugs from the latest review:

1. isWildcardHostPattern was treating *.example.com as an universal wildcard because the check was strings.Contains(token, "*"). Renamed the function to isUniversalWildcard and tightened it to token == "*", so only a bare asterisk counts. *.example.com now correctly returns false.

2. IdentityAgent inside a specific Host block (e.g. Host github.com) was bleeding into the global fallback due to missing host-block tracking. Added a seenAnyHostBlock flag so only IdentityAgent directives that appear before any Host or Match block are accepted as the global default.

Added TestIdentityAgentHostSpecificIsIgnored to lock in that behavior. All 11 tests passing.

mateenali66:
Addressed the latest round of feedback (commit e75329c0):

- Extracted a writeSSHConfig test helper to eliminate the repeated tmpDir / .ssh/config / env setup across all the config-parsing tests
- Added TestIdentityAgentMatchScopeIsIgnored -- Match-scoped IdentityAgent should be ignored same as a host-specific one, this locks that in
- Added package doc comment to push docstring coverage up

12 tests passing.

mateenali66:
That one was real so fixed it (commit 39d81c23).

The parser was checking strings.HasPrefix(lower, "host ") etc, which missed tab-separated keywords (Host\t*) and key=value syntax (IdentityAgent=/path) -- both valid per ssh_config(5).

Replaced the prefix checks with a splitDirective helper that tokenizes on the first whitespace or = character, then switches on the lowercase keyword. Added tests for tab-separated and equals syntax, plus a unit test for splitDirective directly.

15 tests passing now.

